PR #22, opened 2016-10-20
Add auto reset support to Nano32

MakerAsia:
Board Nano32 has auto flash. This pull request is to add auto reset to Nano32 after write flash. This will make users of Nano32 more convenient and do not need to reset by manual after flash.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -94,4 +94,4 @@ tools.esptool.network_cmd.windows=python.exe
 tools.esptool.upload.protocol=esp32
 tools.esptool.upload.params.verbose=
 tools.esptool.upload.params.quiet=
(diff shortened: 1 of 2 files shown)


tuanpmt:
I would like to reset with this option `--reset-after-flash` 😋

MakerAsia:
Which repository is an upstream for esptool using in this project?

projectgus:
Hi,

Upstream esptool.py maintainer here. :) The ESP32 support work-in-progress is housed here at the moment:
https://github.com/themadinventor/esptool/pull/121

Supporting "exit flashing mode" is on the list to support but not yet implemented, sorry! I plan to finish it very soon though.

igrr:
ESP32 support for esptool is being developed here:
https://github.com/themadinventor/esptool/pull/121
Both ESP-IDF and this project get esptool there.

projectgus:
Thanks for being patient as we added this feature. As of about a day ago esp-idf has support to configure reset after flash (defaults to enabled). Once Arduino's copy of esptool.py is updated then it will also have this feature.

PR #26, opened 2016-10-27
Add from original Arduino.h

volca:
The defines are needed by library [Adafruit_SSD1306](https://github.com/adafruit/Adafruit_SSD1306/blob/master/examples/ssd1306_128x64_i2c/ssd1306_128x64_i2c.ino)


me-no-dev:
thanks :)

PR #35, opened 2016-11-08
Add "WEMOS LoLin32" boards support.

wemos:
@projectgus 
@me-no-dev 
Add "WEMOS LoLin32" boards support.
Thank you.


variants/lolin32/pins_arduino.h
```diff
@@ -0,0 +1,23 @@
...
+#define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
+#define digitalPinHasPWM(p)         (p < 34)
+
+#define LED_BUILTIN 2
+#define KEY_BUILTIN 0
+
+static const uint8_t SDA = 21;
+static const uint8_t SCL = 22;
+
+static const uint8_t SS    = 2;
```
me-no-dev: shouldn't SS be GPIO 5?

wemos: The ESP32's "General purpose SPI" can be remap to any GPIO

me-no-dev: true, but it will not work just by defining it here ;)

wemos: I agree with you, changed the SS to GPIO5.
Thank you.

PR #46, opened 2016-11-13
Actually disable ssl when on windows (Cygwin)

ed7coyne:
https://github.com/espressif/arduino-esp32/commit/4865ed0d1382f67d7ed482edc20b81ce6835c0bb says it disables on windows but does the opposite. python in cygwin says it can't find "ssl".

### DIFF
--- a/tools/get.py
+++ b/tools/get.py
@@ -71,13 +71,13 @@ def get_tool(tool):
     local_path = dist_dir + archive_name
     url = tool['url']
     #real_hash = tool['checksum'].split(':')[1]
-    if 'CYGWIN_NT' in sys_name:
+    if 'CYGWIN_NT' not in sys_name:
         ctx = ssl.create_default_context()
         ctx.check_hostname = False
         ctx.verify_mode = ssl.CERT_NONE
     if not os.path.isfile(local_path):
         print('Downloading ' + archive_name);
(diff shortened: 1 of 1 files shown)


tools/get.py
```diff
@@ -71,13 +71,13 @@ def get_tool(tool):
     local_path = dist_dir + archive_name
     url = tool['url']
     #real_hash = tool['checksum'].split(':')[1]
-    if 'CYGWIN_NT' in sys_name:
+    if 'CYGWIN_NT' not in sys_name:
```
me-no-dev: but this will disable SSL check on all other OSs. Maybe you need to install something to get it going?

ed7coyne: Ah I missunderstood what was going on here, I assumed the starting state before https://github.com/espressif/arduino-esp32/commit/4865ed0d1382f67d7ed482edc20b81ce6835c0bb was "ssl is enabled" but I see now you are correct it was disabled beforehand and that commit made it disabled only for windows. My mistake, sorry.

ed7coyne: Also as a datapoint while trying to determine how to get this script working for everyone. The script is broken on a clean install of "babun" (cygwin distro) on windows 10 but works if you follow the "not cygwin" path, i.e. with these changes.

PR #65, opened 2016-11-22
Arduino BLE Advertising

vonzcy:
First steps in making the Arduino BLE library for ESP32 based on current APIs.


vonzcy:
Would this be better now? The BLE library is still under development. I will try to add more functionalities as the bluetooth API grows. For now, many of the data structures and defines are not used, but they would be extended later. Thanks.

me-no-dev:
I spent some time last night with the code and with the example in IDF.
Some things will need to be redone and actually implemented before we merge it here :) no worries though I just about have it done.
```bt_controller_init()``` should be run way earlier, a thread for advertising should be started on the application core, C API should be exposed and CPP should build on that (see Wire, SPI and Serial).
Another thing is that code should properly wait for response which comes in the form of 7 bytes (which should be known what they mean) instead of just delaying 1 second and checking if it can send. If you send data before you get the response, responses get lost and you never get them.
Sending flow should be: send -> wait for response, which takes lots less than one second and allows for better discoverability and so on.
I noticed the 7 byte response has the command sent as bytes 4 and 5. 6 is probably status. Can we get the meaning of those? The whole communication from then on seem to just require proper packets sent through the API

Here is some debug:
```
BTDM CONTROLLER VERSION: 010102
BTDM ROM VERSION 0101
BD_ADDR: 24:0A:C4:00:3E:C6
NVDS MAGIC FAILED
phy_version: 187, Oct 10 2016, 19:23:46, 0
RF Init OK
BLE: Advertising interval: min = 256 ms, max = 256 ms

//reset baseband
BLE: tx[4]: 01 03 0c 00
BLE: tx_ready // API_vhci_host_check_send_available() is true here
BLE: rx[7]: 04 0e 04 05 03 0c 00 // but response comes later

//send advertising parameters
BLE: tx[19]: 01 06 20 0f 3f d4 1c 08 40 5e 75 fc 3f c9 00 00 00 72 0e
BLE: tx_ready
BLE: rx[7]: 04 0e 04 05 06 20 12

//send device name
BLE: tx[36]: 01 08 20 20 11 02 01 06 0d 09 4d 59 5f 42 4c 55 45 54 4f 4f 54 48 00 00 00 00 00 00 00 00 00 00 00 00 00 00
BLE: tx_ready
BLE: rx[7]: 04 0e 04 05 08 20 00

//send company info
BLE: tx[36]: 01 08 20 20 1e 02 01 06 1a ff 4c 00 02 15 74 27 8b da b6 44 45 20 8f 0c 72 0e af 05 99 35 00 00 00 00 c9 00
BLE: tx_ready
BLE: rx[7]: 04 0e 04 05 08 20 00

//start advertising
BLE: tx[5]: 01 0a 20 01 01
BLE: tx_ready
BLE: rx[7]: 04 0e 04 05 0a 20 00

//my phone attempts to connect
BLE: rx[16]: 02 00 20 0b 00 07 00 04 00 10 01 00 ff ff 00 28
BLE: rx[16]: 02 00 20 0b 00 07 00 04 00 10 01 00 ff ff 00 28
BLE: rx[16]: 02 00 20 0b 00 07 00 04 00 10 01 00 ff ff 00 28

//after it had given up, we get this status from baseband
BLE: rx[7]: 04 05 04 00 00 00 13

```

**Comment by me-no-dev** [comment]
Actually does not seem to need advertising thread at all. I wander why it's in the example

**Comment by igrr** [comment]
We have discussed this here a bit... The code will be rewritten to use the ready callback (parsing responses is not strictly needed in the case of advertising). Also @vonzcy will remove the unneeded advertising task.

Regarding writing C wrapper — I don't see much point in this. We are already talking to a C API here, and it is not hardware specific. In the case of peripherals (Wire, Serial, and do on), register layouts are different between chips, so we need a hardware abstraction layer which provides common interface which is used by C++ classes. In this case, VHCI is already such an interface. Wrapping it into another C layer doesn't really help solving any problems.

Regarding handling connections, this is out of scope of this library, for now. It does just advertising, which doesn't require handling of responses. Once bluedroid is integrated, we can extend this to do other things (such as GATT and GAP) based on bluedroid.

**Comment by me-no-dev** [comment]
parsing the response and waiting for it makes advertising take 1ms as opposed to 5 seconds. I think that matters ;) Starting bluetooth earlier (where WiFi usually starts) strips off another second and possible issues at boot. Properly parsing error response when someone trys to connect and it couldn't, notifies us that we are no longer advertising and can start again on that event.
This is as far as I have gotten and things are working properly. Code change is not much

**Comment by me-no-dev** [comment]
while we can start BLE at will (from any thread/core), we currently can not start WiFi at will. What that means is that there is no way decide in setup() which radio you want running. Currently if WiFi is not used throughout the sketch, WiFi will not boot and BLE is good to go, but if you let's say do
```cpp
if(digitalRead(STACK_SELECT_PIN)){
  WiFi.begin(ssid, pass);
} else {
  BLE.begin("ESP32");
}
``` 
WiFi stack will be booted in app_main and BLE will fail.

This is an issue in IDF that should be resolved soon. WiFi should be able to be started at any point, just like BLE.
@vonzcy I have an esp-idf-arduhal repo in gitlab. Please check the main folder where I have a copy of the original files in this PR that I'm working on. We can further discuss it there with the code.

vonzcy:
I am trying to make GATT service and characteristic classes now (first draw the data strcutres we need and have a skeleton). For the 7-byte response, yes, byte 6 is the status (Error Code), I will to add the status checking phase. Btw, in your debug output shown above, the packet setting the ble adv parameters returned an error code 0x12 (invalid packet). This is because you put the wrong pointer in ble_send_cmd() function.

Yes, I found your esp-idf-hal-arduino repo in gitlab. Thanks for your help.I am trying to make GATT service and characteristic classes now (first draw the data strcutres we need and have a skeleton). For the 7-byte response, yes, byte 6 is the status (Error Code), I will to add the status checking phase. Btw, in your debug output shown above, the packet setting the ble adv parameters returned an error code 0x12 (invalid packet). This is because you put the wrong pointer in ble_send_cmd() function.

Yes, I found your esp-idf-hal-arduino repo in gitlab. Thanks for your help.

me-no-dev:
yeah I noticed the error but advertising was working so I did not go much further :) Thanks for pointing it out ;) fixed in gitlab and now it returns 0 as expected

libraries/BLE/src/BLEPeripheral.h
```diff
@@ -0,0 +1,186 @@
...
+     * @param[in] localName  local name to advertise. Choose from:
+     *            BLE_ADVDATA_NO_NAME
+     *            BLE_ADVDATA_SHORT_NAME
+     *            BLE_ADVDATA_FULL_NAME
+     *
+     * @return none
+     *
+     * @note if the nameType is BLE_ADVDATA_NO_NAME, set localName to NULL
+     */
+  void setLocalName(ble_advdata_name_type_t nameType, String localName);
```
vonzcy: Sure, that's true. Thank you.

me-no-dev:
@vonzcy I have just committed a change to not start phy unless WiFi/BLE is used. Please add:
```cpp
#if !CONFIG_ESP32_PHY_AUTO_INIT
    arduino_phy_init();
#endif
```
before ```bt_controller_init();``` :)

PR #99, opened 2016-12-14
Add brief explanation about app_main() function when arduino used as a esp-idf component

eroniki:


README.md
```diff
@@ -112,7 +112,8 @@ Linux 32/64, Linux ARM (like Raspberry Pi, BeagleBone, CubieBoard).
           ```
         - Else you need to implement ```app_main()``` and call ```initArduino();``` in it.
 
-          Keep in mind that setup() and loop() will not be called in this case
+          Keep in mind that setup() and loop() will not be called in this case.
+          If you are writing your code on esp-idf examples, make sure there is just app_main() function in your main.cpp and example files.
```
me-no-dev: I'm not sure that this sentence is clear/correct. There could be much more than app_main in the main file. It needs to be clear that setup() and loop() are not running, so they should not be implemented and expected to be called.

PR #117, opened 2017-01-05
Error Handling for Window10 without CygWin

tobozo:
Added an Exception handling to the offending urlretrieve() call for the issue https://github.com/espressif/arduino-esp32/issues/108

It's not really a fix but the failure message is much more elegant than the initial stack trace as it hints at the solution to the setup problem rather than confusing the user into Windows problems.

### DIFF
--- a/tools/get.py
+++ b/tools/get.py
@@ -80,7 +80,15 @@ def get_tool(tool):
         if 'CYGWIN_NT' in sys_name:
             urlretrieve(url, local_path, report_progress,context=ctx)
         else:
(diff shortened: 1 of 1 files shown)


tobozo:
After some investigation the problem seems to be in urllib
  - python 2.7x
  - win10
  - no Cygwin installed (using git shell from Git For Windows)

Workaround: 
  #> pip install requests
  #> add "import requests" in get.py
  #> use the following code when urlretrieve raises an exception : 

                r = requests.get(url)
                f = open(local_path, 'wb')
                f.write(r.content)
                f.close()

PR #126, opened 2017-01-13
(Real) fix for cygwinless win10 TLS error

tobozo:
Added a new requirement ( pip install request ) and some logic to get it working, I won't blame you if you reject this PR but this is what made it work on my machine


tobozo:
Real fix for #108

me-no-dev:
Thanks!

PR #130, opened 2017-01-16
SDK headers: Use new esp-idf signal name VSPICLK_OUT_IDX

projectgus:
Hi @me-no-dev,

esp-idf commit esp-idf/23455de4 renamed VSPICLK_OUT_MUX_IDX to VSPICLK_OUT_IDX. This PR updates the name to avoid compilation errors when using arduino as an esp-idf component.

FWIW I think this should be safe to change even without rebuilding the binary libraries, as there's no object file difference. But it will break anyone compiling arduino as a submodule in older esp-idf. 

Angus

### DIFF
--- a/cores/esp32/esp32-hal-spi.c
+++ b/cores/esp32/esp32-hal-spi.c
@@ -27,7 +27,7 @@
 #include "soc/gpio_sig_map.h"
 #include "soc/dport_reg.h"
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
I have this prepped as well :) Just did not get to rebuilding, though you are correct to say that it is not required :) simple define change does.

PR #152, opened 2017-01-27
Add WiFiSSLClient

copercini:
Provide SSL/TLS functions to ESP32 with Arduino IDE


me-no-dev:
let's start with moving the files out of the core. Maybe the whole SSL Client and eventually server should be a separate library (again nothing in the core).
Then I'll run some formatting on the code and we will look from there :) how does that sound?

copercini:
Sounds good. 

The files in the core can easily be moved to the same folder of library.  Also,  you can format or edit anything you want =)

libraries/WiFi/src/ssl_drv.h
```diff
@@ -0,0 +1,33 @@
+/* Provide SSL/TLS functions to ESP32 with Arduino IDE
+ * Ported from Ameba8195 (no license specified) to ESP32 by Evandro Copercini - 2017 - Public domain
```
igrr: I'm afraid we need to get some confirmation about the original license of these files before we can consider adding them to the project.

copercini: Yes, you are right.

These files process the input to make a good call to mod_ssl.c

I will rewrite them based on WifiClientSecure of ESP8266 and make this lib compatible with ESP8266 arduino sketches.

Also I will move out of the core folder as @me-no-dev request.

PR #164, opened 2017-02-01
add WiFiClientSecure library

copercini:
-new WiFiClientSecure lib for ESP32
-fixed mbedTLS config file path in platform.txt
-put full bignum.h file insted of include_next (with bug)


libraries/WiFiClientSecure/src/WiFiClientSecure.h
```diff
@@ -0,0 +1,63 @@
...
+	virtual operator bool();
+
+	void setCACert(unsigned char *rootCA);
+	void setCertificate(unsigned char *client_ca);
+	void setPrivateKey (unsigned char *private_key);
+
+	int connect(IPAddress ip, uint16_t port, unsigned char* rootCABuff, unsigned char* cli_cert, unsigned char* cli_key);
+	int connect(const char *host, uint16_t port, unsigned char* rootCABuff, unsigned char* cli_cert, unsigned char* cli_key);
+
+	int startClient(sslclient_context *ssl_client, uint32_t ipAddress, uint32_t port, unsigned char* rootCABuff, unsigned char* cli_cert, unsigned char* cli_key);
```
me-no-dev: all those methods should be private/protected right? The user does not have access to the sslclient_context anyway.

copercini: That's right, any more bugs?

copercini:
It's not working for MQTT too. 

The connection is ok, but after connect the read function is called and the server don't send anything causing a timeout. (I tested with shiftr.io and AWS iot and both have the same problem)

I still don't realize where is the bug...

me-no-dev:
I found out why the header was "missing" from tools/sdk ;) will fix in next commit of idf libs

me-no-dev:
please see [here](https://github.com/espressif/esp-idf/blob/master/docs/style-guide.rst) for formatting your code :)

copercini:
ok, first I am trying to fix the problem with MQTT over TLS. I think the problem is in available() function which doesn't return the correct number of bytes to read, causing timeouts and lost connection.

I tried with `ioctl(ssl_client->socket, FIONREAD, &count);` and `mbedtls_ssl_get_bytes_avail(&ssl_client->ssl_ctx));` but it always return 0, even when there is.

me-no-dev:
can you please switch to my branch and see if you can compile without changing platform.txt and the header
```git pull && git checkout me-no-dev```

copercini:
It compiles fine without changing platform.txt and the header!

Based on this response of Angus http://esp32.com/viewtopic.php?f=13&t=1101 , the problem with MQTT seems fixed too.

me-no-dev:
I was looking at the idf examples yesterday and I noticed that OpenSSL client and server are also added. I think maybe we need to figure out better naming conversion and have both options available. I'll try to do some digging tonight and see if I come up with something.
You can sync this PR in a mean time to not include platform.txt and the bignim header

PR #170, opened 2017-02-05
Add hint

ionciubotaru:
I had a problem when I first try to program ESP32 via serial. You helped me quickly.
I think others will have this problem too, because on ESP8266 you must keep GPIO0 LOW only on boot.
I add this hint at the end of readme.md file.


me-no-dev:
BTW this in general will not be true and it's up to proper board design for you to not need to hold GPIO0 low

PR #177, opened 2017-02-06
Added boards constants for LED_BUILTIN and KEY_BUILTIN

arcao:
**Sources:**
* esp320: http://wiki.sweetpeas.se/index.php?title=ESP320#Developing_using_the_Arduino_IDE
* esp32thing: https://cdn.sparkfun.com/assets/learn_tutorials/5/0/7/esp32-thing-schematic.pdf
* espea32: https://github.com/AprilBrother/ESPea32/blob/master/hardware/ESPea32-schematic.pdf
* hornbill32dev: https://github.com/ExploreEmbedded/Hornbill-ESP32-DEV/blob/master/Design%20Files/Schematic/Hornbill%20Devboard.pdf
* lolin32: values from existing header
* nano32: http://iot-bits.com/closer-look-esp32-nano32/
* node32s: https://github.com/lamloei/node32s-book/blob/master/ch01/node32s-schematic.pdf

_Note: Some board producers have not published schematics yet, so it was little digging work._

### DIFF
--- a/variants/esp320/pins_arduino.h
+++ b/variants/esp320/pins_arduino.h
@@ -9,6 +9,12 @@
 #define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
 #define digitalPinHasPWM(p)         (p < 34)
 
+static const uint8_t LED_BUILTIN = 5;
+static const uint8_t BUILTIN_LED = 5;
+
+static const uint8_t TX = 1;
+static const uint8_t RX = 3;
+
 static const uint8_t SDA = 2;
 static const uint8_t SCL = 14;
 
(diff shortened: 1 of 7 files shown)


variants/esp320/pins_arduino.h
```diff
@@ -9,6 +9,12 @@
 #define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
 #define digitalPinHasPWM(p)         (p < 34)
 
+static const uint8_t LED_BUILTIN = 5;
+static const uint8_t BUILTIN_LED = 5;
```
me-no-dev: since those are the same, wouldn't it be better if one of them is define to forward to the other one?

arcao: Yes, you are right, I will do it. I tried to do it same way as is in ESP8266 core.

arcao:
Done.

me-no-dev:
thanks :)

PR #204, opened 2017-02-12
Fix for implicit declaration of function nvs_flash_init error

arcao:
This error occurs only when `CONFIG_ESP32_PHY_AUTO_INIT=y`.


me-no-dev:
aaahhh :)

PR #205, opened 2017-02-13
Fix for WiFiClientSecure.

wamisnet:
When I tried to use WiFiClientSecure, I got an error: error return statement in function returning non-void.


copercini:
Facepalm for me.

I just tested the most complex functions with certificates, MQTT, stress test and forget the return of basic ones. Thanks for fix it =)

PR #222, opened 2017-02-18
Improvements in EspClass

arcao:
* fixed not working functions for flash chip size, speed and mode
* added function to retrieve chip revision from eFuse
* ~`flashRead` / `flashWrite` supports encrypted flash~ (reverted, see @igrr review)

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -202,12 +202,13 @@ bool EspClass::flashEraseSector(uint32_t sector)
     return spi_flash_erase_sector(sector) == ESP_OK;
 }
 
+// Warning: These functions do not work with encrypted flash
 bool EspClass::flashWrite(uint32_t offset, uint32_t *data, size_t size)
 {
-    return spi_flash_write_encrypted(offset, (uint32_t*) data, size) == ESP_OK;
+    return spi_flash_write(offset, (uint32_t*) data, size) == ESP_OK;
 }
 
 bool EspClass::flashRead(uint32_t offset, uint32_t *data, size_t size)
 {
-    return spi_flash_read_encrypted(offset, (uint32_t*) data, size) == ESP_OK;
+    return spi_flash_read(offset, (uint32_t*) data, size) == ESP_OK;
 }
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -117,7 +117,7 @@ uint32_t EspClass::getFreeHeap(void)
     return esp_get_free_heap_size();
 }
 
-uint8_t EspClass::getCpuRevision(void)
+uint8_t EspClass::getChipRevision(void)
 {
     return (REG_READ(EFUSE_BLK0_RDATA3_REG) >> EFUSE_RD_CHIP_VER_RESERVE_S) && EFUSE_RD_CHIP_VER_RESERVE_V;
 }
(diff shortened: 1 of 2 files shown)


cores/esp32/Esp.h
```diff
@@ -55,6 +57,7 @@ class EspClass
     ~EspClass() {}
     void restart();
     uint32_t getFreeHeap();
+    uint8_t getCpuRevision();
```
igrr: Suggest `getChipRevision`, as this is not really the CPU revision.

arcao:
OK, renamed.

cores/esp32/Esp.cpp
```diff
@@ -199,10 +204,10 @@ bool EspClass::flashEraseSector(uint32_t sector)
 
 bool EspClass::flashWrite(uint32_t offset, uint32_t *data, size_t size)
 {
-    return spi_flash_write(offset, (uint32_t*) data, size) == ESP_OK;
+    return spi_flash_write_encrypted(offset, (uint32_t*) data, size) == ESP_OK;
```
igrr: This will unconditionally encrypt and write to flash, with the restriction that writes must be 16-byte aligned. This may not be always desired, for example a file system like SPIFFS will try to write smaller chunks of data and fail due to this restriction. Also NAND way of writing (i.e. flipping 1s to 0s) will not work with spi_flash_write_encrypted.

I suggest adding APIs which specifically do encrypted writes (or even better, adding a library for this purpose).

cores/esp32/Esp.cpp
```diff
@@ -199,10 +204,10 @@ bool EspClass::flashEraseSector(uint32_t sector)
...
 }
 
 bool EspClass::flashRead(uint32_t offset, uint32_t *data, size_t size)
 {
-    return spi_flash_read(offset, (uint32_t*) data, size) == ESP_OK;
+    return spi_flash_read_encrypted(offset, (uint32_t*) data, size) == ESP_OK;
```
igrr: Same here, with flash encryption enabled, this will always try to decrypt data, even if it wasn't encrypted in the first place. This isn't a problem if data is always encrypted, but as I have mentioned above, unconditionally encrypting data may be problematic due to alignment restrictions and bit flipping not being possible.

cores/esp32/Esp.cpp
arcao: Thanks for info. I'll revert this back and add note that this doesn't work with encrypted flash.

me-no-dev:
I dont think we should venture in flash encryption in arduino yet. I prefer getting the system up unencrypted first then see what will work encrypted and what will need to be adapted. I would like to really spend time with encryption and test its ins and outs first.

cores/esp32/Esp.cpp
igrr: That makes sense. I suppose that another things which is missing in Arduino for flash encryption support are:
- some kind of way to select whether the bootloader needs to perform flash encryption
- some place to store the unique per-device flash encryption key on the PC (and make it available to different Arduino installations on the same computer?)
- a way to say "I want to enable flash encryption for this board", which will cause the key to be generated and store to the aforementioned location. This part may be implemented using an IDE plugin.
- some diagnostics when a user tries to upload sketch into an ESP which has flash encryption enabled, without having a valid encryption key for this ESP.

Overall, this may be a lot of work. I'm not 100% sure that Arduino users need flash encryption in the first place (which is not even very useful without enabling Secure Boot).

PR #238, opened 2017-02-26
Rework WiFiClient

schrod:
Rework WiFiClient to correct error where making a copy of a WiFiClient object resulted in the socket being closed prematurely.

Added loop and select to write to handle/prevent EAGAIN errors.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -26,7 +26,6 @@
 #undef write
 #undef read
 
-#define WIFI_CLIENT_SELECT_TIMEOUT_US (100000)
 
 WiFiClientSocketHandle::~WiFiClientSocketHandle()
 {
@@ -35,12 +34,11 @@ WiFiClientSocketHandle::~WiFiClientSocketHandle()
 
 WiFiClient::WiFiClient():_connected(false),next(NULL)
 {
-    clientSocketHandle = NULL;
 }
 
 WiFiClient::WiFiClient(int fd):_connected(true),next(NULL)
 {
-    clientSocketHandle = new WiFiClientSocketHandle(fd);
+    clientSocketHandle.reset(new WiFiClientSocketHandle(fd));
 }
 
 WiFiClient::~WiFiClient()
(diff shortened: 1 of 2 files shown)


me-no-dev:
There are generally two ways to tackle issues like that. I would rather use shared pointer which will manage the state on it's own, will clean the code and you will not have to worry to ref/unref count.
Please do have ```Added loop and select to write to handle/prevent EAGAIN errors.``` in a separate PR as the change has no relation to the rest.

schrod:
Interesting, I hadn't been aware of shared pointers (I'm mostly a C developer). I agree that looks like it will be a cleaner implementation, I'll work on that change and move the changes to write to a separate PR.

libraries/WiFi/src/WiFiClient.h
```diff
@@ -20,30 +20,21 @@
 #ifndef _WIFICLIENT_H_
 #define _WIFICLIENT_H_
 
+
 #include "Arduino.h"
 #include "Client.h"
+#include <memory>
```
me-no-dev: You can move the whole class in the CPP and only forward declare it here. It's private really and users do not need to bother with it.

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -199,52 +165,21 @@ int WiFiClient::read()
 
 size_t WiFiClient::write(const uint8_t *buf, size_t size)
```
me-no-dev: You need to sync this code now with the one that I already merged :)

libraries/WiFi/src/WiFiClient.h
```diff
@@ -69,21 +87,26 @@ class WiFiClient : public Client
         return !this->operator==(rhs);
     };
 
-    int fd()
+    int fd() const
```
me-no-dev: this should go into the CPP

me-no-dev:
great work :) maybe you want to look at WiFiClientSecure?

schrod:
Yup, looks like WiFiClientSecure has the same issue. I'll see what I can do.

me-no-dev:
udp probably suffers from that same thing :D

PR #240, opened 2017-02-27
Revise WiFiClient::Write to handle EAGAIN

schrod:
The send call may return EAGAIN. This indicates a recoverable error and a retry should be attempted. The current implementation treats this as a fatal error. Further, the current implementation strips the error code, setting it to 0, which prevents the caller from handling it directly. 
This change utilizes select to verify the socket is available prior to calling send and will retry on an EAGAIN condition.


me-no-dev:
nice :)

PR #256, opened 2017-03-10
File system

me-no-dev:


libraries/FS/src/vfs_api.cpp
```diff
@@ -0,0 +1,417 @@
...
+    }
+    char * temp2 = (char *)malloc(strlen(pathTo)+strlen(_mountpoint)+1);
+    if(!temp2) {
+        free(temp1);
+        log_e("malloc failed");
+        return false;
+    }
+    sprintf(temp1,"%s%s", _mountpoint, pathFrom);
+    sprintf(temp2,"%s%s", _mountpoint, pathTo);
+    auto rc = real_rename(temp1, temp2);
```
igrr: can also use `auto rc = ::rename(temp1, temp2);`

me-no-dev: good one! Will give it a go

PR #288, opened 2017-03-27
add temporary WebServer HTTPClient libraries

0x1abin:
transplant form esp8266com


libraries/HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino
```diff
@@ -0,0 +1,72 @@
...
+    USE_SERIAL.println();
+    USE_SERIAL.println();
+
+    for(uint8_t t = 4; t > 0; t--) {
+        USE_SERIAL.printf("[SETUP] WAIT %d...\n", t);
+        USE_SERIAL.flush();
+        delay(1000);
+    }
+
+    wifiMulti.addAP("MasterHax_2.4G", "wittyercheese551");
```
copercini: You forgot to remove your WiFi credentials....

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -0,0 +1,1119 @@
...
+{
+    _transportTraits.reset(nullptr);
+    _port = 443;
+    if (httpsFingerprint.length() == 0) {
+        return false;
+    }
+    if (!beginInternal(url, "https")) {
+        return false;
+    }
+    _transportTraits = TransportTraitsPtr(new TLSTraits(httpsFingerprint));
```
copercini: WiFiClientSecure doesn't support SHA1 Fingerprint anymore, because it's become insecure few days ago... More info in http://www.pcworld.com/article/3173791/security/stop-using-sha1-it-s-now-completely-unsafe.html and https://shattered.io/static/shattered.pdf

terryjmyers: https://github.com/DuSTman31/SHA-3
has an implementation of SHA-3 512.  I compiled the project provided and it worked out of the box.  I compared the results to an online tool (https://emn178.github.io/online-tools/sha3_512.html) and they were identical.  Could we incorporate this library instead of SHA-1?

igrr: Do browsers already show SHA-3 fingerprints of website certificates?

PR #310, opened 2017-04-12
Update IDF to version 3cad00f

me-no-dev:


igrr:
That's odd, did you do git submodule update? `librtc_clk.a` and `librtc_pm.a` shouldn't be in the lib directory anymore.

me-no-dev:
they are not anymore :) check the changes. they are removed

PR #317, opened 2017-04-16
To reduce calls to micros() in loopTask(). Proposal to move micros() call into separate task.

sm-framerelay:
Guys thanks for your work, the arduino-esp32 is the rock!

Maybe I am wrong, but I suppose we can decrease the load on CPU core by removing constant call in infinity loop to micros() in loopTask() which is defined in \cores\esp32\main.cpp

void loopTask(void *pvParameters)
{
    setup();
    for(;;) {
        micros(); //update overflow << -- This one might be moved out into separate task
        loop();
    }
}

The idea is to move micros() call into separate task with high priority, which optionally might be run on second core.

Please check the possible solution:

void vMicrosOverflowTask(void *pvParameters)
{
	// In order no to miss overflow it will be enought to check micros() each half of full overflow
	static const uint32_t taskDelayMs = UINT32_MAX
		/ CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ // one full overflow occurs in uS
		/ 1000 // in millisec
		/ 2 // half of full overflow cycle
		/ portTICK_PERIOD_MS;  // RTOS const to convert in ms
	for (;;){
		micros(); //update overflow
		vTaskDelay(taskDelayMs); // ~ each 9 sec at 240Mhz
	}
}

In this case we will be sure that we won't miss micros() overflow, because standard overflow cycle at 240Mhz is 178956970.625 uS, so the half of them will be enough to check.

Also please note, using existing code in /master without the patch proposed above we have potential issue which might appear if someone try to set vTaskDelay() for more than 17 sec in main loop()

### DIFF
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -14,15 +14,30 @@ void loopTask(void *pvParameters)
 {
     setup();
     for(;;) {
-        micros(); //update overflow
         loop();
     }
+}
+
+void vMicrosOverflowTask(void *pvParameters)
+{
+	// In order no to miss overflow it will be enought to check micros() each half of full overflow
+	static const uint32_t taskDelayMs = UINT32_MAX
+		/ CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ // one full overflow occurs in uS
+		/ 1000 // in millisec
+		/ 2 // half of full overflow cycle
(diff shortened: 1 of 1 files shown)


me-no-dev:
I've tried that and it lead to issues. Also I do not want another task on that CPU that is spun all the time because it will lead to more issues with things like bitbang and others. Current implementation takes under 1us to execute and I do not think that it's a problem or clogging the CPU.

PR #364, opened 2017-05-08
feather variant

ladyada:
Hiya here is the Feather ESP32 variant details!


me-no-dev:
I was just watching your video from earlier :D

ladyada:
yay hopefully the fun one where i check out the first board, not the one where i spent 30 minutes fighting win 7 :)

ladyada:
btw im still seeing some issues with SD cards using the built in library, perhaps. but i'll get you some hardware so you can have something to test against

me-no-dev:
actually both :)

me-no-dev:
Sure thing :) though I can check probably on the hardware I have as well. Now that I have your board definitions :)
Could you maybe describe the issues a bit more?

PR #367, opened 2017-05-09
add support for widora-air board

YuanChenmang:


boards.txt
```diff
@@ -99,6 +99,47 @@ esp32thing.menu.UploadSpeed.512000.upload.speed=512000
...
+
+widora-air.serial.disableDTR=true
+widora-air.serial.disableRTS=true
+
+widora-air.build.mcu=esp32
+widora-air.build.core=esp32
+widora-air.build.variant=widora-air
+widora-air.build.board=WIDORA_AIR
+
+widora-air.build.f_cpu=400000000L
```
me-no-dev: ????? what? 400MHz :D

YuanChenmang: Should be 40M, I changed and then test. :)

YuanChenmang: Should be 240M, I changed and then test

Issue #379, opened 2017-05-17
Add setNoDelay and getNoDelay to WiFiServer class


bbx10:
setNoDelay/getNoDelay are implemented in the ESP8266 version of WiFiServer so I added them to the ESP32 version. This makes it easier to port ESP8266 code to ESP32.

```C
diff --git a/libraries/WiFi/src/WiFiServer.cpp b/libraries/WiFi/src/WiFiServer.cpp
index a10e3b1..1174697 100644
--- a/libraries/WiFi/src/WiFiServer.cpp
+++ b/libraries/WiFi/src/WiFiServer.cpp
@@ -45,8 +45,11 @@ WiFiClient WiFiServer::available(){
   int client_sock = accept(sockfd, (struct sockaddr *)&_client, (socklen_t*)&cs);
   if(client_sock >= 0){
     int val = 1;
-    if(setsockopt(client_sock, SOL_SOCKET, SO_KEEPALIVE, (char*)&val, sizeof(int)) == ESP_OK)
-      return WiFiClient(client_sock);
+    if(setsockopt(client_sock, SOL_SOCKET, SO_KEEPALIVE, (char*)&val, sizeof(int)) == ESP_OK) {
+      val = _noDelay;
+      if(setsockopt(client_sock, IPPROTO_TCP, TCP_NODELAY, (char*)&val, sizeof(int)) == ESP_OK)
+        return WiFiClient(client_sock);
+    }
   }
   return WiFiClient();
 }
@@ -69,6 +72,14 @@ void WiFiServer::begin(){
   _listening = true;
 }
 
+void WiFiServer::setNoDelay(bool nodelay) {
+    _noDelay = nodelay;
+}
+
+bool WiFiServer::getNoDelay() {
+    return _noDelay;
+}
+
 void WiFiServer::end(){
   close(sockfd);
   sockfd = -1;
diff --git a/libraries/WiFi/src/WiFiServer.h b/libraries/WiFi/src/WiFiServer.h
index b2d9996..38433a0 100644
--- a/libraries/WiFi/src/WiFiServer.h
+++ b/libraries/WiFi/src/WiFiServer.h
@@ -29,6 +29,7 @@ class WiFiServer : public Server {
     uint16_t _port;
     uint8_t _max_clients;
     bool _listening;
+    bool _noDelay = false;
 
   public:
     void listenOnLocalhost(){}
@@ -38,6 +39,8 @@ class WiFiServer : public Server {
     WiFiClient available();
[...]

PR #388, opened 2017-05-19
Add setNoDelay and getNoDelay to WiFiServer class

bbx10:
setNoDelay/getNoDelay are implemented in the ESP8266 version of WiFiServer so patch adds them to ESP32. See #379 for discussion.


me-no-dev:
this PR overlaps with the other one :) so code already comited

PR #394, opened 2017-05-22
Add WiFiServer hasClient and WiFiTelnetToSerial example

bbx10:
WiFiTelnetToSerial is also a test for hasClient().


lrmoreno007:
Very good example !!! It's just what I've been looking for a long time. It works great and with a few small touches I can now send commands to the board and receive information from it. THANKS !!!

PR #396, opened 2017-05-23
Add LED_BUILTIN pin

joeybab3:
On my dev module, the built in led was on pin 1, and this also appears to be the case for the one from adafruit as well.


me-no-dev:
pin 1 is used by Serial ;) Those LEDs show activity. Not suitable for use in sketches, because it will break the serial.

joeybab3:
wow... I had no idea, my bad

PR #404, opened 2017-05-30
Additions to WiFiClient and WiFiServer

bbx10:
Required for WebServer and DNSServer libraries. May be use by other libraries and apps.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -224,6 +224,11 @@ size_t WiFiClient::write(const uint8_t *buf, size_t size)
     return res;
 }
 
+size_t WiFiClient::write_P(PGM_P buf, size_t size)
+{
+    return write(buf, size);
+}
+
 int WiFiClient::read(uint8_t *buf, size_t size)
 {
     if(!available()) {
@@ -344,6 +349,34 @@ IPAddress WiFiClient::remoteIP() const
 uint16_t WiFiClient::remotePort() const
 {
     return remotePort(fd());
+}
+
+IPAddress WiFiClient::localIP(int fd) const
+{
+    struct sockaddr_storage addr;
+    socklen_t len = sizeof addr;
+    getsockname(fd, (struct sockaddr*)&addr, &len);
+    struct sockaddr_in *s = (struct sockaddr_in *)&addr;
+    return IPAddress((uint32_t)(s->sin_addr.s_addr));
+}
+
+uint16_t WiFiClient::localPort(int fd) const
+{
+    struct sockaddr_storage addr;
+    socklen_t len = sizeof addr;
+    getsockname(fd, (struct sockaddr*)&addr, &len);
+    struct sockaddr_in *s = (struct sockaddr_in *)&addr;
+    return ntohs(s->sin_port);
+}
+
+IPAddress WiFiClient::localIP() const
+{
+    return localIP(fd());
+}
+
+uint16_t WiFiClient::localPort() const
+{
+    return localPort(fd());
 }
 
 bool WiFiClient::operator==(const WiFiClient& rhs)
--- a/libraries/WiFi/src/WiFiServer.cpp
+++ b/libraries/WiFi/src/WiFiServer.cpp
@@ -21,6 +21,7 @@
 #include <lwip/netdb.h>
 
 #undef write
+#undef close
 
 int WiFiServer::setTimeout(uint32_t seconds){
   struct timeval tv;
@@ -103,8 +104,16 @@ bool WiFiServer::hasClient() {
 }
 
 void WiFiServer::end(){
-  close(sockfd);
+  lwip_close_r(sockfd);
   sockfd = -1;
   _listening = false;
+}
+
+void WiFiServer::close(){
+  end();
+}
+
+void WiFiServer::stop(){
+  end();
 }
 
--- a/libraries/WiFi/src/WiFiClient.h
+++ b/libraries/WiFi/src/WiFiClient.h
@@ -44,6 +44,7 @@ public:
     int connect(const char *host, uint16_t port);
     size_t write(uint8_t data);
     size_t write(const uint8_t *buf, size_t size);
+    size_t write_P(PGM_P buf, size_t size);
     int available();
     int read();
     int read(uint8_t *buf, size_t size);
@@ -84,6 +85,10 @@ public:
     IPAddress remoteIP(int fd) const;
     uint16_t remotePort() const;
     uint16_t remotePort(int fd) const;
+    IPAddress localIP() const;
+    IPAddress localIP(int fd) const;
+    uint16_t localPort() const;
+    uint16_t localPort(int fd) const;
 
     //friend class WiFiServer;
     using Print::write;
--- a/libraries/WiFi/src/WiFiServer.h
+++ b/libraries/WiFi/src/WiFiServer.h
@@ -50,6 +50,8 @@ class WiFiServer : public Server {
     using Print::write;
 
     void end();
+    void close();
+    void stop();
     operator bool(){return _listening;}
     int setTimeout(uint32_t seconds);
     void stopAll();


libraries/WiFi/src/WiFiServer.cpp
```diff
@@ -103,8 +104,16 @@ bool WiFiServer::hasClient() {
 }
 
 void WiFiServer::end(){
-  close(sockfd);
+  lwip_close_r(sockfd);
```
me-no-dev: why?

me-no-dev: oh breaks with the other close...

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -224,6 +224,11 @@ size_t WiFiClient::write(const uint8_t *buf, size_t size)
     return res;
 }
 
+size_t WiFiClient::write_P(PGM_P buf, size_t size)
+{
+    return write(buf, size);
+}
+
```
me-no-dev: this is not official Arduino API is it? We do not really have _P anything ;)

libraries/WiFi/src/WiFiServer.cpp
```diff
@@ -103,8 +104,16 @@ bool WiFiServer::hasClient() {
...
   sockfd = -1;
   _listening = false;
 }
 
+void WiFiServer::close(){
```
me-no-dev: I think we should go with one or the other ;) What's the official API for this?

libraries/WiFi/src/WiFiClient.cpp
bbx10: ESP8266WebServer has *_P methods so I added support for them in the ESP32 version.

```C
void WebServer::send_P(int code, PGM_P content_type, PGM_P content) {
    size_t contentLength = 0;

    if (content != NULL) {
        contentLength = strlen_P(content);
    }

    String header;
    char type[64];
    memccpy_P((void*)type, (PGM_VOID_P)content_type, 0, sizeof(type));
    _prepareHeader(header, code, (const char* )type, contentLength);
    _currentClient.write(header.c_str(), header.length());
    sendContent_P(content);
}

void WebServer::send_P(int code, PGM_P content_type, PGM_P content, size_t contentLength) {
    String header;
    char type[64];
    memccpy_P((void*)type, (PGM_VOID_P)content_type, 0, sizeof(type));
    _prepareHeader(header, code, (const char* )type, contentLength);
    sendContent(header);
    sendContent_P(content, contentLength);
}

void WebServer::send(int code, char* content_type, const String& content) {
  send(code, (const char*)content_type, content);
}

void WebServer::send(int code, const String& content_type, const String& content) {
  send(code, (const char*)content_type.c_str(), content);
}

void WebServer::sendContent(const String& content) {
  const char * footer = "\r\n";
  size_t len = content.length();
  if(_chunked) {
    char * chunkSize = (char *)malloc(11);
    if(chunkSize){
      sprintf(chunkSize, "%x%s", len, footer);
      _currentClient.write(chunkSize, strlen(chunkSize));
      free(chunkSize);
    }
  }
  _currentClient.write(content.c_str(), len);
  if(_chunked){
    _currentClient.write(footer, 2);
  }
}

void WebServer::sendContent_P(PGM_P content) {
  sendContent_P(content, strlen_P(content));
}

void WebServer::sendContent_P(PGM_P content, size_t size) {
  const char * footer = "\r\n";
  if(_chunked) {
    char * chunkSize = (char *)malloc(11);
    if(chunkSize){
      sprintf(chunkSize, "%x%s", size, footer);
      _currentClient.write(chunkSize, strlen(chunkSize));
      free(chunkSize);
    }
  }
#ifdef ESP8266
  _currentClient.write_P(content, size);
#else
  // TODO add ESP32 WiFiClient::write_P
  _currentClient.write(content, size);
#endif
  if(_chunked){
    _currentClient.write(footer, 2);
  }
}
```

me-no-dev: _P methods imply PROGMEM an we do not have PROGMEM on esp32

libraries/WiFi/src/WiFiServer.cpp
bbx10: ESP8266WiFiServer has close() and stop(). ESP32 WiFiServer has end(). ESP8266WebServer calls close. I chose to put all three for maximum backwards compatibility.
```C
void WebServer::close() {
#ifdef ESP8266
  _server.close();
#else
  // TODO add ESP32 WiFiServer::close()
  _server.end();
#endif
}
```

me-no-dev: ESP32 should actually always use stop() as that is the official Arduino API. close and end should e deprecated.

libraries/WiFi/src/WiFiClient.cpp
copercini: PGM_P is a pointer to a string in program space. It's come from here: https://github.com/esp8266/Arduino/pull/753

bbx10: @copercini Compare pgmspace.h on ESP32 vs. ESP8266. There is no support for PROGMEM on ESP32 so pgmspace.h neuters the _P functions to their non _P equivalents.

But to avoid breaking lots of apps and libraries is would be good to translate _P functions to something that works on ESP32. Instead of making this change, libs and apps would have to do this.

#ifdef ESP8266
  _currentClient.write_P(content, size);
#else
  _currentClient.write(content, size);
#endif

If there are many occurrences, perhaps `#define write_P write` would do the job. But it would be less work and fewer breakage issues if write_P() is added even though it does the same thing as write().

me-no-dev: I can not agree to bloat the classes unnecessary just for the sake of compatibility with everything. We have to add all of the _P methods then to Print, Stream and whatever else has them on ESP8266... bad idea!

igrr: How about we create a utility mixin class, EnableProgmemMethods. This class would implement _P versions based on non-_P versions. WiFiClient would be derived from EnableProgmemMethods<WiFiClient>.

This way:
- code of WiFiClient is kept concise
- link time dead code elimination guarantees that unused methods will not end up in the final binary.

me-no-dev: I was searching for exactly that kind of implementation!

libraries/WiFi/src/WiFiServer.cpp
bbx10: OK, I can remove close and end.

libraries/WiFi/src/WiFiClient.cpp
igrr: Ok, scratch that — we wouldn't be able to reuse this on the 8266, as that WiFiClient uses different implementation for _P and non-_P. So it's​ just two methods here, for WiFiClient and WebServer. I'd say just add them, as they will be eliminated if not used.

me-no-dev: And then we'll have this requested on any other place that implements _P. We either do it everywhere or not. IMHO we should not.
I can agree to mark them deprecated and remove at later point.

bbx10: Instead of adding write_P perhaps adding macro defs to cores/esp32/pgmspace.h will work.
I looked around for other _P functions exported by ESP8266 core classes but I am not sure I got them all.

pgmspace.h
```C
#define write_P       write       /* WiFiClient::write_P, SdFile::write_P */
#define printf_P      printf      /* Print::printf_P */
#define writeln_P     writeln     /* SdFile::writeln_P */
```
This exactly equivalent to the ifdef but without developers having to worry about the issue. And without the need to add _P functions to ESP32 classes.

igrr: > I can agree to mark them deprecated and remove at later point.

What approach would you suggest library authors take when it comes to writing code which supports both the ESP8266 and the ESP32?

me-no-dev: What would happen if we separate code like it's with WiFi? will all of the source files be included even if not used?

igrr: At link time, only the functions which were referenced will be left in the program, if -ffunction-section and -gc-sections options are used.

bbx10: Instead of adding write_P, how about adding `#define write_P write` to pgmspace.h?

me-no-dev: ```#define write_P write``` will not lead to proper client.write_P. @igrr correct me if I'm wrong, but that is what I believe the result will be

bestpika:
Can't compile...
```cpp
#include <WiFi.h>

void setup() {}
void loop() {}
```
```
In file included from C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFi.h:37:0,

                 from C:\Users\...\AppData\Local\Temp\arduino_modified_sketch_637679\sketch_jun14c.ino:1:

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:93:15: error: 'IPAddress WiFiClient::localIP() const' cannot be overloaded

     IPAddress localIP() const;

               ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:88:15: error: with 'IPAddress WiFiClient::localIP() const'

     IPAddress localIP() const;

               ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:94:15: error: 'IPAddress WiFiClient::localIP(int) const' cannot be overloaded

     IPAddress localIP(int fd) const;

               ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:89:15: error: with 'IPAddress WiFiClient::localIP(int) const'

     IPAddress localIP(int fd) const;

               ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:95:14: error: 'uint16_t WiFiClient::localPort() const' cannot be overloaded

     uint16_t localPort() const;

              ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:90:14: error: with 'uint16_t WiFiClient::localPort() const'

     uint16_t localPort() const;

              ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:96:14: error: 'uint16_t WiFiClient::localPort(int) const' cannot be overloaded

     uint16_t localPort(int fd) const;

              ^

C:\Users\...\Documents\Arduino\hardware\espressif\esp32\libraries\WiFi\src/WiFiClient.h:91:14: error: with 'uint16_t WiFiClient::localPort(int) const'

     uint16_t localPort(int fd) const;

              ^
```

priyankachepuri09:
Same issue as @bestpika unable to compile.

me-no-dev:
fixed in the latest commit ;)

PR #415, opened 2017-06-03
Add DOIT ESP32 DEVKIT V1 board

bovi:
Add development board from DOIT (doit.am) with ESP32. Similar to the normal Dev Module but with different LED pin.


me-no-dev:
I'm thinking I should maybe make the LED as a menu option rather than adding all those boards...
Will accept the PR though.

PR #432, opened 2017-06-11
set max clients from code

Stefano-Orsolini:
unlock the ability to set from code the maximum number of clients allowed to connect (in my case i must allow only one client at a time)


me-no-dev:
Great! Thanks

PR #445, opened 2017-06-14
added OTA Update S3 example with a sample bin

arvindr21:
OTA Update example with AWS S3. Sample bin added along with sketch for quick validation.


libraries/Update/examples/AWS_S3_OTA_Update/AWS_S3_OTA_Update.ino
```diff
@@ -0,0 +1,280 @@
...
+   Step 2 : Upload it to your Amazon S3 account, in a bucket of your choice
+   Step 3 : Once uploaded, inside S3, select the bin file >> More (button on top of the file list) >> Make Public
+   Step 4 : You S3 URL => http://bucket-name.s3.ap-south-1.amazonaws.com/sketch-name.ino.bin
+   Step 5 : Build the above URL from and fire it either in your browser or curl it `curl -I -v http://bucket-name.ap-south-1.amazonaws.com/sketch-name.ino.bin` to validate the same
+   Step 6:  Plug in your SSID, Password, S3 Host and Bin file below
+
+   Build & upload
+   Step 1 : Preferences > Show verbose output during `compilation`
+   Step 2 : ctrp + r (Verify sketch or build sketch)
+   Step 3 : In the verbose log, you will find a path t0 `sketch-name.iono.bin` on your machine. Example from Apple OSX:
```
me-no-dev: You have a menu ```Sketch -> Export compiled binary``` which puts the bin in the sketch folder. Maybe use that?

libraries/Update/examples/AWS_S3_OTA_Update/AWS_S3_OTA_Update.ino
```diff
@@ -0,0 +1,280 @@
...
+
+  // Execute OTA Update
+  execOTA();
+}
+
+void loop() {
+  // chill
+}
+
+void execOTA() {
```
me-no-dev: I will advise towards moving your functions above setup() and loop(). Reason is that while Arduino IDE and some other tools that use Arduino declare all your functions before compiling, this is not a standard, so if you grab this code and paste it into "make" based build, it will fail with a bunch of errors.

me-no-dev: In this order :) 
```cpp
String getHeaderValue(String header, String headerName);
void execOTA();
void setup();
void loop();
```

libraries/Update/examples/AWS_S3_OTA_Update/AWS_S3_OTA_Update.ino
arvindr21: Thanks. Did not know about that option 👍 . Fixing it.

libraries/Update/examples/AWS_S3_OTA_Update/AWS_S3_OTA_Update.ino
arvindr21: Will Do.

arvindr21:
Thanks.

PR #460, opened 2017-06-21
add python3 support to gen_esp32part.py

briancappello:
Tested with Python 2.7 and Python 3.6; should work on 2.6+ (<2.6 does not support the b'' notation)

### DIFF
--- a/tools/gen_esp32part.py
+++ b/tools/gen_esp32part.py
@@ -12,6 +12,15 @@ import re
 import struct
 import sys
 
+PY2 = sys.version_info[0] == 2
+
+if PY2:
+    to_bytes = lambda string, encoding: string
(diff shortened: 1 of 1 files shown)


me-no-dev:
great! let's give this a go :)

me-no-dev:
Could I bother you to please create this PR in https://github.com/espressif/esp-idf ?
That way everyone will get that better code :)

briancappello:
Hmm the `gen_esp32part.py` in that repo already works with Python 3! So I guess this PR is unnecessary; just this repo needs to be synced with "upstream"?

me-no-dev:
oh... that must have changed recently :) it's due for resync anyway ;)

PR #461, opened 2017-06-22
Two Stop Bit Workaround

claudeheintz:
This incorporates a fix from the latest IDF for the two stop bit issue.
See:
https://github.com/espressif/esp-idf/blob/master/components/driver/uart.c  lines 118-127

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -195,6 +195,13 @@ uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rx
     uartSetBaudRate(uart, baudrate);
     UART_MUTEX_LOCK();
     uart->dev->conf0.val = config;
+    #define TWO_STOP_BITS_CONF 0x3
+	#define ONE_STOP_BITS_CONF 0x1
+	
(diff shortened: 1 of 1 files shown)


me-no-dev:
since this would effectively change all 2 stop bits config, maybe it would be better to adjust their definitions in the header? What do you think?

claudeheintz:
The issue is a bit more complicated than changing the definition in the header.  Apparently, there is a bug in the chip so that it does not correctly handle the two stop bits setting in the conf0 register.  In my tests, I could not get serial input with two stop bits to work correctly using just the conf0 register.  The workaround in the latest IDF looks for two stop bits in the uart_set_stop_bits function.  When a request for two stop bits is detected, it changes the request to a single stop bit for the conf0 register and sets a bit requesting a stop bit delay in the rs485_conf register.  This functionally makes two stop bits work.  But, it involves two registers which makes it difficult to solve in the header.

BTW, the bits in the rs485_conf are for expanding the stop bits settings and are not specifically related to the RS485 function.  They just happen to have been placed in this register.  In some versions of the headers they are more descriptively named than dl1 and dl0 which makes there use more clear.

me-no-dev:
Ah got it!

PR #466, opened 2017-06-25
Change disconnect function

sticilface:
Current implementation does not disconnect the WiFi. Use of `esp_wifi_disconnect()` does.

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.cpp
+++ b/libraries/WiFi/src/WiFiSTA.cpp
@@ -255,6 +255,13 @@ bool WiFiSTAClass::reconnect()
 bool WiFiSTAClass::disconnect(bool wifioff)
 {
     bool ret;
+    wifi_config_t conf;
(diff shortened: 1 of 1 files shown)


me-no-dev:
have you tried calling that function first? 
```cpp
void setup(){
  //....
  WiFi.disconnect(true);
  WiFi.connect("ssid", "pass");
  //....
}
```
It should wipe the settings as well

sticilface:
I was testing the disconnect directly.  Ie configure the ESP to join an AP, then 10 seconds later call WiFi.disconnect().  It returned true, but nothing changed and the ESP still had an IP and was connected.

me-no-dev:
you removed some lines that had a function there ;)

sticilface:
Ah, is disconnect meant to wipe the settings too?

me-no-dev:
what's up with this?

me-no-dev:
great :)

PR #467, opened 2017-06-27
ResetReason: Add an additional method and a test case

Raienryu97:
Method:
--------
Added a verbose print method for ease.

Test Case:
-----------
Putting ESP32 to sleep will give a different reason on wake than the first
time power up.


me-no-dev:
Nice! :)

Raienryu97:
Thanks =D

PR #483, opened 2017-07-05
Disable SimpleBLE when BT not enabled

APokorny:
SimpleBLE would otherwise require a bt header file to be reachable in the include paths.
This forces people to enable bt even when not needed. Which is a pity because that also disables a memory region for BT exclusive use.

Signed-off-by: Andreas Pokorny <andreas.pokorny@siemens.com>

### DIFF
--- a/component.mk
+++ b/component.mk
@@ -1,4 +1,25 @@
-ARDUINO_CORE_LIBS := $(patsubst $(COMPONENT_PATH)/%,%,$(wildcard $(COMPONENT_PATH)/libraries/*/src))
+ARDUINO_CORE_LIBS := libraries/ArduinoOTA/src  \
+    libraries/ESP32/src \
+    libraries/ESPmDNS/src \
+    libraries/FS/src  \
+    libraries/HTTPClient/src \
+    libraries/Preferences/src \
+    libraries/SD/src \
+    libraries/SD_MMC/src \
+    libraries/SPI/src \
+    libraries/Update/src \
+    libraries/Wire/src
+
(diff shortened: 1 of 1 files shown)


copercini:
This is good but will cause issues when install external libraries like ESPWebServer, SPIFFS , Ethernet on ESP32 folder

APokorny:
Issues like?

APokorny:
Ah ok. So you are relying on the wildcard in the components.mk of arduino. So you are creating additional checkouts inside a project? That does not sound like a proper solution. 
I thought there is an EXTRA_COMPONENTS_DIR variable inside the build system to add stuff to your project

me-no-dev:
There must be a better solution than listing all the libs manually ;) seems not right to edit this file each time a lib is added.

APokorny:
Alternatively one could maintain something like that (incomplete example):
```
ARDUINO_CORE_LIBS := $(filter-out libraries/WiFiClientSecure/src,$(filter-out libraries/SimpleBLE/src,$(patsubst $(COMPONENT_PATH)/%,%,$(wildcard $(COMPONENT_PATH)/libraries/*/src))))
```

APokorny:
Does that work for you now?

me-no-dev:
OK, here is wat I will suggest: use ifdef in the source files, so source is not compiled.

me-no-dev:
implemented here: https://github.com/espressif/arduino-esp32/commit/c66a020e8067f657e730d018700e7de5b23175d4

PR #502, opened 2017-07-11
Spi pattern fix

mcgillan:
Fixes problem where not enough data is sent when sending a repeating pattern via SPI that doesn't evenly divide into the hardware FIFO size (64 bytes).

For example, I was sending a 3-byte pattern (RGB data) for a 320 x 480 display, which fits 21 times into the hardware FIFO (3 * 21 = 63 bytes), and by the end of the routine 1/64th of the data (and therefore the image) was missing.  This happened because for every 63 bytes sent, this routine decremented the byte count by 64.

### DIFF
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -1,9 +1,9 @@
-/*
+/* 
  SPI.cpp - SPI library for esp8266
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks :)

PR #503, opened 2017-07-11
Handle partial socket send

bbx10:
send() can return a value > 0 but less than size indicating it was able to accept some of the data in buffer. The caller must try again after updating the buffer pointer and size remaining.

https://github.com/bbx10/WebServer_tng/issues/1


everslick:
I was originally reporting this issue and can confirm, that the PR fixes it. Large pages only have been served partially. Now it works as expected.

me-no-dev:
Good :) Merging

PR #506, opened 2017-07-12
Add ThaiEasyElec's ESPino32

narongrat:
Add new ESPino32 based on ESP-WROOM-32 (ESP32) module from Espressif.

This board is new version from previous ESPino based-on EPS-WROOM-02 (ESP8266).


me-no-dev:
Great :)

PR #509, opened 2017-07-13
Bugfix: Put every SPI access between begin / end Transaction

APokorny:
The two functions sdcard_mount and ff_sd_ioctl had missed the begin/endTransaction calls. 
The change intrdocues a local RAII helper to simplify maintaining begin/end scopes

Signed-off-by: Andreas Pokorny <andreas.pokorny@siemens.com>


APokorny:
Any thoughts?

me-no-dev:
Nice! Really clever approach :)

PR #517, opened 2017-07-16
Ethernet

fesch:
Applied patch to libtcpip_adapter to allow ETH to be used with a static configuration. (Pull request in ESP-IDF has been made but not integrated yet.)

Added a first version of an Ethernet library for LAN8720 based adapters as well a slightly modified copy of the ESP6288 WebServer (while waiting for a common repository for ESP6288 & ESP32).


fesch:
Argh ... I have to lean to use the command line of git ... :-(

PR #518, opened 2017-07-16
Ethernet

fesch:
Applied patch to libtcpip_adapter to allow ETH to be used with a static configuration. (Pull request in ESP-IDF has been made but not integrated yet.)

Added a first version of an Ethernet library for LAN8720 based adapters as well a slightly modified copy of the ESP8266 WebServer (while waiting for a common repository for ESP8266 & ESP32).


me-no-dev:
Why mixing those two libs? Also too quick on the ethernet ;) I have most boards out there with the different chips and will actually need to implement them for all ;) I can not merge the web server only to remove it later ;) What boards did you test ethernet with?

fesch:
- I originally headed for a WebServer but ended up with the need to first code the Ethernet library. Being asked by other users, I build both ...
- Of course Ethernet should be extended to fit other chips too, but as I only have the one with the LAN8720, I thought this would be a good beginning and wanted to share my code (as I was not able to find something working out-of-the-box on the net ...).
- I own an [ESP32-EVB](https://www.olimex.com/Products/IoT/ESP32-EVB/open-source-hardware) from Olimex.

pipi61:
Hi!
on olimex esp32-evb schematic is Lan8710,
the external LAN8720 is connected??

me-no-dev:
I have both olimex boards (big thanks to Olimex) and I also have another unreleased yet board based on the TK110. I will prep support for all and I'm thinking of not using ```Ethernet``` for that or maybe use it through defines so it would not interfere with the normal Ethernet classes for other SPI based LAN chips. I have code running for my 110 board and need to test the Olimex implementations and prepare board definitions for them. All under way :) IPv6 is also about to be merged for ETH

pipi61:
Hi!
This work with (ebay) waveshare LAN8720 module.
In ethernet.h  modify #define ETH_PHY_ADDR  PHY1    //waveshare
Connection diagram: https://sautter.com/blog/ethernet-on-esp32-using-lan8720/
(gpio17 is not connect)
Thank You to Fesch!

Issue #525, opened 2017-07-18
Porting of Esp8266 EEPROM library


pbecchi:
For legacy compatibility it will be very useful to have the equivalent of Esp8266 EEPROM.h library available.
Idf Nvs provide a very powerful tool to write and read data's from Flash memory but do not replace easily Arduino or Esp8266 code written for EEPROM .
Looking to EEPROM.cpp I have seen that , there , all flash related calls are a flash_spi_write and a flash_spi_read call that exist identical in Esp32 IDF. From my ignorant standpoint it should be possible ,having a flash sector address available for that , compile EEPROM library with no modification!
Is this feasible? Any flash sector that could be used without impacting/overlapping other flash utilization?

everslick:
I hacked something up for myself to ease my own migration pains a little..

HEADER
```
#ifndef _EEPROM_H_
#define _EEPROM_H_

#include <stddef.h>
#include <stdint.h>
#include <string.h>

#include <Preferences.h>

#define MAX_EEPROM_SIZE 2048

class EEPROM {

public:

  EEPROM(void);

  void begin(size_t size = MAX_EEPROM_SIZE);
  uint8_t read(int address);
  void write(int address, uint8_t val);
  bool commit();
  void end();

  uint8_t * getDataPtr();

  template<typename T>
  T &get(int address, T &t) {
    if (address < 0 || address + sizeof(T) > _size)
      return t;

    memcpy((uint8_t*) &t, _data + address, sizeof(T));
    return t;
  }

  template<typename T>
  const T &put(int address, const T &t) {
    if (address < 0 || address + sizeof(T) > _size)
      return t;

    memcpy(_data + address, (const uint8_t*) &t, sizeof(T));
    _dirty = true;
    return t;
  }

protected:

  uint8_t *_data;
  size_t   _size;
  bool     _dirty;

private:

  Preferences preferences;

};

#endif
```

IMPLEMENTATION
```
#include "Arduino.h"

#include "eeprom.h"

EEPROM::EEPROM(void) {
  _data  = 0;
  _size  = 0;
  _dirty = false;

  preferences.begin("EEPROM", false);
}

void EEPROM::begin(size_t size) {
  if (size <= 0) return;

  if (size > MAX_EEPROM_SIZE) size = MAX_EEPROM_SIZE;

  if (_data) {
    delete[] _data;
  }

  _data = new uint8_t[size];
  _size = size;

  // read data from flash
[...]

PR #535, opened 2017-07-23
EEPROM_library_ported_Partition_table_updated

pbecchi:
This is the outcome of issue 525.


everslick:
I'm about to test your implementation today. Thanks for providing a clean implementation for the EEPROM emulation. At first glance I noticed, that your PR contains build files like dependency and object files. Those should probably get removed. But maybe the maintainers can cherry pick files? I dunno.

Also 'libraries/EEPROM/library.properties' seems to contain outdated information.

pbecchi:
I have deleted all unnecessary files and folders from example directory and I have somehow  corrected the properties file!

everslick:
Tested. Works as advertised. :-D Thanks a lot!

libraries/EEPROM/library.properties
```diff
@@ -0,0 +1,9 @@
+name=EEPROM
+version=1.0
+author=Ivan Grokhotkov
+maintainer=Ivan Grokhotkov <ivan@esp8266.com>
+sentence=Enables reading and writing data to the permanent FLASH storage, up to 4kb.
+paragraph=
+category=Data Storage
+url=http://arduino.cc/en/Reference/EEPROM
+architectures=esp8266
```
Adminius: architectures=esp32

?

libraries/EEPROM/EEPROM.cpp
```diff
@@ -0,0 +1,155 @@
...
+bool EEPROMClass::commit() {
+  bool ret = false;
+  if (!_size)
+    return false;
+  if (!_dirty)
+    return true;
+  if (!_data)
+    return false;
+
+  noInterrupts();
```
igrr: Not needed on the ESP32

libraries/EEPROM/EEPROM.cpp
```diff
@@ -0,0 +1,155 @@
...
+	  {
+		  ESP_LOGE(TAG, "error in Write");
+	  }
+	  else
+	  {
+		  _dirty = false;
+		  ret = true;
+	  }
+  }
+  interrupts();
```
igrr: likewise

libraries/EEPROM/EEPROM.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+  noInterrupts();
+
+  if (esp_partition_erase_range(_mypart, 0, 4096) != ESP_OK)
+  {
+	  ESP_LOGE(TAG, "partition erase err.");
+  }
+  else
+  {
+	  if (esp_partition_write(_mypart, 0, (void *)_data, _size) == ESP_ERR_INVALID_SIZE)
```
igrr: Indentation does not match the rest of the file

libraries/EEPROM/examples/eeprom_write/eeprom_write.ino
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#include "EEPROM.h"
+
+// the current address in the EEPROM (i.e. which byte
+// we're going to write to next)
+int addr = 0;
+#define N512 64
+void setup()
+{
```
igrr: Please use "Format Sketch" feature in Arduino IDE when preparing example sketches.

libraries/EEPROM/examples/eeprom_write/eeprom_write.ino
```diff
@@ -0,0 +1,63 @@
...
+ * These values will stay in the EEPROM when the board is
+ * turned off and may be retrieved later by another sketch.
+ */
+
+#include "EEPROM.h"
+
+// the current address in the EEPROM (i.e. which byte
+// we're going to write to next)
+int addr = 0;
+#define N512 64
```
igrr: Giving this define some descriptive name would be nice (e.g. `EEPROM_SIZE`?)

libraries/EEPROM/EEPROM.cpp
```diff
@@ -0,0 +1,155 @@
...
+  if (!_size)
+    return false;
+  if (!_dirty)
+    return true;
+  if (!_data)
+    return false;
+
+  noInterrupts();
+
+  if (esp_partition_erase_range(_mypart, 0, 4096) != ESP_OK)
```
igrr: suggest using `SPI_FLASH_SEC_SIZE` for the constant instead

tools/partitions/default.csv
```diff
@@ -3,4 +3,5 @@ nvs,      data, nvs,     0x9000,  0x5000,
 otadata,  data, ota,     0xe000,  0x2000,
 app0,     app,  ota_0,   0x10000, 0x140000,
 app1,     app,  ota_1,   0x150000,0x140000,
-spiffs,   data, spiffs,  0x290000,0x170000,
+eeprom,   data, 0x99,    0x290000,0x1000,
+spiffs,   data, spiffs,  0x291000,0x169000
```
igrr: It may be a better idea to shrink app partition instead. Otherwise SPIFFS partition will need to be re-uploaded after this change.

libraries/EEPROM/EEPROM.cpp
pbecchi: You mean interrupts doesn't need to be stopped? I will take out the relevant lines.

libraries/EEPROM/EEPROM.cpp
pbecchi: Ok will do

libraries/EEPROM/examples/eeprom_write/eeprom_write.ino
pbecchi: Ok

libraries/EEPROM/examples/eeprom_write/eeprom_write.ino
pbecchi: I am not using Arduino Ide I will check it anyhow.

tools/partitions/default.csv
pbecchi: Initially I had the EEprom partition after the Nvs one shrinking Nvs partition 4kbytes.
Then I change it for me-no-dev comment on #525

me-no-dev: Since spiffs is not running yet and the app partitions are purposly enlarged to fit BT and WIFI it is better to do it like that ;)

me-no-dev: Otherwise you are correct

lonerzzz: also giving the 0x99 SubType value a name would be helpful

pbecchi:
I have corrected the example eeprom_write.ino

tools/partitions/default.csv
pbecchi: What shall I do?

copercini: As @igrr say, it will cause some problems in the [Unofficial SPIFFS library](https://github.com/copercini/arduino-esp32-SPIFFS)

me-no-dev: And?

me-no-dev: Else we should cause problems with bigger sketches?

pbecchi:
Why not on upper 4k of nvs partition? So that everybody is happy.

me-no-dev:
Less NVS?

pbecchi:
4k less!

pbecchi:
Is this acceptable?

copercini:
SPIFFS library is quite new and still have few downloads, so this will not be a bigggg problem =)

igrr:
It is not advised to change the size of NVS partition, as part of its data will be lost, leaving WiFi configuration in a potentially semi-defined state. I would say that shrinking sketch space by 4k would be acceptable, but in the end it is @me-no-dev's call. If SPIFFS library is not widely deployed yet, shrinking SPIFFS is also an option.

libraries/EEPROM/EEPROM.cpp
igrr: There's no need to stop the interrupts. Besides, `noInterrupts` and `interrupts` functions are no-op in the current version.

pbecchi:
All modification requested has been done!
Regarding EEprom partition I didn't make any change , you can change it as you like.
It is now overlapping  the SPIFFS partition.

lonerzzz:
@copercini, @igrr, what prevents this branch from being merged at this point and do we have a core group responsible for doing such merges?

copercini:
@lonerzzz I want to know too, because I also have some waiting PRs https://github.com/espressif/arduino-esp32/pull/504

libraries/EEPROM/EEPROM.cpp
igrr: This comment is still valid

igrr:
@lonerzzz There is one change which still needs to be done (remove noInterrupts/interrupts) and then it is up to @me-no-dev to approve and merge this.

pbecchi:
Last correction done.

libraries/EEPROM/EEPROM.cpp
```diff
@@ -0,0 +1,150 @@
...
+    return false;
+  if (!_dirty)
+    return true;
+  if (!_data)
+    return false;
+
+
+  if (esp_partition_erase_range(_mypart, 0, SPI_FLASH_SEC_SIZE) != ESP_OK)
+  {	  
+     ESP_LOGE(TAG, "partition erase err.");
```
me-no-dev: please use Arduino's log facility :) ```log_e("partition erase err.");

pbecchi:
corrected!

me-no-dev:
Done! :)

pbecchi:
Thanks!

PR #583, opened 2017-08-16
Increase BT task stack size

copercini:
For the correct work of some BLE examples


tools/sdk/include/config/sdkconfig.h
```diff
@@ -64,7 +64,7 @@
 #define CONFIG_TASK_WDT_TIMEOUT_S 5
 #define CONFIG_INT_WDT_TIMEOUT_MS 300
 #define CONFIG_ESPTOOLPY_FLASHMODE "dout"
-#define CONFIG_BTC_TASK_STACK_SIZE 2048
```
me-no-dev: this is in sdkconfig, so it will not work that way :) need to be done at build time ;) I'll add the change

me-no-dev:
wil be 8K next IDF update ;)

copercini:
Nice! thx!

PR #585, opened 2017-08-18
Update README.md with install instructions for openSUSE

joba1:


joba1:
Tested on Leap 42.2 right now. Works.

joba1:
Just discovered, this works only, if python version 3 is the default python installation on the system.
If you still use python 2, other packages need to be installed:
sudo zypper install git python-pip python-pyserial
Simply install them as well? Check current default version? Probably it is the same problem on the other distros...

me-no-dev:
How about this:
```bash
if [ `python --version 2>&1 | grep '2.7' | wc -l` = "1" ]; then \
sudo zypper install git python-pip python-pyserial; \
else \
sudo zypper install git python3-pip python3-pyserial; \
fi && \
sudo usermod -a -G dialout $USER && \
mkdir -p ~/Arduino/hardware/espressif && \
cd ~/Arduino/hardware/espressif && \
git clone https://github.com/espressif/arduino-esp32.git esp32 && \
cd esp32/tools/ && \
python get.py
```

joba1:
looks good to me

2017-08-18 14:14 GMT+02:00 Me No Dev <notifications@github.com>:

me-no-dev:
OK :) please add it and I will merge

me-no-dev:
@joba1 it's been a while :) will you add my workaround for python3 to the PR so I can merge it?

PR #590, opened 2017-08-22
Change to retry setting up I2C command rather than waiting indefinitely

lonerzzz:
With a large I2C bus, I encountered a situation where the following line in esp32-hal-i2c.c, method i2cInitFix hung indefinitely.

`
    while(!i2c->dev->command[2].done);
`

Adding the included changes resolved the problem in my situation, by allowing a retry.

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -445,7 +445,8 @@ void i2cInitFix(i2c_t * i2c){
     i2c->dev->fifo_data.data = 0;
     i2cSetCmd(i2c, 1, I2C_CMD_WRITE, 1, false, false, false);
     i2cSetCmd(i2c, 2, I2C_CMD_STOP, 0, false, false, false);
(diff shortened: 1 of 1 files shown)


me-no-dev:
Can we please have this tested with a few devices? What comes to mind is OLEDs for the big transfer size, EEPROM for holding the SDA while operation is executed and some more general sensors.

lonerzzz:
I have tested with MCU to MCU communication so far. Will be testing to sensors and with 1K transfers shortly.

everslick:
I tested this patch and it breaks EEPROM communication for me.

unfortunately i cannot say more about the specific reason, at this point, other then that.

i tested it with one of these cheep RTC / EEPROM modules.

me-no-dev:
I think it should maybe use some sort of timeout ;) have to see what is the max time that regular eeproms can keep the line LOW though

lonerzzz:
I have a fix that has proven more stable for handling lockup so I am closing this issue in favor of my newer submission.

PR #601, opened 2017-08-31
Missing xcrun fix.

fluentart:
Had to run "xcode-select --install" before your script worked.

More info here: https://stackoverflow.com/questions/32893412/command-line-tools-not-working-os-x-el-capitan-macos-sierra


me-no-dev:
thanks :)

PR #605, opened 2017-09-03
Fix support for some arduino libraries which don't put source files under a src directory

imxieyi:
There are many libraries which don't put their source files under *src* directory.
For example, nearly all sensor libraries from Adafruit don't have a *src* directory.
In order to compile with these libraries, we should manually create a *src* directory and move all source files into it, which is really annoying.
If the library also includes source files under sub directories, it would result in linker errors.
**However, only changing a line in *component.mk* can solve all these issues.**


me-no-dev:
I guess you are using the core's libraries folder to store the libs for your sketches :) you could just add another component and do it there ;) Anyway :) I will merge this! Thanks

PR #606, opened 2017-09-03
allow configTime() to handle timezone and dst begin/end info

MarkyAD:
This allows configTime() to use timezone info in the same manner as esp-idf. configTime() can now be called with a character array describing the timezone and daylight saving time begin and end.

configTime("EST5EDT4,M3.2.0/02:00:00,M11.1.0/02:00:00", "my.ntp.server", NULL, NULL);


me-no-dev:
we need to keep the old version for compatibility ;) so let's call this method something else :)
maybe ```configTzTime```

MarkyAD:
ok, closing this one.

me-no-dev:
why? :) you could have just edited it and pushed the update :)

MarkyAD:
lol sorry, anyway it is there.

PR #608, opened 2017-09-04
add configTzTime() to setup sntp using TZ environment variable

MarkyAD:
leave original configTime() intact, add ability to setup sntp using only TZ environment variable and ntp server.


me-no-dev:
Done :) thanks

PR #621, opened 2017-09-11
add Microduino-esp32

radiumray:
add boards.txt and pins_arduino.h of Microduino-ESP32
https://wiki.microduino.cn/
qq 20170911153736


me-no-dev:
Thanks :)

PR #628, opened 2017-09-12
Add example to update ESP32 via SD card

Curclamas:
This example demonstrates how the Update library can also be used for offline uploads via SD card.  This provides an update path where Internet is not available or OTA won't work and no serial programmer is at hand.


libraries/Update/examples/SD_Update/SD_Update.ino
```diff
@@ -0,0 +1,111 @@
...
+
+      updateBin.close();
+
+      //remove any previous leftovers
+      if (fs.exists("/update_old.bin")) {
+         fs.remove("/update_old.bin");
+      }
+      
+      //rename the file to prevent an infinite loop of updates
+      fs.rename("/update.bin", "/update_old.bin");
```
me-no-dev: why not just delete it?

me-no-dev:
Please remove the bin from the example ;) I have left a couple other comments too

Curclamas:
@me-no-dev thanks for the comments. I can surely remove the .bin from the repo and just delete the .bin from the SD card. Just two cents why I did it that way:
I've added the binary to the repo to provide an easy way to test the app (like in the AWS Update example). The renaming of the binary happens because it gives an Indication of success while not "harshly" deleting it. But you are right, just deleting it would make the code leaner and easier to read.

me-no-dev:
every implementation that I have seen deletes the bin on success and some leave a text file with log.

Curclamas:
For the sake of simplicity I modified it to just delete the file. What about the included binary in the repo?

krzychb:
> What about the included binary in the repo?

Bin is h/w specific and my not match e.g. the XTAL frequency of user's board.

Curclamas:
@krzychb that is actually a good point, I will remove it and adjust the file accordingly

Curclamas:
@me-no-dev @krzychb I've removed it, does it look fine now? :)

me-no-dev:
@Curclamas see I have another comment about checking if "/update.bin" is actually a file. New API will return "true" on fs.open if the path is a directory as well. So it needs to check that it is a file.

me-no-dev:
@Curclamas what's going on? :) are you going to fix this issue here?
Please note:

on ESP32 you need to make sure that it's a file
```cpp
void updateFromFS(fs::FS &fs) {
   File updateBin = fs.open("/update.bin");
   if (updateBin) {
     if(updateBin.isDirectory()){
       Serial.println("Update is not a file");
       updateBin.close();
       return;
     }
```

Curclamas:
Hey @me-no-dev , thanks for the hint! Will implement it in the next days since I'm quite busy right now!

Curclamas:
@me-no-dev, just added your recommended snippet to the code with a slight modification and commited it.

PR #629, opened 2017-09-12
Very slight difference between LoRa kit and WiFi kit.

pnewb:
Looks like the LoRa kit pinout was copied in for the WiFi kit.

If you reference http://esp32.net/images/Heltec/WIFI-Kit-32/Heltec_WIFI-Kit-32_DiagramPinout.jpg, you'll see a few differences, this PR should address them.  I've only tested out the SPI pins and the LED_BUILTIN specifically.

### DIFF
--- a/variants/heltec_wifi_kit_32/pins_arduino.h
+++ b/variants/heltec_wifi_kit_32/pins_arduino.h
@@ -11,7 +11,7 @@
 #define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
 #define digitalPinHasPWM(p)         (p < 34)
 
-static const uint8_t LED_BUILTIN = 2;
+static const uint8_t LED_BUILTIN = 25;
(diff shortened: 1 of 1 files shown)


variants/heltec_wifi_kit_32/pins_arduino.h
```diff
@@ -52,8 +55,8 @@ static const uint8_t T4 = 13;
 static const uint8_t T5 = 12;
 static const uint8_t T6 = 14;
 static const uint8_t T7 = 27;
-static const uint8_t T8 = 33;
-static const uint8_t T9 = 32;
```
me-no-dev: Actually this is correct ;) 32 is T9 and 33 is T8

pnewb: Well that's interesting.  The pinout diagram states gpio 33 = t8, but the espressif PDF states gpio 32 = t8.

me-no-dev:
Awesome :) thanks

PR #430, opened 2017-06-10
Porting ESP8266WebServer to ESP32 (Issue #425)

salqadri:
Solving issue #425 by porting ESP8266WebServer to ESP32. Also ported most examples, except WebUpdate, SDWebServer and FSBrowser.

PR #635, opened 2017-09-13
ESP8266Webserver --> ESP32Webserver 

PaulPetring:


me-no-dev:
This already exists as [PR](https://github.com/espressif/arduino-esp32/pull/430) and the plan in general is totally different. You can read on it in the PR :)

PR #658, opened 2017-09-23
Add SD total  and used disk space like SPIFFS

luc-github:
Add example of long file name
Add example of fatfs API using shortname

### DIFF
--- a/libraries/SD/src/SD.h
+++ b/libraries/SD/src/SD.h
@@ -17,6 +17,7 @@
 #include "FS.h"
 #include "SPI.h"
 #include "sd_defines.h"
+#include "ff.h"
 
 namespace fs
 {
@@ -32,6 +33,8 @@ public:
     void end();
     sdcard_type_t cardType();
     uint64_t cardSize();
+    uint64_t totalBytes(const char * logical_drive="0:", FRESULT * error=NULL);
+    uint64_t usedBytes(const char * logical_drive="0:", FRESULT * error=NULL);
 };
 
 }
--- a/libraries/SD_MMC/src/SD_MMC.h
+++ b/libraries/SD_MMC/src/SD_MMC.h
@@ -17,6 +17,7 @@
 #include "FS.h"
 #include "driver/sdmmc_types.h"
 #include "sd_defines.h"
+#include "ff.h"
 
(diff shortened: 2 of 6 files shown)


me-no-dev:
I think this can be added to SD_MMC as well?

luc-github:
I guess but I cannot test to confirm behavior on mmc
I add to mmc and trust fate ? :smile:

me-no-dev:
both run on fatfs ;) and I have the means to test :P

luc-github:
ok no problem then, I will add it

libraries/SD/src/SD.h
```diff
@@ -32,6 +33,8 @@ class SDFS : public FS
     void end();
     sdcard_type_t cardType();
     uint64_t cardSize();
+    uint64_t totalBytes(const char * logical_drive="0:", FRESULT * error=NULL);
+    uint64_t usedBytes(const char * logical_drive="0:", FRESULT * error=NULL);
```
me-no-dev: Please forward declare ```FRESULT``` and not expose ```ff.h``` here :)

libraries/SD_MMC/src/SD_MMC.h
```diff
@@ -32,6 +33,8 @@ class SDMMCFS : public FS
     void end();
     sdcard_type_t cardType();
     uint64_t cardSize();
+    uint64_t totalBytes(const char * logical_drive="0:", FRESULT * error=NULL);
+    uint64_t usedBytes(const char * logical_drive="0:", FRESULT * error=NULL);
```
me-no-dev: Same thing :)

libraries/SD/src/SD.h
luc-github: Ok I can add 
```
enum FRESULT : unsigned int;
```
instead but the compilation failed because FRESULT is not explicitly typed as unsigned int ( it failed also if no type is provided)
Sorry I am not familiar with enum declaration forwarding ....


Also in SD_test.ino I then need to add `#include "ff.h"`
because of `FR_OK` and `FILINFO fno;`

luc-github:
I close it and redo new one removing parameters as discussed with @me-no-dev

PR #673, opened 2017-09-27
Add totalBytes / usedBytes to SD/SDMMC to be same as SPIFFS

luc-github:
Add totalBytes / usedBytes to SD/SDMMC to be same as SPIFFS


me-no-dev:
Good :) merging!

luc-github:
thank you

PR #678, opened 2017-09-28
I2c reset functionality

lonerzzz:
Because of two situations, reboot and bus error, in which the ESP32 I2C state machine can get into a non-recoverable state where the busy flag remains set, potentially causing lockup or preventing I2C from running, the reset function has been added at the HAL layer and made available from Wire.

An attempt was made to time out operations and have i2cReset called internally, but that ended up creating more problems than it solved because the slave may need to release the line(s) first. At this point anyone who runs into bus errors that leave the I2C unusable after SDA and SCL go high can call the reset() function to reset the hardware I2C elements.

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -18,6 +18,7 @@
 #include "freertos/task.h"
 #include "freertos/semphr.h"
 #include "rom/ets_sys.h"
+#include "driver/periph_ctrl.h"
 #include "soc/i2c_reg.h"
 #include "soc/i2c_struct.h"
 #include "soc/dport_reg.h"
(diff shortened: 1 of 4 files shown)


cores/esp32/esp32-hal-i2c.c
```diff
@@ -445,7 +460,25 @@ void i2cInitFix(i2c_t * i2c){
     i2c->dev->fifo_data.data = 0;
     i2cSetCmd(i2c, 1, I2C_CMD_WRITE, 1, false, false, false);
     i2cSetCmd(i2c, 2, I2C_CMD_STOP, 0, false, false, false);
+    if (i2c->dev->status_reg.bus_busy) // If this condition is true, the while loop will timeout as done will not be set
+    {
+        //log_e("Busy at initialization!");
+    }
```
me-no-dev: maybe leave the error uncommented? otherwise what is the point of the check? :P

lonerzzz: The fact that you commented means it served its point. :) I left it commented out since I wanted to highlight where the error comes into play and because of the fact most errors don't have error logging enabled in the read and write methods so was unsure when we want to show and not show.

I can certainly add it any with any others we think should be there.

Any other concerns?

me-no-dev: I think we can uncomment the logs, since when I wrote it there was no way to control the log level :P now there is :)

lonerzzz:
Perfect, thank you.

PR #691, opened 2017-10-02
Print an error when serial number is invalid

copercini:


cores/esp32/HardwareSerial.cpp
```diff
@@ -14,6 +14,10 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
     if(_uart) {
         end();
     }
+    if(0 > _uart_nr || _uart_nr > 2) {
+        log_e("Serial number is invalid, please use 0, 1 or 2");
+        end();
```
me-no-dev: no need for end(). this if can be the first thing in the function

copercini: Done :D

me-no-dev:
and done :)

PR #707, opened 2017-10-09
Update pgmspace.h

stickbreaker:
Fix redefines of strXXX, original was using strnXXX with length set to 0x7fffffff.  This caused problems with strcpy and strcat. The destination buffer was NULL'd to 0x7fffffff which killed the app.


me-no-dev:
Awesome :) thanks!

PR #711, opened 2017-10-09
Changed update example to use different control flow structure

Kevintjeb:
I saw that an update example got added, felt that the example could be made tidier when not using goto.


me-no-dev:
Good idea :) Thanks!

PR #719, opened 2017-10-12
Added argument to HardwareSerial.begin to specify whether the logic l…

ewasscher:
…evels of the UART rx and tx lines should be inverted


me-no-dev:
good idea :)

PR #750, opened 2017-10-20
Fix Early return when Address NAK is received

stickbreaker:
The i2cWrite() function was returning to the app before the i2c transaction had completed.  This caused the next Wire() call to return a I2C_ERROR_BUSY.


me-no-dev:
Arghhh... thanks :)

PR #756, opened 2017-10-22
Fixes for zero length packet bug, buffer overflow in parseInt(), added end() method

davruet:
ArduinoOTA would stop receiving any packets if the port received a zero-length UDP packet, commonly sent by network scanners like nmap. Fixed to flush() after every call to parsePacket(), even if read length is 0.

Additionally, added length checking to fix a potential buffer overflow in parseInt().

Finally, added an end() method that stops the OTA listener and releases resources.


davruet:
Fixing annoying whitespace changes... will resubmit.

PR #760, opened 2017-10-22
Add ICACHE_RAM_ATTR compatibility macro with esp8266

copercini:
This help the ESP8266 libraries to run on ESP32

### DIFF
--- a/cores/esp32/esp32-hal.h
+++ b/cores/esp32/esp32-hal.h
@@ -45,6 +45,7 @@ void yield(void);
 
 #define ESP_REG(addr) *((volatile uint32_t *)(addr))
 #define NOP() asm volatile ("nop")
+#define ICACHE_RAM_ATTR IRAM_ATTR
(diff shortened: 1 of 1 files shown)


copercini:
or maybe put it here:  https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/include/esp32/esp_attr.h

me-no-dev:
let's add a file called "esp8266-compat.h" and put that and ICACHE_FLASH_ATTR and maybe more in near future :) include that header into Arduino.h

copercini:
Nice idea!

The strangest thing is that there are some ICACHE_FLASH_ATTR calls in the actual ESP32 core:
https://github.com/espressif/esp-idf/search?q=ICACHE_FLASH_ATTR

me-no-dev:
it evaluates to nothing and prolly not called anywhere to get the undefined error :)

PR #781, opened 2017-10-26
Unicode support in get.py

Patas007:
Localized versions of Linux systems use Unicode characters in the names of standard directories like "Downloads".


me-no-dev:
Great :) Thanks

PR #799, opened 2017-10-31
Fix Heltec Wifi-LoRa kit pinout

RX14:
Based on http://esp32.net/images/Heltec/WIFI-LoRa-32/Heltec_WIFI-LoRa-32_DiagramPinoutFromBottom.jpg


me-no-dev:
thanks :)

PR #807, opened 2017-11-04
Now SmartConfig works!

allex1978:


bill-orange:
I added:

sta_conf->bssid_set = 0;

to the .cpp file and recompiled.  It made no differece.  If I reset the ESP32 is searches forever until I use the phone app.  Is there any else I need to do to effect the change?

As a stop-gap I added the following.  It look a bit awkward,

  if (!WiFi.begin()) 
  {
    Serial.println("Unable to reconnect! Starting SmartConfig");
    setupWiFi ();
  }

allex1978:
Hi bill-orange, you used just a SmartConfig sample code. 
line WiFi.beginSmartConfig(); is always disconnect from old AP and start SmartConfig process. 
If you not want to start SmartConfig after reset you have to rebuild the logic.

bill-orange:
“Have to rebuild the logic”. Could you expand on that a bit. Also, not to be picky by shouldn’t this run on an Esp32 the same way it runs on an Esp8266 for code portability.

beegee-tokyo:
@bill-orange that is what I do
1) try to connect via stored SSID and password
2) if this fails (after timeout) start SmartConfig

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -625,6 +625,7 @@ void WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
     } else if (status == SC_STATUS_LINK) {
         wifi_sta_config_t *sta_conf = reinterpret_cast<wifi_sta_config_t *>(result);
         log_d("SSID: %s", (char *)(sta_conf->ssid));
+        sta_conf->bssid_set = 0;
```
me-no-dev: that is rather strange to make any difference

bill-orange:
This also seems to work.  Still though, shouldn't ESP32 and ESP8266 code work the same for something like this?  

void setupWiFi () {
  /* Set ESP32 to WiFi Station mode */
  WiFi.mode(WIFI_AP_STA);
  /* start SmartConfig */

  if (WiFi.reconnect()) {
    Serial.println( "reconnected from stored credentials");
    WiFi.printDiag(Serial); Serial.println(WiFi.localIP());
    Serial.println("Connected to wifi");
    return;
  }

  WiFi.beginSmartConfig();

  /* Wait for SmartConfig packet from mobile */
  Serial.println("Waiting for SmartConfig.");
  while (!WiFi.smartConfigDone()) {
    delay(500);
    Serial.print(".");
  }
  Serial.println("");
  Serial.println("SmartConfig done.");

  /* Wait for WiFi to connect to AP */
  Serial.println("Waiting for WiFi");
    delay(500);
    Serial.print(".");
  }
  Serial.println("WiFi Connected.");
  Serial.print("IP Address: ");
  
}

beegee-tokyo:
@bill-orange you are right, in my code is a mistake, should be WiFi.reconnect() instead of WiFi.begin(). But it worked for me before. Thanks for pointing out my mistake.

PR #812, opened 2017-11-05
fix NodeMCU-32S .build.board property

CONSULitAS:
The .build.board=xx proptery is used for generating a macro named "ARDUINO_xxx".

C macro names should never have a "-" sign.

* current: ARDUINO_NodeMCU-32S
* should be: ARDUINO_=NodeMCU_32S

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -830,7 +830,7 @@ nodemcu-32s.serial.disableRTS=true
 nodemcu-32s.build.mcu=esp32
 nodemcu-32s.build.core=esp32
 nodemcu-32s.build.variant=nodemcu-32s
-nodemcu-32s.build.board=NodeMCU-32S
+nodemcu-32s.build.board=NodeMCU_32S
 
 nodemcu-32s.build.f_cpu=240000000L
 nodemcu-32s.build.flash_mode=dio


stickbreaker:
What about the `variant` field? Shouldn't it also be changed?
Chuck.

CONSULitAS:
> What about the `variant` field? Shouldn't it also be changed?

Not necessary for correct board recognition in Arduino.

Compare ESP8266/Arduino boards.txt:

```
...
d1_mini.build.mcu=esp8266
d1_mini.build.f_cpu=80000000L
d1_mini.build.board=ESP8266_WEMOS_D1MINI
d1_mini.build.core=esp8266
d1_mini.build.variant=d1_mini
...
```

You can detect that board type in your code by
``` c++
  // correct board in IDE selected?
  #ifndef ARDUINO_ESP8266_WEMOS_D1MINI
    #error Please select board "WeMos D1 R2 & mini"!
  #endif
```

With the current config you need 
``` c++
// NodeMCU ESP-32S - V1.1 LED is HIGH active
#ifdef ARDUINO_NodeMCU-32S
  #define LEDPIN 2
#endif
```
which is really weird because of the "-" sign in the used macro 😁 !

This really works (just tested half an hour ago with a generic blink sketch) but you could think its equivalent to `ARDUINO_NodeMCU - 32S` with spaces which is confusing and horribly error prone...

But I think it would be nice to clean up the config a bit.

stickbreaker:
Yes, I agree.  fix all the small things and the big things will take care of themselves!

CONSULitAS:
> What about the `variant` field? Shouldn't it also be changed?

Just one more thing:

The variant field is used to build the path to the variant config. For the mentioned WeMos D1 mini you get `"-I/Users/xxx/Library/Arduino15/packages/esp8266/hardware/esp8266/2.4.0-rc2/variants/d1_mini"` as an compiler option. Just switch on the detailed output while compiling.

Here you see the effect of the build.board parameter, too:
*`-DARDUINO_ESP8266_WEMOS_D1MINI`
*`-DARDUINO_BOARD="ESP8266_WEMOS_D1MINI"` (just forgot the ARDUINO_BOARD macro 😄 should be there, too)

For NodeMCU-32S we get:
* `-DARDUINO_NodeMCU-32S`
* `-I/Users/xxx/Documents/Arduino/hardware/espressif/esp32/variants/nodemcu-32s"`
* `-DARDUINO_BOARD="NodeMCU-32S"` **is missing** if you compare to ESP8266/Arduino (see `platform.txt`) :arrow_right: other PR to come

BTW: The compiler throws `<command-line>:0:16: warning: ISO C++11 requires whitespace after the macro name` for each call. This seems to be because of the minus sign.

**Conclusion:** No, we should not change the `variant` field or we have to change the folder structure, too.

PR #813, opened 2017-11-05
Fix compiler flags to be compatible with ESP8622/Arduino

CONSULitAS:
see also "fix NodeMCU-32S .build.board property" (https://github.com/espressif/arduino-esp32/pull/812#issuecomment-341990261)

`-DARDUINO_BOARD={build.board}` is missing

With this fix you get a macro `ARDUINO_BOARD` with the name of the selected board (e.g. `"NodeMCU-32S"`).

https://github.com/esp8266/Arduino/blob/master/platform.txt#L75
https://github.com/esp8266/Arduino/blob/master/platform.txt#L78
https://github.com/esp8266/Arduino/blob/master/platform.txt#L81


me-no-dev:
Nice :) don't know how I missed that

PR #820, opened 2017-11-10
Add missing `import ssl`

rtwfroody:


me-no-dev:
why is this? OS?

rtwfroody:
For Windows, cygwin. On line 83 of the original the ssl module is used, but it is never imported.

me-no-dev:
OK, but for Windows you have EXE to execute, and while I was compiling it on Windows, I had no issues or errors that class is not imported. Did you get errors?

rtwfroody:
Yes. I got an error that the ssl module wasn't defined. I forget the exact
syntax. I ran get.py using python from a cygwin terminal. I'm not sure if
that's even supposed to work.

PR #822, opened 2017-11-10
fix enable ADC on pin37,38

radiumray:
fix enable ADC on gpio37,38
gpio37,38 will enable when make this change.


me-no-dev:
Thanks :)

palomarcoder:
Will these pins actually work though, on boards that have the capacitors hard-wired?  Do we need to cut the traces to the capacitors?  or is there a way to disable the pre-amp?  
I've never really seen good documentation on which of these pings can be used and which cannot if the capacitors are in place.

Issue #832, opened 2017-11-13
turning off interrupts doesn't work


rtwfroody:
The following code shows that disabling interrupts is not implemented for this platform:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/Arduino.h#L83
```
#define sei()
#define cli()
#define interrupts() sei()
#define noInterrupts() cli()
```

This is causing problems for me because I'm trying to read a DHT22 temperature sensor. The library contains a critical section (https://github.com/adafruit/DHT-sensor-library/blob/master/DHT.cpp#L149) where it needs to measure the length of high and low pulses on a digital input pin. Interrupts going off throw off the measurement. I've modified the code to watch timers instead of counting cycles (https://github.com/adafruit/DHT-sensor-library/pull/93), but I still see a 5-10% failure rate.

It looks like FreeRTOS has functions to use in critical sections: http://www.freertos.org/taskENTER_CRITICAL_taskEXIT_CRITICAL.html. Using those macros leads to compilation errors. Is it possible to use FreeRTOS functions from Arduino code?

This all may be related to https://github.com/espressif/arduino-esp32/issues/811

beegee-tokyo:
Not sure why you get compilation errors (I am on PlatformIO using the staging version of arduino-esp32), but I use 
**`portMUX_TYPE mux = portMUX_INITIALIZER_UNLOCKED;`**
**`portENTER_CRITICAL(&mux);`**
**`portEXIT_CRITICAL(&mux);`**
for reading a DHT11 sensor (not the Adafruit library but [markruys/arduino-DHT](https://github.com/markruys/arduino-DHT/blob/master/DHT.cpp)) and it works well.

Here is the modified readSensor() function:

```
void DHT::readSensor()
{
  // Make sure we don't poll the sensor too often
  // - Max sample rate DHT11 is 1 Hz   (duty cicle 1000 ms)
  // - Max sample rate DHT22 is 0.5 Hz (duty cicle 2000 ms)
  unsigned long startTime = millis();
  if ( (unsigned long)(startTime - lastReadTime) < (model == DHT11 ? 999L : 1999L) ) {
    return;
  }
  lastReadTime = startTime;

  temperature = NAN;
  humidity = NAN;

  // Request sample

  digitalWrite(pin, LOW); // Send start signal
  pinMode(pin, OUTPUT);
  if ( model == DHT11 ) {
    delay(18);
  }
  else {
[...]

PR #835, opened 2017-11-14
Implement sei/cli using port{EXIT,ENTER}_CRITICAL

rtwfroody:
This fixes the library I use to read temperature/humidity from DHT22.

One problem is that the FreeRTOS functions work when nested, while I believe the Arduino sei() enables interrupts regardless of how many times cli() was invoked. Another problem is that these FreeRTOS functions don't work when called from interrupt handlers. (There are others for that.)

These caveats aside, this does seem to work in my thermostat project.

This is probably not ready to merge, but might be a good starting point for a more complete solution. It helps with issue #832.

### DIFF
--- a/cores/esp32/esp32-hal-timer.c
+++ b/cores/esp32/esp32-hal-timer.c
@@ -25,6 +25,8 @@
 #define HWTIMER_LOCK()      portENTER_CRITICAL(timer->lock)
 #define HWTIMER_UNLOCK()    portEXIT_CRITICAL(timer->lock)
 
+portMUX_TYPE arduino_mux = portMUX_INITIALIZER_UNLOCKED;
+
 typedef struct {
     union {
         struct {
--- a/cores/esp32/esp32-hal-timer.c
+++ b/cores/esp32/esp32-hal-timer.c
@@ -25,8 +25,6 @@
 #define HWTIMER_LOCK()      portENTER_CRITICAL(timer->lock)
 #define HWTIMER_UNLOCK()    portEXIT_CRITICAL(timer->lock)
 
-portMUX_TYPE arduino_mux = portMUX_INITIALIZER_UNLOCKED;
-
 typedef struct {
     union {
         struct {
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -80,8 +80,6 @@
 #define degrees(rad) ((rad)*RAD_TO_DEG)
 #define sq(x) ((x)*(x))
 
-#define sei()   portEXIT_CRITICAL(&interrupt_mux);
-#define cli()   portENTER_CRITICAL(&interrupt_mux);
 #define interrupts() sei()
 #define noInterrupts() cli()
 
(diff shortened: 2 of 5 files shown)


cores/esp32/esp32-hal-timer.c
```diff
@@ -25,6 +25,8 @@
 #define HWTIMER_LOCK()      portENTER_CRITICAL(timer->lock)
 #define HWTIMER_UNLOCK()    portEXIT_CRITICAL(timer->lock)
 
+portMUX_TYPE arduino_mux = portMUX_INITIALIZER_UNLOCKED;
+
```
me-no-dev: first mayb e move those to esp32-hal.h and esp32-hal-misc.c ? should not be in the timers ;)

atanisoft: Perhaps also rename arduino_mux to interrupt_lock or interrupt_mux ?

gtalusan:
I'm a bit confused by this.... wouldn't you want to actually disable interrupts instead of locking on a mutex?  I see there are calls for `xt_ints_off` and `xt_ints_on` that let you directly mask the interrupt.  There's also `ets_intr_lock` and `ets_intr_unlock` that big hammer all interrupts.

me-no-dev:
http://esp-idf.readthedocs.io/en/latest/api-guides/freertos-smp.html#critical-sections-disabling-interrupts

Question is: What do you want to achieve by cli/sei? Protect memory? Make sure nothing interrupts particular code?

timsifive:
@gtalusan Actually disabling interrupts would be better, although then you might lose things like the watchdog. I suppose that's just inevitable if you're trying to implement Arduino's simple view of the world in a more complex system.

@me-no-dev The DHT sensor code needs to disable interrupts so it can count how long its output signal 
is low/high. The sensor generates a sequence of high pulses on its signal line without a clock, and the reader needs to compare the duration of the high pulses to a reference pulse. So the code disables interrupts, and polls the signal in a tight loop, counting the number of loop iterations that the signal is high. Obviously interrupts interfere with that.

Strangely, now that I've migrated my project to use the Arduino layer as a component in esp-idf, reading the temperature seems to work often enough without applying this patch at all. It works about 93% of the time without this patch. With the patch it works 100% of the time.

stickbreaker:
@rtwfroody I did a similar  modification to the OneWire library #951 . It seems to work just fine.

Chuck.

luc-github:
Sorry to comment, and forgive me if I am wrong 

`cli()` is same as `noInterrupts()` & `sei()` is same as `interrupts()` so should be according https://github.com/espressif/arduino-esp32/blob/600f4c4130f2411e8ff33396d0ef40a7ec257cd6/tools/sdk/include/freertos/freertos/portmacro.h#L213-L244 
`portDISABLE_INTERRUPTS()` and `portENABLE_INTERRUPTS()` per my understanding

`portENTER_CRITICAL(&interrupt_mux);` and `portEXIT_CRITICAL(&interrupt_mux);`  which call `vTaskEnterCritical(mux)` and `vTaskExitCritical(mux)`  have different purpose , no ?

me-no-dev:
@luc-github you are correct :) closing this. Feel free to add one with the proper defines

PR #840, opened 2017-11-16
I2C ReWrite for ISR, and TIMEOUT

stickbreaker:
This PR is mainly for Comment.  

I don't believe it should be unconditionally merged. 
I re-purposed the README.md as a change explanation file. So, something else needs to be done with README.md.

A direct pull will cause all existing Arduio code that uses sendStop=false and checks the return code to Fail with a NEW error code I2C_ERROR_CONTINUE.

Chuck.


atanisoft:
Perhaps rename README.md to README-i2c.md?

stickbreaker:
@atanisoft But, I wanted it to be auto-Displayed on the main GitHub of my fork.

stickbreaker:
Before this is pull is merged, many debug messages and structures need to be removed, basically the Dump*, intBuff[] and associated operations.

Also, there is a partially implemented Async structure.  I am planning on developing an Async feature where an I2C transaction is handed off to the ISR and the ISR notifies the foreground process via EventGroup Flags.  I haven't yet thought through exact implementation and operation details, it might take a while.  Before I get to it though, I want SLAVE mode support.  That was my original reason for messing with I2C, I want SLAVE mode. 
Chuck.

stickbreaker:
@me-no-dev Travis is failing in Platformio, complaining about a missing file"WiFiClientSecure.h" I have no Idea.  

chuck.

me-no-dev:
PIO issues ;) dissregard

me-no-dev:
in master now :)

PR #857, opened 2017-11-22
Add Wemos Lolin32-lite variant

henricazottes:
Same pins file except that the builtin led is on gpio 22 instead of gpio 5.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -335,6 +335,49 @@ lolin32.menu.UploadSpeed.512000.upload.speed=512000
 
 ##############################################################
 
+lolin32-lite.name=WEMOS LOLIN32-Lite
+
+lolin32-lite.upload.tool=esptool
+lolin32-lite.upload.maximum_size=1310720
+lolin32-lite.upload.maximum_data_size=294912
+lolin32-lite.upload.wait_for_upload_port=true
+
+lolin32-lite.serial.disableDTR=true
+lolin32-lite.serial.disableRTS=true
+
+lolin32-lite.build.mcu=esp32
+lolin32-lite.build.core=esp32
+lolin32-lite.build.variant=lolin32-lite-lite
+lolin32-lite.build.board=LOLIN32-LITE
+
+lolin32-lite.build.f_cpu=240000000L
+lolin32-lite.build.flash_mode=dio
+lolin32-lite.build.flash_size=4MB
+lolin32-lite.build.boot=bootloader
+lolin32-lite.build.partitions=default
+
+lolin32-lite.menu.FlashFreq.80=80MHz
+lolin32-lite.menu.FlashFreq.80.build.flash_freq=80m
+lolin32-lite.menu.FlashFreq.40=40MHz
+lolin32-lite.menu.FlashFreq.40.build.flash_freq=40m
+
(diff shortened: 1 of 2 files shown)


me-no-dev:
sure, but there is no board definition that uses this variant?

henricazottes:
I admit I didn't check.
But even if there is one compatible, wouldn't a user look for the variant corresponding to the board he is using ?

For my case, I tried the Lolin 32 variant, it didn't work, I checked what could be the problem and found it was just the LED pin definition. I kinda know what I'm doing, maybe less skilled users would just have conclude that their board is dead :/

me-no-dev:
Without a board definition that uses variant ```lolin32-lite``` this file is never used ;)
Either add a board definition or close this :)

me-no-dev:
..... ```lolin32``` already exists! maybe you meant to name it ```lolin32-lite```? Please pay attention!

henricazottes:
Thanks for the tip :)

henricazottes:
Anything else to correct/change ?

me-no-dev:
you are running an old code base ;) boot=bootloader is no longer correct! Please update your fork and change this to boot=dio!!!!

you are running an old code base ;) boot=bootloader is no longer correct! Please update your fork and change this to boot=dio!!!!

henricazottes:
Done :)

henricazottes:
@me-no-dev anything else ?

boards.txt
```diff
@@ -335,6 +335,49 @@ lolin32.menu.UploadSpeed.512000.upload.speed=512000
...
+lolin32-lite.upload.maximum_size=1310720
+lolin32-lite.upload.maximum_data_size=294912
+lolin32-lite.upload.wait_for_upload_port=true
+
+lolin32-lite.serial.disableDTR=true
+lolin32-lite.serial.disableRTS=true
+
+lolin32-lite.build.mcu=esp32
+lolin32-lite.build.core=esp32
+lolin32-lite.build.variant=lolin32-lite-lite
```
wunderlins: This is wrong, the variant folder is `lolin32-lite` in this patch. Therefore the correct value would be:
`lolin32-lite.build.variant=lolin32-lite`

boards.txt
```diff
@@ -335,6 +335,49 @@ lolin32.menu.UploadSpeed.512000.upload.speed=512000
...
+lolin32-lite.build.board=LOLIN32-LITE
```
wunderlins: on my system, with arduino 1.8.5 (linux-x64) it only works if I change this to `lolin32-lite.build.board=LOLIN32`

evark:
It is not just the LED_BUILTIN different.
Pin 21 (default SDA) , does not exist in lolin32-lite. 
I2C pins must relocated from 21/22 to another location, eg 22/23.

boards.txt
rhyslbw: Correct in #1076

boards.txt
rhyslbw: Correct in #1076

rhyslbw:
Another PR with the variant here #1076

DavidAntliff:
@me-no-dev why did you close this?

PR #864, opened 2017-11-23
Bugfix I_DELAY macro

robotrovsky:
When compiling 

> const ulp_insn_t program[] = {
> I_DELAY(1)
> };

with the xtensa-esp32-elf-g++ compiler i always got the error:

> sorry, unimplemented: non-trivial designated initializers not supported
> 
>        };

This was due to the different order in the macro and the struct. The struct has another order of the fields (opcode, unused, cycles) vs (cycles, unused, opcode):
>    struct {
>        uint32_t cycles : 16;       /*!< Number of cycles to sleep */
>        uint32_t unused : 12;       /*!< Unused */
>        uint32_t opcode : 4;        /*!< Opcode (OPCODE_DELAY) */
>    } delay;                        /*!< Format of DELAY instruction */

After updating the order in the macro it is possible to compile with the g++ compiler.

### DIFF
--- a/tools/sdk/include/ulp/esp32/ulp.h
+++ b/tools/sdk/include/ulp/esp32/ulp.h
@@ -266,9 +266,9 @@ _Static_assert(sizeof(ulp_insn_t) == 4, "ULP coprocessor instruction size should
  * Delay (nop) for a given number of cycles
  */
 #define I_DELAY(cycles_) { .delay = {\
-    .opcode = OPCODE_DELAY, \
+    .cycles = cycles_, \
     .unused = 0, \
-    .cycles = cycles_ } }
+    .opcode = OPCODE_DELAY } }
 
 /**
  * Halt the coprocessor.


me-no-dev:
This is great, but this change should be actually done [here in esp-idf](https://github.com/espressif/esp-idf/blob/master/components/ulp/include/esp32/ulp.h) :) the tolls/sdk folder actually contains the headers and compiled form of esp-idf, so if the change is made there, it will be pulled here automatically every new rebuild of the libs :)

me-no-dev:
@igrr @Spritetm PTAL

robotrovsky:
@me-no-dev If you want i can make a pull request to the IDF. Also i haven't checked if other macros are also mixed up, i just noticed, that the I_DELAY macro is messing up my build :)

me-no-dev:
👍 please make PR in idf :) we can all benefit from the fix then.

robotrovsky:
OK

robotrovsky:
Done. https://github.com/espressif/esp-idf/pull/1310

PR #873, opened 2017-11-24
Use real flash mode and frequency for bootloader

ivankravets:


me-no-dev:
will you change anything here since qio/qout upload is done in dio/dout?

ivankravets:
I updated PR. Could you review?

me-no-dev:
looks good :) merging...

PR #874, opened 2017-11-25
Set HSPI ports by default when HSPI is selected

zipiju:
When user selected HSPI with SPIClass name(HSPI) ESP was, by default,
still using VSPI ports (the ones defined in pins_arduino.h).
With this change when user selects HSPI then HSPI default ports will be
used.
If user won't specify HSPI then VSPI default ports will be used.
If user will specify SCLK, MOSI, MISO and SS with SPI.begin() then user
defined ports will be used no matter if VSPI or HSPI is selected.
With this change fe. SD library can use default HSPI ports. It was
possible to
pass HSPI SPI instance to SD lib, however even then it was using VSPI
ports which were (probably) GPIO matrixed to HSPI.

### DIFF
--- a/libraries/SPI/src/SPI.cpp
+++ b/libraries/SPI/src/SPI.cpp
@@ -49,10 +49,17 @@ void SPIClass::begin(int8_t sck, int8_t miso, int8_t mosi, int8_t ss)
         return;
     }
 
-    _sck = sck;
-    _miso = miso;
-    _mosi = mosi;
-    _ss = ss;
+    if(sck == -1 && miso == -1 && mosi == -1 && ss == -1) {
+        _sck = (_spi_num == VSPI) ? SCK : 14;
+        _miso = (_spi_num == VSPI) ? MISO : 12;
+        _mosi = (_spi_num == VSPI) ? MOSI : 13;
+        _ss = (_spi_num == VSPI) ? SS : 15;
+    } else {
+        _sck = sck;
+        _miso = miso;
+        _mosi = mosi;
+        _ss = ss;
+    }
 
     spiAttachSCK(_spi, _sck);
     spiAttachMISO(_spi, _miso);
--- a/libraries/SPI/src/SPI.h
+++ b/libraries/SPI/src/SPI.h
@@ -52,7 +52,7 @@ private:
 
 public:
     SPIClass(uint8_t spi_bus=HSPI);
-    void begin(int8_t sck=SCK, int8_t miso=MISO, int8_t mosi=MOSI, int8_t ss=-1);
+    void begin(int8_t sck=-1, int8_t miso=-1, int8_t mosi=-1, int8_t ss=-1);
     void end();
 
     void setHwCs(bool use);


me-no-dev:
Cool! Keep in mind that the default SPI port might change in near future ;) PSRAM is using VSPI

rmetzner49:
Hello, I'm sort of a newbie at this, so please bear with what may be an obvious question to someone that eats and sleeps code.  I have an application where I'm at wit's end sharing VSPI with an RA8875 Display.  There is a documented Silicon bug where this chip does NOT tri-state MISO.  I have tried every which way to isolate the hardware but even just putting a 74LVC125 with OE grounded makes it not work.  

So my answer is to use both HSPI and VSPI.  Would the correct way to go about this be to create TWO separate Classes, one called HSPIClass and one called VSPIClass? From what I could tell the Library creates an object using either port, but not both.  

Thank you in advance for your help.

stickbreaker:
@rmetzner49 have you tried connecting the OE to the RA8875's CS pin? if the RA8875 does not tristate MISO then adding a tri-state buffer(74LVC125) between the RA8875's SO and the MISO input of the ESP32 should work?

Chuck.

rmetzner49:
You would think so, but even with just that one Buffer delay, the display stops working.  I initially did connect it to the CS pin but that broke it, so my thought was just connecting it to GND should work provided it's the only device on the bus.  It should just look like a buffer, but no dice.  I suppose I could try running every pin through a buffer just to see if that changes but I'm not sure that's the correct road.  There is a LOT of activity on the SPI bus running the display because I also used the Touch feature.  The other device I want to run is a TDC1000 Ultrasonic Active Front End.  Without having run THAT, I don't know what overhead it has beyond just initializing.  My plan was to use the PulseIn function to measure the actual start/stop events on REAL pins.  Do you think making two classes will allow both ports to work?

dpharris:
A quick and dirty solution might be a diode on that line, in the
appropriate polarity?

David

stickbreaker:
that buffer is spec'd to only delay the signal 4.3ns at 3.3V. from OE to output is 2.4ns.

What is your bus speed? 160mhz? I'm being factious. :) if you are using 4MHz or 8MHz I cannot see how the delay from a single buffer would affect your communication.  More likely you have a SPI MODE mismatch.   Looking at the [RA8875 datasheet](https://cdn-shop.adafruit.com/datasheets/RA8875_DS_V19_Eng.pdf) from AdaFruit, the Device is using MODE3.

Chuck.

rmetzner49:
Good question.  I had tried to up the speed months ago without much luck.  I came across the following in RA8875.cpp which tells me Adafruit is mucking with the settings.  It was initially at 2Mhz but I bumped it to 4.  REM are my initials and comments.  I'll try to peek at which MODE is being sent, though I never tweaked that.  I remember trying 4, 8, 10, & 12 and 4 is about the fastest reliable setting.  I'm using Adafruit's RA8875 drivers.



 // REM CHANGED FROM 2000000 
  SPI.begin();
  spi_speed = 4000000;  //  4MHZ IS ALL SHE WROTE!! WON'T DO 10MHZ EITHER


RA8875 start
SPI_speed: 4000000
FuelGauge 0

stickbreaker:
@rmetzner49 look through their code for `SPI.beginTransaction();`, `SPISettings();` or `setDataMode()`

hopefully you will find `SPI.beginTransaction(SPISetting(2000000, MSBFIRST, SPI_MODE3));`

Chuck.

rmetzner49:
// max speed!
        SPI.beginTransaction(SPISettings(spi_speed, MSBFIRST, SPI_MODE0));
WELL THEN!  they are calling it with the wrong MODE.

rmetzner49:
Though if I read the RA8875 Data Sheet correctly, it says "read mode <SysClock/6" which matches my empirical tests; that it dies at any setting > 4MHz because 20M/6 = 3.333.

rmetzner49:
dpharris:  I doubt a simple diode; even a Schottky is the answer, since this thing seems really fussy on timing.

stickbreaker:
Here is a screen grab of the pertinent SPI timing:
ra8875 spi

clock is inactive Positive (CPOL==1), Data is sampled on the Rising Edge of clock (CPHA==1) there for
 0b(CPOL)(CPHA) = 0b11 = 3

Try just changing to Mode3 and see what happens.

I can see where Mode0 (clock inactive Low, Sampled on Falling edge) would be very timing sensitive.

Chuck.

rmetzner49:
I'll do that right after I replace the LVC125 which seems to have gone to lunch.  Gotta haul out the microscope.  ;-)  The other thing I was thinking is I could maybe pass both the SCLK and MOSI through a buffer and see if that fixes it.  Thank you for at least making me examine the issue.  I still may try to make TWO classes and see what happens.  It's going to be a bunch of work though.  I wish I could run it past 3.33 MHz because it's dorky to switch between one screen and another.  Takes about 3-4 seconds.  Both the TDC1000 and the Display CAN work on the same bus only because there isn't any traffic on the TDC1000 once getting it initialized.  Thanks, Bob M.

rmetzner49:
Now I'm really confused.  I haven't tried MODE3 yet but I seem to be getting MISO activity while CS is high, which I don't think is supposed to happen.  I have a brand new OWON scope and I think one of the features I stumbled on in the menus is it has a BUS menu for I2C and SPI.  Time to get the manual out and understand that feature.  Thanks, and stay  tuned.  I did BTW get the LVC125 replaced and I can now ground OE and still have the display work.  As soon as I determine what's going on with CS, I'll know more.   Bob M.

stickbreaker:
@rmetzner49 
Are you doing individual `SPI.transfer()` calls, each call en curse a `MUTEX` call,  there is a `transaction` feature set that is much more efficient for bigger transfers.

Chuck.

rmetzner49:
I'm doing what EVER the Adafruit Library does.  I haven't messed with anything else, and I would expect therefore the display should work, and it does sans the problem with the RA8875 we discussed above.  Unless I'm confused and SPI works differently than say, any other addressed device, there shouldn't be MISO (or MOSI)  activity in the absence of CS being LOW.  Else how would a device know it's being selected for communication?  Once I understand how to properly set up my new 'scope using its SPI trigger feature, I might see sensible activity.  Right now though I can see why when I connect the LVC125's OE pin to CS it quits working.  In my head, there shouldn't be activity on MISO unless CS is LOW.  Yet there is.   -- Bob M.

rmetzner49:
Will hit this again tomorrow with a clear head.

stickbreaker:
@rmetzner49 Yep, the MISO line should be tri-stated when it is inactive. 

Good luck.!
Chuck.

rmetzner49:
Well this has been interesting:  Whilst poking around all of my "hal" files to understand what's going on, I had included esp32-hal-adc files even though I decided against using the onboard A/D.  I had already thrown in an I2C, AD7994 to take care of my battery monitoring functions and I was also going to use it for Ultrasonic Flow Detection also.  So I ripped those files out that deal with the onboard ADC and low and behold, MOSI now works when tri-stated through CS on the SPI bus.  The picture attached shows this.  The top trace is MOS, the botom is CS and you can see about 1V when CS is inactive.  Please excuse the shabby triggering, but the display now works.

stickbreaker:
@rmetzner49 Sounds like you found something, but alas, I can't see your 'attached' picture 😦. 

It was MISO right? not MOSI?  Did you change the SPI Mode?

Chuck.

rmetzner49:
Yes, it was MISO and I tried MODE3 but NO GO.  MODE0 is all that works.  I'll try attaching the picture again, if not, I'll attach it to gmail.
wp_20180115_14_53_45_pro

rmetzner49:
Worked that time.

rmetzner49:
I love my new Christmas Present to myself. (Scope).  Seems every bit as capable as my Tek-2465 and it was $438 including shipping!

stickbreaker:
@rmetzner49 your are dating your self.  I have a Tek-2236, couldn't afford the 2465 in 1984! , I'm looking at a Rigol MSO1074z for $699. I can't believe I spent $2,400 on a scope in 1984 and now, I am questioning if I want to spend $699.00.  I just ran a [inflation calculator](https://data.bls.gov/cgi-bin/cpicalc.pl?cost1=2400&year1=198401&year2=201712). $2,400 in 1984 is equivalent to $5,806 in today's dollars!  What a ripoff!
Chuck.

stickbreaker:
@rmetzner49 That display look great! I'm a little younger than you but, hardly enough that matters. 
The primary reason I'll looking at it is the 16 channel logic analyzer. Though it is only 8bit resolution with 4 inputs.  If you enable the logic analyzer you loose channel 3 and 4.  It just stuffs those 16bits into the trace buffer where channel 3 and 4's data would go. 

Do you think you will trace down the interaction with the 'hal-adc'? 

Chuck.

rmetzner49:
TBH, I don't know that was the only factor.  I *did* also grab both new "hal-spi" files from this article.  I also remember Espressif changed a bunch of HAL code between 10/14/17 and 1/1/18 because  about the time I added esp32-hal-i2c to the project, it no longer compiled until I updated esp32-hal.  This is why I ended up on this thread trying to run both SPI ports.  I originally had a 32U4 with its timers setting up the Ultrasonic and I had it about half working, but every wire I moved gave different results.  My plan was to use i2c to talk between processors.  
[LINDHOLM.zip](https://github.com/espressif/arduino-esp32/files/1633060/LINDHOLM.zip)

So although I might have eventually gotten it to work, using the TDC1000 seemed like the way to go.  Of course it's going to be a chore to port the setup of that device to this project, but I have a good starting example from Lindholm using it with a TDC7200 stopwatch chip.  Now that I can have it on VSPI, I can at least see light at the end of the tunnel.

stickbreaker:
@rmetzner49 You might run into some issues with the I2C code.  The I2C HAL currently in the main branch espressif/arduino-esp32 are very unreliable, they use a polling interface with the I2C hardware and are very timing sensitive. 

I rewrote the I2C 'HALs' in my fork stickbreaker/ardunio-esp32. My version uses an ISR model instead, It has proven to be robust.  The maintainer for the main branch @me-no-dev was working on incorporating my fork into main branch. But, I have seen no activity for ~30days. 

I'm working on SlaveMode now. I am confusing myself with spinlock's and Mutex's between the ISR and foreground apps.   Currently I'm almost up to the point to stimulate the hardware and see how it twitches.  Hopefully my prognostication will match reality. 

Chuck.

rmetzner49:
Not to worry, the only thing the i2C will be doing is the AD7993 and the DS3232 RTC and I've already had both working.  On the other hand, my need for a good A/D will be gone, so I may revive the ESP32's on board.  In theory, I could also use the ESP32's RTC but I didn't see much info on how it runs on battery or any examples of an RTC using the internal.   I successfully used i2c talking to the 32U4 using Nick Gannon's messaging method as explained on his site.  Most of my pre-retirement work was Hardware, signal conditioning and motor control.  We had a whole group that wrote the main apps in Windows VB, so I'm really green at coding "C".  On the flip side, no one in the VB group knew anything about embedded coding so I eventually ended up writing some drivers.

stickbreaker:
@rmetzner49 sounds good, EXCEPT the ESP32 RTC stands for Real Time CoProcessor.  It is lowpower lowspeed cpu.  It could be programmed to act as a Real Time Clock, but you will have to add a 32KHz accurate clock, and code.

Chuck.

rmetzner49:
Yeah, that’s not very attractive.  For what little it costs, the DS3232 is the better choice, though even with that, I had to “roll my own” since no one had.  I think I used the DS1307 as an  example and just massaged the appropriate registers.

Out of curiosity, which Dev System do you use?  (More than likely not Arduino).  Which ESP32 Board?  I started with Sparkfun’s “Thing” and moved on to the less expensive “VROOM” boards.  I flip between Arduino and Visual Studio ported to Arduino.  Obviously the caveat being if someone changes a Core or Library, I’m at their mercy.  Then you also have to watch whose stuff you grab, because some of it just flat doesn’t work.  @me-no-dev seems reliable as is Stoffregen but he is not “open” with his source or Hex files.  You can really see where some people’s head is at when you see how differently they approach TIME stuff like BinToBCD and BCDtoBin on clock libraries.  

 

Bob Metzner

stickbreaker:
@rmetzner49 I just use the Arduino environment 1.8.5 with NotePad++ as an external editor.  So I just edit my code in NotePad++, save it and touch  Arduino checkmark or Right Arrow to compile and download. Currently I'm using a couple of clone WeMos Bluetooth + battery boards. 

I have a messy I2C bus (18" of 0.050" ribbon connecting 6 24LCxx EEproms, a DS1307, 2 MCP23008 (one as a 4x4 keypad driver, one driving a 20x4 LCD using a modified LiquidCrystal library), And of course a UNO.  I'm using the UNO as an I2C Master to poke the ESP32 as a slave. 

A couple of DS18b20 as temp sensors.(I had to modify the OneWire.h library to make it work on the ESP32);

A Micro SD card directly connected as the SD.h readme.md shows. (but it is unstable right now).

Nothing Special.  After I get I2C stable I'm planning on embedding ESP-VRoom-32 modules on some custom boards.

Chuck.

rmetzner49:
Chuck;

 

I’m also using the VROOM module on my final board.  Using PCB123 from Sunstone because it’s FREE.  I’ve road tested other Board Layout Software but decided that NONE of them have a decent autorouter, so I do all my routing by hand.

 

Of course as you know the kicker with ANY such package is you have to get really good at making your own library parts.  Even though they tie into DigiKey, anything newer than about 2 years you have to make yourself.  You’re probably using Eagle.  I’ve made a VROOM32 block for PCB123 if that’s what you use.

 

Bob

rmetzner49:
I also wanted to mention the display is from buydisplay.com which sells on fleaBay.  Adafruit’s RA8875 libraries worked out of the box with that.  Although I’ve zapped quite a few modules because they didn’t put a buffer on the 8875’s lines and I think I didn’t fully understand the power sequencing issue when running things at 3.3V.  I ripped the 3.3V regulator off the DEV board so I can run at 3.3V since then.  So I went out and bought the Adafruit board, plugged the flat cable into THEIR display and BINGO.  I’m going to leave it on there before I buy another display, since unbuffered, it’s not a robust game.

 

Bob M.

stickbreaker:
@rmetzner49 Yea I use Eagle 7.4 Not gonna upgrade to monthly payments. When AutoCad bought them out they changed licensing styles.  I have been looking at KiCAD..  I have used IteadStudios (china) to manufacture PCB, a 10cm x 10cm 2 layer Qty10 costs about $35 with shipping(3 to 4 weeks).
Chuck.

dpharris:
Chuck et al. --

Did your (significant) upgrade to I2C result in multimaster capability --
ie, if a master-send loses arbitration does it fall back to slave-receive
mode?

Thanks for all your work,
David


On Jan 16, 2018 08:52, "chuck todd" <notifications@github.com> wrote:

@rmetzner49 <https://github.com/rmetzner49> Yea I use Eagle 7.4 Not gonna
upgrade to monthly payments. When AutoCad bought them out they changed
licensing styles. I have been looking at KiCAD.. I have used IteadStudios
(china) to manufacture PCB, a 10cm x 10cm 2 layer Qty10 costs about $35
with shipping(3 to 4 weeks).
Chuck.

—

You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/espressif/arduino-esp32/pull/874#issuecomment-358027818>,
or mute the thread

stickbreaker:
@dpharris No, Currently it only support Master Mode communications.  I am working on Slave functions.  Multiprocessor, multi thread, interrupt servicing, callbacks, spinlocks, mutex's Have me running in circles with my hair on fire! 

I'm having design issues with deadlocking between the ISR and foreground tasks modifying buffers.
Here is my current dilemma:
* SLAVE ISR triggered on Core0, which will add characters to a buffer thru a pointer and offset + count perhaps dispatch the completed transaction via Callback()
* Core1 in routine to add Master transaction to work queue, this addition causing a reallocate() expansion of the buffer control array.  This reallocate() ends up moving the buffer control block to a new position in memory.
* Can I assume `xPortEnter_CriticalfromISR()` will freeze the other Core?
* Can I assume `xPortEnter_Critical()` will freeze the already entered ISR on the Other Core?
 
My problem is that I can see no way to test these edge condition because they are so timing critical. The ISR has to accessing the buffer control block while the other Core is reallocating the control block! I am tending towards a two level control structure, the Active structure is only modified when the ISR Is pinned down, or maybe the ISR manages the Staged to Active merging.  The Slave Mode I2C on the Arduino co-exists with Master Mode I2C transparently.  That is what I am trying to emulate. 

Chuck.

rmetzner49:
Hey, how’s it going?  

 

So I finally got my TDC1000 and TDC7200 objects working after “Arduinoizing” Lindholm’s code and figuring out in which CPP file to create the Instance in so it would compile.  Obviously also my Hardware is way different; I ran out of I/O pins so I stuck a PCF8574 on there as well.  I now have BOTH buses going on including the Display’s MISO bug you helped me with.

 

(This runs on SPI every ½ second):

 

TDC1000.regWrite(0x00, 0xAA);

TDC7200.regWrite(0x01, 0x55);

Serial.println("========================");  

uint8_t temp1 = TDC1000.regRead8(0x00); 

uint8_t temp2 = TDC7200.regRead8(0x01); 

uint32_t temp3=TDC1000.readClockFrequencyIn();

Serial.print("ClockFrequencyIn: ");  

Serial.println(temp3);  

Serial.println("==========================");  

 

Printout:

========================

TDC1000_/CS:  15    Register:  0   Read Data  AA

TDC7200_/CS:  27    Register:  1   Read Back  55

ClockFrequencyIn: 0

 

I2C is working great:  I check for every device during setup:

 

Waiting for I2C...

AD7994  A/D should be at 0x21 

PCF8574 I/O should be at 0x38 

DS3232  RTC should be at 0x68 

Scanning I2C Bus for Devices ...

I2C device found at address 0x21 or 33 DEC  !

I2C device found at address 0x38 or 56 DEC  !

I2C device found at address 0x68 or 104 DEC  !

Done

 

Take Care,

Bob Metzner

stickbreaker:
@rmetzner49 I haven't done much with SPI, your code looks valid, I just reread this thread. 
Back at the beginning, the first issue I brought up was that MISO might not be tri-stating. 
But, My recommendation of a '125 buffer caused failures.

How did you get it to work?

I think a '125 with a weak 10k .. 100k pullup on MISO should work.  the pullup will keep a float pin oscillation from happening. 

Chuck.

stickbreaker:
@rmetzner49 Ya, Stopping and Starting should work, as long as your CS pins stay disabled during the switch.  The devices won't know any difference.

Chuck.

rmetzner49:
Such a bitchy thing this RA8875 Display Chip.  Reading some more of the Blogs caused me to try something else:  I ended up putting the 74LVC124 on all THREE pertinent signals; MISO, MOSI and CLK.  When I looked at the signals more closely and putting the MAX35101 in a tight “write_reg”  followed by a “read_reg”  1mS later, something from the 8875 was hanging on to MISO during the “0” data state.

 

So from the ESP32 to the RA8875 connect the input of the LVC125 to the CPU and the output to the RA8875 on the CLK and MOSI lines.

Connect the input of the third section to MISO of the RA8875 and the output to the CPU and it started working as expected.

 

You were so helpful early on, I just thought I’d share with you in case you have others frustrated with sharing an SPI channel with the RA8875.

 

Best Regards,

Bob Metzner

stickbreaker:
@rmetzner49 Nice to see you got it to work.  Isn't it nice :grimacing: when "interface standards" are adhered to?

Chuck.

PR #875, opened 2017-11-25
Sync flags with platform.txt; Custom partitions per board

ivankravets:
Do you have ideas why `-fstack-protector` generates a warning? Should be it removed from platform.txt?

```
cores/esp32/IPv6Address.cpp:1:0: warning: -fstack-protector not supported for this target
```


me-no-dev:
-fstack-protector is a new feature. Are you running the latest toolchain?

ivankravets:
You are right, the problem was with toolchain. Please merge, I'm working on updating of toolchains.

PR #881, opened 2017-11-26
Minor fixes for PlatformIO

ivankravets:


cyberman54:
How do i get this fix in my platformIO environment? 
Need it urgently, because my Windows10 username contains a blank :-(

ivankravets:
@cyberman54 Thanks! Fixed in https://github.com/espressif/arduino-esp32/commit/7fa8cafb27a1e28e3da74c5d44d4b0c50a252f05

Please switch to staging version http://docs.platformio.org/en/latest/platforms/espressif32.html#using-arduino-framework-with-staging-version

PR #890, opened 2017-11-29
implement FSPreferences, VSF (SD, SD_MMC, SPIFFS) compatible 'Preferences-lib' alternative

Curclamas:
## Background
Preferences.h lib is very useful but has some shortcomings. To name two: limited space and likely limited wear leveling. When we are able to use 8GB or 16GB SD-Cards space and wear leveling become much less of a problem.

## Possible side-effects
None that I know, it's a separate library that maps the Preferences functions on top of VSF API.

## What to test
Different memory devices should be tested, I confirmed it with 8GB and 16GB cards as SD and SD_MMC.

## Where to start reviewing
First look at the constructors and destructors. Then one should definitely look at the private functions:
```cpp
        size_t putBytes(const char* key, uint8_t type, const void* value, size_t len);
        size_t getBytes(const char* key, uint8_t type, void* buf, size_t maxLen = 0, size_t minLen = 0);     
```
These are used to write to the VFS API, the other functions merely call these with different type/length params.

One exception is:
```arduino
String getString(const char* key, String defaultValue)
```

because I wanted to avoid unnecessary double buffering. The double buffering is used in the other functions so that we only write to the provided buffer/pointer when everything was read successfully. This could lead to stack overflow when reading very large chunks of data. To mitigate this, one could also implement a non-buffered method in the future.

### DIFF
--- a/libraries/FSPreferences/examples/Showcase/Showcase.ino
+++ b/libraries/FSPreferences/examples/Showcase/Showcase.ino
@@ -1,0 +1,64 @@
+/*
+ ESP32 FSPreferences showcase
+
+ This is a simple demonstration of different features of the FSPreferences library
+
+ created for arduino-esp32
+ by Frederik Merz (Curclamas) <Frederik.merz@novalight.de>
+*/
+
+#include <SD.h>
+#include <SD_MMC.h>
+#include <SPIFFS.h>
+
+#include <FSPreferences.h>
+
+/* Select your FS */
+// #define MY_FS SD
+#define MY_FS SD_MMC
+// #define MY_FS SPIFFS
+
+FSPreferences preferences(MY_FS);
+
+void setup() {
+  Serial.begin(115200);
+  Serial.println();
+
+  MY_FS.begin();
+
+  preferences.begin("my-app1", false);
+  preferences.putInt("myInt", 1337);
+  preferences.putBool("myBool", true);
+  preferences.putBool("myBool2", true);
+  preferences.remove("myBool2");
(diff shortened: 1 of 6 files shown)


igrr:
Not a maintainer of this repo, just bringing some context from ESP-IDF... As you know, Preferences is built on top of NVS, and NVS APIs so far only supported flash as storage. We are starting some work to allow other types of storage as backends for NVS. So instead of `nvs_flash_open` there could be e.g. `nvs_file_open` which would map to anything that can be opened using `open` call. `nvs_*_open` would also return a handle to the NVS instance (at the moment the instance is global), which one would pass to `nvs_open`, allowing for multiple NVS instances of different type to exist in the same device. Theoretically, someone could then use this new functionality in Preferences library, to let it target different kinds of storage.

Curclamas:
@igrr that sounds quite interesting! One thing though: imagine we would pass a fatfs (or a fatfs file, or SD handle?) to nvs, how would the actual data look on the SD card, respectively on its file system? Do we have a file mynvs.bin and that cointains all of the stored nvs stuff in a similar fashion like its done on the SPI Flash nvs partition?

Curclamas:
One more remark: if you have a look at my proposal you see that the structure of the preferences is nicely visible by inspecting the folders/files, while your proposal sounds rather "monolithic" if I got it correctly. @me-no-dev what would be your opinion on that?

igrr:
It would not have to be a monolithic file (nvs.bin), but could be a human-readable text format. From logistics perspective, we would probably use the same CSV format as the one used by NVS binary generator tool from IDF.

PR #897, opened 2017-12-03
Allow PSK instead of passphrase in WiFiSTA::begin

mascii:
Allow PSK, which generated by `wpa_passphrase` command or the PBKDF2 function.

I referred to the following the PR: https://github.com/esp8266/Arduino/pull/1850

> In WPA protocol, the maximum length of the passphrases is 64 characters in order to distinguish them from the actual PSK which is 64 ASCII characters long, so in most systems if a 64 chars string is passed, it is assumed to be a PSK, otherwise is treated as a passphrase and is used to compute the PSK.

To calc the PSK, please use http://jorisvr.nl/wpapsk.html


me-no-dev:
Thanks :)

PR #934, opened 2017-12-19
Ensure errors are generated if the wrong method is called to retrieve the SSL data stream.

lonerzzz:
Currently, the getStream() function on the httpClient will return the stream for the non-SSL without generating errors and this will cause problems when someone tries to read the stream as there is nothing to read. The changes check whether the SSL mode is in use or not and generate an error if the method is called while SSL is in use.


lonerzzz:
@me-no-dev Thank you.

PR #941, opened 2017-12-22
Rename package_esp32_index.template.json to package_esp32_index.json

olicooper:
I am trying to use the 'Visual Micro' plugin for Visual Studio and it complain that this file has the 'wrong filename' because of the ".template" bit. Is it possible to change it  so I can use it? And maybe upload a link to this site: https://github.com/arduino/Arduino/wiki/Unofficial-list-of-3rd-party-boards-support-urls so everyone knows you can use Arduino for ESP32 too ?


olicooper:
To get it working properly with Visual Micro, I had to make a number of changes within the file itself too. Changes can be found here: https://github.com/olicooper/arduino-esp32/commit/0131aae69fb2f14a7c45e86260afeb153b8f7aa6

me-no-dev:
yup that is why this is just a template :)

PR #945, opened 2017-12-27
WiFiClientSecure::lastError() method

lbernstone:
Added a lastError method to WiFiClientSecure so you can present errors to end users for certificate troubleshooting.


copercini:
Great stuff ;D

PR #946, opened 2017-12-28
Update windows.md

vppillai:
adding an additional step that is required to make sure the BLE libs are synced


me-no-dev:
thanks!

PR #948, opened 2017-12-28
Add progress callback to Update::writeStream().

everslick:
This is a more or less verbatim port from the ArduinoOTA library to the UpdateClass. Usage example:

```
  Update.onProgress(
    [](size_t current, size_t total) {
      Serial.printf("\r  received %d of %d bytes           ", current, total);
    }
  );
```
Very useful for possible inclusion in the HttpUpdate library.


me-no-dev:
nice!

PR #958, opened 2017-12-31
tools: Update get.py to work with python 3

patrickelectric:
Python3 use unicode as default


patrickelectric:
Working on ArchLinux

me-no-dev:
```
$ python get.py
Traceback (most recent call last):
  File "get.py", line 27, in <module>
    file_name = null
NameError: name 'null' is not defined
```

PR #1011, opened 2018-01-19
DNS Server : bug fix and prettifying

Loufylouf:
I had problem with the last version of the DNS server lib, there were lots of malformed packets on Wireshark and I could not get the library to work. So I checked what was done in the ESP8266 library, integrated some of it and prettified a bit the whole thing. Everything should be a little clearer for someone trying to integrate new functionalities. Tested it with success with WiFiManager.

### DIFF
--- a/libraries/DNSServer/src/DNSServer.cpp
+++ b/libraries/DNSServer/src/DNSServer.cpp
@@ -60,6 +60,8 @@ void DNSServer::processNextRequest()
     if (_buffer == NULL) return;
     _udp.read(_buffer, _currentPacketSize);
     _dnsHeader = (DNSHeader*) _buffer;
+    _type       = (uint16_t) ( (_buffer[_currentPacketSize - 4] << 8) + _buffer[_currentPacketSize - 3] ) ; 
+    _class      = (uint16_t) ( (_buffer[_currentPacketSize - 2] << 8) + _buffer[_currentPacketSize - 1] ) ; 
 
     if (_dnsHeader->QR == DNS_QR_QUERY &&
         _dnsHeader->OPCode == DNS_OPCODE_QUERY &&
--- a/libraries/DNSServer/src/DNSServer.cpp
+++ b/libraries/DNSServer/src/DNSServer.cpp
@@ -5,7 +5,7 @@
 
 DNSServer::DNSServer()
 {
-  _ttl = htonl(60);
+  _ttl = htonl(DNS_DEFAULT_TTL);
   _errorReplyCode = DNSReplyCode::NonExistentDomain;
   _dnsHeader = NULL;
   _buffer = NULL;
--- a/libraries/DNSServer/src/DNSServer.cpp
+++ b/libraries/DNSServer/src/DNSServer.cpp
@@ -123,14 +123,29 @@ void DNSServer::replyWithIP()
   if (_buffer == NULL) return;
   _dnsHeader->QR = DNS_QR_RESPONSE;
   _dnsHeader->ANCount = _dnsHeader->QDCount;
-  _dnsHeader->QDCount = 0;
+  //_dnsHeader->QDCount = 0;
 
   _udp.beginPacket(_udp.remoteIP(), _udp.remotePort());
   _udp.write(_buffer, _currentPacketSize);
+
+  // Use DNS name compression : instead of repeating the name in this RNAME occurence,
+  // set the two MSB of the byte corresponding normally to the length to 1. The following
+  // 14 bits must be used to specify the offset of the domain name in the message 
+  // (<255 here so the first byte has the 6 LSB at 0) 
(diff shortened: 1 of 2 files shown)


me-no-dev:
uhmm I do not really agree. here is what happens in the code:
- packet is received and checked and domain name is parsed
- if we use wildcard domain or domain matches what we have in settings, continue
- start new packet and copy the contents of the received one
- change Qcount to 0 and Acount to 1 to show that there is one answer in the packet
- add answer to the end of the packet (after the question)

Now this is all wrong!
Because the question structure is just like the starting of an answer, I just add TTL and data to the end of it, which makes it properly formatted answer.

With the code you included, it did not work for me and nslookup reported malformed packet

Loufylouf:
I beg to differ :) 
According to the spec (https://tools.ietf.org/pdf/rfc1035.pdf), on page 25, we have the following format for a DNS message : 

> +---------------------+
>  | Header |
>  +---------------------+
>  | Question | the question for the name server
>  +---------------------+
>  | Answer | RRs answering the question
>  +---------------------+
>  | Authority | RRs pointing toward an authority
>  +---------------------+
>  | Additionall | RRs holding additional information
>  +---------------------+
So it seems that we agree on the first two, the header and the question. The thing is, you stop there, add a TTL and some data. Okay, but what does the spec say ? Straight to page 29 : 
> The answer, authority, and additional sections all share the same
> format: a variable number of resource records, where the number of
> records is specified in the corresponding count field in the header.
> Each resource record has the following format:
> 
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
>  / NAME /
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
>  | TYPE |
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
>  | CLASS |
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
>  | TTL |
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
>  | RDLENGTH |
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--|
>  / RDATA /
>  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
So, in the reply, if you just copy the question, then add the TTL, RDLength and RData, you just miss 3 fields : the name, the type and the class. You could copy the question again, but to have a complete message, you should, all things considered, copy the question twice instead of once like you do.

My version, which is also the ESP8266 version, use the complete format, with a little trick named message compression described on page 30 (and in my code). I haven't seen anywhere your solution consisting of you just answering with the question padded with the TTL, RDLength and RData, so if you have any reference for that, I'd gladly read it.

BillyNate:
Hi me-no-dev & Loufylouf,

I've checked out both the official master and Loufylouf's.

On my devices (one smartphone and a windows desktop) the captive portal _does not work_ when I'm on `espressif:master`, but _does work_ when I'm on `Loufylouf:master`.
I did not do any Wiresharking, but did notice different results when doing an `nslookup google.com`.

I feel it is safe to say that the captive portal / DNSServer does not fully work for all devices connected to the ESP, and it might be worth looking into this a bit more 😉

Loufylouf:
@BillyNate Can you be a bit more specific on the last part of your message ? Did my version work on all of your devices ? Or are you referring to the test of me-no-dev with my version ?

BillyNate:
I added your repository to my remotes, fetched and pulled your `master`. Once pulled the DNS started to work properly on both devices I tested on.

Does this make it more clear?

Loufylouf:
Yeah thanks ! 

@me-no-dev I now get what you're doing ! I overlooked the Qcount = 0 part. The thing is, while browsing normally on my computer and wiresharking the DNS messages, all query responses carry the question as well ("dns && dns.count.queries == 0" filter returns nothing), so while I agree your solution should be working (after all, there's an ID to match an answer with a question), this does not seem to be the case at least for me (on Ubuntu and Android) and @BillyNate (windows desktop) .

me-no-dev:
I think the best option here is to create a new answer and do not copy stuff over ;) that whole copy/paste thing sucks :D

Loufylouf:
Well, in my code, that's exactly what is done for the answer. Since the format of the question doesn't change, I don't see the problem copy-pasting it. Well, except for the part where we modify the DNS header and then paste the whole buffer, that can be a bit unsettling at first.

tablatronix:
#1037

Loufylouf:
I've added a commit that creates the DNS question structure and build the response from scratch instead of a dirty copy-paste

tablatronix:
@me-no-dev have you had a chance to look this over or document whatever issues you were having?
I would really like to at least see captive portal problems resolved. I could not find ant packet issues with osx, maybe you can detail reproducing your dns issues? What kind of implementation?

PR #1034, opened 2018-01-25
Added EEPROMClass and Examples

fedy0:
[1] Use multiple sector flash partitions defined by the name column in the partition table
[2] Added 'eeprom_class' example to show how to use the 'EEPROMClass'
[3] Included additional methods to the EEPROM library
[4] Added 'eeprom_extra' example to show how to use the additional methods
[5] Included in eeprom_class example how to use ESP32 Partition Manager [https://github.com/francis94c/ESP32Partitions](url)

Thanks for the opportunity to contribute in this wonderful repo :)


eman6576:
@fedy0 Could an `update` method also be added that mirrors what the Arduino library has? Also, could this get merged soon? 😀

fedy0:
@eman6576 
The 'update' method's function as used in Arduino library has been taken care of within the EEPROM 'begin', 'write' & 'commit' methods of esp32. See code snippet below used within the write method to implement update.

// Optimise _dirty. Only flagged if data written is different.
  uint8_t* pData = &_data[address];
  if (*pData != value)
  {
    *pData = value;
    _dirty = true;
  }

When 'begin' method is called, it loads into the RAM (_data) the EEPROM memory block whose size is specified by its argument (max_size 4kB ~ 0x1000 or 4096)

Also when the 'commit' method is called, it performs the actual write to EEPROM memory if there is a change (ie. _dirty == true) in the buffered _data array in the RAM; thus reducing write cycles. Introducing an explicit update or even within the commit method would impact the time of executing a commit. This implementation would only be justified if commit method is used sparingly. I'd work on that and I hope it maybe helpful. 

For now, to prolong the lifespan of the flash sector apportioned to the EEPROM, one calls commit() within the program only when necessary

Note: Without a call to commit and esp32 looses power supply, all data buffered in RAM(_data) is lost. So for an application that depends on one (battery) or two supplies (regular PSU & battery), any sense in power outage, a commit should be performed or before deep-sleep a commit should be performed too.

fedy0:
In the example 'eeprom_class.ino' read the introduction comments on how to use it.

Typically, you have to
install (download zip, extract, copy and paste inside '.../Documents/Arduino/tools)
this tool
(https://github.com/francis94c/ESP32Partitions)
ONCE in Arduino IDE to help you manage your partitions.

Also see it's ReadMe

If you don't have a 'tools' folder in your '.../Documents/Arduino/' directory, create one first then install the tool inside that folder. Restart your Arduino IDE and navigate to 'Tools' menu to access the 'esp32 partition manager'.

@vorrias thanks for the compliments and sorry for the late reply

Issue #1037, opened 2018-01-27
captiveportal example not working on some platforms


tablatronix:
### Hardware:
Board:							doit v1, ESP32 Dev Module
Core Installation/update date:			today
IDE name:							platformio
Flash Frequency:					?40Mhz?
Upload Speed:						?115200?

### Description:
using captiveportal example
No captive portal launches on windows, linux, android.
surprisingly it works on osx and ios.

Testato:
Bug confirmed here on Arduino IDE also

tablatronix:
I see lots of malformed dns packets in wireshark.

```
iturd-Pro:tools shawn$ dig google.com
;; Warning: Message parser reports malformed message packet.

; <<>> DiG 9.9.7-P3 <<>> google.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 42797
;; flags: qr rd ad; QUERY: 0, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1
;; WARNING: recursion requested but not available

;; Query time: 8 msec
;; SERVER: 192.168.4.1#53(192.168.4.1)
;; WHEN: Sun Jan 28 18:09:06 CST 2018
;; MSG SIZE  rcvd: 12
```

`[Expert Info (Error/Malformed): Malformed Packet (Exception occurred)]`

tablatronix:
`Source mac must not be a group address` and lg and ig bits present, does that sound right?

tablatronix:
Looks like dig always fails, since it sends additional OPT every time, I noticed OPT and PTRs fail.

Ill look into this again tomorrow and new machines and see what I see.

Testato:
Probably i found the problem:
The DNSserver redirect all to 192.168.1.1, instead of the captive portal IP (192.168.4.1)

![capture](https://user-images.githubusercontent.com/6893841/35515760-4916585e-050a-11e8-8db5-8737144601c6.JPG)

![capture2](https://user-images.githubusercontent.com/6893841/35515766-4cf8dd5c-050a-11e8-96a1-3ec850846f5a.JPG)

Log made on Win10

tablatronix:
Are you using the captiveportal example, it uses that address instead.

tablatronix:
This fixes it, the dns repsones are not including the question along with the answer.
This is from esp8266, not sure why @me-no-dev left it out.

DNSServer.cpp
```
[...]

PR #1045, opened 2018-01-29
fix captive portal #1037

tablatronix:
Not sure if this was left out on purpose or not


tablatronix:
❗️ I have no idea if this is correct, I just copied from esp8266 to get it working

me-no-dev:
already implemented :)

Issue #339, opened 2017-04-29
Sketch too big, but identical sources are compiling and flashing under esp-idf environment


croisez:
I am writing a program which uses following capabilities of the ESP32: Blufi protocol, reception of BLE advertisings, HTTPS post (WiFiClientSecure), and WiFi.

I have written such program using Eclipse/esp-idf and it is compiling/flashing fine.
I wanted to try the Arduino way, and I am stuck with size of sketch too big error.

<PRE>
Multiple libraries were found for "WiFi.h"
 Used: /Users/lmc/Dropbox/arduino/hardware/espressif/esp32/libraries/WiFi
 Not used: /Applications/Arduino.app/Contents/Java/libraries/WiFi
Using library Blufi at version 1.0 in folder: /Users/lmc/Dropbox/arduino/hardware/espressif/esp32/libraries/Blufi 
Using library WiFiClientSecure at version 1.0 in folder: /Users/lmc/Dropbox/arduino/hardware/espressif/esp32/libraries/WiFiClientSecure 
Using library WiFi at version 1.0 in folder: /Users/lmc/Dropbox/arduino/hardware/espressif/esp32/libraries/WiFi 

Sketch uses 1,095,399 bytes (104%) of program storage space. Maximum is 1,044,464 bytes.
Global variables use 70,684 bytes (23%) of dynamic memory, leaving 224,228 bytes for local variables. Maximum is 294,912 bytes.
processing.app.debug.RunnerException: Sketch too big; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing it.
</PRE>

The functionalities are very identical, I don't include unused #includes.
As concern Blufi, I have written a lib for ESP32-Arduino from the Blufi example found in esp-idf.

How could I workaround this error and make my sketch flashable?

me-no-dev:
bluetooth and wifi use a whole lot of flash and in Arduino's partition config there are 2 sketch partitions 1MB each. I guess you went over that, which means that it will step into the next partition. I've seen report before and I have hit the limit myself. Taking more space for sketch means that it will be twice that much less available for SPIFFS when it comes. I might try to 1.25MB and have 0.5MB less SPIFFS

torntrousers:
I have the same problem when using BLE and WiFi. Is tweaking the partition size something simple to do which I can try locally?

me-no-dev:
Here is the partition definition: https://github.com/espressif/arduino-esp32/blob/master/tools/partitions/default.csv
[...]

PR #1052, opened 2018-01-31
Fix upload.maximum_size as current partition size 0x140000 for esp32vn-iot-uno & heltec boards

erikwilson:
It appears that all boards which use the current partition size of 0x140000 for apps should also have upload.maximum_size=1310720 in boards.txt

Also see: https://github.com/espressif/arduino-esp32/issues/339


me-no-dev:
cool :) Thanks for the contribution!

PR #1056, opened 2018-01-31
TwoWire: add method setBus

cspiel1:
This was necessary for getting i2c master mode working. More concretely I used Adafruit_BME280 and Adafruit_SSD1306 from Sming library, which are working when setting:

`Wire.setBus(1);`


me-no-dev:
Hi :)

You have included also a few Adafruit libraries and I do not think that this is what you wanted. Also would you please argument the need for that new method? Each bus is intended to have it's own class and not to be changeable after.

cspiel1:
On Sun 04.03.18  19:33, Me No Dev wrote:
> Hi :)
> 
> You have included also a few Adafruit libraries and I do not think that this is what you wanted. Also would you please argument the need for that new method? Each bus is intended to have it's own class and not to be changeable after.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/espressif/arduino-esp32/pull/1056#issuecomment-370256650

Indeed, the pull-request was only meant for commit "TwoWire: add method setBus"
4f66f97. Sorry for the following commits!

What do you mean with "it's own class"? The class is TwoWire and there is an
object Wire of this class.

The librariesAdafruit_SSD1306 and Adafruit_BME280_Library use the object Wire.
If I don't want to make a code change in these I have to change the bus number
of the object Wire. That's the purpose of method setBus.

regards,
chris.

stickbreaker:
@cspiel1 The `Wire()` class is created in Wire.cpp with:
```c++
TwoWire Wire = TwoWire(0);
```
if you want to access the second i2c hardware peripheral, create a new object:
```c++
TwoWire Wire1 = TwoWire(1);
```
* note: this won't work with my library yet, stickbreaker\arduino-esp32 the released version only handles the first hardware peripheral.

Chuck.

cspiel1:
On Mon 05.03.18  11:17, chuck todd wrote:
> @cspiel1 The `Wire()` class is created in Wire.cpp with:
> ```c++
> TwoWire Wire = TwoWire(0);

TwoWire is the class. Wire is the object, an instance of TwoWire.


> ```
> if you want to access the second i2c hardware peripheral, create a new object:
> ```c++
> TwoWire Wire1 = TwoWire(1);

Yes. You are right for Adafruit_BME280_Library. One could use the method

bool Adafruit_BME280::begin(TwoWire *theWire)

and pass e.g. your Wire1 object. But Adafruit_SSD1306 uses the object Wire
directly. So, there is no chance to replace it by another TwoWire object. So
why shouldn't setBus added? Will it break something?

regards, chris.

lonerzzz:
My 2 cents in playing with the original code internals is that the classes aren't set up for a dynamic change of bus and changing the bus would create operating issues. If you were setting the bus, you would have to ensure a reset of all internal states within the method. I am not sure how or if @stickbreaker intends to address it in his code.

stickbreaker:
@cspiel1 the bus value is used to setup the interrupts, register access, buffers.  To achieve what you want. `TwoWire()` needs a destructor.  Then you could:
```c++
Wire.~TwoWire();
Wire=TwoWire(1);
Wire.begin(scl,sda);
```

Currently the `TwoWire()` object does not contain a custom destructor, so if you used this code, the Interrupt, memory buffers, peripheral would still be configure.

Implementing a custom destructor is not near the top of my to-do list.  

You can just modify your copy of Wire.cpp and change the `Wire()` object initialization.
  
Chuck.

cspiel1:
On Mon 05.03.18  20:46, chuck todd wrote:
> @cspiel1 the bus value is used to setup the interrupts, register access, buffers.  To achieve what you want. `TwoWire()` needs a destructor.  Then you could:
> ```c++
> Wire.~TwoWire();
> Wire=TwoWire(1);
> Wire.begin(scl,sda);
> ```
> 
> Currently the `TwoWire()` object does not contain a custom destructor, so if you used this code, the Interrupt, memory buffers, peripheral would still be configure.
> 
> Implementing a custom destructor is not near the top of my to-do list.  
> 
> You can just modify your copy of Wire.cpp and change the `Wire()` object initialization.
>   
> Chuck. 

Ok. Then I understood why setBus method is a bad idea. I will have to fix
the Adafruit library.

regards,
chris.

me-no-dev:
Great :) closing this now :) I will look into such option though

PR #1059, opened 2018-02-01
FIX: getEfuseMac can return random data

tferrin:
esp_efuse_mac_get_default() only stores into the bottom 6 bytes of it's 8-byte argument, so must initialize the arg to zero to avoid trash on the stack.


me-no-dev:
nice find :) Thanks for the contribution!

Issue #1061, opened 2018-02-01
[arduino-esp32 as component in idf hello world] one character 0xFF allways receive and sucks


ESP32DE:
hi

after a few time i try esp32-arduino core  
( last actually version with 490 commits ) with ESP-IDF ( last ESP-IDF version with 4630 commits )
[as component](https://github.com/espressif/arduino-esp32/blob/master/docs/esp-idf_component.md) for a fun project

i noted that i get allways some rubbish character after boot
the code is simple:

```
//file: main.cpp
#include "Arduino.h"

void setup(){
  Serial.begin(115200);
}

void loop(){
  Serial.println("loop");
  delay(1000);
}
```

the output:
![grafik](https://user-images.githubusercontent.com/16070445/35678853-37c80d7a-0755-11e8-9737-820659793e9e.png)

i change the code then with a delay for enough time give for setup the peripherals

```
//file: main.cpp
#include "Arduino.h"

void setup(){
  Serial.begin(115200);
  delay(500);
}

void loop(){
  Serial.println("loop");
  delay(1000);
}
```

the output is little better only one character on begin sucks 

![grafik](https://user-images.githubusercontent.com/16070445/35679066-0506742a-0756-11e8-93fb-52f7e9903faf.png)

i try then few things but this one character goes not away

```
//file: main.cpp
#include "Arduino.h"

void setup(){
  Serial.begin(115200);
   delay(5000);
}

void loop(){

  delay(5000);
  while (-1) {
   Serial.println("loop");
   delay(1000);
  }
} 
```

but this character sucks allways again

![grafik](https://user-images.githubusercontent.com/16070445/35679212-81b01b66-0756-11e8-814a-d52ba1093ca3.png)

![grafik](https://user-images.githubusercontent.com/16070445/35679237-9246378a-0756-11e8-8093-b1f02761ecd2.png)

the character is 0xFF ( on baud 115200 ) 

![grafik](https://user-images.githubusercontent.com/16070445/35679286-bdb9a564-0756-11e8-9214-f993e52e260d.png)

from where this character comes?
which task has this sign?
why is it here?
can we eliminate it?

thanks
best wishes

me-no-dev:
ughhh... things like that are usually caused by unstable UART lines that detect fake data if there is some voltage fluctuation on the pins (you can start UART1 on other pins and try touching them. you will see the same thing to happen).
What is the UART chip on that board? What board is that?

ESP32DE:
:)

[...]

PR #1065, opened 2018-02-02
bugfix(uart + i2c): fix signal glitch on tx , i2c sda and i2c scl pin

ESP32DE:
see https://github.com/espressif/arduino-esp32/issues/1061

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -73,7 +73,8 @@ i2c_err_t i2cAttachSCL(i2c_t * i2c, int8_t scl)
         return I2C_ERROR_DEV;
     }
     digitalWrite(scl, HIGH); // optional, tested successful in actually version 
-    pinMode(scl, OUTPUT_OPEN_DRAIN | PULLUP);
+    // pinMode(scl, OUTPUT_OPEN_DRAIN | PULLUP);
+    pinMode(scl, OPEN_DRAIN | PULLUP); // open for a test on weekend
     // digitalWrite(scl, HIGH); // successful tested in a prev version, but fails in the actually so use optional!
     pinMatrixOutAttach(scl, I2C_SCL_IDX(i2c->num), false, false);
     pinMatrixInAttach(scl, I2C_SCL_IDX(i2c->num), false);
@@ -97,7 +98,8 @@ i2c_err_t i2cAttachSDA(i2c_t * i2c, int8_t sda)
         return I2C_ERROR_DEV;
     }
     digitalWrite(sda, HIGH); // optional, tested successful in actually version
-    pinMode(sda, OUTPUT_OPEN_DRAIN | PULLUP);
+    // pinMode(sda, OUTPUT_OPEN_DRAIN | PULLUP);
+    pinMode(sda, OPEN_DRAIN | PULLUP); // open for a test on weekend
     // digitalWrite(sda, HIGH); // successful tested in a prev version, but fails in the actually so use optional!
     pinMatrixOutAttach(sda, I2C_SDA_IDX(i2c->num), false, false);
     pinMatrixInAttach(sda, I2C_SDA_IDX(i2c->num), false);
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -72,7 +72,9 @@ i2c_err_t i2cAttachSCL(i2c_t * i2c, int8_t scl)
     if(i2c == NULL){
         return I2C_ERROR_DEV;
     }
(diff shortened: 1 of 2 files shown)


stickbreaker:
@ESP32DE look at my comments in your pull request on my repo [stickbreaker/arduino-esp32 Pull 19](https://github.com/stickbreaker/arduino-esp32/pull/19).

If you make the same changes to this pull, it will solve the I2C begin Glitch also.

Chuck.

lonerzzz:
Good find. Is this something worth pushing to the IDF folks in case it has not been addressed there?

stickbreaker:
@lonerzzz I looked at the IDF code, maybe here?
[IDF driver/i2c.c](https://github.com/espressif/esp-idf/blob/f2b2d045bace116bb6cfdaf9549a8950f7844839/components/driver/i2c.c#L504)
```c
gpio_set_direction(scl_io, GPIO_MODE_OUTPUT_OD);
gpio_set_direction(sda_io, GPIO_MODE_OUTPUT_OD);
```
Maybe mode should be: `GPIO_MODE_DEF_OD` instead of `GPIO_MODE_OUTPUT_OD`?

I can't test it because I don't use IDF directly.

Chuck.

lonerzzz:
I don't use the IDF directly either so it may be just a matter of creating an issue and telling them how it is solved in Arduino I2C based on your findings. They should be back next week so you could tag @me-no-dev to get his input on passing it along.

ESP32DE:
@stickbreaker 
@lonerzzz 
https://github.com/stickbreaker/arduino-esp32/pull/19

i did optional order the setup for pinMode, cause in older version of arduino esp32 core 
the order after open drain setup is successfull but not before.
and in actually version the order must come before open drain setup.

interesting is the second found, to setup the open drain as "OPEN_DRAIN" instead "OUTPUT_OPEN_DRAIN" nice found! i will check this in the esp-idf too
cause we have in esp-idf in i2c a problem :) .. let me check on weekend, so i have more time for this.
txs!
best wishes
rudi ;-)

ESP32DE:
cross thinking
a I²C master did clk the slaves
make it a different if we set the clk pin on master as "OUTPUT_OPEN_DRAIN"
and on the slaves we set clk pin as "OPEN_DRAIN" 
did you test this way too @stickbreaker ?
( i wil check this way on esp-idf ) 

i think for sda we need "OPEN_DRAIN"
cause the ACK is read ( no OUTPUT_OPEN_DRAIN )  by master on this
so "OPEN_DRAIN" makes sence here, good found :+1: 

best wishes
rudi ;-)

negativekelvin:
@stickbreaker  @ESP32DE  no I don't think that is a problem because that is inside bus reset function and at the end i2c_set_pin gets called which sets GPIO_MODE_INPUT_OUTPUT_OD

stickbreaker:
@ESP32DE 
>>cross thinking
a I²C master did clk the slaves
make it a different if we set the clk pin on master as "OUTPUT_OPEN_DRAIN"
and on the slaves we set clk pin as "OPEN_DRAIN"
did you test this way too @stickbreaker ?
( i wil check this way on esp-idf )

The glitch seems to be created when the output driver is assigned to the pin.  Inside [cores/esp32/esp32-hal-gpio.c](https://github.com/espressif/arduino-esp32/blob/a153f6ad280e5932be9c80341556348ff505d586/cores/esp32/esp32-hal-gpio.c#L159) the OPEN_DRAIN option is the last configuration set.  I think this ordering is causing the problem.  

But, I do not understand the thought process/functional sequence for configuring the pin/port mux hardware on the ESP32. I am very Leary about changing something that is so fundamental.  I do not know why that OPEN_DRAIN is configured as the last step.  

Another unexplored issue:  What are the ramifications of assigning a pin that has been previously configured?  This sequence works from a reset condition.  I have not tried setting either pin to another mode(input/output/adc) before assigning it to the I2C peripheral.  

I do not yet have a functional SLAVE mode ESP32 operational yet, so I haven't done any testing as a slave.  Both SCL and SDA are input/output. SCL can be driven by Slave I2c devices to pause communications. (SCL clock Stretching). I do not see any benefit using OUTPUT_OPEN_DRAIN on SCL.  The OUTPUT option causes the glitch. 

Chuck.

me-no-dev:
yes let's fix this in the source instead. @stickbreaker there is no logic behind why OPEN_DRAIN is last step. I did not expect the order to actually cause an issue. Different bits in different registers are set, so unless the ESP somehow overwrites certain registers on change, the order should not matter.
Can you test adjusting that in the hal driver?

@ESP32DE I will close this request and move towards fixing in the pinMode source :)

stickbreaker:
@me-no-dev Where can I find documentation of the GPIO matrix?  

Chuck.

PR #1073, opened 2018-02-03
bugfix(i2c): fix signal glitch on sda and scl  pin

ESP32DE:
see https://github.com/espressif/arduino-esp32/issues/1061
here a tested version and an optional suggestion

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -72,7 +72,9 @@ i2c_err_t i2cAttachSCL(i2c_t * i2c, int8_t scl)
     if(i2c == NULL){
         return I2C_ERROR_DEV;
     }
+    // digitalWrite(scl, HIGH); // optional before
     pinMode(scl, OUTPUT_OPEN_DRAIN | PULLUP);
+    digitalWrite(scl, HIGH); // tested 
     pinMatrixOutAttach(scl, I2C_SCL_IDX(i2c->num), false, false);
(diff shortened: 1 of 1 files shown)


ESP32DE:
note:
i tested the patch with an earlier version successful
if you use the actually version you must use the option 
to make the pin HIGH before you setup the pin as OPENDRAIN
( see the comment in the code ) 

so perhabs would be better, if we use the optional version then, 
cause i think, there is nobody using an earlier version ?

just in time tested with the actually version too now
i did use the optional and set Pin Level HIGH before i set it as OPENDRAIN
successful too.

perkint4:
On 3 Feb 2018 19:19, "rudi ;-)" <notifications@github.com> wrote:

note:
i tested the patch with an earlier version successful
if you use the actually version you must use the option
to make the pin HIGH before you setup the pin as OPENDRAIN
( see the comment in the code )

so perhabs would be better, if we use the optional version then,
cause i think, there is nobody using an earlier version ?

just in time tested with the actually version too now
i did use the optional and set Pin Level HIGH before i set it as OPENDRAIN
successful too.

ESP32DE:
i will close this PR cause this is only work in a prev version
we must use the optional in the actually version to kick the glitch on sda and scl on wire.begin().
so you will get asap the optional for better merging :)

ESP32DE:
see PR https://github.com/espressif/arduino-esp32/pull/1065
i added the optional to this.

PR #1076, opened 2018-02-03
Added support for WEMOS LOLIN32 LITE.

sria91:
Added support for WEMOS LOLIN32 LITE.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -335,6 +335,37 @@ lolin32.menu.UploadSpeed.512000.upload.speed=512000
 
 ##############################################################
 
+lolin32-lite.name=WEMOS LOLIN32 LITE
+
+lolin32-lite.upload.tool=esptool
+lolin32-lite.upload.maximum_size=1310720
+lolin32-lite.upload.maximum_data_size=294912
+lolin32-lite.upload.wait_for_upload_port=true
+
+lolin32-lite.serial.disableDTR=true
+lolin32-lite.serial.disableRTS=true
+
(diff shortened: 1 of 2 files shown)


DavidAntliff:
Espressif: it would be good to get this (or #857) merged, please?

Meliox:
Thanks.
I'm actually unable to compile for lolin32 lite with this board when I try in use the wire.h lib. Despite calling being with SDA and SCL. My finding is that they are not defined in pin_arduino.h. As all pins can be used it's not a huge problem, but they are defined here, hence a default refernce is needed.: https://github.com/espressif/arduino-esp32/blob/master/libraries/Wire/src/Wire.cpp#L63.

`static const uint8_t SDA = 21;
static const uint8_t SCL = 22;

static const uint8_t SS    = 5;
static const uint8_t MOSI  = 23;
static const uint8_t MISO  = 19;
static const uint8_t SCK   = 18;`

There might also be other things missing, as added these did not fix the missing SDA, SCL, which I don't know the cause of.

Furthermore, does it only support UploadSpeed=115200?

Anyways, I actually solved the issue just by using the LOLIN32.

henricazottes:
@Meliox so you don't have the proper builtin led pin ? Or do you use Lolin32 + pins modification ?

Meliox:
@henricazottes I am not exactly sure. My code is here: https://github.com/Meliox/Arduino/blob/master/weatherStation/weatherStation.ino . Compiled out of the box and is working perfectly with LOLIN32.

PR #1082, opened 2018-02-05
Python 3 fails,  unicode is not needed or supplied

alecthegeek:
See  https://stackoverflow.com/questions/6812031/how-to-make-unicode-string-with-python3

May unicode an alias for str()


me-no-dev:
Thanks :)

Issue #962, opened 2018-01-02
MDNS.addService not working


natcl:
Please fill the info fields, it helps to get you faster support ;)
### Hardware:
Board:							Adafruit ESP32 Feather
Core Installation/update date:			
IDE name:							Arduino 1.8.5 on MacOS
Flash Frequency:					?80Mhz?
Upload Speed:						

### Description:
Using the mDNS_Web_Server example sketch, I can't get the addService method to work.  The sketch runs fine, I can access my ESP32 using the esp32.local URL but it doesn't get advertised on the network when checking with different mdns browsers.  Similar code works with the esp8266.  Also browsing services seems to work.

baggior:
I confirm this problem

baggior:
#387

beegee-tokyo:
Works without problems for me, but I am not using MDNS.begin, instead I am using ArduinoOTA to start the mDNS service.
```cpp
ArduinoOTA.setHostname("ESP32-Test-2481CE9C");
ArduinoOTA.begin();
// addd some more service text
MDNS.addServiceTxt("_arduino", "_tcp", "service", "MHC");
MDNS.addServiceTxt("_arduino", "_tcp", "type", "TestBoard");
MDNS.addServiceTxt("_arduino", "_tcp", "id", "ESP32-Test");
```
Result of Service Browser on my Android tablet:
![screenshot_2018-02-04-16-56-39](https://user-images.githubusercontent.com/512690/35775875-8557cc64-09cc-11e8-9079-ea5bb87f7bb0.png)

natcl:
But that’s to advertize the Arduino service. If you want to advertize a generic service (like OSC over UDP) it doesn’t work.

baggior:
No the service is not shown even in Service Browser forAndroid
The same code works for ESP8266 but not for ESP32
I did 

WiFi.setHostname(..)
MDNS.begin()
MDNS.addService(..)

beegee-tokyo:
Maybe it works for me, because I am using the "staging" version, but when I do 
```cpp
	WiFi.setHostname("ESP32-Test-mDNS");
	MDNS.begin("ESP32-Test-mDNS");
	MDNS.addService("_osc","_udp",4500);
	MDNS.addServiceTxt("_osc","_udp","does","it");
	MDNS.addServiceTxt("_osc","_udp","work","?");
	MDNS.addServiceTxt("_osc","_udp","yes","!");
```
The result in Android service browser is
[...]

Issue #387, opened 2017-05-19
ESPmDNS Seems not to work.


Kraay89:
_Not sure if this is even close to allowed here (my first ever issue created on GH)._

After adding the following to my program in the appropriate place in a slightly altered wifi server example (after WiFi.begin... and before server.begin)

`WiFi.setHostname("MyDevice");`

My program compiles and everything works as advertised, however:
I can't reach the device from a browser as i was expecting to... I cant ping the device either from powershell. Is this correct behavior and did i misinterpret how ESPmDNS works and what it provides or am i missing something?

erenfro:
I'm having this same issue. In my case, I'm using the ArduinoOTA which does ultimately set the mDNS name itself, and it too started failing since a recent git pull. It worked at one point prior, but stopped. I just wish I recalled which checkout tag I was on at the time I noticed it working.

copercini:
@Kraay89 try with all letters in lowercase ;)

zsiroslaci:
I just tried ESPmDNS example. I was compiled without any error, but it not work.
I suppose that MDNS.begin("esp32"); should include IPAddress as well , or something to bond to it, but no exaple what and how. In case of  Esp8266, the WiFi.localIP() is included an that is working. in ESP32 it does not work, giving compiling error. What is the solution?

baggior:
I can confirm that mdns is not working to me too
no error shown but the service is not dicovered.
board = esp32dev

(The same code on ESP8266 works)

beegee-tokyo:
As commented [here](https://github.com/espressif/arduino-esp32/issues/962) there are differences between the ESP8266 version of the library and the ESP32 version
- Service name must be all lower case and start with an '_'
- Protocol must be all lower case and start with an '_'
If you call the same way as in ESP8266
```cpp
MDNS.addService("osc", "udp", 4500);
MDNS.addServiceTxt("osc", "udp", "board", "ESP8266");
```
it will fail (without error) on ESP32. You need to call with
```cpp
MDNS.addService("_osc", "_udp", 4500);
MDNS.addServiceTxt("_osc", "_udp", "board", "ESP32");
```

everslick:
[...]

PR #1085, opened 2018-02-06
Make ESPmDNS compatible to ESP8266mDNS

beegee-tokyo:
The ESP8266 version of the mDNS library expected service name and protocol **without** leading `_` (underscore). The ESP32 version expected service name and protocol **with** leading `_` (underscore).

This small change makes the ESP32 version compatible with the ESP8266 version and keeps backward compatibility with existing code by accepting service name and protocol either **with or without** leading `_` (underscore).

See issues #962 #387

Tested with following combinations:    
```cpp
  MDNS.addService("test1","tcp",9998);
  MDNS.addServiceTxt("test1", "tcp", "id", "mDNS-Test");

  MDNS.addService("_test2","tcp",9998);
  MDNS.addServiceTxt("_test2", "tcp", "id", "mDNS-Test");

  MDNS.addService("test3","_tcp",9998);
  MDNS.addServiceTxt("test3", "_tcp", "id", "mDNS-Test");

  MDNS.addService("_test4","_tcp",9998);
  MDNS.addServiceTxt("_test4", "_tcp", "id", "mDNS-Test");
```

### DIFF
--- a/libraries/ESPmDNS/src/ESPmDNS.cpp
+++ b/libraries/ESPmDNS/src/ESPmDNS.cpp
@@ -59,6 +59,7 @@ bool MDNSResponder::begin(const char* hostName){
     }
     WiFi.onEvent(_on_sys_event);
     _hostname = hostName;
+	_hostname.toLowerCase();
     if(mdns_hostname_set(hostName)) {
         log_e("Failed setting MDNS hostname");
         return false;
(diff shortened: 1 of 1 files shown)


baggior:
thank you

libraries/ESPmDNS/src/ESPmDNS.cpp
```diff
@@ -59,6 +59,7 @@ bool MDNSResponder::begin(const char* hostName){
     }
     WiFi.onEvent(_on_sys_event);
     _hostname = hostName;
+	_hostname.toLowerCase();
```
atanisoft: shouldn't this be merged with the line above as:
`_hostname = hostName.toLowerCase();`

baggior: I don't think
String.toLowerCase() make side effect on the self object and doesn't returns
void toLowerCase(void);

https://www.arduino.cc/reference/en/language/variables/data-types/string/functions/tolowercase/

atanisoft: You are right, I was looking at it and it looked odd. Normally a method like this would return the modification rather than do the modification on the object itself. It appears that https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/WString.h#L255 is just doing a direct modification on the object in place.

beegee-tokyo:
@me-no-dev - Anybody working on these pull requests? There are 27 open pull requests, and it seems nobody is working on them.

baggior:
How can I contribute working on the pull requests?

What do I have to do to commit on this repository?

me-no-dev:
Sorry guys, I had lots of unrelated to Arduino work to cover :) I surely am back though

PR #1087, opened 2018-02-06
HTTPClient: make clean up faster (fixes #828)

### LINKED ISSUE #828: "HTTPClient::end(void)" runs are too long
In /libraries/HTTPClient/src/HTTPClient.cpp, function "void HTTPClient::end(void)", line 231:

        if(_tcp->available() > 0) {
            log_d("still data in buffer (%d), clean up.", _tcp->available());
            while(_tcp->available() > 0) {
                _tcp->read();
            }
        }

If the unread data is very large, it will take a long time. If changed to:

        if(_tcp->available() > 0) {
            log_d("still data in buffer (%d), clean up.", _tcp->available());
            _tcp->flush();
        }

You can achieve what you want.

h3ndrik:
If the unread data is very large, cleaning up byte by byte will take a long time.
on a http radio stream easily >=60s observed.

so just flush tcp buffer instead of reading it byte by byte.

thanks to @xiruilin and @Uksa007

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -232,9 +232,7 @@ void HTTPClient::end(void)
     if(connected()) {
         if(_tcp->available() > 0) {
             log_d("still data in buffer (%d), clean up.", _tcp->available());
-            while(_tcp->available() > 0) {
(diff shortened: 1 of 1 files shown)


me-no-dev:
nice :)

PR #1091, opened 2018-02-08
Update get.py

RyLeeHarrison:
crude solution to Python3.6 error: 

> 
> Traceback (most recent call last):
>   File "get.py", line 27, in <module>
>     current_dir = os.path.dirname(os.path.realpath(unicode(__file__)))
> NameError: name 'unicode' is not defined

### DIFF
--- a/tools/get.py
+++ b/tools/get.py
@@ -24,7 +24,14 @@ else:
 if 'Windows' in platform.system():
     import requests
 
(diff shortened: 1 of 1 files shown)


me-no-dev:
`AttributeError: 'NoneType' object has no attribute 'startswith'`
Fails on Python2 :) maybe detect the version first?

PR #1096, opened 2018-02-10
Fixes SPI mode setting (Issue #1094)

### LINKED ISSUE #1094: SPI mode inconsistency
Hi!

I guess there is an inconsistency in the SPI mode assignment.

In the function `spiSetDataMode` we have:

https://github.com/espressif/arduino-esp32/blob/a153f6ad280e5932be9c80341556348ff505d586/cores/esp32/esp32-hal-spi.c#L333-L340

And in the function `spiTransaction` we have:

https://github.com/espressif/arduino-esp32/blob/a153f6ad280e5932be9c80341556348ff505d586/cores/esp32/esp32-hal-spi.c#L714-L721

The `user.ck_out_edge` value is different.

Here is the `spiGetDataMode` function:

https://github.com/espressif/arduino-esp32/blob/a153f6ad280e5932be9c80341556348ff505d586/cores/esp32/esp32-hal-spi.c#L308-L319

kroggen:
Fixes #1094

### DIFF
--- a/cores/esp32/esp32-hal-spi.c
+++ b/cores/esp32/esp32-hal-spi.c
@@ -309,9 +309,9 @@ uint8_t spiGetDataMode(spi_t * spi)
     bool outEdge = spi->dev->user.ck_out_edge;
     if(idleEdge) {
         if(outEdge) {
-            return SPI_MODE3;
+            return SPI_MODE2;
         }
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks :)

PR #1103, opened 2018-02-13
Unicode not supported in Python 3

afreeland:
Python 3 renamed the `unicode` type to `str`, the old `str` type has been replaced by `bytes`

Worked for Python 3 but have not tested in Python 2...wasnt sure if there were any test to run.

https://stackoverflow.com/questions/19877306/nameerror-global-name-unicode-is-not-defined-in-python-3

### DIFF
--- a/tools/get.py
+++ b/tools/get.py
@@ -15,16 +15,19 @@ import sys
 import tarfile
 import zipfile
 import re
+_unicode = None
 if sys.version_info[0] == 3:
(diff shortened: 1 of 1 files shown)


me-no-dev:
Already resolved in https://github.com/espressif/arduino-esp32/commit/2718d69c19a34c56abff770a02e1eb5ec84ac689 :)

Issue #1128, opened 2018-02-20
reason2str wrong?


tablatronix:
Why do the upper reason codes not align with the bottom.

`[W][WiFiGeneric.cpp:298] _eventCallback(): Reason: 202 - ASSOC_FAIL`

    WIFI_REASON_BEACON_TIMEOUT              = 200,
    WIFI_REASON_NO_AP_FOUND                     = 201,
    WIFI_REASON_AUTH_FAIL                            = 202,
    WIFI_REASON_ASSOC_FAIL                          = 203,
    WIFI_REASON_HANDSHAKE_TIMEOUT       = 204,

EDIT: ohhhh
Ok so maybe its just
reason2str that is wrong..

Should this maybe be 0 indexed instead, so (r-176)-1
`#define reason2str(r) ((r>176)?system_event_reasons[r-177]:system_event_reasons[r-1])`

copercini:
Closed due it's fixed here: https://github.com/espressif/arduino-esp32/pull/1132

Nice PR!

PR #1132, opened 2018-02-20
fix event reason logging > 176

tablatronix:
off by one

#1128


me-no-dev:
thanks :)

tablatronix:
off by one bug is actually caused by index 12 missing from  system_event_reasons causing misalignment of indexes >11 < 24

PR #1323 fixes this and reverts this one

Issue #1022, opened 2018-01-22
SPIFFS +  ESP32 Sketch Data Upload Problem (mkspiffs)


smacyas:
### Hardware:
Board:							?ESP32 Dev Module?
Core Installation/update date:			latest
IDE name:							Arduino IDE
Flash Frequency:					40Mhz?
Upload Speed:						115200

### Description:
Hello.
I need help.
I have WROOM-32 ESP32 module.
In folder of my ino-project I created folder ‘data’ and put into it some files: css, html, txt by ESP32 Sketch Data Upload (https://github.com/me-no-dev/arduino-esp32fs-plugin), I have uploaded them successfully to my module.
When I try to read file from SPIFFS (previously mounted SPIFFS, of course) (latest arduino-esp32, https://github.com/espressif/arduino-esp32), listDir(SPIFFS, "/", 0) - OK; but while I use readFile(SPIFFS, "/style.css"); there is some trash, but no correct data. When I use write and read file, which was NOT uploaded by ESP32 Sketch Data Upload, it is OK.

By the way, I noticed string: flash read err, 1000 ets_main.c 371

Any ideas?
Thanks.

### Debug Messages:
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
**flash read err, 1000**
ets_main.c 371 
ets Jun  8 2016 00:22:57

rst:0x10 (RTCWDT_RTC_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0018,len:4
load:0x3fff001c,len:956
load:0x40078000,len:0
load:0x40078000,len:13076
entry 0x40078a58

Starting device...
ESP32 Chip ID: 9238022
      FreeHeap: 126324
      ChipRevision: 1
      CpuFreqMHz: 240
      SdkVersion: v3.1-dev-239-g1c3dd23f-dirty
      FlashChipSize: 4194304
      FlashChipSpeed: 40000000
      FlashChipMode: 2

SPIFFS Mount succesfull: 
Listing directory: /
  FILE: /door.png  SIZE: 846
  FILE: /favicon.ico  SIZE: 239
  FILE: /index.html  SIZE: 4875
[...]

PR #1159, opened 2018-02-28
Updated to download mkspiffs 0.2.2

rfestag:
Updated the package template (used by tools/get.py) to download the new mkspiffs tool. This is in reference to https://github.com/espressif/arduino-esp32/issues/1022,  and is necessary because of a change implemented in Dec 2017. Currently, the mkspiffs that is installed by get.py does not work with arduino-esp32, causing problems for https://github.com/me-no-dev/arduino-esp32fs-plugin.

### DIFF
--- a/package/package_esp32_index.template.json
+++ b/package/package_esp32_index.template.json
@@ -43,7 +43,7 @@
             {
               "packager": "esp32",
               "name": "mkspiffs",
-              "version": "0.2.1"
+              "version": "0.2.2"
             }
           ]
         }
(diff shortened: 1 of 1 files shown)


copercini:
I think it will be necessary to update the SHA-256 hashes of checksum and filesize too, otherwise, the get.py will detect it as corrupt file

rfestag:
It looks like the sha256 hash checking is commented out, and seems to work properly right now. However, in case that changes in the future, it certainly makes sense to update. When I get back this evening I'll put those together and push an update.

rfestag:
I just updated the hashes and file sizes. It also turns out that the windows filename changed. I only tested the linux64 version. I also uncommented the hash checking, and it appears to work correctly. Not sure why it is currently commented out, so I  didn't commit that change.

me-no-dev:
Thanks!

PR #1205, opened 2018-03-12
Corrected Flash Mode

nickoloss:
Changed Flashmode to DIO due board only supports DIO


me-no-dev:
hmmm... how come? I run my picos in qio mode

me-no-dev:
I'm kinda inclined to believe that this is the case... maybe I just thought that I am running QIO

PR #1214, opened 2018-03-14
Allow update to function over HTTPS due to peek() missing from mbedTLS.

brokentoaster:
Removed redundant check of magic byte from writeStream()

* The magic byte is checked in _writeBuffer() before any flash is altered.
* The check using peek() earlier will at best only save a wasted read of one buffer length of data being read if data is missing the magic byte
* As we are already prepared to read an entire buffers worth of data anyway this saving doesn't matter.

### DIFF
--- a/libraries/Update/src/Updater.cpp
+++ b/libraries/Update/src/Updater.cpp
@@ -200,18 +200,6 @@ bool UpdateClass::_writeBuffer(){
     return true;
 }
 
-bool UpdateClass::_verifyHeader(uint8_t data) {
-    if(_command == U_FLASH) {
-        if(data != ESP_IMAGE_HEADER_MAGIC) {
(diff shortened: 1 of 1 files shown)


me-no-dev:
hmm.. I thought we added peek to SSL... @copercini am I wrong?

copercini:
It's on my fork, https://github.com/copercini/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L116 but I don't have a server for test, so it's only theoretically implemented

me-no-dev:
Can close this now :)

brokentoaster:
The use of peek() in Updater was still redundant and can/should/could still be removed from Updater.cpp although it is no longer a requirement in order to simply compile.

plerup:
This patch is the only way to get the https update working as far as I can tell. Please merge

PR #1220, opened 2018-03-15
Update WiFiMulti.cpp

c0zm0z:
added 5 sec timeout from ESP8266


stickbreaker:
@c0zm0z you are hard coding a 5 seconds timeout in the `.run()` function, could to change the function parameters from `void` to `uint32_t timeOut=5000`? 
It would accomplish the same thing, but, give some control if a shorter or longer timeout was necessary.

Chuck.

PR #1222, opened 2018-03-16
Add Core Debug Level for WeMosBat

stickbreaker:
Should we add this declaration option for all boards?


JamesGKent:
adding it for all boards would simplify asking people to turn on logging for debug purposes...
i'd vote in favour of adding it, even though same thing can be achieved by using the ESP Dev module board option...

stickbreaker:
@JamesGKent As long as the pin definitions and Flash speeds are the same.

Chuck.

JamesGKent:
@stickbreaker i hadn't considered that, in which case i'd definitely advocate that the debug menu should be added to all boards

me-no-dev:
I think that a better way should be found to organize the boards better. the config file is a zoo of boards that are actually all the same...

PR #1239, opened 2018-03-20
Preserver custom pin assigments with Wire.begin(custom_sda,custom_scl)

stickbreaker:
This code allows Wire.begin() to assign the default values of SDA, and SCL only if they have not been previously configured.  Arduino libraries that use Wire() usually re-init the I2C interface in their initialization code with a call to Wire.begin().  If a user app sets custom pins assignment in setup(); These assignments will be overwritten with the default values  whenever Wire.begin() is called.


me-no-dev:
done :)

PR #1242, opened 2018-03-20
M5Stack variant update (boards.txt)

reaper7:
When M5Stack board is selected then
M5Stack variant (pins_arduino.h) file is not automatically included,
because boards.txt m5stack-core-esp32.build.variant=esp32
This commit changes m5stack-core-esp32.build.variant from esp32 to m5stack_core_esp32
also rename M5Stack variant directory from m5stack-core-esp32 to m5stack_core_esp32


me-no-dev:
Done :)

PR #1243, opened 2018-03-21
Can not invalidate global serial instance

KEDARUMA-FANTASTIC:
I am in trouble because I can not disable instances of wide area serial
This implementation was included in ESP 8266, so I ported it


me-no-dev:
I was a bit unsure wether this would be used on ESP32 so I did not propagate it everywhere :)

PR #1269, opened 2018-03-28
fixes #1249 softApDisconnect fails always

### LINKED ISSUE #1249: softAPdisconnect failure, no error return
I am getting failures with softAPdisconnect returning false.

debugging 
`esp_err_t err = esp_wifi_set_config(WIFI_IF_AP, &conf);` returns error base 11
So no error code..

Any ideas how to debug this further ?

tablatronix:
same as esp8266 PR I submitted.
`esp_wifi_set_config(WIFI_IF_AP, &conf)`  returns blank error code, if authmode is not open when password is not set. 

NOT documented anywhere I can tell.


me-no-dev:
good find!

Issue #1282, opened 2018-04-02
resource concumpition BluetoothSerial.cpp


AlphaLima:
BluetoothSerial.cpp
Works fine for me, but it needs 50% of the ESP32 Resources, together with my sketch it is over 100%, so it becomes useless. Is there a way to reduce the library size?

----------------------------- Remove above -----------------------------

### Hardware:
Board:							Node32
Core Installation/update date:			?11/jul/2017?
IDE name:							?Arduino IDE? ?Platform.io? ?IDF component?
Flash Frequency:					?40Mhz?
Upload Speed:						?115200?

### Description:
Describe your problem here

### Sketch:
```cpp

//Change the code below by your sketch
#include <Arduino.h>

void setup() {
}

void loop() {
}
```

### Debug Messages:
```
Enable Core debug level: Debug on tools menu of Arduino IDE, then put the serial output here 
```

lbernstone:
You can change the partition layout of your device. https://desire.giesecke.tk/index.php/2018/01/30/change-partition-size/

Issue #1194, opened 2018-03-08
Increasing BLE sketch size of ~156k after adding BluetoothSerial lib


reaper7:
### Hardware:
Board:							ESP32 Dev Module
Core Installation/update date:			Branch_b4b9a79 05/03/2018
IDE name:							Arduino IDE
Flash Frequency:					80Mhz
Upload Speed:						921600

### Description:
after adding the BluetoothSerial library (https://github.com/espressif/arduino-esp32/commit/b4b9a79eead317d899fe4f199b4d536b99c25112), sketches which use BLE library (https://github.com/nkolban/ESP32_BLE_Arduino/tree/6bad7b42a96f0aa493323ef4821a8efb0e8815f2) increase finally size by **~156 kbytes!**

Lets try for example library: BLE -> example: BLE_uart:

- branch with previous commit (https://github.com/espressif/arduino-esp32/commit/835268c326e0a6c95a36e66f5b6a6d24c2c51f8f) finally sketch size: 931869 bytes (71%)

- branch with BluetoothSerial library (https://github.com/espressif/arduino-esp32/commit/b4b9a79eead317d899fe4f199b4d536b99c25112) finally sketch size: 1088621 bytes (83%)

For this reason my sketches with BLE and several additions do not fit esp32 flash any more,
even though I do not use new BluetoothSerial library in these cases (increase size from 1265006 to **1422022**)

What is interesting, I did not notice the increase final sketch size when BLE library is not used.

In the current situation, creation a little more advanced programs with BLE and for e.g. display, becomes impossible due to lack of space.

### Sketch:
library: BLE -> example: BLE_uart

beegee-tokyo:
The Bluetooth stack is very large. You need to change you partition sizes as shown in the [Wiki](http://desire.giesecke.tk/index.php/2018/01/30/change-partition-size/).

reaper7:
@beegee-tokyo - tnx! this is a option but...  
...it's looks like these two libraries use two separate bluetooth stacks?

beegee-tokyo:
I do not know if there are two different stacks for BLE and Bluetooth, because that is somewhere down below in the ESP-IDF. But I started with using BLE only first and added Bluetooth Serial later (check my [blog](http://desire.giesecke.tk/index.php/2018/02/27/esp32-bluetooth-serial-and-ble-server-running-together/)) the program size didn't double, so I guess there is only one stack.

reaper7:
[...]

PR #1302, opened 2018-04-09
Additional partition scheme min_spiffs

reaper7:
with minimal SPIFFS partition size for bigger apps and OTA support 

useful for problems:
https://github.com/espressif/arduino-esp32/issues/1194
https://github.com/espressif/arduino-esp32/issues/1282
but leaving the OTA functionality available (unlike no_ota scheme)

addition for:
https://github.com/espressif/arduino-esp32/pull/1296/commits/535a55dd4ca19d7fa10d10e47367ed8edd7b866b
https://github.com/espressif/arduino-esp32/pull/1296/commits/75b08d542f560d1b588f47a43100acd43a869064

P.S. is there a possibility to add partitions select for all available boards? (not only for esp32 dev board)

### DIFF
--- a/Kconfig.projbuild
+++ b/Kconfig.projbuild
@@ -90,6 +90,8 @@ config ARDUHAL_PARTITION_SCHEME_MINIMAL
     bool "Minimal (for 2MB FLASH)"
 config ARDUHAL_PARTITION_SCHEME_NO_OTA
     bool "No OTA (for large apps)"
+config ARDUHAL_PARTITION_SCHEME_MIN_SPIFFS
+    bool "Minimal SPIFFS (for large apps with OTA)"
 endchoice
 
 config ARDUHAL_PARTITION_SCHEME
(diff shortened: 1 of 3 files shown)


me-no-dev:
yeah... to add partitions to all boards we should figure out either a board generator, or combine the boards and change the variant as a submenu. Boards are too many anyway and most are the same

reaper7:
Another problem is a hardcoded upload.maximum_size.
For large application we select partition scheme for "big apps", 
partitions will be properly prepared but at the compilation end we still get exceeded size error, 
because we compare compiled sketch size with hardcoded upload.maximum_size 
but it should be compared with the right entry from selected partition.

ReanimationXP:
^ This is a problem I'm running into. All sketches larger than 1310720 bytes report the following, regardless of partition size:
`Sketch uses 1560446 bytes (119%) of program storage space. Maximum is 1310720 bytes.`

beegee-tokyo:
Until this is fixed, you can still use the old method to [Change partition size](https://desire.giesecke.tk/index.php/2018/04/20/change-partition-size-arduino-ide/).
Works fine for me. Output from example sketch with Bluetooth and BLE included:

> Archiving built core (caching) in: C:\Users\BeeGee\AppData\Local\Temp\arduino_cache_147130\core\core_espressif_esp32_esp32max_FlashMode_qio,FlashFreq_80,FlashSize_4M,UploadSpeed_921600,DebugLevel_none_af83d5f1a3fa262478237a470cbae092.a
> Sketch uses 1567406 bytes (79%) of program storage space. Maximum is 1966080 bytes.
> Global variables use 70308 bytes (23%) of dynamic memory, leaving 224604 bytes for local variables. Maximum is 294912 bytes.
> Sending invitation to 192.168.0.112 ..........

reaper7:
as @beegee-tokyo wrote, at this moment we have new partition file
and we have to make a change only in the boards.txt

PR #1316, opened 2018-04-15
Convert the few remaining cr/lf files to use lf for eol

leres:
If you develop on windows and need cr/lf files, see this:

    https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace

    Git can handle this by auto-converting CRLF line endings into LF
    when you add a file to the index, and vice versa when it checks out
    code onto your filesystem. You can turn on this functionality with
    the core.autocrlf setting. If you're on a Windows machine, set it
    to true - this converts LF endings into CRLF when you check out code:

    $ git config --global core.autocrlf true


me-no-dev:
nice... :) I was planning to do this but I guess have not found the time so far!

PR #1323, opened 2018-04-16
WiFi disconnect reason message updated

rmshub:
wifi_err_reason_t doesn't have reason code 12. After reason code 11, we are getting wrong wifi disconnected reason message. Added "UNSPECIFIED"  in place of reason code 12.
And above 201 reason code, it should be r-176 instead of r-177. Because messages are started from zeroth offset.


tablatronix:
ahh this probably explains my PR

https://github.com/espressif/arduino-esp32/pull/1132/commits/486b813bc7a54726ea63fef8dd646afea82e61d9

Shouldn't this PR contain the esp_wifi_type.h change also ?

tablatronix:
I mean they should probably match for clarity ?

tablatronix:
@me-no-dev this PR is the proper fix for  #1132 
I faileld to notice index 12 is missing from system_event_reasons causing this one off issue.

index 12 is also missing from wifi_err_reason_t, but is not addressed by this.

rmshub:
Yeah. My PR is not fix WiFI reason code in esp_wifi_type.h because we don't know exactly it is missed accidentally or they defined WiFi error code with some valid cause. If it is missed accidentally, then it is required to address esp_wifi_type.h and then update Wifi reason messages accordingly.

tablatronix:
Maybe we should insert a comment there..

me-no-dev:
thanks!

PR #1326, opened 2018-04-17
Fix compile error using asm

Pro:
In function 'xPortGetCoreID':
error: expected ')' before ':' token
         :"=r"(id));

It is recommended to use `__asm__` instead of `asm`

See:
https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords


me-no-dev:
This change need to go to IDF and not here. We just use the headers from there :)

Pro:
Done https://github.com/espressif/esp-idf/pull/1965

PR #1344, opened 2018-04-22
Add Schedule and Functional/Scheduled Ticker

hreintke:
Identical implementation as in ESP8266

### DIFF
--- a/cores/esp32/Schedule.cpp
+++ b/cores/esp32/Schedule.cpp
@@ -1,0 +1,78 @@
+#include "Schedule.h"
+
+struct scheduled_fn_t
+{
+    scheduled_fn_t* mNext;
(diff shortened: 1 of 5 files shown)


me-no-dev:
How about this to get written in more ESP32 and RTOS style? Use a thread (start it on first call of `schedule`) instead of attaching always to the loop ;)

me-no-dev:
I understand, but this is a dual-core chip which runs an actual OS, so let's make it run as it should. ESP8266 runs NonOS like most other MCUs.

hreintke:
@me-no-dev See PR https://github.com/espressif/arduino-esp32/pull/1516

fabiuz7:
no hope about this PR? It would be great porting some old esp8266 program without messing up with different "API"

PR #1353, opened 2018-04-25
Fix wrong range for sigma delta duty cycle

mmone:
The esp-idf expects duty values for the the sigma delta modulator in the range of -128 to 127
The arduino framework is supposed to use the range 0-255 thus the offset calculation was wrong.


me-no-dev:
Hey :) Thanks!

PR #1374, opened 2018-05-02
Add note about sketch book location

marcelstoer:
I had no idea what `~/Documents/Arduino` represents as _my_ sketch book location is different. Also fixed a list formatting glitch.


me-no-dev:
thanks :)

PR #1382, opened 2018-05-04
Selectable (from menu) partitions for m5stack 

reaper7:
addition for m5stack (as is prepared for lolin32 board: espressif#1379)

discussion here: espressif#1378


boards.txt
```diff
@@ -1267,6 +1267,15 @@ m5stack-core-esp32.menu.FlashFreq.80.build.flash_freq=80m
...
 m5stack-core-esp32.menu.FlashFreq.40.build.flash_freq=40m
 
+m5stack-core-esp32.menu.PartitionScheme.default=Default
+m5stack-core-esp32.menu.PartitionScheme.default.build.partitions=default
+m5stack-core-esp32.menu.PartitionScheme.no_ota=No OTA (Large APP)
+m5stack-core-esp32.menu.PartitionScheme.no_ota.build.partitions=no_ota
+m5stack-core-esp32.menu.PartitionScheme.no_ota.upload.maximum_size=2097152
+m5stack-core-esp32.menu.PartitionScheme.min_spiffs=Minimal SPIFFS (Large APPS with OTA)
+m5stack-core-esp32.menu.PartitionScheme.min_spiffs.build.partitions=min_spiffs
+m5stack-core-esp32.menu.PartitionScheme.min_spiffs.upload.maximum_size=1966080
```
domints: Also I think it's important to consider if your board may contain 2MB flash version of ESP32. If so, it would be good to add definition for minimal partition scheme. 
I didn't add one to LOLIN32, as I think all of lolins come with 4MB flash.

reaper7: Currently, the original m5stack flash size also has only 4M,
but of course we have to remember about this possibility

PR #1388, opened 2018-05-06
Fix WiFiClient error handling

chemicstry:
`WiFiClient::connected()` incorrectly checks `errno`. If any other function before this call sets `errno`, the `connected()` returns false even though `recv()` result was 0. This is because `recv` function does not reset `errno` variable on successful operation and old unknown value is used in switch.

This behaviour can be observed when using multiple `WiFiClient` objects and one of them disconnects with an error. Subsequent `connected()` call on a healthy connection would return false due to a previous set `errno`.

I propose a change to only check for `errno` if `recv()` call fails.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -313,7 +313,7 @@ uint8_t WiFiClient::connected()
     if (_connected) {
         uint8_t dummy;
         int res = recv(fd(), &dummy, 0, MSG_DONTWAIT);
-        if (res <= 0) {
+        if (res < 0) {
             switch (errno) {
                 case ENOTCONN:
                 case EPIPE:
@@ -328,7 +328,6 @@ uint8_t WiFiClient::connected()
             }
         }
         else {
-            // Should never happen since requested 0 bytes
             _connected = true;
         }
     }


Saulzi:
I am having this issue. Looking forwards to when this is merged.

me-no-dev:
@Saulzi no need to wait anymore

everslick:
This commit broke the [webserver](https://github.com/bbx10/WebServer_tng). The very first GET is fast but each subsequent GET is delayed exactly 2000ms. Which makes page loads which have more then one GET painstakingly slow. I zoomed in on this commit using git bisect.

everslick:
@sansillusion do you load more then ONE resource on page loads?

if yes, are your page loads still fast when you SHIFT reload the page (this fetches the resources from the server, omitting the local browser cache).

I double checked by reverting this one commit and pages load fast again afterwards.

sansillusion:
@everslick  you are right, I had minimize request to a minimum in my project. If I reload right after a request there do seem to be a 2000ms delay. But single page loads seem fast after that 2000ms delay, while before I would get random delays BEFORE ANY page loads.

sansillusion:
looking at webserver.h (https://github.com/bbx10/WebServer_tng/blob/master/src/WebServer.h)
on line 50 there is a 2000ms delay for conection closing, could it be that the connection does not get closed and webserver has to wait for it's timeout ?

everslick:
yes, that could be, more specific it seems to be [this](https://github.com/bbx10/WebServer_tng/blob/master/src/WebServer.cpp#L219-L227) code:

```
  if (_currentStatus == HC_WAIT_CLOSE) {
    if (millis() - _statusChange > HTTP_MAX_CLOSE_WAIT) {
      _currentClient = WiFiClient();
      _currentStatus = HC_NONE;
    } else {
      yield();
      return;
    }
  }
```
question remains, what's the right thing to do. :-/

sansillusion:
@everslick I have tried changing the timeout value from 2000ms to 200ms and it seem to have sped-up the whole thing. I have not seen any error arise from it too. I have no idea why there should be a delay there but it works with a much smaller delay here.

brainsucker-na:
Yeah, sorry, that merge breaks a lot (surprised no one reported that before) :(

PR #1397, opened 2018-05-09
Port SSL fingerprint checking from ESP8266 WiFiClientSecure to ESP32

### LINKED ISSUE #278: WifiClientSecure has no verify method
I'm porting some esp8266 code and discovered that the ESP32 version of the WiFiClientSecure class does not seem to have any methods to help verify any certs or fingerprints. Is this something that is on the roadmap?

chemicstry:
This PR ports `verify` function from ESP8266 WiFiClientSecure which verifies remote certificate fingerprint.

Since sha1 is no longer secure, I switched to sha256.

Fixes #278


me-no-dev:
Nice :)

Issue #1393, opened 2018-05-08
WiFi.persistent(false) does not disable persistent WiFi config.


everslick:
Currently wifi settings are written to the flash regardless of calling WiFi.persistent(false) or not.

In my opinion [WiFiGeneric.cpp#L123](https://github.com/espressif/arduino-esp32/blob/70d0d4648766cea047613062ed50a9ecfc9de31f/libraries/WiFi/src/WiFiGeneric.cpp#L123) should be like:

```
if (_persistent) {
  esp_wifi_set_storage(WIFI_STORAGE_FLASH);
} else {
  esp_wifi_set_storage(WIFI_STORAGE_RAM);
}
```
Ref:
https://github.com/espressif/arduino-esp32/issues/548

Any thoughts on that?

everslick:
I propose the Fix like in #1404.

everslick:
fix meged to master.

tablatronix:
This is problematic now and can break code and is not a full feature implementation as it has issues with calling persistent during runtime.

* If calling persistent(false) before wifistart, credentials are not loaded, say in a library constructor
* If changing persistent before begin(), credentials are not saved, wifista begin() esp_wifi_set_config, then calls esp_wifi_start() <-- oh and no persistent flag here

Not really sure how this should be engineered to deal with these issues. But I feel this is wrong

Also not sure why the decision was made to use passing arguments for persistent states, seems certain to introduce bugs, should be steady state

Perhaps

* `_persistent` should be used and not passed around in `wifiLowLevelInit` and `espWiFiStart`
* persistent() != _persistent should change esp_wifi_set_storage() everytime, this might be a problem with the way lowlevel init is tracked not sure. But it is the only way to deal with control over set_config
* get_config could have wrappers to handle as esp8266 equiv, otherwise get_config will be persistent following and should probably always prefer flash if we ever want autoconnect to work properly.
* also wifiLowLevelDeinit needs to be working, maybe that will help switching wifiLowLevelInit back and forth.. shrug

tablatronix:
Ok so using any wifi functions in a constructor is nogo, wifi is not init, in particular nvs is not available yet.
So that is another issue by itself.

everslick:
What is the use case for enabling/disabling persistent configuration storage during runtime in the first place?

[...]

Issue #548, opened 2017-07-31
Wifi.persistent(true) is not working


salqadri:
I tried doing a WiFi.persistent(true); before my WiFi.begin and unfortunately when I restart my device Wifi.SSID() is still empty. This behavior is different to what I am used to with the ESP8266. Has WiFi.persistent basically not been implemented yet?

me-no-dev:
pretty much :)

salqadri:
Ok thanks for the update; I've workaround the problem using Preferences. Keeping the issue open until it gets fixed.

marian-craciunescu:
any news about this ?

marian-craciunescu:
@salqadri  do you have an example for Preferences ?A workaround should be nice.

salqadri:
Sure. Here are some apis I wrote for myself to save and load strings that works for both ESPs:
`
bool IsStringSaved(uint16_t startAt)
{
#ifdef ESP32
  Preferences* preferences = new Preferences();
  preferences->begin("my-app", false);
  String token = preferences->getString((String("") + startAt).c_str());
  preferences->end();
  delete preferences;
  if (token.length() == 0) {
    return false;
  }
#else
  EEPROM.begin(EEPROM_SIZE);
  char s = EEPROM.read(startAt);
  char c = EEPROM.read(startAt+1);
  EEPROM.end();
  return s == 'S' && c == ':';
#endif
}

// Saves string to EEPROM. Returns index position after end of string.
int SaveString(uint16_t startAt, const String& id)
{
#ifdef ESP32
  Preferences* preferences = new Preferences();
  preferences->begin("my-app", false);
  int ret = preferences->putString((String("") + startAt).c_str(), id.c_str());
  preferences->end();
  delete preferences;
  return ret;
#else
  uint16_t len = id.length();
  EEPROM.begin(EEPROM_SIZE);
  EEPROM.write(startAt, 'S');
  EEPROM.write(startAt + 1, ':');
  for (uint16_t i = 0; i <= len; i++)
  {
    EEPROM.write(i + startAt + 2, (uint8_t) id.c_str()[i]);
  }
  EEPROM.write(startAt + len + 2, 0);
  EEPROM.end();
  return startAt + len + 2;
#endif
}

// Loads string from EEPROM. Returns length of string read.
int LoadString(uint16_t startAt, char* buffer, size_t maxLen)
{
#ifdef ESP32
  Preferences* preferences = new Preferences();
  preferences->begin("my-app", false);
[...]

PR #1404, opened 2018-05-12
Fix wifi persistent

everslick:
WiFi config is currently written to NVS regardless of if `WiFi.persistent(false);` was called before or not. This PR moves some of the static functions in WiFiGeneric.cpp into the WiFiGenericClass to give them access to the _persistent member. Additionally `WiFi.getMode()` and `WiFi.mode()` have been made regular methods instead of static. Now `esp_wifi_set_storage(WIFI_STORAGE_RAM);` is called when persistent storage is switched off.


everslick:
Rer: #548 and #1393

everslick:
this change breaks ETH, will come up with a better fix....

PR #1406, opened 2018-05-12
Fix WiFi.persistent(false); having no effect.

everslick:
This PR calls `esp_wifi_set_storage(WIFI_STORAGE_RAM);` when persistent WiFi configuration is not desired.

according to the ESP32 API reference guide the default is to store WiFi configuration in NVS.

esp_err_t esp_wifi_set_storage(wifi_storage_tstorage)
Set the WiFi API configuration storage type.

Attention
1. The default value is WIFI_STORAGE_FLASH

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -108,7 +108,7 @@ void tcpipInit(){
     }
 }
 
-static bool wifiLowLevelInit(){
+static bool wifiLowLevelInit(bool persistent){
     static bool lowLevelInitDone = false;
     if(!lowLevelInitDone){
         tcpipInit();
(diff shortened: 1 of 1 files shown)


everslick:
To my understanding, it will not erase the config that is already stored in NVS. There is a esp_wifi_restore() function, that will reset the config to its defaults, but this is never used.

Personally I'd prefer if the ESP32 would only do what it is told to. i.e. NOW connect to SSID, NOW start AP, NOW disconnect from SSID, NOW disable AP. Never store anything in NVS on your own, never connect to anything on your own. But from the API it is hard to deduce which call to which methods in which order will do that.

tablatronix:
This is problematic now and can break code and is not a full feature implementation as it has issues with calling persistent during runtime.

* If calling persistent(false) before wifistart, credentials are not loaded, say in a library constructor
* If changing persistent before begin(), credentials are not saved, wifista begin() esp_wifi_set_config, then calls esp_wifi_start()

Not really sure how this should be engineered to deal with these issues.

tablatronix:
Perhaps

* `_persistent` should be used and not passed around in `wifiLowLevelInit` and `espWiFiStart`
* persistent() != _persistent should change esp_wifi_set_storage() everytime, this might be a problem with the way lowlevel init is tracked not sure. But it is the only way to deal with control over set_config
* get_config could have wrappers to handle as esp8266 equiv, otherwise get_config will be persistent following and should probably always prefer flash if we ever want autoconnect to work properly.
* also wifiLowLevelDeinit needs to be working, maybe that will help switching wifiLowLevelInit back and forth.. shrug

tablatronix:
Going to move these to #1393

Issue #662, opened 2017-09-24
How can I reset Serial2 / UART2?


jestebanes:
Hardware:

Board:	Own design based on ESP-WROOM-32
Core Installation/update date:	23/sept/2017
IDE name:	Arduino IDE
SDK version: v3.0-dev-745-gc4e65d6a
Flash Frequency:	80Mhz
Upload Speed:	512000

Description:

I am using OTA to upgrade my firmware but after a software update Serial2 doesn't work properly
Apparently, if I make a hardware reset, the serial port works fine but if ESP.restart() is used the serial port doesn't work.

Doing a little investigation, ESP.restart call to esp_restart(); defined in system.h and I found, the library header. Then the problem is clear UART1 and UART2 weren't reset.

how can I reset those UARTs?

/**
  * @brief  Restart PRO and APP CPUs.
  *
  * This function can be called both from PRO and APP CPUs.
  * After successful restart, CPU reset reason will be SW_CPU_RESET.
  * Peripherals (except for WiFi, BT, UART0, SPI1, and legacy timers) are not reset.
  * This function does not return.
  */
void esp_restart(void) __attribute__ ((noreturn));

Best wishes

me-no-dev:
What do you see as errors? Old data? Generally SerialX.begin should reinit the serials.

kurtow:
Board:  ESPRESSIF ESP32-DEVKITC  based on ESP-WROOM-32
Core Installation/update date: 23/sept/2017
IDE name: Arduino IDE

Hello,
I have the same problem.
After a hardware reset, the Serial2 runs with a probability of about 50% 
(sometimes  I had to reset three or four times before it starts)

After restarting from deepsleep Serial2 runs with a probability of 20%.
(In most cases its hanging after restart)
I mean (but I am not sure), that this problem did not occur 
at my last arduino-core download of 21.08.

kindly regards

jestebanes:
Hello @me-no-dev , I see a lot of crap

This is a communication after one hard reset or powering up the device, using UART2 connected to pins 16 & 17 at 19200 bauds, the first line is a question sent by ESP32 to the serial client and the second line is the response of the client on the serial bus, the numbers before : are the timeline. The data is obtained with an external hardware that sniffs the serial bus. Every line ends with an ACK, client sent ACK to ESP32 and ESP32 send ACK to the client

P300138824:031D58C0156F5C 1A
P300138833:03C059C015C9E9 CA 

[...]

Issue #1314, opened 2018-04-14
Cannot use UART0/UART1/UART2 concurrently on ESP32 using Arduino IDE


zamorae:
### Hardware:
Board:							ESP-WROOM-32
Core Installation/update date:	       11/jul/2017
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
Upload Speed:				       115200

### Description:
Hi everyone,

I am working on a project in which I need to have an ESP32 connected via serial to another ESP32 and at the same time to a Nextion HMI. I also need to communicate from the central ESP32 to these 2 devices concurrently ( having the serial communications in the "central" ESP32 open to both devices and issuing commands simultaneously)
I have been making some testing in preparation to the interfacing to these devices by replacing the Nextion with an ESP826MOD:

Central ESP32: ESP-WROOM-32 devkit, connected to a USB port of a PC
Client Device A: ESP-WROOM-32 devkit connected to USB port of another PC
Client Device B: ESP8266MOD devkit connected to a USB port of the same PC as Central ESP32

I load a simple sketch (Based on serial passthrough example by Erik Nyquist) on the central ESP32 to open UART1 and UART2. The idea is to open the Serial Monitor connected via the USB port to each device and issue simple strings, which I expect reflected on any of the other Serial Monitors from A and B.
 For devices A and B,  I setup UART1 for the ESP8266MOD (Device B - TX/RX pins) and UART2 on the 2nd ESP32 (Device A - pins 16 & 17).  I then proceed with the following sketches (Based on serial passthrough example by Erik Nyquist):

### Sketch:
```cpp

//Central ESP32

#include <HardwareSerial.h>

HardwareSerial Serial1(1);
HardwareSerial Serial2(2);

void setup() {
  // initialize serial:
  Serial.begin(115200);
  Serial1.begin(115200,SERIAL_8N1, 4, 2);    //Baud rate, parity mode, RX, TX
  Serial2.begin(115200,SERIAL_8N1, 16, 17);
}

void loop() {
  
   if(Serial.available())                   //if user typed something through the USB...
    {    
      Serial1.write(Serial.read());         //Push it through port 1 to Device B
      Serial2.write(Serial.read());         //Push it through port 2 to Device A
     
      Serial.flush();                       //erase any data on serial port after operation 
    }
 
   if (Serial1.available())                 //If data is received through Serial 1 from Device B...
    {    
       Serial.write(Serial1.read());        //Show data on Serial Monitor
[...]

PR #1408, opened 2018-05-14
Port UART reset fix from ESP-IDF

chemicstry:
`uartFlush()` does not work after soft reset due to a hardware bug in ESP32. This was fixed in ESP-IDF by reading out all data from the buffers instead of using reset bits.

Using reset bits on UART1 also corrupts data in UART2. From technical reference manual:
> UART2 doesn’t have any register to reset Tx_FIFO or Rx_FIFO, and the UART1_TXFIFO_RST and UART1_RXFIFO_RST in UART1 may impact the functioning of UART2. Therefore, these two registers in UART1 should only be used when the Tx_FIFO and Rx_FIFO in UART2 do not have any data. (R/W)

Hence UART should only be flushed by reading out.

ESP-IDF commits:
https://github.com/espressif/esp-idf/commit/4052803e161ba06d1cae8d36bc66dde15b3fc8c7
https://github.com/espressif/esp-idf/commit/492db0591d0c5643cb863b76600ab0bf5a204b64

Fixes: #662

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -315,11 +315,14 @@ void uartFlush(uart_t* uart)
     UART_MUTEX_LOCK();
     while(uart->dev->status.txfifo_cnt);
 
-    uart->dev->conf0.txfifo_rst = 1;
-    uart->dev->conf0.txfifo_rst = 0;
+    //Due to hardware issue, we can not use fifo_rst to reset uart fifo.
+    //See description about UART_TXFIFO_RST and UART_RXFIFO_RST in <<esp32_technical_reference_manual>> v2.6 or later.
 
-    uart->dev->conf0.rxfifo_rst = 1;
-    uart->dev->conf0.rxfifo_rst = 0;
+    // we read the data out and make `fifo_len == 0 && rd_addr == wr_addr`.
+    while(uart->dev->status.rxfifo_cnt != 0 || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
+        READ_PERI_REG(UART_FIFO_REG(uart->num));
+    }
+
     UART_MUTEX_UNLOCK();
 }
 


chemicstry:
And probably this one: #1314

Rob58329:
The above "Port UART reset fix from ESP-IDF #1408 [chemicstry]" fixed my ESP32 issue with corrupt Serial-ports after an “ESP.restart()” – very many thanks!

[PS. FYI:  I was basically just using “while (Serial2.available()) Serial.write(Serial2.read()); while (Serial.available())  Serial2.write(Serial.read());” to transfer data to & from a GSM-module.]

me-no-dev:
@chemicstry will you please make the adjustment that I requested so I can merge this? Thanks :)

chemicstry:
@me-no-dev I do not see any comment where you requested adjustments. Could you repeat?

me-no-dev:
Not seeing this?

screen shot 2018-06-01 at 10 47 28

negativekelvin:
@me-no-dev 

> Before you submit your review, your line comments are pending and only visible to you

chemicstry:
@me-no-dev At first I tried using `uart->dev->fifo.rw_byte` but compiler optimises dummy reading and it does not work unless I write some ugly assembler stuff. `READ_PERI_REG` macro is deeply integrated into IDF and I was unable to trace back how it actually works. Is there an example in arduino code on how to dummy read register or maybe you have a suggestion on how to do it?

Regarding the interrupts @sysizlayan. When `uartFlush()` is called during UART initialization, ISR is not enabled yet and it should not spit any data out. However, `HardwareSerial::flush()`, which uses `uartFlush()` has different behaviour and I think different function should be used. `flush` is intended to wait for data transfer to complete instead of reseting. Maybe `uartFlush()` should really be `uartReset()`.

me-no-dev:
@negativekelvin holly cow! I was unaware of this... thanks for pointing out!

@chemicstry maybe use some dummy holder for the bytes?
```cpp
uint8_t dummy;
while(uart->dev->status.rxfifo_cnt != 0 || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
  dummy = uart->dev->fifo.rw_byte;
}
```

brianrho:
As @chemicstry said, Serial.flush() behaves differently than expected of the Arduino API, clearing not only the TX buffer but also the RX buffer. Are there any plans to change this?

PR #1413, opened 2018-05-16
ADD: SPISlave code

andrewjaykeller:
SPISlave code that is the same as esp8266 implementation.


me-no-dev:
Hi!
There are a few issues in this code :)
First it should be part of the SPI class itself (an SPI bus can either be master or slave and not both).
Then you have some calls to IDF API while Arduino counterparts exist.
Also I did not see anywhere the option to specify which bus to use (it's hardcoded).
In the Master example you need to use `beginTransaction` and `endTransaction` and not just `transfer` to make the code thread safe. There is probably more, but let's start with those first :)

PR #1416, opened 2018-05-17
Fix starting WiFi in wrong mode, when persistance is disabled.

everslick:
This commit does two things:

+ It completely disables NVS when Wifi.persistent(false) was called.
+ It fixes getMode() to return WIFI_MODE_NULL as long as esp_wifi_start() was not called. This prevents enableSTA() and enableAP() from setting the wrong mode.


everslick:
again, i did not realize, that tcpipInit() was called from outside, because its declaration was not to be found in the header. i added it back.

PR #1424, opened 2018-05-22
Use esp_timer_get_time as time source for micros and mills

igrr:
[`esp_timer_get_time`](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/esp_timer.html#obtaining-current-time) returns monotonic time in microseconds, as a 64-bit number. It can be called from tasks and interrupts, does not use any critical sections/mutexes, and is thread safe. Calls to it can be made while cache is disabled. Since IDF commit https://github.com/espressif/esp-idf/commit/0d65f3b7f9a1dace8853719871d920d80aa3604b, the time returned by esp_timer is also updated after light sleep: it appears to the application as if the timer was running during light sleep.


me-no-dev:
Awesome! How long does it take to call `esp_timer_get_time`? Also please remove this line: https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/main.cpp#L17

PR #1436, opened 2018-05-25
Added IIC pins

timkoers:
Added IIC pins. I can confirm that this is working!


timkoers:
@me-no-dev As said above, I can confirm that the IIC protocol is working. I've tested it with a sensor and it is detecting the sensor.
If you'd like to read more about the IIC (and view the Olimex example) 
[Click here](https://github.com/OLIMEX/ESP32-GATEWAY/blob/master/SOFTWARE/MOD-RTC/main/main.c)

me-no-dev:
Thanks :)

PR #1480, opened 2018-06-09
add support for TXT records in mDNS query responses

horihiro:
This patch is for adding methods which get TXT record values from mDNS result.


horihiro:
same as https://github.com/esp8266/Arduino/pull/3107

me-no-dev:
Thanks @horihiro ! Merged

PR #1506, opened 2018-06-17
Add "ODROID ESP32" boards support

seungchar:
Add new ODROID-GO board based on ESP-WROVER-32 (128MiB ROM).

https://bit.ly/2lgWvGA

We are going to release it next week.


me-no-dev:
Thanks :) Merged!

PR #1508, opened 2018-06-18
Add component CMakeLists.txt file for use with CMake-based build system

projectgus:
ESP-IDF has a new CMake-based build system currently in preview release.

[More details here](https://esp32.com/viewtopic.php?f=13&t=5559&p=24109)

This PR adds a CMakeLists.txt file so arduino-esp32 component can be used with CMake-based IDF projects.

The one unusual restriction here is that the directory containing the component must be named `arduino-esp32`. Will fix this restriction before the CMake branch is made the default branch in ESP-IDF.


PerMalmberg:
@projectgus Using file-globbing is not recommended as per the CMake documentation.

https://cmake.org/cmake/help/latest/command/file.html

> Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

projectgus:
@PerMalmberg That's right. However, within the IDF build system we do some tricks to make globbing more viable:
https://docs.espressif.com/projects/esp-idf/en/feature-cmake/api-guides/build-system.html#file-globbing-incremental-builds

That said, I have forgotten to use any of those tricks here - thanks for the reminder. Tomorrow I'll amend this PR to either use `git_describe` to track the arduino-esp32 git revision, or name all the source files individually.

PerMalmberg:
@projectgus Ah, yes that does make a difference. 👍

me-no-dev:
@projectgus ping me please when you are ready :)

projectgus:
@PerMalmberg Redone with individual file names, as I realised there's no guarantee that arduino-esp32 will be a git repo in all cases.

@me-no-dev PTAL. Will send a follow-up MR once we have support in IDF for overriding component names, but this should work for now.

projectgus:
@me-no-dev Now with Travis checks to make sure the source file list is up to date.

me-no-dev:
@projectgus please remove the script calling line from travis.yml because it will cause merge conflicts with the update branch that I am working on. We have moved scripting out of yml and I can add the line manually to the appropriate script :) thanks!

PR #1516, opened 2018-06-19
freeRTOS based schedule

hreintke:
This PR implements a FunctionQueue class which provides scheduling functionality for ESP32.

Each FunctionQueue executes as an independent RTOS task, pinned to core 1, priority 1 (equal to loop).
The actual queue is implemented as a RTOS queue.
bool FunctionQueue::scheduleFunction(std::function<void(void)> sf) is used for scheduling.

```
FunctionQueue myFQ();
myFQ.scheduleFunction([](){Serial.printf("Scheduled print\r\n");
```
If instantiated with bool parameter = true, `FunctionQueue myFQ(true);` the scheduled functions will be executed in sync with loop (loop - scheduled - loop - scheduled). That provides the same functionality as in ESP8266.

The Ticker class is updated to have the scheduled versions use the FunctionQueue.

Comments welcome


cores/esp32/FunctionQueue.cpp
```diff
@@ -0,0 +1,109 @@
...
+    	}
+    	else
+    	{
+    		_this->processQueueItem(); // Will block when queue is empty
+    	}
+    }
+    vTaskDelete(NULL);
+}
+
+FunctionQueue FQ;
```
chrissicool, line 108: Other libraries allow to not automatically instantiate global variables.

Do we need to obey NO_GLOBAL_INSTANCES here?

PR #1524, opened 2018-06-22
Exclude {min, max} macros when using <functional>.

Crazor:
Those defines not only break `<vector>`, but `<functional>` as well.


Crazor:
Why is this even redefined? Wouldn't it be better to just `#include <algorithm>`?

me-no-dev:
It's a remanence from AVR :) let's go with 8266 way ;) seems that they had the same issue and solved it after esp32 was created.

yoursunny:
> let's go with 8266 way

#1783 implements this solution.
I've been fed up with this problem for several years.

PR #1525, opened 2018-06-22
Initial version of rmt driver

david-cermak:
This is an initial version of RMT driver which supports
* initialise/allocate channel buffers based on the data needed
* sending data spanning over multiple channel data field
* sending more data within one channel data buffer with interrupt scrolling (continuous mode)
* synchronous and asynchronous reception

Please look mainly at the interface if this is viable and convenient for users; Please comment.

Still missing feature for using carrier frequency and a lot of testing needed...


lbernstone:
The header file needs a define at the beginning to prevent reloading.
```
#ifndef MAIN_ESP32_HAL_RMT_H_
#define MAIN_ESP32_HAL_RMT_H_
```

libraries/ESP32/examples/RMT/RMTWriteNeoPixel/RMTWriteNeoPixel.ino
```diff
@@ -0,0 +1,84 @@
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "freertos/event_groups.h"
+#include "Arduino.h"
+
+#include "esp32-hal.h"
+
+#define LED_MATRIX_SIZE 8*4*24
```
me-no-dev: Let's make `8*4` to be `NUMBER_PIXELS`.

libraries/ESP32/examples/RMT/RMTWriteNeoPixel/RMTWriteNeoPixel.ino
```diff
@@ -0,0 +1,83 @@
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "freertos/event_groups.h"
+#include "Arduino.h"
+
+#include "esp32-hal.h"
+
+#define NR_OF_LEDS   8*4
+#define NR_OF_PIXELS 24*NR_OF_LEDS
```
me-no-dev: This should be number of bits. each led is a pixel

libraries/ESP32/examples/RMT/RMTWriteNeoPixel/RMTWriteNeoPixel.ino
```diff
@@ -0,0 +1,83 @@
...
+void loop() 
+{
+    // Init data with only one led ON
+    int led, col, bit;
+    int i=0;
+    for (led=0; led<NR_OF_LEDS; led++) {
+        for (col=0; col<3; col++ ) {
+            for (bit=0; bit<8; bit++){
+                if ( (color[col] & (1<<(8-bit))) && (led == led_index) ) {
+                    led_data[i].val = 0x80070006;
```
me-no-dev: what is this value? please use the struct fields

libraries/ESP32/examples/RMT/RMTWriteNeoPixel/RMTWriteNeoPixel.ino
```diff
@@ -63,9 +63,15 @@ void loop()
         for (col=0; col<3; col++ ) {
             for (bit=0; bit<8; bit++){
                 if ( (color[col] & (1<<(8-bit))) && (led == led_index) ) {
-                    led_data[i].val = 0x80070006;
+                    led_data[i].level0 = 1;
+                    led_data[i].duration0 = 8;
+                    led_data[i].level1 = 0;
+                    led_data[i].duration1 = 1;
                 } else {
```
stickbreaker: These values are different,. a direct conversion between the `.val` and fields would be
```c++
led_data[i].level0 = 1;
led_data[i].duration0 = 7;
led_data[i].level1 = 0;
led_data[i].duration1 = 6;
```
Did you want to change the values?

Chuck

david-cermak: Thank you chuck, good catch!

Indeed I meant to update the values (to be in line with the diagrams), but not that much. wanted to use pulses 8/4 instead of 7/6 (the led's are quite tolerant so it's accepted)

PR #1533, opened 2018-06-25
ArduinoOTA endless loop

lbernstone:
udp.read returns a negative number on errors.  Since value was defined as a char, it has no way to trap this, so it continues to add FF to the value in an endless loop.


me-no-dev:
thanks :)

PR #1535, opened 2018-06-26
Allow using argument with attachInterrupt

bertmelis:
split __attachArgument and __attachArgumentArg like on ESP8266.
it could pave the way for functional Interrupt.

### DIFF
--- a/cores/esp32/esp32-hal-gpio.c
+++ b/cores/esp32/esp32-hal-gpio.c
@@ -70,6 +70,7 @@ const DRAM_ATTR esp32_gpioMux_t esp32_gpioMux[GPIO_PIN_COUNT]={
 };
 
 typedef void (*voidFuncPtr)(void);
+typedef void (*voidFuncPtrArg)(void*);
 static voidFuncPtr __pinInterruptHandlers[GPIO_PIN_COUNT] = {0,};
 
 #include "driver/rtc_io.h"
@@ -208,7 +209,11 @@ static void IRAM_ATTR __onPinInterrupt(void *arg)
         do {
             if(gpio_intr_status_l & ((uint32_t)1 << pin)) {
                 if(__pinInterruptHandlers[pin]) {
-                    __pinInterruptHandlers[pin]();
+					if(arg){
+						((voidFuncPtrArg)__pinInterruptHandlers[pin])(arg);
+					} else {
+						__pinInterruptHandlers[pin]();
+					}
                 }
             }
         } while(++pin<32);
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-gpio.c
```diff
@@ -244,6 +253,10 @@ extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int intr_type)
     esp_intr_enable(gpio_intr_handle);
 }
 
+extern void __attachInterrupt(uint8_t pin, voidFuncPtr userFunc, int intr_type) {
+	__attachInterruptArg(pin, (voidFuncPtrArg)userFunc, NULL, intr_type);
```
weyoui: The casting (voidFuncPtrArg) looks redundant.

cores/esp32/esp32-hal-gpio.c
```diff
@@ -208,7 +209,11 @@ static void IRAM_ATTR __onPinInterrupt(void *arg)
         do {
             if(gpio_intr_status_l & ((uint32_t)1 << pin)) {
                 if(__pinInterruptHandlers[pin]) {
-                    __pinInterruptHandlers[pin]();
+					if(arg){
```
weyoui: It looks too much indentation in here. Are there tabs or gitub change like this?

weyoui:
Good job. looks good to me. Left a minor comments.

cores/esp32/esp32-hal-gpio.c
bertmelis: yes, but then I get a compiler warning.

me-no-dev:
no alias for __attachInterruptArg? how do you call it?

bertmelis:
My idea was to have it not directly available because it is also not the case for ESP8266 nor AVR-Arduino.
If you want to use it, you'll have to `extern "C"` the function definition.

It should pave the way for FunctionalInterrupt.

me-no-dev:
@bertmelis why not add an example then :)

bertmelis:
Hehe, it doens't work. Obviously. I used it with only 1 interrupt. And the code as it stands now shares the ISR but it also shares the argument. So it only takes the argument tht is used first.

Solution would be to expand the __pinInterruptHandlers-array with the argument (make it a struct) and use it in the ISR.

bertmelis:
I made a very simple example but somethings not working: when I detach from within an interrupt it doesn't detach. Now I just "simulate" a button by touching the pins with a wire to 1 press are like 56 presses. But I suspect the detachInterrupt is not working in the master codebase neither, when called from within an ISR.

Shouldn't detachInterrupt be marked IRAM_ATTR, and does this solve the issue?

me-no-dev:
nooo you can not call it in ISR ;) it calls `esp_intr_disable`/`esp_intr_enable`. Maybe add `detachInterruptFromISR` that does not call them. And please add `void attachInterruptArg(uint8_t pin, voidFuncPtrArg handler, void* arg, int mode);` to the API like the rest

bertmelis:
> And please add void attachInterruptArg(uint8_t pin, voidFuncPtrArg handler, void* arg, int mode); to the API like the rest

Done!

hreintke:
Now he attachInterruptArg is a "public" function instead of a "core only" function, 
I cannot use the knowledge that the arg is a pointer to InterruptArgStructure which holds the functionalInterrupt routine. 
Need to find another solution for FunctionalInterrupt.

bertmelis:
Can't you just use the same logic as for esp8266? There's not that much different. And the alias is "weak", so you can override.

bertmelis:
@hreintke When releasing a functional interrupt, you can just remove the "extern" alltogether.

hreintke:
@bertmelis see #1728

PR #1539, opened 2018-06-26
 Update IDF to aaf1239

me-no-dev:


bperrybap:
Hey this merge includes so changes in Wire that breaks the Wire library for many existing applications.
This merge includes a new endTransmission(bool) method. While having a bool function would be nice, unfortunately, since the current values being passed into endTransmissiont(uint8_t) are 0 or 1 this becomes ambiguous with the new endTransmission(bool) function.
No other core has implemented this endTransmission(bool) function.
So unfortunately, In order to maintain compatibility with the existing Arduino Wire API, and existing applications that use the current endTransmission() function as endTransmission(0) or endTransmission(1) the new endTransmission(bool) function will need to be removed.
Issue for this is here: https://github.com/espressif/arduino-esp32/issues/1725

PR #1544, opened 2018-06-28
Added OTAWebupdated

karan6190:
Added example related to  OTA through Web Page
User can able to update the firmware by uploading the binary file of the code through Web page (user and device must be same LAN )
esp32login
esp32upload


libraries/ArduinoOTA/examples/OTAWebUpdater/OTAWebUpdater.ino
```diff
@@ -0,0 +1,160 @@
+#include <WiFi.h>
+#include <WiFiClient.h>
+#include <ESP32WebServer.h>
```
me-no-dev: it is just WebServer. No ESP32 in front

libraries/ArduinoOTA/examples/OTAWebUpdater/OTAWebUpdater.ino
```diff
@@ -0,0 +1,160 @@
...
+#include <ESPmDNS.h>
+#include <Update.h>
+
+const char* host = "ESP32";
+const char* ssid = "xxx";
+const char* password = "xxxx";
+
+ESP32WebServer server(80);
```
me-no-dev: same here

libraries/ArduinoOTA/examples/OTAWebUpdater/OTAWebUpdater.ino
karan6190: yes, WebServer is required for Zero config routing protocol (http://hostname.local), just to detect your device in LAN connectivity. if WebServer not used then everytime the user has to detect Ip of esp32 
And eventually, Firmware will be uploaded in esp32 only.

me-no-dev: Yes... but the name of the lib is not ESP32WebServer, but just WebServer ;) I get why it's needed :D

me-no-dev: https://github.com/espressif/arduino-esp32/blob/master/libraries/WebServer/src/WebServer.h

me-no-dev:
Thanks :)

JoanMCD:
The user and PW is visible in the html page... (rightclick, 'view page source' in Browser). It suggests a false security !

Issue #1304, opened 2018-04-10
BluetoothSerial shows up as audio device on macOS, immediately disconnects


gohai:
### Hardware:
Board:							Adafruit ESP32 Feather
Core Installation/update date:			April 10 2018
IDE name:							Arduino 1.8.5
Flash Frequency:					80MHz
Upload Speed:						921600

### Description:
Running the SerialToSerialBT example shows up as an "audio" device on macOS 10.13.4. The device disconnects on its own soon after pairing. Thus, it cannot be used currently. (See screenshot for details.)

<img width="780" alt="screen shot 2018-04-10 at 9 46 42 am" src="https://user-images.githubusercontent.com/4945451/38571327-0cd8149a-3ca5-11e8-90df-bdcb6c1e59fc.png">

beegee-tokyo:
ESP32 bluetooth is showing as speakers on Android as well, but Bluetooth Serial is working fine as shown in my little [tutorial](https://desire.giesecke.tk/index.php/2018/04/10/esp32-wifi-setup-over-bluetooth-serial/). Cannot not test on MAC or iOS though.

gohai:
@copercini This define appears to set the default via `p_bta_dm_cfg`, but shouldn't it be possible to re-configure this e.g. through [BTM_SecAddDevice](https://github.com/espressif/esp-idf/blob/f620a3c4f35cb715e5b6e37e8556ec2278628660/components/bt/bluedroid/stack/btm/btm_dev.c#L41) or [BTM_SetDeviceClass](https://github.com/espressif/esp-idf/blob/f620a3c4f35cb715e5b6e37e8556ec2278628660/components/bt/bluedroid/stack/btm/btm_devctl.c#L552)?

gohai:
I looked at the output of macOS' _Console_ application while I attempted to connect to the device: the process `bluetoothaudiod` attempts a couple of times to creates an audio device, realizes that `No protocols known for device or services are nil`, and finally gives up, at which point `bluetoothd` removes the device. So this confirms the suspicion that the device class plays a role here.

lbernstone:
You can update the code yourself if you want to test.  https://github.com/espressif/arduino-esp32/issues/1142

gohai:
@lbernstone Thank you for the pointer. I just tried this, following your instructions - would you mind helping me a bit further?

[...]

PR #1556, opened 2018-06-28
BluetoothSerial: set COD to be compatible with macOS (#1304)

gohai:
Device now shows up like so, and - most importantly - can be used on macOS machines. Perhaps COD can be tweaked for a bit more modern (mobile?) icon, but I'll leave this to others

screen shot 2018-06-28 at 1 31 28 pm


gohai:
pinging @copercini

me-no-dev:
it has been working on Mac for me the whole time
and comes up as BT device and not a computer. What other CODs are available?

gohai:
@me-no-dev Please see #1556

me-no-dev:
OK, but can we find a better icon? I do not find that one to be a good choice. There should be some icon to signify serial port or other bus

gohai:
Please change COD to your liking. I was only using this particular one because it was suggested elsewhere as making BluetoothSerial work.

copercini:
Here is the list: https://www.bluetooth.com/specifications/assigned-numbers/baseband

me-no-dev:
OK How about:

`cod.major = 0b00001;` - Phone (cellular, cordless, pay phone, modem, ...)
`cod.minor = 0b000100;` - Wired modem or voice gateway
Service I am not sure... maybe 22?

me-no-dev:
I tried this and I tried a few others... shows the standard icon for BT device (the blue logo)

gohai:
But you also saw this icon before @me-no-dev, didn't you? I can probably test your suggested values tomorrow, and report back

me-no-dev:
@gohai I have the feeling that Mac OS caches the icon in a way, but I can not prove it.

gohai:
@me-no-dev I was able to test the values you suggested. They seem to work for me on macOS 10.13 - seeing a Bluetooth icon also. Updated the commit accordingly.

dalazarev-telegram:
Hello, before this fix my iphone could recognize, but could not connect to the device. After this fix iphone cannot detect the device. Android can pair successfully. According to the sample app, iphone should be able to pair with the device (sample sketch is used)

PR #1625, opened 2018-07-12
Spurious Interrupts Temporary fix 20180711

stickbreaker:
the 'eject' ERROR is and indication of an interrupt triggering without an source.  I am working to eliminate these unserviceable interrupts.  This update increase stability on a HelTek Wifi Lora 32 board. with a SSD1306 OLED.  This update fixes a glaring error in the interrupt allocation code, the Interrupt mask was wrong.  I also dynamically adjust the FiFo thresholds based on Bus clockrate. The change to FiFo thresholds has reduced the number for 'eject' events.  I also change 'eject' from and ERROR to DEBUG.  An 'eject' event does not compromise i2c transmissions. It happens after a transaction has completed. 

Chuck.

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -491,6 +491,18 @@ static void IRAM_ATTR fillTxFifo(i2c_t * i2c)
 
     if(!full || (a >= i2c->queueCount)) { // disable IRQ, the next dq will re-enable it
         i2c->dev->int_ena.tx_fifo_empty=0;
+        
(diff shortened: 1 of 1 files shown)


me-no-dev:
Merged :)

helmut64:
Hi Chuck, can it be that something serious got wrong lately with the I2C support, I updated Arduino ESP32 support and my I2C devices (e.g. ds3231, SI7021) don’t work anymore.

helmut64:
Just a side node, instead of the Heltec ESP32 Board it is way better use our ECO Power Board, see www.radioshuttle.de.

PR #1631, opened 2018-07-13
Code linking with Arduino1.8.5 failed

rmshub:
While linking the code, it invoking esptool.exe from esptool folder ( tools/esptool/esptool.exe). However, there is no folder "esptool" under "tools" folder in the code package. We getting an error exec: "C:\\Users\\Username\\Documents\\Arduino\\hardware\\espressif\\esp32\\tools\\esptool\\esptool.exe" file doesn't exist. Kindly check the issue.
OS: Windows 10


me-no-dev:
you must have not read through the instructions and you missed to run get.exe ;) that will put esptool where it needs to be

PR #1635, opened 2018-07-13
Remove unused variable

ciband:
From platformio:

WiFiMulti.cpp:129:27: warning: variable 'mac' set but not used [-Wunused-but-set-variable]


me-no-dev:
it is used when debug is on ;) a couple of lines below the one you removed.

ciband:
Sorry about that.  I completely missed that.

PR #1650, opened 2018-07-17
expose post args during upload

luc-github:
Currently post args are only available at the end of upload but they are already listed - this PR just expose them with minimal changes
It also set a define for post args array size originaly set to 32


me-no-dev:
looks good @luc-github :)

PR #1653, opened 2018-07-17
Selective compilation

nathanjel:
1. This allows to unselect specific modules from build. Enables speedup of compilation time and also enables and reduces size where whole archive is needed for late linking, e.g. when using something like `COMPONENT_ADD_LDFLAGS := -Wl,--whole-archive -larduino-esp32 -Wl,--no-whole-archive` 
2. Some modules (e.g. Azure IoT) will not eventually build, as the source path wildcards are not deep enough. This is fixed in this code.

### DIFF
--- a/Kconfig.projbuild
+++ b/Kconfig.projbuild
@@ -106,8 +106,147 @@ config AUTOCONNECT_WIFI
     bool "Autoconnect WiFi on boot"
     default "n"
     depends on AUTOSTART_ARDUINO
+    depends on ARDUINO_SELECTIVE_WiFi
     help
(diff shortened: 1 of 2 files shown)


me-no-dev:
I think you are working on an old component.mk. see the conflict above ;) it has to do with the Azure issue you mention.
While it is nice to have such feature, we are moving towards Cmake which might make this task easier to implement and support in the future.

nathanjel:
Will this be accepted post conflict resolution? It would greatly help us, as we need to compile the whole library as I mentioned above.

PR #1655, opened 2018-07-18
feature/selective-compilation

nathanjel:
Enable selective compilation of Arduino libraries. Makes a big difference for projects which use entire Arduino library for late linking purposes (as in `COMPONENT_ADD_LDFLAGS := -Wl,--whole-archive -l$(COMPONENT_NAME) -larduino-esp32 -Wl,--no-whole-archive`)

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -32,6 +32,10 @@
 #define DEBUG_OUTPUT Serial
 #endif
 
+#ifndef WEBSERVER_MAX_POST_ARGS
+#define WEBSERVER_MAX_POST_ARGS 32
+#endif
+
 static const char Content_Type[] PROGMEM = "Content-Type";
 static const char filename[] PROGMEM = "filename";
 
@@ -383,8 +387,9 @@ bool WebServer::_parseForm(WiFiClient& client, String boundary, uint32_t len){
   client.readStringUntil('\n');
   //start reading the form
   if (line == ("--"+boundary)){
-    RequestArgument* postArgs = new RequestArgument[32];
-    int postArgsLen = 0;
+	 if(_postArgs) delete[] _postArgs;
+	 _postArgs = new RequestArgument[WEBSERVER_MAX_POST_ARGS];
+     _postArgsLen = 0;
     while(1){
       String argName;
       String argValue;
@@ -445,7 +450,7 @@ bool WebServer::_parseForm(WiFiClient& client, String boundary, uint32_t len){
             DEBUG_OUTPUT.println();
 #endif
 
-            RequestArgument& arg = postArgs[postArgsLen++];
+            RequestArgument& arg = _postArgs[_postArgsLen++];
             arg.key = argName;
             arg.value = argValue;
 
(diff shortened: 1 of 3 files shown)


me-no-dev:
I love the idea, but I am wondering if there is any better way to do this. The list of libs will need to be maintained in both files

nathanjel:
Hear You well, and hate it as well ;-) It will be fairly easy to conver this to Kconfig file maintenance only. But this is still not perfect. Any advice from the constructors of IDF build system, if there is a reasonable way to update Kconfig from the makefiles? While I am happy to explore it, not sure if I have ample time to reverse engineer the build process to achieve that.

me-no-dev:
Agreed. I am looking for some more obvious approach. I do not think that here is a way to alter kconfig from makefiles in IDF. But modularisation is needed so we move towards cmake. If you can make it so it requires only Kconfig maintenance, then that is still a win IMO.

Kconfig.projbuild
```diff
@@ -106,8 +106,147 @@ config AUTOCONNECT_WIFI
     bool "Autoconnect WiFi on boot"
     default "n"
     depends on AUTOSTART_ARDUINO
+    depends on ARDUINO_SELECTIVE_WiFi
```
me-no-dev: I do not think that I use `AUTOCONNECT_WIFI` in code... do i? And why the option in general depend on setting that might not even be available if `ARDUINO_SELECTIVE_COMPILATION` is off. Or am I not getting something?

component.mk
```diff
@@ -1,4 +1,128 @@
-ARDUINO_CORE_LIBS := $(patsubst $(COMPONENT_PATH)/%,%,$(sort $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/*/))))
+ARDUINO_CORE_LIBS := $(patsubst $(COMPONENT_PATH)/%,%,$(sort $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/*/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/*/*/*/*/*/*/))))
+
+SPECIFIC_LIBS :=
+
+ifneq "y" "$(CONFIG_ARDUINO_SELECTIVE_ArduinoOTA)"
+MODULE := ArduinoOTA
+SPECIFIC_LIBS += $(patsubst $(COMPONENT_PATH)/%,%,$(sort $(dir $(wildcard $(COMPONENT_PATH)/libraries/$(MODULE)/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/$(MODULE)/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/$(MODULE)/*/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/$(MODULE)/*/*/*/*/)) $(dir $(wildcard $(COMPONENT_PATH)/libraries/$(MODULE)/*/*/*/*/*/))))
```
me-no-dev: can this be made into function that accepts module name as argument and not repeat this long line so many times :) that will help

Kconfig.projbuild
nathanjel: Autoconnect WIFI was there in Kconfig, so I left it as is. The point of dependency there is to ensure You will not autoconnect WIFI if it's not compiled in. In default setting that specific key is enabled (default = all arduino goes in). Indeed, it's never used, so maybe just remove it so that people will not rely on it?

component.mk
nathanjel: I think we both know it can. And also it can be driven by the list of folders and then we will have no need to re-code this file and only need to list components in Kconfig. It also needs time.

nathanjel:
Updated with new commits, respecting comments. Closing this PR.

PR #1659, opened 2018-07-20
UDP Packet with zero data length

Schuemi:
If you receive a package with a data length of zero, parsePacket returns zero, but rx_buffer will be created. So if you perform another parsePacket you will get a zero, even if something could be read. This example would not work: https://www.arduino.cc/en/Reference/EthernetUDPParsePacket

Also I added a check if rx_buffer exist when you try to flush it.


me-no-dev:
Thanks! :)

PR #1663, opened 2018-07-22
Add TTGO LoRa32 with OLED Version 1.x Board

CaptIgmu:


variants/ttgo-lora32-v1/pins_arduino.h
```diff
@@ -0,0 +1,64 @@
...
+static const uint8_t LED_BUILTIN = 2;
+#define BUILTIN_LED  LED_BUILTIN // backward compatibility
+
+static const uint8_t KEY_BUILTIN = 0;
+
+static const uint8_t TX = 1;
+static const uint8_t RX = 3;
+
+static const uint8_t SDA = 21;
+static const uint8_t SCL = 22;
```
stickbreaker: Is, SDA and SCL accurate?  How is the OLED connected? How is the LORA sx1276 connected?  If the OLED is not using 21,22 I would recommend you add additional defines

```c++
// OLed
static const uint8_t OLED_SCL = 15;
static const uint8_t OLED_SDA = 4;
static const uint8_t OLED_RST = 16;

// LoRA
static const uint8_t LORA_SCK = 5;
static const uint8_t LORA_MOSI = 27;
static const uint8_t LORA_MISO = 19;
static const uint8_t LORA_CS = 18;
static const uint8_t LORA_RST = 14;
static const uint8_t LORA_IRQ = 26;
```

It think it would be useful.
Chuck.

cyberman54: And we have LED and button on board:

// LED (attn: is active low!)
static const uint8_t LED_BUILTIN = 2;

// button ("PRG")
static const uint8_t BUTTON = 0;

CaptIgmu: @stickbreaker  The Version 1 boards use pins 4,15,16 for OLED so makes more sense to get these definitions into pins_arduino.h along with the LoRa radio defs. I had these in my example programs at CaptIgmu/Arduino/tree/master/esp32/TTGO_LoRa32_OLED. Thank you Chuck, I will add some commits!

CaptIgmu: @cyberman54 These are defined in pins_arduino.h
```
static const uint8_t LED_BUILTIN = 2;

static const uint8_t KEY_BUILTIN = 0;
```
I don't need the PRG button in my applications but you could re-assign the constant in your program!  Thanks.

stickbreaker:
@CaptIgmu I like those changes, now for the official repo managers to agree.

Chuck.

me-no-dev:
We need to do something about this boards :D the file is becoming unmaintainable

PR #1665, opened 2018-07-22
Fix for spurious interrupts during I2C communications

stickbreaker:
This version no longer needs an interrupt for each byte transferred. It only needs interrupts for START, STOP, FIFO empty/Full or error conditions.  This dramatically reduces the interrupt overhead.  I think the prior version was causing an interrupt overload condition where the ISR was not able to process every interrupt as they happened.

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -33,6 +33,22 @@
 #define DR_REG_I2C_EXT_BASE_FIXED               0x60013000
 #define DR_REG_I2C1_EXT_BASE_FIXED              0x60027000
 
+/* Stickbreaker ISR mode debug support
+
+ENABLE_I2C_DEBUG_BUFFER
+  Enable debug interrupt history buffer, setting this define will result in 1544 bytes of RAM
+  being used whenever CORE_DEBUG_LEVEL is higher than WARNING. Unless you are debugging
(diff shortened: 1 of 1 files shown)


stickbreaker:
@cyberman54 It looks like you are using an old file.  The error message in this file is changed.  You are using rc3.  To test this change copy this file over top of the one in rc3.

on my computer (win7 pro)
>C:\Users\user\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0-rc3\cores\esp32\esp32-hal-i2c.c

Chuck.

cyberman54:
@stickbreaker I thought i did copy the file to the right place. Looks like i'm doing wrong with my platformIO installation. I am not using Arduino IDE, but arduino-espressif32 framework on platformIO. I copied the file to this path:

> ...\.platformio\packages\framework-arduinoespressif32@src-537c58760dafe7fcc8a1d9bbcf00b6f6\cores\esp32\esp32-hal-i2c.c

cyberman54:
@stickbreaker i managed to replace the file (just renamed the old one before, but that doesn't work since it is still compiled). Now it works, the speed is back and no more errors!

Looks like this change in the i2c code was a heavy lifting, thank you for your efforts!

me-no-dev:
always happy to merge such fixes :D

PR #1669, opened 2018-07-23
[OTA Timeout] Added ability set OTA timeout in the OTA client

tsathishkumar:
Currently the timeout for OTA packets is set as 1 sec, which is not good enough in a slower connection. This PR enables the clients to set custom timeout for OTA packets.


tsathishkumar:
Changed to method instead of #define.

me-no-dev:
I also like this 🎉

PR #1674, opened 2018-07-25
fix for "realtime uart" in esp32-hal-uart.c

smartinick:
if you are doing uart stuff that depends on timing, the original code will leave you in the "blind" (Serial.availabe() reads ZERO) until 112 bytes have been received.
Setting this value to 1 will cause 112 times more interrupts, but still use the buffer and not leave your code blind in regards of timing.

i'm aware that his is no real solution, ideally someone (@me-no-dev, @igrr ?) would add some code to HardwareSerial and esp32-hal-uart to expose a new method or optional parameter for serial.begin which could affect this setting.

i'm happy for my project with it and did not encounter side-issues with other projects - there's enough cpu resources left to handle the 112 times higher rx-interrupt rate...

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -96,7 +96,8 @@ static void IRAM_ATTR _uart_isr(void *arg)
 void uartEnableInterrupt(uart_t* uart)
 {
     UART_MUTEX_LOCK();
-    uart->dev->conf1.rxfifo_full_thrhd = 112;
+    //uart->dev->conf1.rxfifo_full_thrhd = 112; //original
+    uart->dev->conf1.rxfifo_full_thrhd = 1; //fix for "realtime" uart
     uart->dev->conf1.rx_tout_thrhd = 2;
     uart->dev->conf1.rx_tout_en = 1;
     uart->dev->int_ena.rxfifo_full = 1;


me-no-dev:
oh no :D you get bytes from the buffer either when the fifo threshold is reached or when there is sufficient pause between bytes on the UART to trigger TOUT. For your code to wait it means that the transmission is still ongoing. I can not accept a PR that will cause Interrupts 112 times more ;) This will be so wrong on many levels if baudrate is high.
I would accept a PR that exposes a function which lets you change the fifo full threshold though (with some explanation on what it does and what it causes).

stickbreaker:
@me-no-dev have Serial.available() initiate a flush of the fifo if the Buffer is empty?  That way if a APP wants to hammer the UART with continuous `available()` calls no harm is done, ISR driver is still waiting for FifoFull unless `available()` is called. The `available()` flushes the fifo into Buffer and `available()` just feeds from Buffer.  Unless the APP is faster than the Serial rate, `available()` would only trigger the flush intermittently. 

If this idea worked, the same code could be added to `read()`, if Buffer is empty trigger a FifoFlush.

It would just have to interleave with the ISR on fifo Empty calls, perhaps a MUTEX?   On second thought, I can see race conditions and deadlocks. I wonder what would happen if APP level, `Serial.available()` got the MUTEX, but before it started emptying Fifo, the FifoFull interrupt trigger?  I don't know the UART ISR code.   Does a `spinlock()` require less overhead then a FreeRTOS MUTEX(semaphore)?  Could the `interrupt_rxFifoEmpty()` be directly called from APP level Code?

I don't have time to work on it,  but it seems to be an interesting problem.

Chuck.

me-no-dev:
@stickbreaker you came and just complicated things :D I might have an idea, caused by your idea

stickbreaker:
@me-no-dev You're welcome, I'm always happy to cause others more work! :man_dancing: :fireworks: :champagne: 

(arn't emojis useless!)

Chuck.

smartinick:
Wow, i'm happy about the responses and thanks to you people with way more insight/knowledge of the sources thinking about solutions!

if there's no better fix, i can change, test and create another pull request with changes to hardwareserial/esp32-hal-uart to either add a new method or extend the parameters of serial.begin.
just let me know, but that might take some time to finish.

the reason for my change is simple: i have uart communication where i have a timeframe of approx. 4-5ms to send data. and with 115200bps and 112 bytes until the app "sees" a new byte that's no good....

shall we create a issue for this or leave communications on here...?  i don't care.... but this PR is obsolete in either way ;)

Thanks!

PR #1677, opened 2018-07-25
WebServer Speed Up

MaValki:
Functions _uploadReadByte and _parseForm were modified in order to speed up uploading data. 
Now there is no need to call time consuming client.connected() method.


me-no-dev:
WooHoo! You made it :D

PR #1690, opened 2018-07-28
Fix WiFiMulti Logs

copercini:


psykokwak-com:
Please, add ".c_str()" on line 128 : log_d("[WIFI] SSID: %s", WiFi.SSID().c_str());

copercini:
@psykokwak-com done =)

PR #1727, opened 2018-08-05
Update Arduino/hardware path

tnoborio:
Changed Adruino direcgtory path because it was explanation directory for mac


me-no-dev:
Thanks for finding this :)

PR #1728, opened 2018-08-05
Functional interrupt

hreintke:
Add the functional option to attachInterrupt.
Implementation is similar to ESP8266


me-no-dev:
Please add example :)

hreintke:
I will make the example.
Where do you want me to put it ? 
I only see example directories with libraries.

me-no-dev:
https://github.com/espressif/arduino-esp32/tree/master/libraries/ESP32/examples
the ESP32 library is used exactly for examples related to the core :) add a GPIO folder there, then move the GPIOInterrupt sketch and folder there, renaming both to BasicGPIOInterrupt, then add your example with new name. If you do it all good, both examples will show in the IDE under File->Examples->ESP32->GPIO

hreintke:
@Elijahg : 
I don't know what you want to achieve but ...

The sketch in your post works for me. 
Be aware that after 10 seconds the button1 interrupt is disabled,  from that moment onwards only button2 will be in use.

You post this in the Functional Interrupt PR, however your sketch doesn't use that functionality.

m0x72:
@Elijahg Commit 879388e170452f056a2a63bceb83f23f63ce9a06 should have fixed the error you're experiencing with `attachInterruptArg`.

PR #1730, opened 2018-08-05
Updated ISSUE_TEMPLATE to try to get better postings

lbernstone:


copercini:
What do you think about ask for PSRAM use too?

me-no-dev:
I agree for PSRAM :) @lbernstone will you please add it?

PR #1767, opened 2018-08-17
Improve I2C bus recovery of Interrupted READ cycle

stickbreaker:
If the esp32 is reset during a i2c read cycle the slave device may be in control of the SDA line.  

If the SDA line is held low, the esp32 cannot issue a START or STOP to recover the bus. 

The previous code did not correctly configure the SCL output pin, and it cycled SCL 9 times with SDA Low.  Since the slave device was in a READ cycle, it just continued outputting the bits of the current byte.  When the ACK/NAK bit space occurred, The low output value of SDA was interpreted as ACK so the slave device continued with the next byte.  It never terminated the READ cycle. 

This new code will correctly recover from an interrupted READ

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -1350,33 +1350,32 @@ static void i2cReleaseISR(i2c_t * i2c)
 }
 
 static bool i2cCheckLineState(int8_t sda, int8_t scl){
-    if(sda < 0 || scl < 0){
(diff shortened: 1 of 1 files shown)


vkurpad:
@stickbreaker I am still running into the issue "Guru Meditation Error: Core  1 panic'ed (InstrFetchProhibited). Exception was unhandled." after making a HTTP request and trying to read the MPU data again. I tried it with MQTT and didn't have any luck. It appears to be similar to issue #1352.

PR #1776, opened 2018-08-20
attachInterruptArg should take voidFuncPtrArg as argument

hreintke:
@bertmelis 
When working on an example for the FunctionalInterrupt, I took yours from GPIOInterrupt as base.
However, could not get it to run correctly.

I think the signature of attachInterruptArg is wrong.
With the patch from this PR it works as expected.

Can you take a look ? 

PS : Did not notice when creating FunctionalInterrupt as that implements it's own attach.


bertmelis:
How could I have missed that?

me-no-dev:
so sorry for the delay guys :(

PR #1780, opened 2018-08-21
make ESP32_PHY_CALIBRATION_AND_DATA_STORAGE optional

Curclamas:
### Background
When using arduino-esp32 as IDF component we would like to disable `ESP32_PHY_CALIBRATION_AND_DATA_STORAGE` since we sometimes have different antennae on our boards and we would like to perform a full calibration on default.


me-no-dev:
ahh damn... how can we make it so it's on by default?

Issue #1744, opened 2018-08-09
Cannot enable WPA2 Enterprise using arduino libraries


daniel-dsouza:
### Hardware:
Board:							Adafruit Feather ESP32
Core Installation/update date:			?8/9/2018?
IDE name:						Arduino IDE and Platform.io
Flash Frequency:					80Mhz
Upload Speed:						921600

### Description:
Fails on `esp_wifi_sta_wpa2_ent_enable(&config);`

### Sketch:
```cpp
#include "esp_wpa2.h"
#include <WiFi.h>

const char* ssid = "eduroam";
const char* eap_username = "hunter";
const char* eap_password = "hunter123";

void setup() {
  Serial.begin(115200);
  delay(10);

  Serial.print("Connecting to network: ");
  Serial.println(ssid);

  WiFi.disconnect(true);  //disconnect form wifi to set new wifi connection
  esp_wifi_sta_wpa2_ent_set_identity((uint8_t *)eap_username, strlen(eap_username));
  esp_wifi_sta_wpa2_ent_set_username((uint8_t *)eap_username, strlen(eap_username));
  esp_wifi_sta_wpa2_ent_set_password((uint8_t *)eap_password, strlen(eap_password));
  esp_wpa2_config_t config = WPA2_CONFIG_INIT_DEFAULT();
  esp_wifi_sta_wpa2_ent_enable(&config);

  WiFi.begin(ssid);

  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }

  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address set: ");
  Serial.println(WiFi.localIP()); //print LAN IP
}

void loop() {}
```

### Debug Messages:
```
Enable Core debug level: Debug on tools menu of Arduino IDE, then put the serial output here

Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.
Core 1 register dump:
PC      : 0x400e6f51  PS      : 0x00060530  A0      : 0x800d1ca6  A1      : 0x3ffb1f40  
A2      : 0x3ffb1f74  A3      : 0x3ffc5890  A4      : 0x00000000  A5      : 0x526312a6  
A6      : 0x00000001  A7      : 0x00000000  A8      : 0x00000000  A9      : 0x3ffb1f40  
[...]

Issue #1381, opened 2018-05-04
WPA2 Enterprise PEAP-MSCHAPv2 not working


averri:
### Hardware:
Board:							esp32dev
Core Installation/update date:		04/05/2018
IDE name:						Platform.io
Flash Frequency:					80Mhz
Upload Speed:					115200

### Description:
The ESP32 cannot connect to WiFi using WPA2 Enterprise PEAP/MSCHAPv2. There is an error message reported, please see below.

Error messages when executing the sketch:

```
.E (3449) wpa: Method private structure allocated failure

..E (7006) wpa: Method private structure allocated failure

..E (10559) wpa: Method private structure allocated failure
```

### Sketch:
```cpp
#include "esp_wpa2.h"
#include <WiFi.h>

const char *ssid = "mySSID";
#define EAP_ID "myID"
#define EAP_USERNAME ""
#define EAP_PASSWORD "myPassword"

void setup() {

    Serial.begin(9600);
    delay(10);

    Serial.println();
    Serial.print("Connecting to ");
    Serial.println(ssid);

    WiFi.disconnect(true);
    
    esp_wifi_sta_wpa2_ent_set_identity((uint8_t *)EAP_ID, strlen(EAP_ID));
    //esp_wifi_sta_wpa2_ent_set_username((uint8_t *)EAP_USERNAME, strlen(EAP_USERNAME));
    esp_wifi_sta_wpa2_ent_set_password((uint8_t *)EAP_PASSWORD, strlen(EAP_PASSWORD));

    esp_wpa2_config_t config = WPA2_CONFIG_INIT_DEFAULT(); 
    esp_err_t code = esp_wifi_sta_wpa2_ent_enable(&config);

    Serial.print("Code: ");
    Serial.println(code);
    
    WiFi.begin(ssid);

    while (WiFi.status() != WL_CONNECTED) {
        delay(2000);
        Serial.print(".");        
    }

    Serial.println("");
    Serial.println("WiFi connected");
    Serial.println("IP address: ");
    Serial.println(WiFi.localIP());
}

void loop() {
    // Your code here.
}
```

### Debug Messages:
```
Connecting to mySSID
[D][WiFiGeneric.cpp:293] _eventCallback(): Event: 2 - STA_START
Code: 0
.E (3569) wpa: Method private structure allocated failure

.[D][WiFiGeneric.cpp:293] _eventCallback(): Event: 5 - STA_DISCONNECTED
[...]

PR #1782, opened 2018-08-21
Fix WifiClientEnterprise - STA mode set

martinius96:
Few testers let me know, they were unable with connection to Eduroam network under PEAP+MsCHAPv2 methods.
They were trying many modifications. Solution is very absurd. 
Use function Wifi.mode(); with parameter WIFI_STA for init hardware!
One tester from Italy was able to connect to Eduroam network under TTLS + MsCHAPv2 with that sketch too!

Sketch was tested and worked almost for all testers with that problem (Doesn't worked for EAP-TLS --> where system want certificate from client).

Refer to: 
https://github.com/espressif/arduino-esp32/issues/1381#issuecomment-412073401
https://github.com/espressif/arduino-esp32/issues/1381#issuecomment-414174639
https://github.com/espressif/arduino-esp32/issues/1744#issuecomment-411974121


martinius96:
Merge? :)

martinius96:
@me-no-dev

me-no-dev:
sorry! was quite busy :(

PR #1810, opened 2018-08-28
Add BananaPi-BIT Development Board Support

yelvlab:


lbernstone:
Why not include the ADC pins (A0-A19) or the touch pins (T0-T9)?

yelvlab:
ADC pin is included in P0-P19.Such as P1 is IO32(ADC1_CH4) P2 is IO33(ADC1_CH5) and so on

stickbreaker:
@yelvlab so if I want to do an `analogRead(A0);` it will automagically know to use the correct pin?  Or will it generate a compiler error?

Chuck.

yelvlab:
@lbernstone I know it ,but the user of our board definition is relatively primitive. We are considering avoiding the user's doubts (the same IO multiple definitions A0 and P0), and unifying the pin name definition, then the Analog function or the Digital function of each pin will have The document is written.Later, we will also release a less customized development board, continuing the ESP32 Ax or Dx definition method and Arduino UNO appearance, but we are not ready yet, you can preview it first, the following is github https://github.com/yelvlab/BPI-UNO32

variants/bpi-bit/pins_arduino.h
```diff
@@ -0,0 +1,63 @@
+#ifndef Pins_Arduino_h
+#define Pins_Arduino_h
+
+#include <stdint.h>
+
+#define EXTERNAL_NUM_INTERRUPTS 16
+#define NUM_DIGITAL_PINS 40
+#define NUM_ANALOG_INPUTS 16
+
+#define analogInputToDigitalPin(p) (((p) < 20) ? (esp32_adc2gpio[(p)]) : -1)
```
stickbreaker: This define is used to convert from A[n] to gpio pin.  Since you are choosing not to define analog input pins should you change both `NUM_ANALOG_INPUTS` and `analogInputtoDigitalPin()` to reflect this choice?
```c++
#define NUM_ANALOG_INPUTS 0
#define analogInputToDigitalPin(p) -1
```

Chuck.

yelvlab: I think the definition here can be retained, because we not only have to consider the primary users, but also the people with certain Arduino and ESP32 programming development foundations, they can not use our prepared P in their own code ( n), from the line definition A (n) such a pin, and, our board does not give up the ADC function.

yelvlab

yelvlab: I think the definition here can be retained, because we not only have to consider the primary users, but also the people with certain Arduino and ESP32 programming development foundations, they can not use our prepared P(n) in their own code , from the line definition A (n) such a pin, and, our board does not give up the ADC function.

me-no-dev:
Done! Where can I get one of those boards? I collect all kinds of ESP based things (not TTGOs and similar though). Your board seems to follow the MicroBIT thing :) don't have such yet

yelvlab:
you can find it on AliExpress.

yelvlab:
Hi ,@me-no-dev 

You can find it on AliExpress.[AliExpress Banana pi bit](https://www.aliexpress.com/item/Banana-PI-Bit-board-with-EPS32-for-STEAM-education/32861694375.html?spm=2114.search0104.3.1.183223bd54GQZf&ws_ab_test=searchweb0_0,searchweb201602_1_10065_10068_5017611_10546_10059_10548_10696_100031_10084_10083_10103_451_452_10618_10307_5017711,searchweb201603_45,ppcSwitch_4&algo_expid=4aa10caf-3f98-4658-b641-33375823623e-0&algo_pvid=4aa10caf-3f98-4658-b641-33375823623e&priceBeautifyAB=0)

me-no-dev:
thanks @yelvlab will get one :)

Issue #1824, opened 2018-09-01
serial cannot receive the last byte


piaoyezju:
### Hardware:
Board:							ESP32 Dev Module
Core Installation/update date:			latest
IDE name:							Arduino IDE

### Description:
A few minutes after booting, the serial port will not receive the last byte unless there is new data behind it, and then the new data is still missing the last byte.
The code is as follows, connect P16, P17
please help me ,thanks!

### Sketch:
```cpp
//#include "SPIFFS.h"
#include <WiFi.h>
HardwareSerial USE_SERIAL1(1);
HardwareSerial USE_SERIAL2(2);
#define DEBUG_SERIAL Serial
#define UART1_RX          19
#define UART1_TX          18
#define UART2_RX          16
#define UART2_TX          17
#define UART0_RX          23
#define UART0_TX          22

byte state = 0x00;
unsigned long expireTime = 0;

void startQuery() {
  if (state == 0x00) {   // new query
    if(USE_SERIAL2.available() > 0) {
      DEBUG_SERIAL.println("new query, clear data:" + String(USE_SERIAL2.available()));
      while (USE_SERIAL2.available() > 0) {
        USE_SERIAL2.read();
      }
    }
    expireTime = millis() + 300;
    state = 0x01;
    byte cmd[9] = {0x2b, 0x00, 0x01, 0x00, 0x01, 0xa1, 0x00, 0x00, 0x7e};
    USE_SERIAL2.write(cmd, 9);
  } else if (state == 0x01) {
    if(USE_SERIAL2.available() >= 9) {
      while (USE_SERIAL2.available() >= 9) {
        byte head = USE_SERIAL2.read();
        if (head == 0x2b) {
          byte buf[8] = {0};
          USE_SERIAL2.readBytes(buf, 8);
          state = 0x00;
          DEBUG_SERIAL.print(".");
        }else{
          DEBUG_SERIAL.print("error first byte:");
          DEBUG_SERIAL.print(head, HEX );
          DEBUG_SERIAL.println("");
        }
      }
    }else if (expireTime < millis()) { // timeout
      DEBUG_SERIAL.println("timeout, clear data:" + String(USE_SERIAL2.available()));
      while (USE_SERIAL2.available() > 0) {
        USE_SERIAL2.read();
      }
      state = 0x00;
    }
  }
}

void setup() {
//  SPIFFS.begin(true);
  WiFi.begin();
[...]

PR #1849, opened 2018-09-07
_uart_isr use wr_addr != rd_addr as test for internal queue not empty

qt1:
This is a possible fix for uart fifo problem.

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -80,7 +80,7 @@ static void IRAM_ATTR _uart_isr(void *arg)
         uart->dev->int_clr.rxfifo_full = 1;
         uart->dev->int_clr.frm_err = 1;
         uart->dev->int_clr.rxfifo_tout = 1;
-        while(uart->dev->status.rxfifo_cnt) {
+        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
             c = uart->dev->fifo.rw_byte;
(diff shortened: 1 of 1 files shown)


lbernstone:
Fixes #1824

cv007:
https://github.com/espressif/arduino-esp32/blob/85032b226c7775bae510e954112823a2ae32181a/cores/esp32/esp32-hal-uart.c#L83

the above change worked for me 
(ran several days with no problem- previously the rxfifo_cnt would 'lose' about a dozen bytes a day- the bytes were never missing, its just the rxfifo_cnt was returning 0 when there were bytes available, which the rd/wr addr compare now solves)

but maybe this can simply be reduced to-
`while(uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr) {`
as I don't see any need to check the 'unreliable' rxfifo_cnt
if rxfifo_cnt == 0, then wr_addr/rd_addr has to be compared to catch the possibly 'bad' rxfifo_cnt
if rxfifo_cnt != 0, then wr_addr != rd_addr will also be true
so just check wr_addr/rd_addr
(I also tried this for a few days with no problem)


related-

https://github.com/espressif/arduino-esp32/blob/85032b226c7775bae510e954112823a2ae32181a/cores/esp32/esp32-hal-uart.c#L342

in uartFlush-

this also could be reduced to-
while(uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr) {

although in both cases the rxfifo_cnt check is harmless if left as-is

It seems the reading of rxfifo_cnt only occurs in this file, so any problem rxfifo_cnt seems to have should now be taken care of :)

ouraborous:
Implementing this fix helped in my code, seems like a serious silicon issue that is not documented in the errata.

PR #1852, opened 2018-09-10
Add T-Beam Board Support

lewisxhe:
The [T-Beam](http://www.lilygo.cn/prod_view.aspx?TypeId=50010&Id=1006) board is online


me-no-dev:
Thanks :) board is in!

PR #1855, opened 2018-09-11
Adds functionality allowing size of HardwareSerial's RX buffer to be changed at runtime.

coddingtonbear:
This adds functionality analogous to the ESP8266's Arduino core's `HardwareSerial::setRxBufferSize`, providing users functionality allowing them to increase the size of the rx buffer as a way of preventing lost characters when using high baud rates.  Given that the queue size is only 256 characters by default, it is terribly easy for fast serial connections to exceed the default buffer length.

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -46,6 +46,10 @@ void HardwareSerial::end()
     _uart = 0;
 }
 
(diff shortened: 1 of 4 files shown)


me-no-dev:
Hey :) I will merge this but I am thinking of reworking those buffers to be able to hold way more than a K of data.

Issue #1725, opened 2018-08-04
Update IDF to aaf1239 breaks Wire library endTransmission()


bperrybap:
The Update IDF to aaf1239 here: https://github.com/espressif/arduino-esp32/commit/a59eafbc9dfa3ce818c110f996eebf68d755be24#diff-f1344d8f8dd31c4411770e150a4425fc
Includes some updates to the Wire library. One of the updates is a new endTransmission(bool) function.
While it would be nice to have this function use a bool instead of a uint8_t , adding it along side of the existing uint8_t functions breaks existing code that calls the existing endTransmission(uint8_t) function with 0 or 1.
i.e. existing applications or libraries that make calls to endTransmission(0) and endTransmission(1) will break since it is ambiguous which endTransmission() function should be called the bool or the uint8_t version.

I have not seen any other Arduino core implement a endTransmission(bool) function.
In order to maintain compatibility with existing 3rd party i2c applications and 3rd party i2c libraries the endTransmission(bool) function will need to be removed.

bperrybap:
While 3rd party s/w can work around this issue, since it is an issue that has been created in the ESP32 Wire code, the proper place to fix it is in ESP32 vs all the 3rd party s/w.

stickbreaker:
Official Arduino `Wire.endTransmission()`https://www.arduino.cc/en/Reference/WireEndTransmission

bperrybap:
While that reference page uses the words "boolean" and "true" the AVR code that Arduino.cc supplies does not use a C/C++ bool type for endTransmission()
https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/Wire/src/Wire.h
However Arduino.cc do seem to use a bool type for SAMD/DUE:
https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/Wire/Wire.h
So even they have issues - but they don't have compilation issues when using endTransmission(0) or endTransmission(1) like ESP32 does.

While DUE uses a bool type, pretty much all other cores use uint8_t for the type rather than bool.

What may also create some issues/confusion is that "boolean" is not a C/C++ type it is an Arduino proprietary type.
In the past "boolean" was a uint8_t but in recent IDEs it has been retyped to be an actual bool.

But the bottom line is that the ESP32 core Wire library now includes both endTransmission(bool) and endTransmission(uint8_t) 
I've not seen any other core or 3rd party i2c library have both of these.
[...]

PR #1888, opened 2018-09-22
Wire endTransmission() fix for issue #1725

bperrybap:
This patch will fix issue #1725 
There is no need to have the uint8_t version of the endTransmission() function.
Having both endTransmission(uint8_t) and endTransmission(bool) creates problems. The biggest problem being that endTransmission(0) and endTransmission(1) won't compile like it does on all the other cores.
Simply removing endTransmission(uint8_t) solves the problem.

### DIFF
--- a/libraries/Wire/library.properties
+++ b/libraries/Wire/library.properties
@@ -1,5 +1,5 @@
 name=Wire
-version=1.0
+version=1.0.1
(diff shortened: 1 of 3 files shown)


me-no-dev:
Hey thanks! Sorry for the delay :)

PR #1911, opened 2018-09-30
Deep-sleep example-sketches reported wrong wakeup-reason

WereCatf:
Incorrect values used, use the proper defines from header-files instead.


MarkyAD:
@WereCatf the deep sleep examples for gpio and touchpad wakeup need the same fix.

PR #1916, opened 2018-10-01
WiFiClientEvents.ino: Added a message for all event types

robdejonge:
Previously, the `switch` statement covered only 2 different events. I took the time to add them all in here.


lbernstone:
Can you change the title to indicate what component was affected (WiFiClientEvents.ino)

robdejonge:
Will do. 

I’m actually realizing I should change the comments written to the serial output, and make them more descriptive. One thing I’m not understanding is the Ethernet events. This is a WiFi library. How does it handle Ethernet events? If someone could explain this, i’ll Update comments and revise the pull request.

PR #1917, opened 2018-10-02
WiFi: improve WiFiEvent list in WiFiClientEvents example

### LINKED ISSUE #1875: Super minor issue, missing comment
This is a super minor issue, but it threw me for a loop for a minute.

In this example: https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/examples/WiFiClientEvents/WiFiClientEvents.ino#L25-L26

There is a missing event in the comments: SYSTEM_EVENT_AP_STAIPASSIGNED

The code works fine. I was expanding the WiFiEvent() function to include all of the events and event number 17 kept coming up without a matching case statement.

yoursunny:
fixes #1875

### DIFF
--- a/libraries/WiFi/examples/WiFiClientEvents/WiFiClientEvents.ino
+++ b/libraries/WiFi/examples/WiFiClientEvents/WiFiClientEvents.ino
@@ -3,34 +3,35 @@
  *
 */
 
-/* 
+/*
 * WiFi Events
 
-SYSTEM_EVENT_WIFI_READY               < ESP32 WiFi ready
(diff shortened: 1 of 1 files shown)


yoursunny:
I added the missing `SYSTEM_EVENT_AP_STAIPASSIGNED` event using the following reference:
https://github.com/espressif/esp-idf/blob/11b444b8f493165eb4d93f44111669ee46be0327/components/esp32/include/esp_event.h#L29

I also prepend numbers before each event, so that a human reader can easily lookup a number.

PR #1925, opened 2018-10-04
make WiFi.softAP() more robust

### LINKED ISSUE #1924: WiFi.softAP("apname", "") not working
WiFi.softAP only works when passphrase is 0, in esp8266 it also works when passphrase is "".

In WiFiAP.cpp on line 104 should change from:

    if(passphrase && (strlen(passphrase) > 63 || strlen(passphrase) < 8)) {

to:

    if(passphrase && (strlen(passphrase) > 63 || strlen(passphrase) < 8 || strlen(passphrase) == 0)) {

MarkyAD:
Make WiFi.softAP() fail less easily, attempt to fix #1924

### DIFF
--- a/libraries/WiFi/src/WiFiAP.cpp
+++ b/libraries/WiFi/src/WiFiAP.cpp
@@ -97,32 +97,27 @@ bool WiFiAPClass::softAP(const char* ssid, const char* passphrase, int channel, 
         return false;
     }
 
-    if(!ssid || *ssid == 0 || strlen(ssid) > 31) {
-        // fail SSID too long or missing!
-        return false;
-    }
-
-    if(passphrase && (strlen(passphrase) > 63 || strlen(passphrase) < 8)) {
-        // fail passphrase to long or short!
+    if(!ssid || *ssid == 0) {
+        // fail SSID missing!
         return false;
     }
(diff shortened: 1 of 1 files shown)


luc-github:
IMHO accepting password  length< 8 and >0, and ignoring it. is not right, should be failed.
```
if(!passphrase || strlen(passphrase) < 8) {
        conf.ap.authmode = WIFI_AUTH_OPEN;
        *conf.ap.password = 0;
    }
```
people think they secure their AP and actually it become open

ESP8266 use code that seems correct to me : 
```
if(passphrase && strlen(passphrase) > 0 && (strlen(passphrase) > 63 || strlen(passphrase) < 8)) {
        // fail passphrase to long or short!
        DEBUG_WIFI("[AP] fail passphrase to long or short!\n");
        return false;
    }
```

MarkyAD:
The issue with how it was originally written is it fails silently and only returns a bool with no way of knowing why. Also, it should've been checking for strlen(passphrase) > 0 in the test for too long or short, allowing 0 length passphrases to fall through to WIFI_AUTH_OPEN. Added a couple log_e() statements and kept the truncating of long ssid/pass.

PR #1935, opened 2018-10-07
Update WString.h

terryjmyers:


terryjmyers:
Companion change: https://github.com/espressif/arduino-esp32/pull/1936

PR #1943, opened 2018-10-09
Add 16MB partition large APPS w/ OTA & SPIFFS

Kadah:
* 7.25MB APPS partitions w/OTA
* Same SPIFFS partition size as default

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -48,6 +48,9 @@ esp32.menu.PartitionScheme.min_spiffs.build.partitions=min_spiffs
 esp32.menu.PartitionScheme.min_spiffs.upload.maximum_size=1966080
 esp32.menu.PartitionScheme.fatflash=16M Fat
 esp32.menu.PartitionScheme.fatflash.build.partitions=ffat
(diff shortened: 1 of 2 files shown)


Kadah:
I attempted to make the menu option brief but also convey the increased size.

I currently don't have one, but an 8MB flash option would complete the set of standard sizes the ESP-WROOM-32D/U and ESP32-WROVER come in.

lbernstone:
Is this a generally useful configuration?  I don't think we should clutter up the menus.  Are your apps getting to 7.25MB?  I made the app partitions on ffat.csv 2MB each.  Should this be larger?

Kadah:
(Ignore the random PR close, Github was only showing the "close and comment button. For some reason it has regularly not shown any of the green buttons this week.)

me-no-dev:
yeah... why so large apps? why not more FS instead?

Flo082002:
Maybe it is possible to add an option to create custom partition sizes.

lbernstone:
> Maybe it is possible to add an option to create custom partition sizes.

https://github.com/francis94c/ESP32Partitions

negativekelvin:
https://github.com/espressif/esp-idf/issues/1184

PR #1959, opened 2018-10-12
Added loadCert methods to WiFiClientSecure

### LINKED ISSUE #1957: WifiClientSecure - Enhancement - add load methods.
Hi All,
Apologies in advance if this is the wrong place to ask this question.

Im porting an application from a ESP8266 (Wemos D1 Mini) to the ESP32 (Lolin D32 Pro) and have problems with the new WiFiClientSecure library on the ESP32.

In the ESP8266 I have stored a root CA certificate a device certificate and a private key in SPIFFS which I then load into WiFiClientSecure using the loadCACert, loadCertificate and loadPrivateKey methods which accept a Stream as a parameter and load from SPIFFS no problem.

The new library doesnt have these methods and the setCACert only accepts const char* .

Can the library be enhanced to match the ESP8266 one as this will help support code written for the ESP8266 working on the ESP32 without changes being required and will make development easier for beginners like me ;)

Thank you
Kevin

lbernstone:
Fixes #1957.
Allows a file object to be used as a certificate in WiFiClientSecure.

### DIFF
--- a/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
+++ b/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
@@ -230,6 +230,51 @@ bool WiFiClientSecure::verify(const char* fp, const char* domain_name)
     return verify_ssl_fingerprint(sslclient, fp, domain_name);
 }
 
+char *WiFiClientSecure::_streamLoad(Stream& stream, size_t size) {
+  char *dest = (char*)malloc(size);
+  if (!dest) {
+    return nullptr;
+  }
+  if (size != stream.readBytes(dest, size)) {
+    free(dest);
+    return nullptr;
+  }
+  char ret[size+1];
+  snprintf(ret, size, "%s", dest);
+  free(dest);
+  return ret;
+}
+
+bool WiFiClientSecure::loadCACert(Stream& stream, size_t size) {
+  char *dest = _streamLoad(stream, size);
+  bool ret = false;
+  if (dest) {
+    setCACert(dest);
+    ret = true;
+  }
+  return ret;
+}
+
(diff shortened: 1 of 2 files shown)


stickbreaker:
@lbernstone   I have some question on these `load*` function using `_loadStream()`.  

from the original issue #1957 the OP said he used these three function to configure the secureClient:

>In the ESP8266 I have stored a root CA certificate a device certificate and a private key in SPIFFS which I then load into WiFiClientSecure using the loadCACert, loadCertificate and loadPrivateKey methods which accept a Stream as a parameter and load from SPIFFS no problem.

My question is that `_loadStream()` uses a `static char *` for the stream buffer. So, every call to `_loadStream()` reuses the same pointer,  It free's the buffer if necessary before each use, **BUT** prior calls to `load*` stored the address that is now being free'd?  Am I totally confused(very possible). 

 If I understand the OP's usage, he will use:
```c++
WifiClientSecure client;
client.loadCACert(file);// which reads from Stream and stores into heap buffer address also held in static char *
client.loadCertificate(file2);// which frees buffer containing CACertificate? then reallocates buffer for Certificate
client.loadPrivateKey(file3); // which frees buffer containing Certificate?
```
Now:
client._CA_cert point to freed memory/invalid
client._cert points to freed memory/invalid
client.private_key points to private_key/valid
  
I am working #3679 
Chuck.

stickbreaker:
@lbernstone I'm thinking that these pointers need a type field, something to mark constant or heap.  So that the destructor can clean up, right now the `WifiSecureClient()` destructor ignores these pointers.

Chuck.

lbernstone:
Yes.  Messy, messy.  I'll be afk for a couple days, but will take a look this weekend.  Maybe just needs the destructor to manually free the memory.

PR #1961, opened 2018-10-13
Added baudrate detection to esp32-hal-uart and HardwareSerial

Jeroen88:
Added function uartDetectBaudrate() and wrapped it into HardwareSerial::begin(). With this change it is possible to detect the baudrate on a serial port. Of course there must be incoming data on the serial port for detection to be possible. The uart can not detect other parameters like number of start- or stopbits, number of data bits or parity. For detection, if data is inverted is not important (of course for reading inversion is important).
To detect baudrate pass a 0 ('zero') for the baudrate and a timeout as the last parameter. Serial data must appear before the timeout, otherwise the Serial port will be left uninitialized. To detect if baudrate detection was successful test with HardwareSerial::baudRate() to return a value not equal to 0.
Pay attention: HardwareSerial::baudRate() may return an approximate baudrate. This has nothing to do with this PR but is due to rounding in the library. E.g. a baudrate of 115200 returns 115201.

Example usage:
```
void setup() {
  Serial.begin(115200);
  delay(100);
  Serial.println();

  Serial1.begin(0, SERIAL_8N1, -1, -1, false, 11000UL); // 0 for baudrate is detection, last parameter is a timeout

  unsigned long detectedBaudRate = Serial1.baudRate();
  if(detectedBaudRate) {
    Serial.printf("Detected baudrate is %lu\n", detectedBaudRate);
  } else {
    Serial.println("No baudrate detected, Serial1 will not work!");
  }
}
```


lbernstone:
might be good to put units in the header variable (timeout_ms)

Jeroen88:
@lbernstone Done!

Jeroen88:
Mixed up two different branches in the last commit. Will correct soon!

Jeroen88:
Reverted accidentally changed files to master

me-no-dev:
really nice :)

PR #1973, opened 2018-10-16
Feature/http client

Jeroen88:
Add two begin() methods to HTTPClient() to pass a WiFiClientSecure client with setCACert() and/or setCertificate() and setPrivateKey() already set. Include an example BasicHttp**s**Client.ino to show the usage. This adaptation aligns this library with ESP8266HTTPClient. The 5 begin() methods that already existed are set deprecated and the library version is updated to 1.2.
Currently backwards compatibility is maintained, but the idea is to drop all code enclosed within #ifdef HTTPCLIENT_1_1_COMPATIBLE and #endif, when a new version of the core is released.
When using this library make sure that the passed in WiFiClientSecure 'lives' longer than the HTPPClient _including it's destructor!_ In the example this is achieved by adding an extra scoping block { ... } for HTTPClient, to make sure it is destructed at the end of the block in stead of at the end of the function.

### DIFF
--- a/libraries/HTTPClient/examples/BasicHttpsClient/BasicHttpsClient.ino
+++ b/libraries/HTTPClient/examples/BasicHttpsClient/BasicHttpsClient.ino
@@ -104,31 +104,36 @@ void loop() {
   if(client) {
     client -> setCACert(rootCACertificate);
 
-    HTTPClient https;
+    {
+      // Add a scoping block for HTTPClient https to make sure it is destroyed before WiFiClientSecure *client is 
+      HTTPClient https;
   
-    Serial.print("[HTTPS] begin...\n");
-    if (https.begin(*client, "https://jigsaw.w3.org/HTTP/connection.html")) {  // HTTPS
-      Serial.print("[HTTPS] GET...\n");
-      // start connection and send HTTP header
-      int httpCode = https.GET();
+      Serial.print("[HTTPS] begin...\n");
+      if (https.begin(*client, "https://jigsaw.w3.org/HTTP/connection.html")) {  // HTTPS
+        Serial.print("[HTTPS] GET...\n");
+        // start connection and send HTTP header
+        int httpCode = https.GET();
   
-      // httpCode will be negative on error
-      if (httpCode > 0) {
-        // HTTP header has been send and Server response header has been handled
-        Serial.printf("[HTTPS] GET... code: %d\n", httpCode);
+        // httpCode will be negative on error
+        if (httpCode > 0) {
+          // HTTP header has been send and Server response header has been handled
+          Serial.printf("[HTTPS] GET... code: %d\n", httpCode);
   
-        // file found at server
-        if (httpCode == HTTP_CODE_OK || httpCode == HTTP_CODE_MOVED_PERMANENTLY) {
-          String payload = https.getString();
-          Serial.println(payload);
+          // file found at server
+          if (httpCode == HTTP_CODE_OK || httpCode == HTTP_CODE_MOVED_PERMANENTLY) {
+            String payload = https.getString();
+            Serial.println(payload);
+          }
+        } else {
+          Serial.printf("[HTTPS] GET... failed, error: %s\n", https.errorToString(httpCode).c_str());
         }
-      } else {
-        Serial.printf("[HTTPS] GET... failed, error: %s\n", https.errorToString(httpCode).c_str());
-      }
   
-      https.end();
-    } else {
-      Serial.printf("[HTTPS] Unable to connect\n");
+        https.end();
+      } else {
+        Serial.printf("[HTTPS] Unable to connect\n");
+      }
+
+      // End extra scoping block
     }
   
     delete client;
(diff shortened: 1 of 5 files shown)


Jeroen88:
See PR #1979 for comments on setTimeout. This PR is included in #1979.

Jeroen88:
> But it appears that the "80sPlanet" is being parsed as Port 80

No, port 80 is the default port for a http connection, this is added by the client because your url starts with http.

> [E][WiFiGeneric.cpp:659] hostByName(): DNS Failed for streaming.shoutcast.com

This is the place where things go wrong

> [D][HTTPClient.cpp:208] begin(): mix up of new and deprecated api

But this should not happen. This error appeared before, because _client was not inititialized to nullptr. This is corrected in PR #2097 

Could you merge the latest core and next this PR and try again? Please report back if your issue was solved. If not, please create a new issue in stead of a comment in a PR

usrdes:
@Jeroen88 

Thank you for your response.

"Could you merge the latest core and next this PR and try again?" 
I am using the release: Latest commit 273196d  7 days ago

This is still generating the issue. 
This release was on 11/18, I did a complete install on 11/24. 

For now, I have rolled back to the previous release and it appears to be working. 
Not sure when I can get back and create a new issue.

Jeroen88:
@usrdes did you merge PR #2097 too? This is essential!

Jeroen88:
Do you re-install using git? Then you might read the instructions at e.g https://gist.github.com/adam-p/15413fabef6cffecd897

If you install using a zip file, it is probably easier to edit the file yourself and change it. In line 211 of file libraries/HTTPClient/src/HTTPClient.h you should change:

`WiFiClient* _client;`

into:

`WiFiClient* _client = nullptr;`

usrdes:
@Jeroen88,

Thank you very much, I edited the file and that fixed it!
Thank you, Thank you

Jeroen88:
@usrdes You're welcome!

libraries/HTTPClient/examples/BasicHttpsClient/BasicHttpsClient.ino
```diff
@@ -0,0 +1,147 @@
...
+"-----END CERTIFICATE-----\n";
+
+// Not sure if WiFiClientSecure checks the validity date of the certificate. 
+// Setting clock just to be sure...
+void setClock() {
+  configTime(0, 0, "pool.ntp.org", "time.nist.gov");
+
+  Serial.print(F("Waiting for NTP time sync: "));
+  time_t nowSecs = time(nullptr);
+  while (nowSecs < 8 * 3600 * 2) {
```
marcelstoer: Two questions:
- Why not define (or calculate) this static value _outside_ the loop?
- Why exactly > 1.1.1970 4pm? Shouldn't it be larger than let's say 1546300800 (2019-01-01T00:00:00+00:00)? In my experience 16h is not enough.

Jeroen88: @marcelstoer do you mean 8 * 3600 * 2 with 'this static value'? If so, you shouldn't bother, the compiler will replace it with a static value.

Basically the value is quite arbitrary, any big enough value will do. time(nullptr) will return the number of seconds since the last reset, until it is overwritten by the NTP time sync. After that time(nullptr) is far bigger than the static test value and the while loop exits.

atanisoft: time_t starts counting at the unix Epoch which you can read up about [here](https://en.wikipedia.org/wiki/Unix_time).

If the time sync takes more than 16 hours there is a serious issue with network connectivity and time sync!

marcelstoer: > do you mean 8 * 3600 * 2 with 'this static value'?

Errhm, yes. Sorry. I should have _said_ "constant" when I _mean_ "constant".

> the compiler will replace it with a static value

True, the comment was more a style nitpick (and in that quit subjective I know).

@Jeroen88 @atanisoft for my second comment I was drawing from ESP8266 (i.e. NONOS SDK) experience, maybe it doesn't apply here anymore. For devices that spent most of their life in deep sleep we had to ensure that the constant value was a somewhat recent timestamp - at least from the current century. Otherwise, we might get `nowSecs` values that are way off. `while` loops a handful of times. Hence, it has nothing to do with "time sync takes more than 16 hours" or so.

Jeroen88: @marcelstoer I do not have any experience with deep sleep, but if time(nullptr) keeps running in deep sleep, the while() will break inmediately. I can imagine that in that case the ESP time and the real (network / NTP time) have drifted, and that you do want to have a time sync. The only thing I can think of here is to force a NTP time sync.

@atanisoft I didn't test it, but a freshly booted ESP32 could only initiate it's time after it has connected to the network, I guess? So some kind of waiting should be introduced to ensure synchronization has finished?

PR #1986, opened 2018-10-22
leave possible endless loop

FotoFieber:
Getting weather data from wunderground, I got stuck in an endless loop in HTTPClient.getString()

Others seem to have the same problem:
https://forum.arduino.cc/index.php?topic=533282.0


me-no-dev:
nice catch!

PR #1994, opened 2018-10-23
Feature/path arguments

Bmooij:
Adding path arguments to WebServer.
Same as for the [esp8266](https://github.com/esp8266/Arduino/pull/5214)

```
server.on("/led/2/actions/{}", []() {
    String action = server.pathArg(0);
    ...
 });
```


me-no-dev:
Thanks! Is that a recent change in 8266?

d-a-v:
Soon to be

Issue #827, opened 2017-11-11
base64 encode resets esp32 on certain strings


euquiq:
Hi there! 

I think I found a bug:  Sometimes my DOIT ESP32 DEVKIT V1 starts resetting itself when encoding a string by using the base64 library. .

When trying to encode certain quantity of characters, my ESP32 will give the following error:

```
CORRUPT HEAP: Bad tail at 0x3ffc9aac. Expected 0xbaad5678 got 0xbaad5600
 assertion "head != NULL" failed: file "/Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/heap/./multi_heap_poisoning.c", line 199, function: multi_heap_free
 abort() was called at PC 0x400d2af3 on core 1
 
 Backtrace: 0x400870ac:0x3ffcee50 0x400871ab:0x3ffcee70 0x400d2af3:0x3ffcee90 0x40086dd9:0x3ffceec0 0x40082796:0x3ffceee0 0x40082d19:0x3ffcef00 0x4000bec7:0x3ffcef20 0x400d0905:0x3ffcef40 0x400d0931:0x3ffcef80 0x400d0880:0x3ffcefa0 0x400dc84f:0x3ffceff0
 
 Rebooting...
```

I can reproduce the error with the following code:

```
#include <base64.h>
base64 b;

void setup()
{
	Serial.begin(115200);
	Serial.println("12345678 = " + b.encode("12345678"));
	delay(1000);
	Serial.println("123456 = " + b.encode("123456"));
	delay(1000);
	Serial.println("1234567 = " + b.encode("1234567")); //THIS LINE WILL CRASH ESP32
}

void loop()
{

}
```

I also tried this simple code in two other ESP32 boards, the geekworm EASY KIT ESP32-01 (revision 0 silicon) and another one with revision 1 silicon, which says ESP32_Core_board-V2. 

Same behavior in the three of them.

Regards,

Enrique

me-no-dev:
how about ```Serial.println(String("123456 = ") + b.encode("123456"));```

euquiq:
Hi me-no-dev!

It is not that line of code that breaks the Arduino ESP32: it is the following one. 

`Serial.println("1234567 = " + b.encode("1234567")); //THIS LINE WILL CRASH ESP32`

And the use of the String function helps not:

```
Opening port
Port open

12345678 = MTIzNDU2Nzg=
123456 = MTIzNDU2
[...]

PR #2007, opened 2018-10-30
use libbase64 macro to calculate base64 length

pepe79:
The calculation of the base64 encoded string length is incorrect as pointed out in #827. libbase64 already provides a macro that can be used to calculate the base64 encoded string length.

You also have to increment the calculated size by 1 for the trailing zero character.


me-no-dev:
thanks!

Issue #1707, opened 2018-08-01
Add getSketchSize & getFreeSketchSpace API


JoanMCD:
### Hardware:
Board:							?ESP32 Dev Module?
Core Installation/update date:			?1/Aug/2018?
IDE name:							?Arduino IDE?

### Description:
- Feature request -
Please implement the well known (ESP8266) API functions  for the ESP32

getSketchSize()
getFreeSketchSpace()

and if it makes sence for the ESP32:
getFlashChipRealSize()
getCoreVersion()
getBootVersion()
getBootMode()

thx, Joan

lbernstone:
https://docs.espressif.com/projects/esp-idf/en/v3.3.3/api-reference/system/system.html#chip-version

CriPstian:
> https://docs.espressif.com/projects/esp-idf/en/v3.3.3/api-reference/system/system.html#chip-version

This chip version API described at that link does not provide any of the info that was exposed by the functions mentioned above.
Is this a `NO` to my question?
I need a way to obtain the actual size of my chip flash size, not the sum of the existing partitions, like the `getFlashChipRealSize` did before.

lbernstone:
It simply means that I am not google.  This information is readily available if you look for it.
https://github.com/espressif/esp-idf/blob/release/v3.3/components/spi_flash/include/esp_spi_flash.h#L54

CriPstian:
> It simply means that I am not google. This information is readily available if you look for it.
> https://github.com/espressif/esp-idf/blob/release/v3.3/components/spi_flash/include/esp_spi_flash.h#L54

This is actually what I was saying. There's a clear note above that function that states:
```This value does not necessarily match real flash size.```
But no worries, I'll investigate further, hopefully come back here with an answer or PR.

lbernstone:
https://github.com/espressif/esptool can certainly get at that info.  Look there.

VTeselkin:
getFlashChipRealSize 
=
`String(spi_flash_get_chip_size())`

PR #2028, opened 2018-11-07
Added ESP:: functions for sketch size

lbernstone:


lbernstone:
Fixes #1707

Jeroen88:
Nice, could you add getSketchMD5() too? Then I will incorprate these in my PR #1979 :)

lbernstone:
Metadata for esp32 images includes SHA-256 hash.  Is that useful?

Jeroen88:
@lbernstone Before updating, the ESP client adds a header for the server with the MD5 of the sketch _running_. I think this header is not used much by servers, but I think that the idea was to give the server the opportunity to match the sketch currently running with data on the server for some business logic. To keep it compatible with the ESP8266, I do not need the MD5 of the sketch being downloaded, but the MD5 of the present sketch. That function is not available yet, is it?

lbernstone:
MD5 would have to be calculated, which would likely take a minute of 100% cpu and blocked SPI.  SHA256 is stored in the metadata, and is directly available through an [IDF function](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/spi_flash.html#_CPPv224esp_partition_get_sha256PK15esp_partition_tP7uint8_t).

Jeroen88:
I double checked ESP8266HTTPUpdate and EspClass::getSketchMD5, and for the ESP8266 indeed the MD5 of the running sketch is added to the request headers. Also, it is computed on the fly, and not stored in some flash header or so, so at least for the ESP8266 it is fast enough. 
To be compatible with the ESP8266 it would be better to add MD5, but my impression is that it is not used very often. A SHA-256 would be a good alternative then. 
When I have some time, I will add skech size and SHA-256 to the HTTPUpdate class of PR #1979

Issue #2041, opened 2018-11-09
secureClient.connect(SERVER_DOMAIN, 443) takes long time at handshake. 


a-c-sreedhar-reddy:
### Hardware:
Board:							ESP32 Dev Module
IDE name:					       Arduino IDE
Upload Speed:					115200
Computer OS:                                        Windows 10

### Description:
I would like to POST data to a server every 10 sec. But after some requests the program enters this loop
```
    log_v("Performing the SSL/TLS handshake...");
    while ((ret = mbedtls_ssl_handshake(&ssl_client->ssl_ctx)) != 0) {
        if (ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE) {
            return handle_error(ret);
        }
	vTaskDelay(10 / portTICK_PERIOD_MS);
    }
```
in [ssl_client.cpp](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/ssl_client.cpp) and it never exits.
So is it good to have a timeout for this WHILE loop?

a-c-sreedhar-reddy:
@Jeroen88  Thanks. But  I am using wifiSecueClient which does not support reuseconnection.

Jeroen88:
reuse is handled by HTTP client, independent of the underlying transport provided by WiFiClient(-Secure). You can also send the headers yourself, and handle the (re-)connection yourself, but HTTP Client is far more easy. You could also take a look at my PR #1979, which changes the HTTP Client API (in line with the same changes I did for the ESP8266) to be used more easily with different WiFi clients

copercini:
@a-c-sreedhar-reddy what core version are you using? a similar bug was fix in the end of 2017

About your project design: is it not better to use MQTT? this keeps the connection open for you send and receive data all the time

a-c-sreedhar-reddy:
@copercini My Esp sdk version is v3.1-dev-661-gf586f5e6

a-c-sreedhar-reddy:
After adding a timeout to the WHILE loop it works good. 

But there is another problem. If at any point secureClient.connect(host, port) returns 0 then all subsequent connects (retries) fail for for many hours. Then  connect starts working automatically sometimes or I need to restart the device.  Should we be clearing anything in the memory or buffer before we retry?

copercini:
it seems that the old connections are not being closed properly, then at some point the memory will run out, causing all these problems (including the timeout issue)

[...]

PR #2044, opened 2018-11-10
handshake in ssl_client.cpp

a-c-sreedhar-reddy:
issue #2041

### DIFF
--- a/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
+++ b/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
@@ -289,5 +289,5 @@ int WiFiClientSecure::lastError(char *buf, const size_t size)
 
 void WiFiClientSecure::setHandshakeTimeout(unsigned long handshake_timeout)
 {
-    sslclient->handshake_timeout = handshake_timeout;
+    sslclient->handshake_timeout = handshake_timeout * 1000;
 }


lbernstone:
You should use a #define value rather than a hardcoded timeout value.

atanisoft:
Instead of a #define it would be even better to have this timeout value configured in sslclient_context so it can be adjusted by the library users code.

me-no-dev:
agreed with @atanisoft ! @a-c-sreedhar-reddy will you make the change?

a-c-sreedhar-reddy:
> agreed with @atanisoft ! @a-c-sreedhar-reddy will you make the change?

@me-no-dev changes made.

libraries/WiFiClientSecure/src/WiFiClientSecure.h
```diff
@@ -62,6 +62,7 @@ class WiFiClientSecure : public WiFiClient
     bool loadCertificate(Stream& stream, size_t size);
     bool loadPrivateKey(Stream& stream, size_t size);
     bool verify(const char* fingerprint, const char* domain_name);
+    void setHandshakeTimeout(unsigned long handshake_timeout);
```
atanisoft: It probably should be noted that this is expected to be seconds and not milliseconds.

a-c-sreedhar-reddy: yeah I converted them to millis in the function.

atanisoft: most likely some sort of note here will help others not to pass the wrong type. Maybe even in the method itself do some bounds checking to ensure it doesn't wait forever (ie: cap it at 3600 seconds)

a-c-sreedhar-reddy: @atanisoft  default timeout will be 2 min. It is user wish if he wants to change it.

PR #2046, opened 2018-11-10
uartResizeRxBuffer() returns no value

Bascy:
uartResizeRxBuffer() is declared to return a size_t but does not return anything when uart is not defined


libraries/ESPmDNS/src/ESPmDNS.cpp
```diff
@@ -82,7 +82,7 @@ void MDNSResponder::setInstanceName(String name) {
 
 void MDNSResponder::enableArduino(uint16_t port, bool auth){
     mdns_txt_item_t arduTxtData[4] = {
-        {(char*)"board"         ,(char*)ARDUINO_VARIANT},
```
me-no-dev: there is a reason for this ;) please revert the change

me-no-dev:
please do not change mDNS config! revert that and I will accept the other change :)

PR #2067, opened 2018-11-17
Add FreeRTOS exsample

lutfisan:


libraries/FreeRTOS/FreeRTOS.ino
```diff
@@ -0,0 +1,114 @@
...
+  
+  while (!Serial) {
+    ; // wait for serial port to connect. Needed for native USB, on LEONARDO, MICRO, YUN, and other 32u4 based boards.
+  }
+
+  // Now set up two tasks to run independently.
+  xTaskCreatePinnedToCore(
+    TaskAnalogReadA0
+    ,  "AnalogReadA0"   // A name just for humans
+    ,  128  // This stack size can be checked & adjusted by reading the Stack Highwater
```
chegewara: In esp-idf docs stack highwater is wrong description. Vanilla FreRTOS is using dword for stack size, but esp32 stack size is in bytes, so 128 bytes i way to small stack. I would start with 1kB just in case.

chegewara:
Its just a bad copy/paste example.

lutfisan:
> Am I missing something here?
> 
> Comments in line 7 and lines 56 to 81 hint that the first task will be about blinking but the code for both tasks is an analog read.

sorry for my mistake, task 1 and task 2 just read ADC

chegewara:
Again, 128 bytes its to few for task stack size. App will crash. You copied code from freeRTOS examples, but freeRTOS is using dwords for stack size and esp-idf is using bytes. Minimum stack size is about 750 byes so i suggest to set it to 1 or 2kB.

Also, like i wrote before, tasks with priority 1 and 2 and no delay in loop or any other function that will suspend task does not allow to switch to other tasks.

PR #2072, opened 2018-11-18
Fix Partition Calculation error for min SPIFFS

luc-github:
I have kept the original APPS partintion  size on purpose as 40KB SPIFFS is useless IMHO
So I used the gap from original miscalculation for SPIFFS


bdring:
I agree with @luc-github. Large apps typically involve a web server. SPIFFS are the best places to store the web pages and 40k is too small.

huynh213:
Silly question, what address I should set for the min_spiffs.bin (which I generated from min_spiffs using PY) on the ESP Download Tool?

PR #2076, opened 2018-11-19
Add ESP32-POE board

alibabzo:
The manufacturer also advises

```c
#define ETH_CLK_MODE ETH_CLOCK_GPIO17_OUT
#define ETH_PHY_POWER 12
```

Should this go in `pins_arduino.h` as well?

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1665,6 +1665,36 @@ esp32-gateway.menu.UploadSpeed.115200.upload.speed=115200
 
 ##############################################################
 
+esp32-poe.name=OLIMEX ESP32-POE
+
+esp32-poe.upload.tool=esptool
+esp32-poe.upload.maximum_size=1310720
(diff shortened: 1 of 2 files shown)


me-no-dev:
please rebase to the latest boards config :)

alibabzo:
Beaten to it! Thanks for your all your ESP32 Arduino work :)

Stanimir-Petev:
Sorry for replying so late but I just saw Alibabzo's suggestion and it turns out it is working great with the ETH_LAN8720 demo by default. Otherwise these two lines:
#define ETH_CLK_MODE ETH_CLOCK_GPIO17_OUT
#define ETH_PHY_POWER 12
should be added at the top of the sketch.

I don't know if it is too late since the files has been updated already but if it is possible you can add the lines in the header file. Otherwise I will just open another pull request.

PR #2112, opened 2018-11-27
Clean warnings when all warning enabled

luc-github:
Not used variables / functions due to debug log

Dual define with different values :
cores\esp32/binary.h
#define B110 6
#define B1000000 64

tools/sdk/include/newlib/sys/termios.h
#define B110        3
#define B1000000   23

Local variable returned in WiFiclient Secure

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -248,7 +248,7 @@ static void IRAM_ATTR i2cDumpCmdQueue(i2c_t *i2c)
 
 /* Stickbreaker ISR mode debug support
  */
- #if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)
+#if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)
 static void i2cDumpDqData(i2c_t * i2c)
 {
 #if defined (ENABLE_I2C_DEBUG_BUFFER)
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -248,9 +248,10 @@ static void IRAM_ATTR i2cDumpCmdQueue(i2c_t *i2c)
 
 /* Stickbreaker ISR mode debug support
  */
+ #if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)
 static void i2cDumpDqData(i2c_t * i2c)
 {
-#if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)&&(defined ENABLE_I2C_DEBUG_BUFFER)
+#if defined (ENABLE_I2C_DEBUG_BUFFER)
     uint16_t a=0;
     char buff[140];
     I2C_DATA_QUEUE_t *tdq;
@@ -306,9 +307,12 @@ static void i2cDumpDqData(i2c_t * i2c)
         }
         a++;
     }
+#else
+    log_i("Debug Buffer not Enabled");
 #endif
 }
-
+#endif
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO
 static void i2cDumpI2c(i2c_t * i2c)
 {
     log_e("i2c=%p",i2c);
@@ -332,11 +336,12 @@ static void i2cDumpI2c(i2c_t * i2c)
         i2cDumpDqData(i2c);
     }
 }
+#endif
 
+#if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)  
 static void i2cDumpInts(uint8_t num)
 {
-#if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO) && (defined ENABLE_I2C_DEBUG_BUFFER)  
-
+#if defined (ENABLE_I2C_DEBUG_BUFFER)
     uint32_t b;
     log_i("%u row\tcount\tINTR\tTX\tRX\tTick ",num);
     for(uint32_t a=1; a<=INTBUFFMAX; a++) {
@@ -349,9 +354,10 @@ static void i2cDumpInts(uint8_t num)
     log_i("Debug Buffer not Enabled");
 #endif
 }
+#endif
 
-static void IRAM_ATTR i2cDumpStatus(i2c_t * i2c){
 #if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)&&(defined ENABLE_I2C_DEBUG_BUFFER)
+static void IRAM_ATTR i2cDumpStatus(i2c_t * i2c){
     typedef union {
         struct {
             uint32_t ack_rec:             1;        /*This register stores the value of ACK bit.*/
@@ -377,11 +383,11 @@ static void IRAM_ATTR i2cDumpStatus(i2c_t * i2c){
     sr.val= i2c->dev->status_reg.val;
     
     log_i("ack(%d) sl_rw(%d) to(%d) arb(%d) busy(%d) sl(%d) trans(%d) rx(%d) tx(%d) sclMain(%d) scl(%d)",sr.ack_rec,sr.slave_rw,sr.time_out,sr.arb_lost,sr.bus_busy,sr.slave_addressed,sr.byte_trans, sr.rx_fifo_cnt, sr.tx_fifo_cnt,sr.scl_main_state_last, sr.scl_state_last);
-#endif
 }
+#endif
 
-static void i2cDumpFifo(i2c_t * i2c){
 #if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)&&(defined ENABLE_I2C_DEBUG_BUFFER)
+static void i2cDumpFifo(i2c_t * i2c){
 char buf[64];
 uint16_t k = 0;
 uint16_t i = fifoPos+1;
@@ -422,8 +428,8 @@ if(i != fifoPos){// actual data
       }
     }while( i!= fifoPos);
 }
-#endif
 }
+#endif
 
 static void IRAM_ATTR i2cTriggerDumps(i2c_t * i2c, uint8_t trigger, const char locus[]){
 #if (ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO)&&(defined ENABLE_I2C_DEBUG_BUFFER)
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -440,7 +440,7 @@ uint8_t WiFiClient::connected()
         uint8_t dummy;
         int res = recv(fd(), &dummy, 0, MSG_DONTWAIT);
         if(res < 0) {
-            log_e("%d", errno);
+            log_e("RES: %d", res);
         }
         switch (errno) {
             case EWOULDBLOCK:
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -439,6 +439,9 @@ uint8_t WiFiClient::connected()
     if (_connected) {
         uint8_t dummy;
         int res = recv(fd(), &dummy, 0, MSG_DONTWAIT);
+        if(res < 0) {
+            log_e("%d", errno);
+        }
         switch (errno) {
             case EWOULDBLOCK:
             case ENOENT: //caused by vfs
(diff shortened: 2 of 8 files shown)


libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -439,6 +439,9 @@ uint8_t WiFiClient::connected()
     if (_connected) {
         uint8_t dummy;
         int res = recv(fd(), &dummy, 0, MSG_DONTWAIT);
+        if(res < 0) {
+            log_e("%d", errno);
```
atanisoft: logging the errno is not meaningful without a message to say why it is being logged. It would also be beneficial to log the "res" value here possibly.

cores/esp32/esp32-hal-i2c.c
```diff
@@ -306,9 +307,12 @@ static void i2cDumpDqData(i2c_t * i2c)
         }
         a++;
     }
+#else
+    log_i("Debug Buffer not Enabled");
```
atanisoft: it would be better to not log anything at all here so as not to clutter the console

stickbreaker: Except, if debug level is at "info" I assume you actually want to see information? 
The debug buffer in hal-i2c takes up 2570 bytes of ram if enabled. 
That is why this message exists. If you set debug_level at or above "info" I expect you have a problem you are trying to solve.

atanisoft: @stickbreaker this function isn't called from what I can tell if the level isn't at least DEBUG.

libraries/WiFi/src/WiFiClient.cpp
luc-github: Yes you are right should be res and not errno - will update it

luc-github: changed to : ` log_e("RES: %d", res);`

cores/esp32/esp32-hal-i2c.c
luc-github: I have tried to follow same as others functions - should I change it ?

stickbreaker: My bad, I got confused with the add/subtract highlighting. I just though you were questioning the "enable" message.  My only excuse is that I hadn't had my coffee before I read your Post :smiley: 

I don't have any problems with your changes.

atanisoft: coffee is a must!

stickbreaker: One other question on the debug subject: I added a an app level control for debug output with `Wire.setDebugFlags( uint32_t setBits, uint32_t resetBits);`, but I did not document how/when to use it.  Is there any standard for such documentation?

I used it around calls to a "recalcitrant" i2c devices when I get unexpected responses. I don't know if it would be good to put "debug" conditionals around all levels of this code (i.e. Wire() and hal-i2c).

What do you think?

luc-github: @stickbreaker @atanisoft  I followed same as line 354 - it means, I need to remove original 354 as well ? Or I misunderstood ?
thanks for your review

stickbreaker: @luc-github Let me spend some time looking through it.  The "debug_buffer" only houses interrupt activity(used inside the ISR). I just scanned through `i2cDumpDqData()` and don't know why I had it conditional on "debug_buffer"?

stickbreaker: For some reason github won't let me create an edited version of your version? So, I'm going to create an edited version the current espressif version in my repo? (might work? very confusing to me anyway!)

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -118,17 +118,19 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout)
...
 
-                IPAddress ip;
-                uint8_t * mac;
                 switch(status) {
                 case 3:
-                    ip = WiFi.localIP();
-                    mac = WiFi.BSSID();
+                    {
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_DEBUG
+                    IPAddress ip = WiFi.localIP();
```
me-no-dev: you can not declare new variables inside the case. Add another `#if` for their definitions above.

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -683,7 +683,9 @@ void WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
     smartconfig_status_t status = (smartconfig_status_t) st;
     log_d("Status: %s", sc_status_strings[st % 5]);
     if (status == SC_STATUS_GETTING_SSID_PSWD) {
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_DEBUG
         smartconfig_type_t * type = (smartconfig_type_t *)result;
```
me-no-dev: you should also include the call to log_d in the `#if`. Same for the one below. Makes the reason for the `#if` clearer.

libraries/WiFi/src/WiFiMulti.cpp
atanisoft: the variable is not necessary at all even, the log_d line can be updated to just:
```
log_d("[WIFI] IP: %s", WiFi.localIP().toString().c_str());
```

atanisoft: same for BSSID probably with WiFi.BSSIDstr()

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -232,18 +232,17 @@ bool WiFiClientSecure::verify(const char* fp, const char* domain_name)
 }
 
 char *WiFiClientSecure::_streamLoad(Stream& stream, size_t size) {
-  char *dest = (char*)malloc(size);
+  static char *dest = nullptr;
+  if(dest)free(dest);
```
me-no-dev: :D Please at least try to put some style. 
```cpp
if(dest) {
  free(dest);
}
```

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -694,7 +696,9 @@ void WiFiSTAClass::_smartConfigCallback(uint32_t st, void* result) {
         _smartConfigDone = true;
     } else if (status == SC_STATUS_LINK_OVER) {
         if(result){
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_DEBUG
             ip4_addr_t * ip = (ip4_addr_t *)result;
+#endif
```
atanisoft: move the log_d line inside the if/endif block as well

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -232,18 +232,17 @@ bool WiFiClientSecure::verify(const char* fp, const char* domain_name)
...
+  dest = (char*)malloc(size);
   if (!dest) {
     return nullptr;
   }
   if (size != stream.readBytes(dest, size)) {
     free(dest);
```
me-no-dev: set dest to null here, or free will be called again on this region and bad things might happen :P

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
atanisoft: if dest is initialized to nullptr this line won't ever be evaluated as true

tools/sdk/include/vfs/esp_vfs.h
```diff
@@ -27,7 +27,7 @@
 #include <sys/reent.h>
 #include <sys/stat.h>
 #include <sys/time.h>
-#include <sys/termios.h>
+//#include <sys/termios.h>
```
atanisoft: this comes from ESP-IDF and should be fixed on that side instead

libraries/WiFi/src/WiFiMulti.cpp
luc-github: Ok I will do 
For my personnal knowledge - as I have set brace, why  it is not correct to declare new local variables inside the case ?

me-no-dev:
we can not touch sdk files :) the binary strings are still used here and there in libs :( if we get rid of them I expect other issues to come

me-no-dev:
so is this ready to be merged now? all i2c things resolved?

luc-github:
No I still wait for @stickbreaker 

What could be the workaround  for #define conflict - it is source of potential issue also I think

me-no-dev:
where is esp_vfs included that it causes this issue?

luc-github:
@me-no-dev  it is here : https://github.com/espressif/arduino-esp32/blob/master/libraries/SD/src/sd_diskio.cpp#L19

stickbreaker:
@luc-github keep the debug_enable conjunction, that way just setting debug level doesn't cause serial monitor spamming. The two step process, hopefully we wont use i2c  debugging constantly.

I had forgotten my reasoning for the double level activation.

Chuck

me-no-dev:
@luc-github in this case you undefine the offending arduino bin definitions before you include the file. That file does not care for Arduino binary definitions ;)

luc-github:
so we left as it is ? as a known issue ?If it is agreed I am fine with it ^_^ - so PR is done
```
In file included from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0/tools/sdk/include/vfs/esp_vfs.h:30:0,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.cpp:19:

C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0/tools/sdk/include/newlib/sys/termios.h:97:0: warning: "B110" redefined

 #define B110        3

 ^

In file included from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\cores\esp32/Arduino.h:40:0,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.h:17,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.cpp:14:

C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\cores\esp32/binary.h:65:0: note: this is the location of the previous definition

 #define B110 6

 ^

In file included from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0/tools/sdk/include/vfs/esp_vfs.h:30:0,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.cpp:19:

C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0/tools/sdk/include/newlib/sys/termios.h:118:0: warning: "B1000000" redefined

 #define B1000000   23

 ^

In file included from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\cores\esp32/Arduino.h:40:0,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.h:17,

                 from C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\libraries\SD\src\sd_diskio.cpp:14:

C:\Users\luc\AppData\Local\Arduino15\packages\esp32\hardware\esp32\1.0.0\cores\esp32/binary.h:277:0: note: this is the location of the previous definition

 #define B1000000 64

 ^
```

me-no-dev:
@luc-github what will happen if before https://github.com/espressif/arduino-esp32/blob/master/libraries/SD/src/sd_diskio.cpp#L19 you add `#undef B1000000` and so on?

atanisoft:
> @luc-github what will happen if before https://github.com/espressif/arduino-esp32/blob/master/libraries/SD/src/sd_diskio.cpp#L19 you add `#undef B1000000` and so on?

is esp_vfs.h even used? from a cursory glance at sd_diskio.cpp it doesn't appear to be

luc-github:
@me-no-dev  seems avoid issue
```
#undef B110 
#undef B1000000
```

@atanisoft  seems not as commented compilation succeed
```
extern "C" {
    #include "diskio.h"
    #include "ffconf.h"
    #include "ff.h"
    //#include "esp_vfs.h"
    #include "esp_vfs_fat.h"
    char CRC7(const char* data, int length);
    unsigned short CRC16(const char* data, int length);
}
```
@me-no-dev @atanisoft  I think commenting `#include "esp_vfs.h"` is the way to go 
what do you think ?

me-no-dev:
sure! go ahead :)

luc-github:
I think everything is clear now - thank you  guys ^_^

PR #2117, opened 2018-11-27
Mark as after Release 1.0.1

stickbreaker:
change version to reflect post Release 1.0.1 status

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -1,5 +1,5 @@
-name=ESP32 Arduino
-version=0.0.1
+name=ESP32 Arduino 1.0.1 DEV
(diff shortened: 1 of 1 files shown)


platform.txt
```diff
@@ -1,5 +1,5 @@
-name=ESP32 Arduino
-version=0.0.1
+name=ESP32 Arduino 1.0.1 DEV
```
me-no-dev: revert this line please :)

stickbreaker: But, the version number by itself does not show up anywhere. Without it, this mod is useless.

PR #2123, opened 2018-11-28
Increased git clone depth

pacucha42:
**git clone --depth** parameter increased to 100 commits (50 by default) to cover all previous tags - fix for automatic release issues regarding Release 1.0.1

NOTE: this problem might reappear according to number of commits in between single releases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ python:
 os:
   - linux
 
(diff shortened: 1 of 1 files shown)


pacucha42:
after discussion with Hristo: depth parameter shall be omitted completely

PR #2133, opened 2018-12-01
WiFiClientSecure: add support for PSK (pre-shared key) ciphers

tve:
This PR adds support for using pre-shared keys in TLS connections. MQTT servers seem to start to support this, at least mosquitto does. This is an alternative to generating a self-signed cert and then distributing the cert to all mobile devices. With PSK one can now generate a long random hex string and simply use that as key. It also makes it easy to have multiple identity->key pairs, basically they act like passwords.
I would be happy to write some docs for this, but am not sure where to add given that WiFiClientSecure doesn't have any docs.

### DIFF
--- a/libraries/WiFiClientSecure/README.md
+++ b/libraries/WiFiClientSecure/README.md
@@ -1,0 +1,67 @@
+WiFiClientSecure
+================
+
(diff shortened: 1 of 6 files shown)


tve:
README added

Issue #1623, opened 2018-07-11
SPI.transfer(buffer, size) is missing


PaulStoffregen:
ESP32's SPI library is missing overloaded function SPI.transfer(buffer, size).

Here is Arduino's documentation.

https://www.arduino.cc/en/Reference/SPITransfer

Arduino sketches and libraries using this function will not compile for ESP32.  This is a very easy fix, just add this line in SPI.h

      void transfer(uint8_t * data, uint32_t size) { transferBytes(data, data, size); }

me-no-dev:
why not make a PR 😄

me-no-dev:
don't implement it in the header like that though

iafilius:
Hi, just a "me too" message (for libaries 1.0.0 and 1.0.1-rc1).  The quick fix by adding the oneliner to SPI.h worked for me.

PaulStoffregen:
While not strictly necessary to get Ethernet to work, you will see slightly improved performance on the W5500 chip if you also add SPI.transfer(const void *txBuffer, void *rxBuffer, size_t count) and define SPI_HAS_TRANSFER_BUF.

Here's where it's actually used by the Ethernet library:
https://github.com/arduino-libraries/Ethernet/blob/master/src/utility/w5100.cpp#L372

If SPI.h does not have this function, Ethernet will revert to using single byte transfers.

iafilius:
@me-no-dev  Thanks for accepting accepting the commit

@PaulStoffregen ,

you mean something like:

SPI.h:
class SPIClass
<>
public:
<>
void transfer(const void * txBuffer, const void * rxBuffer, uint32_t size);
<>


SPI.cpp:

// SPI_HAS_TRANSFER_BUF - is defined to signify that this library supports
// a version of transfer which allows you to pass in both TX and RX buffer
// pointers, either of which could be NULL
#define SPI_HAS_TRANSFER_BUF 1
SPI.transfer(const void *txBuffer, void *rxBuffer, size_t count) 
{
	transferBytes(txBuffer, rxBuffer, size);
}

If i see it correctly, no corresponding transferBytes() is available with the right types (void*), and no void* type parameter functions are available yet. What do you have in mind?
Regards

Popcron07:
Hi, I ran into a problem which I think might have something to do with your changes.
When I try to compile my code I get this error message:

[...]

PR #2136, opened 2018-12-02
Fixed Arduino SPI/Ethernet compile issue as described in issue #1623

iafilius:
Hi,
implemented missing SPI.transfer, required to compile Arduino SPI based Ethernet (W5500)
Splitted the suggested change in Issue #1623, in a header and source part. It compiles, works, but i'm not completely confident with exact location, and if this would be the best solution.
Please review, and accept the proposal. 
Not completely feeling confident with providing the data parameter twice to the transferBytes function, please, just copied it form the suggested solution in #1623 . 

Regards,

Arjan


me-no-dev:
thanks :)

Issue #2139, opened 2018-12-02
WiFiClient error [WiFiClient.cpp:443] connected(): RES: -1


Aietes:
### Hardware:
Board:							ESP32 Dev Module
Core Installation/update date:		latest
IDE name:						Platform.io with Visual Studio Code
Flash Frequency:					?40Mhz?
PSRAM enabled:                                       no
Upload Speed:					921600
Computer OS:                                           Mac OSX

### Description:
I'm trying to connect and send a message via MQTT over WiFi. Code that used to run flawlessly now fails with error [WiFiClient.cpp:443] connected(): RES: -1. I'm not sure what is causing it. I'm using the PubSubClient library to send and receive MQTT message.

I'm connecting to WiFi in the standard way:

```cpp
bool myWiFi::connect(String *ssid, String *pass)
{
  Serial.print("Connecting to WiFi");

  WiFi.mode(WIFI_STA);
  while (WiFi.status() == 255)
    ; //wait till WiFi is available

  if (!WiFi.reconnect() )
  {
    WiFi.begin(ssid->c_str(), pass->c_str());
  }

  unsigned long millis_rto = millis();
  unsigned long millis_cto = millis();
  while (WiFi.waitForConnectResult() != WL_CONNECTED)
  {
    Serial.print(".");
    if (millis() - millis_rto >= RETRY_TIMEOUT)
    {
      //reset timer
      millis_rto = millis();
      //start new try
      Serial.print("\nConnecting to WiFi");
      WiFi.begin(ssid->c_str(), pass->c_str());
    }
    if (millis() - millis_cto >= CONN_TIMEOUT)
    {
      Serial.println("\nWiFi connection timed out, open AP.");
      getConfig(ssid, pass);
      return false;
    }
    delay(500);
  }

  Serial.print("\nConnected to WiFi with IP ");
  Serial.println(WiFi.localIP());
  return true;
}
```

The connection works, the error message appears after a successful connection and receiving an IP. 
I'm tracing the WiFi events to find out what's going on with WiFi.onEvent(WiFiEvent). I'm getting the expected sequence of events:

```
Connecting to WiFi
WiFi Event: 0
WiFi Event: 2
WiFi Event: 5
Connecting to WiFi.
WiFi Event: 4
WiFi Event: 7
Connected to WiFi with IP 192.168.178.57
```

But then the following error occurs:

```
[E][WiFiClient.cpp:443] connected(): RES: -1
```
[...]

PR #2140, opened 2018-12-02
 removing log

luc-github:
fix https://github.com/espressif/arduino-esp32/issues/2139

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -439,9 +439,8 @@ uint8_t WiFiClient::connected()
     if (_connected) {
         uint8_t dummy;
         int res = recv(fd(), &dummy, 0, MSG_DONTWAIT);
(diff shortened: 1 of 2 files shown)


bertmelis:
My opinion about this PR:

I did some digging after receiving numerous messages during an OTA (using the repo example).
After some digging I think I found a possible cause: OTA calls `WiFiClient::connected()` repeatedly. The implementation uses `recv()` from sockets.h and in the docs I read this:

> Upon successful completion, recv() returns the length of the message in bytes. If no messages are available to be received and the peer has performed an orderly shutdown, recv() returns 0. Otherwise, -1 is returned and errno is set to indicate the error. 

So it returns -1 because no data is available but the socket is still connected. So it signals an error which isn't a real error. Hence, reporting `res < 0` as an error doesn't make sense and should be deleted as in this PR. Furthermore, should there be an error, it is reported in the lines below.

PR #2141, opened 2018-12-02
I2C ReSTART returns Success 

stickbreaker:
This change restores compatibility with preexisting Arduino libraries that use Wire().  The ReSTART queuing operation is now hidden.  `Wire.endTransmission(id,len,FALSE);` now returns `I2C_ERROR_OK`(0) to indicate success, though, `Wire.lastError()` will correctly return `I2C_ERROR_CONTINUE` to indicate the last `Wire()` operation was added to the command queue, and not executed.  Since `Wire.lastError()` does not exist in main branch Arduino, this dichotomy is not obvious.

The number of queued operation is limited by the depth of the `Wire()` Tx/Rx buffers, currently 128 bytes each.  Each `Wire.Write()` operation uses one byte of the Tx buffer, `Wire.requestFrom(id,LEN,false);` uses `LEN` bytes of the 128byte Rx buffer.  If either buffer is exhausted before a STOP condition is issued, `I2C_ERROR_MEMORY`  will be returned.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -170,7 +170,7 @@ uint8_t TwoWire::endTransmission(bool sendStop)  // Assumes Wire.beginTransactio
     txIndex = 0;
     txLength = 0;
     transmitting = 0;
-    return last_error;
+    return (last_error == I2C_ERROR_CONTINUE)?I2C_ERROR_OK:last_error; // Don't return Continue for compatibilty.
 }
 
 /* @stickBreaker 11/2017 fix for ReSTART timeout, ISR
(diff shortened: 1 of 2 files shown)


digi-z:
Hello Stickbreaker
I use your wire.cpp and wire,h but get compilation erros for nodeMCU: i2cDebug and i2cGetStatus not declared. If I try to include esp32-hal-i2c.h I get faults that #includes are nested too deeply. How do I have to configure the libraries to be able to compile them?
Regards!

stickbreaker:
@digi-z If you are using V1.0.1, either from an Arduino Boards Manager install, or the Cloned Github repo,  All you need to do is include the `Wire.h` library in your sketch:
```c++
#include <Wire.h>

void setup(){
Serial.begin(115200);
Wire.begin(); // default 100khz, SDA and SCL as defined in pins.h
}

void loop(){
// use it
}
```
Chuck.

PR #2151, opened 2018-12-04
Fix WiFiClientSecure read() and available() logic to properly include peeked byte

rdowning-triax:


libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -196,19 +195,16 @@ int WiFiClientSecure::read(uint8_t *buf, size_t size)
         buf[0] = _peek;
         _peek = -1;
         size--;
-        if(!size || !available()){
-            return 1;
-        }
-        buf++;
-        peeked = 1;
+        int ret = read(buf+1, size-1);
```
me-no-dev: I fail to see what this fixes? Also calling a function on itself? Why? Note that `size` has already been subtracted. Every time there is a peek byte, you will get 1 byte less.

me-no-dev:
Please add descriptions to you pull requests. What they do, what they fix, and so on.

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
rdowning-triax: Ah yes, the double subtraction was my mistake. The changes to read(...) just clean up and remove a bunch of redundant code.

What's really meant to be fixed is available(). Here it has the same problem WiFiClient had, where if not connected it would return 0, even if there was data available. Another problem was, if the data_to_read result res was less than 0, it would just return that negative value, even if there is a peeked byte available.

rdowning-triax: Actually there was a problem in read(...) too, where if you have a peeked byte, but then res is a negative number, you just return the negative, whereas it should be returning 1 to indicate you *read* the peeked byte at the very least.

me-no-dev: Still I would rather you revert this recursive call and fix the "returns -1 if read failed but there was a peek" when `get_ssl_receive` returns negative.

me-no-dev:
Please rebase :) there is a duplicate change for WiFiClient. Also note this PR: https://github.com/espressif/arduino-esp32/pull/2155

PR #2155, opened 2018-12-05
Fix bug in PR #2148: correcting available() introduced an bug!

Jeroen88:
If ::available() is called before ::connect() _rxBuffer is not initialised

PR #2161, opened 2018-12-05
Fix HTTPUpdate flash size check and add SPIFFS size check

Jeroen88:
HTTPUpdate checked if the new sketch would fit into the _current_ partition free space, which was wrong. Now it checks if it fits into the _next_ flash partition.

A size check for a SPIFFS update is added, this was stil 'to do' from the previous PR.

Still to do: the SHA256 hash could be checked if a server sends it in the request header.

This PR includes PR #2155.

### DIFF
--- a/libraries/HTTPUpdate/src/HTTPUpdate.cpp
+++ b/libraries/HTTPUpdate/src/HTTPUpdate.cpp
@@ -251,14 +251,14 @@ HTTPUpdateResult HTTPUpdate::handleUpdate(HTTPClient& http, const String& curren
                     startUpdate = false;
                 }
             } else {
-                const esp_partition_t* _partition = esp_ota_get_next_update_partition(NULL);
-                if(!_partition){
+                int sketchFreeSpace = ESP.getFreeSketchSpace();
+                if(!sketchFreeSpace){
                     _lastError = HTTP_UE_NO_PARTITION;
                     return HTTP_UPDATE_FAILED;
                 }
 
-                if(len > _partition->size) {
-                    log_e("FreeSketchSpace to low (%d) needed: %d\n", _partition->size, len);
+                if(len > sketchFreeSpace) {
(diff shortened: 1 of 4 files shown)


libraries/HTTPUpdate/src/HTTPUpdate.cpp
```diff
@@ -366,6 +379,10 @@ bool HTTPUpdate::runUpdate(Stream& in, uint32_t size, String md5, int command)
         }
     }
 
+// To do: the SHA256 could be checked if the server sends it
+
+    delay(0);
```
me-no-dev: why?

libraries/HTTPUpdate/src/HTTPUpdate.cpp
```diff
@@ -229,14 +231,25 @@ HTTPUpdateResult HTTPUpdate::handleUpdate(HTTPClient& http, const String& curren
...
+                }
+
+                if(len > _partition->size) {
+                    log_e("spiffsSize to low (%d) needed: %d\n", _partition->size, len);
+                    startUpdate = false;
+                }
             } else {
-                if(len > (int) ESP.getFreeSketchSpace()) {
-                    log_e("FreeSketchSpace to low (%d) needed: %d\n", ESP.getFreeSketchSpace(), len);
+                const esp_partition_t* _partition = esp_ota_get_next_update_partition(NULL);
```
me-no-dev: please move this logic to Esp.cpp and replace the one in `ESP.getFreeSketchSpace`. What is currently there makes no sense for any use.

libraries/HTTPUpdate/src/HTTPUpdate.cpp
Jeroen88: Removing...

me-no-dev:
Nice :) thanks!

Jeroen88:
You're welcome! :)

Issue #2193, opened 2018-12-15
ARDUINO_VARIANT cannot be cast to (char*) when compiling in Eclipse/windows


Bascy:
### Hardware:
Board:							ESP32 Dev Module
Core Installation/update date:		commit ec3a97b598f2a8516297fee873ce78492f5d7e64 (dec 15)
IDE name:						Eclipse / Sloeber 4.3 plugin
Computer OS:                                         Windows 10

### Description:
On compilation of the ESPmDNS component the following code results in an error 
` error: 'esp32' was not declared in this scope`
because the macro ARDUINO_VARIANT is replaced by esp32 and not by the string "esp32"

```
void MDNSResponder::enableArduino(uint16_t port, bool auth){
    mdns_txt_item_t arduTxtData[4] = {
        {(char*)"board"         ,(char*)ARDUINO_VARIANT},
        {(char*)"tcp_check"     ,(char*)"no"},
        {(char*)"ssh_upload"    ,(char*)"no"},
        {(char*)"auth_upload"   ,(char*)"no"}
    };

```

Bascy:
Guess this should be solved as described here:
https://www.guyrutenberg.com/2008/12/20/expanding-macros-into-string-constants-in-c/

atanisoft:
This is more likely an issue with how you are compiling in Eclipse. In Arduino IDE (https://github.com/espressif/arduino-esp32/blob/master/platform.txt#L68) and PIO (https://github.com/espressif/arduino-esp32/blob/master/tools/platformio-build.py#L94) it is passed in as a quoted string.

So, how are you building in Eclipse?

Bascy:
Using Eclipse plugin from Sloeber, The following definition of ARDUINO_VARIANT is used:

`-DARDUINO_VARIANT="esp32"  `

I'm guessing that Eclipse / windows is loosing the double quotes somewhere

lbernstone:
Can you add something about eclipse env quoting to the issue title so the next person finds this?

atanisoft:
@Bascy the difference is how the quotes are declared, can you try:
```
-DARDUINO_VARIANT="\"esp32\""
```
The same issue could also be seen with ARUDINO_BOARD as well if you dig into it further.

Bascy:
Yes this also solves the problem for me, but as i understand it the problem only occurs when compiling on Eclipse and/or Windows OS. 
So I make a pull request with a solution to this, implemented only when using Eclipse

[...]

PR #2197, opened 2018-12-15
Add quotes around macro value (Issue #2193)

Bascy:


me-no-dev:
I like how you solved this :)

PR #2217, opened 2018-12-18
Increase UART concurrency

stickbreaker:
While data is beginning received, it is stuck in the RX Fifo until the fifo fills to `rxfifo_full_thrhd`, or there is a break in the continuous data stream longer than the configured `rx_tout_thrhd`.  This fix changes processing so that if a `read()` sees an empty queue, or `available()` is called, the RX fifo is flushed.  This will decrease any latency between reception at the port and data handling in `Serial()`.

This mod has not been exhaustively tested, but I have encountered no issues with my per-existing code.

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -66,9 +66,23 @@ static uart_t _uart_bus_array[3] = {
 };
 #endif
 
+static void _uart_rx_flush( uart_t * uart){
+    uint8_t c;
+    UART_MUTEX_LOCK();
+    UBaseType_t spaces=0;
+    if(uart->queue != NULL) {
+        spaces=uxQueueSpacesAvailable( uart->queue);
+    }
+    while(((spaces--) >0)&&(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr))) {
+        c = uart->dev->fifo.rw_byte;
+        xQueueSend(uart->queue, &c,0);
+    }
+	UART_MUTEX_UNLOCK();
+}
+
 static void IRAM_ATTR _uart_isr(void *arg)
 {
-    uint8_t i, c;
+    uint8_t i,c;
     BaseType_t xHigherPriorityTaskWoken;
     uart_t* uart;
 
@@ -77,15 +91,15 @@ static void IRAM_ATTR _uart_isr(void *arg)
         if(uart->intr_handle == NULL){
             continue;
         }
+		while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
+        	c = uart->dev->fifo.rw_byte;
+        	if(uart->queue != NULL && !xQueueIsQueueFullFromISR(uart->queue)) {
+            	xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
+        	}
         uart->dev->int_clr.rxfifo_full = 1;
         uart->dev->int_clr.frm_err = 1;
         uart->dev->int_clr.rxfifo_tout = 1;
-        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
-            c = uart->dev->fifo.rw_byte;
-            if(uart->queue != NULL && !xQueueIsQueueFullFromISR(uart->queue)) {
-                xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
-            }
-        }
+    	}
     }
 
     if (xHigherPriorityTaskWoken) {
@@ -96,14 +110,14 @@ static void IRAM_ATTR _uart_isr(void *arg)
 void uartEnableInterrupt(uart_t* uart)
 {
     UART_MUTEX_LOCK();
-    uart->dev->conf1.rxfifo_full_thrhd = 112;
-    uart->dev->conf1.rx_tout_thrhd = 2;
+    uart->dev->conf1.rxfifo_full_thrhd = 112; // needs speed compensation?
+    uart->dev->conf1.rx_tout_thrhd = 32; // no need to force rxFiFo empty, because uartRead() will empty if necessary
     uart->dev->conf1.rx_tout_en = 1;
     uart->dev->int_ena.rxfifo_full = 1;
     uart->dev->int_ena.frm_err = 1;
     uart->dev->int_ena.rxfifo_tout = 1;
     uart->dev->int_clr.val = 0xffffffff;
-
+// can this share an interrupt?
     esp_intr_alloc(UART_INTR_SOURCE(uart->num), (int)ESP_INTR_FLAG_IRAM, _uart_isr, NULL, &uart->intr_handle);
     UART_MUTEX_UNLOCK();
 }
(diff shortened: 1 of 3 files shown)


cores/esp32/esp32-hal-uart.c
```diff
@@ -296,6 +331,11 @@ uint8_t uartPeek(uart_t* uart)
     uint8_t c;
     if(xQueuePeek(uart->queue, &c, 0)) {
         return c;
+    } else {
+        _uart_rx_flush(uart);
+        if(xQueueReceive(uart->queue, &c, 0)) {
```
stickbreaker: This is wrong,  Should have used `xQueuePeek()`

cores/esp32/esp32-hal-uart.c
```diff
@@ -82,19 +82,19 @@ static void _uart_rx_flush( uart_t * uart){
...
         if(uart->intr_handle == NULL){
             continue;
         }
-		while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
+	while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
         	c = uart->dev->fifo.rw_byte;
         	if(uart->queue != NULL && !xQueueIsQueueFullFromISR(uart->queue)) {
-            	xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
+        	    	xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
         	}
```
atanisoft: one extra indent here possibly?

stickbreaker: maybe I got fixed.

stickbreaker: although it still looks wrong here?
ARG!

stickbreaker: maybe I got it now, No TABS, only Spaces!

cores/esp32/esp32-hal-uart.c
```diff
@@ -66,27 +66,41 @@ static uart_t _uart_bus_array[3] = {
...
+    UART_MUTEX_LOCK();
+    UBaseType_t spaces=0;
+    if(uart->queue != NULL) {
+        spaces=uxQueueSpacesAvailable( uart->queue);
+    }
+    while(((spaces--) >0)&&(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr))) {
+        c = uart->dev->fifo.rw_byte;
+        xQueueSend(uart->queue, &c,0);
+    }
+	UART_MUTEX_UNLOCK();
```
atanisoft: looks like an extra tab/space is here as well

stickbreaker: your correct, another TAB, Maybe fixed ;)

cores/esp32/esp32-hal-uart.c
```diff
@@ -66,27 +66,41 @@ static uart_t _uart_bus_array[3] = {
 };
 #endif
 
+static void _uart_rx_flush( uart_t * uart){
```
me-no-dev: is "flush" the best name here? you are moving the data to the queue, not actually flushing it. Maybe `_uart_rx_read_fifo` is better?

cores/esp32/esp32-hal-uart.c
```diff
@@ -66,27 +66,41 @@ static uart_t _uart_bus_array[3] = {
...
             c = uart->dev->fifo.rw_byte;
             if(uart->queue != NULL && !xQueueIsQueueFullFromISR(uart->queue)) {
                 xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
             }
         }
-    }
+        uart->dev->int_clr.rxfifo_full = 1;
+        uart->dev->int_clr.frm_err = 1;
+        uart->dev->int_clr.rxfifo_tout = 1;
+ 	}
```
me-no-dev: delete one tab here

cores/esp32/esp32-hal-uart.c
```diff
@@ -96,14 +110,14 @@ static void IRAM_ATTR _uart_isr(void *arg)
...
-    uart->dev->conf1.rx_tout_thrhd = 2;
+    uart->dev->conf1.rxfifo_full_thrhd = 112; // needs speed compensation?
+    uart->dev->conf1.rx_tout_thrhd = 32; // no need to force rxFiFo empty, because uartRead() will empty if necessary
     uart->dev->conf1.rx_tout_en = 1;
     uart->dev->int_ena.rxfifo_full = 1;
     uart->dev->int_ena.frm_err = 1;
     uart->dev->int_ena.rxfifo_tout = 1;
     uart->dev->int_clr.val = 0xffffffff;
-
+// can this share an interrupt?
```
me-no-dev: why?

me-no-dev:
I am thinking that the HAL read functions should return int -1 on error. What do you think?

me-no-dev:
oopsss :D

cores/esp32/esp32-hal-uart.c
stickbreaker: ok, changed to spaces

stickbreaker:
>@me-no-dev 
> I am thinking that the HAL read functions should return int -1 on error. What do you think?

I agree `read()` needs a EOF indication. I'll can change `uartRead()` `uartPeek()` from `uint8_t` to `int16_t`.

**BUT** it will have too propagate clear through `HardwareSerial.cpp`.

Do you what this PR to include that change?

Chuck.

cores/esp32/esp32-hal-uart.c
stickbreaker: I had to share interrupts in I2C because of interrupt exhaustion problems identified at Release 1.0.0.  If it is possible to share interrupts then this change would free two.  

Maybe I'm miss-informed, is the default shared or individual?  When I specified `ESP_INTR_FLAG_EDGE` I created a conflict with a pin interrupt or timer. I had to explicitly ask for a shared interrupt with:
```
uint32_t flags = ESP_INTR_FLAG_IRAM |  //< ISR can be called if cache is disabled
          ESP_INTR_FLAG_LOWMED |   //< Low and medium prio interrupts. These can be handled in C.
          ESP_INTR_FLAG_SHARED; //< Reduce resource requirements, Share interrupts
```

I may be barking at a problem that does not exist.
 
I just see you only specifying: `ESP_INTR_FLAG_IRAM` without any shared option.
```
esp_intr_alloc(UART_INTR_SOURCE(uart->num), (int)ESP_INTR_FLAG_IRAM, _uart_isr, NULL, &uart->intr_handle);
```

me-no-dev:
Yes @stickbreaker please add that serial change here :) seems proper time

cores/esp32/esp32-hal-uart.c
me-no-dev: I vote for trying and also very important... let's change the arg in `esp_intr_alloc` to not be null but instead to point to the uart struct so loop through all interfaces will not be necessary when ISR fires.

cores/esp32/esp32-hal-uart.c
```diff
@@ -276,28 +306,38 @@ uint32_t uartAvailableForWrite(uart_t* uart)
     return 0x7f - uart->dev->status.txfifo_cnt;
 }
 
-uint8_t uartRead(uart_t* uart)
+int16_t uartRead(uart_t* uart)
```
me-no-dev: `int` is fine :) we are on 32 bit platform

stickbreaker:
Going to rebase(i think, so That I can compare to my local mods)

Issue #2033, opened 2018-11-08
WiFi memory leak after disconnect (caused by log_d) ? 


kasskas:
### Hardware:
Board:							 ESP32-WROOM
Core Installation/update date:			2018 Sep
IDE name:							Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                          no
Upload Speed:						115200
Computer OS:                                            Windows 10?

### Description:
Every time WiFi client connects and disconnects, free memory in heap is reduced. After some time this causes a crash of the MCU. The issue happens using WiFiCLientBasic example code.  Building without debug output seems to solve the issue. Could there be a possible leak in log_d output? 
After some test with none debug out, it seems that log_d in WiFi begins leaks memory. Is there any known fixes ? 

### Sketch:  
```
#include <WiFi.h>
#include <esp_wifi.h>

const char* ssid = "ssid";
const char* password = "pass";

void setup() {
	Serial.begin(115200);
	WiFi.begin(ssid, password);

	while (WiFi.status() != WL_CONNECTED) {
		delay(500);
		Serial.println("Connecting to WiFi..");
	}
	Serial.println("WiFi connected");
	Serial.println("IP address: ");
	Serial.println(WiFi.localIP());
	WiFi.disconnect(true);
	delay(500);
}

void loop() {

	WiFi.begin();
	/*while (WiFi.status() != WL_CONNECTED)
		;*/
	delay(10000);
	WiFi.disconnect(true);

	//Serial.print("after:  ");
	Serial.println(xPortGetFreeHeapSize());
	delay(10000);
}
```

### Debug Messages:
With log_d output
```
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 2 - STA_START
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 4 - STA_CONNECTED
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 7 - STA_GOT_IP
[D][WiFiGeneric.cpp:386] _eventCallback(): STA IP: 192.168.1.103, MASK: 255.255.255.0, GW: 192.168.1.1
after:  234380
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 3 - STA_STOP
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 3 - STA_STOP
[D][WiFiGeneric.cpp:342] _eventCallback(): Event: 0 - WIFI_READY
[...]

PR #2233, opened 2018-12-22
fix #2232 and #2033

### LINKED ISSUE #2232: Memory leak in log_printf (when printing exactly 64 characters plus trailing \0)
Same problem as reported (and fixed) here: #2201 

Buffer is being allocated if `vsnprintf` reports 64 or more characters (due to the trailing \0):
https://github.com/espressif/arduino-esp32/blob/310e78e6fded9414411eb48dcf212018c3791483/cores/esp32/esp32-hal-uart.c#L440-L441
But the buffer is only being released if *more than 64 characters* had been reported by `vsnprintf`:
https://github.com/espressif/arduino-esp32/blob/310e78e6fded9414411eb48dcf212018c3791483/cores/esp32/esp32-hal-uart.c#L459-L460

Probably related: #2033 (happens only when debug is on and free heap drops by 64 bytes each time)

Code to reproduce:
```
log_printf("1234567890123456789012345678901234567890123456789012345678901234");
```

atanisoft:
@felixstorm reported one more location where a buffer is created but not freed when the size is exactly 64 bytes.

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -456,7 +456,7 @@ int log_printf(const char *format, ...)
     ets_printf("%s", temp);
 #endif
     va_end(arg);
-    if(len > 64){
+    if(len > sizeof(loc_buf)){
         free(temp);
     }
     return len;


felixstorm:
@atanisoft Thanks for the PR!

PR #2249, opened 2018-12-28
Fix for regression of millis(). It wraps back to zero every 4295 seconds otherwise.

lyusupov:
Regression has been noticed in between release 1.0.0 and current git version of the Core.
When millis() is in use for system uptime measurement, it now wraps every 1+ hour  (4295 seconds) -
which is not good for this purpose.
This fix brings the maximum calculated uptime value (based on millis())  back to 1000+ hours.

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -145,7 +145,7 @@ unsigned long IRAM_ATTR micros()
 
 unsigned long IRAM_ATTR millis()
 {
-    return (unsigned long) (micros() / 1000);
+    return (unsigned long) (esp_timer_get_time() * _sys_time_multiplier / 1000);
 }
 
 void delay(uint32_t ms)


me-no-dev:
does this happen on cpu frequencies below 80MHz or always?

lyusupov:
I've noticed this regression on 80 MHz.
Double casting from uint64 into uint32 through micros() causes the issue.

me-no-dev:
ok :) I will close this PR though, because I am working on the cpu code and it will cause a nasty git conflict. Will include the fix in the new code :)

me-no-dev:
Initial changes are here: https://github.com/espressif/arduino-esp32/commit/daea578841059fef36f98d72cef15e22733bdc3e

Please have a look :) and thanks for catching this.

lyusupov:
> Initial changes are here: daea578

Looking good to me. Thank you!
Are there any time estimates for the code to become committed soon ?

me-no-dev:
yes, I just need to integrate the clock change in a couple of drivers and will push it to master :)

lyusupov:
Good!

Should I close this PR or you will ?

me-no-dev:
let's close it :)

PR #2250, opened 2018-12-28
[WIP] Handle APB frequency change

me-no-dev:


lbernstone:
No way to get an event generated when the frequency changes?  Seems like tick rate is far more important to FreeRTOS than anything in arduino-esp32.

me-no-dev:
there is code for adjusting tick rate in IDF's PM code, so that can be done. But no, there is no event from PM when it changes frequency (it does adjust tick rate but does not tell drivers to adjust their rates if APB changed)

cores/esp32/esp32-hal-cpu.c
```diff
@@ -0,0 +1,169 @@
...
+    }
+    capb = calculateApb(&cconf);
+    apb = calculateApb(&conf);
+    log_i("%s: %u / %u = %u Mhz", (conf.source == RTC_CPU_FREQ_SRC_PLL)?"PLL":((conf.source == RTC_CPU_FREQ_SRC_APLL)?"APLL":((conf.source == RTC_CPU_FREQ_SRC_XTAL)?"XTAL":"8M")), conf.source_freq_mhz, conf.div, conf.freq_mhz);
+    if(capb != apb && apb_change_callbacks){
+        triggerApbChangeCallback(APB_BEFORE_CHANGE, capb, apb);
+    }
+    rtc_clk_cpu_freq_set_config_fast(&conf);
+    _cpu_freq_mhz = conf.freq_mhz;
+    _sys_time_multiplier = 80000000 / apb;
```
atanisoft: perhaps a call to the IDF PM update function here so that RTOS Ticks are updated, or update it like the PM module does: https://github.com/espressif/esp-idf/blob/master/components/esp32/pm_esp32.c#L316 (why it does it this way instead of a call to _xt_tick_divisor_init() I'm not sure)

cores/esp32/esp32-hal-cpu.c
```diff
@@ -0,0 +1,169 @@
...
+    if(cconf.freq_mhz == cpu_freq_mhz && _cpu_freq_mhz == cpu_freq_mhz){
+        return true;
+    }
+    if(!rtc_clk_cpu_freq_mhz_to_config(cpu_freq_mhz, &conf)){
+        log_e("CPU clock could not be set to %u MHz", cpu_freq_mhz);
+        return false;
+    }
+    capb = calculateApb(&cconf);
+    apb = calculateApb(&conf);
+    log_i("%s: %u / %u = %u Mhz", (conf.source == RTC_CPU_FREQ_SRC_PLL)?"PLL":((conf.source == RTC_CPU_FREQ_SRC_APLL)?"APLL":((conf.source == RTC_CPU_FREQ_SRC_XTAL)?"XTAL":"8M")), conf.source_freq_mhz, conf.div, conf.freq_mhz);
```
atanisoft: this may be better at log_d instead of log_i but I'd defer to your judgement on specific level to log this at.

cores/esp32/esp32-hal-cpu.c
```diff
@@ -0,0 +1,169 @@
...
+    apb_change_t * r = apb_change_callbacks;
+    while(r != NULL){
+        r->cb(r->arg, ev_type, old_apb, new_apb);
+    }
+    xSemaphoreGive(apb_change_lock);
+}
+
+bool addApbChangeCallback(void * arg, apb_change_cb_t cb){
+    initApbChangeCallback();
+    apb_change_t * c = (apb_change_t*)malloc(sizeof(apb_change_t));
```
atanisoft: is there a template class that can be used for linked lists that is generic enough to plug in here, maybe even something from ESP-IDF that can be leveraged instead of managing the list here (and later in remove)?

cores/esp32/esp32-hal-cpu.c
me-no-dev: yes yes :) already mentioned in my initial post that this needs to be done

cores/esp32/esp32-hal-cpu.c
me-no-dev: will remove it all together but it's a good test point to ensure that UART flushes before cpu can be changed. nd I often keep log at info so I do not get all unnecessary debug :P

cores/esp32/esp32-hal-cpu.c
me-no-dev: I don't know of one, but I have been thinking of such thing for some time. This is C though... so IDK

me-no-dev:
well... I do not see IDF to be updating peripheral clocks when APB changes anyway... so they might benefit from such functionality and Arduino can benefit as well. But I can't just let PM lose here or we will get a lot of complaints about "garbage in serial" and more

cores/esp32/esp32-hal-cpu.c
atanisoft: is there a reason to stick with C only here, there are already a few C++ pieces here and there...

atanisoft: you can also use something like this to mix C++ into C...
`template<int signalGenerator>
void IRAM_ATTR signalGeneratorTimerISR(void)
`
it has C linkage but you can call it with different values to get different versions of it (this is part of how I do ISRs using one method but two diff data sources)

me-no-dev: the reason why it is C is so the whole HAL can be used in C-only libs and extensions. Not really sure if it's worth it, but it's how it started :)

me-no-dev:
So I had a talk with Ivan last night on PM and clock changes. The reason why PM does not have callbacks is performance. That is quite understandable as waiting on peripherals to switch clock might take long time. One thing that we can do is switch UART and LEDC to use REF_TICK which is stable when APB/CPU clocks are changed, but it's currently hard defined as 1MHz which is not optimal for UART. I will investigate what the consequences are if it is changed and make a PR to IDF if possible and add it to menuconfig.

me-no-dev:
For this PR it should be sufficient for now to change RTOS tick and esp_timer to behave properly and then we can go with implementing the callbacks in the peripherals (and test them). That way we can at least have a working manual CPU/APB switching. Another point is that for WiFi to work, APB must be at 80MHz and since APB can not be higher than CPU, min CPU for WiFi use must be 80MHz

copercini:
for now, just a small trash on Serial during the change (from default clock to 80Mhz)
image
code: https://gist.github.com/copercini/25265bdde2db4fd111db48fba8d96b44

me-no-dev:
@copercini this looks like a different issue ;) it happens too early. maybe buffer overrun or something

atanisoft:
@me-no-dev it could be related to the change if there are still characters in the uart tx buffer, it may be a good idea to flush the buffers prior to change.

me-no-dev:
@copercini uart baudrate switching is not implemented (though in your case APB will not change so WTF?)

me-no-dev:
@copercini you should try this latest code. I do not see bad chars in the serial anymore.
test code:
```cpp
void setup() {
    Serial.begin(115200);
    Serial.setDebugOutput(true);
    Serial.println();

    Serial.printf("CPU Freq: %uHz\n", getCpuFrequency() * 1000000);
    Serial.printf("APB Freq: %uHz\n", getApbFrequency());

    setCpuFrequency(3);

    Serial.printf("CPU Freq: %uHz\n", getCpuFrequency() * 1000000);
    Serial.printf("APB Freq: %uHz\n", getApbFrequency());

    Serial.printf("1: %lu, %lu\n", micros(), millis());
    delayMicroseconds(1000000);
    Serial.printf("2: %lu, %lu\n", micros(), millis());
    delay(1000);
    Serial.printf("3: %lu, %lu\n", micros(), millis());
    delay(10000);
    Serial.printf("4: %lu, %lu\n", micros(), millis());
}
```

Did not work below 3MHz though...

atanisoft:
@copercini with @me-no-dev latest code the Serial.flush() call may not be necessary.

me-no-dev:
yup, no need to call flush :)

copercini:
ok, I saw the internal flush here https://github.com/espressif/arduino-esp32/blob/cpu-clock-integration/cores/esp32/esp32-hal-uart.c#L383

but `getLocalTime()` takes some time to answer, still bugging even with the internal flush
image

me-no-dev:
@copercini this is bugging too early for the code to be at fault

me-no-dev:
could it be something else?

cores/esp32/esp32-hal-cpu.c
```diff
@@ -0,0 +1,183 @@
...
+    xSemaphoreTake(apb_change_lock, portMAX_DELAY);
+    if(apb_change_callbacks == NULL){
+        apb_change_callbacks = c;
+    } else {
+        apb_change_t * r = apb_change_callbacks;
+        if(r->cb != cb || r->arg != arg){
+            while(r->next){
+                r = r->next;
+                if(r->cb == cb && r->arg == arg){
+                    goto unlock_and_exit;
```
stickbreaker: Memory Leak if callback added twice.

Chuck.

me-no-dev: good catch!

me-no-dev:
guys :) all peripheral support is added :) give it a shot and check for errors

me-no-dev:
ok, let all hell loose :D

cyberman54:
Could we remove the `log_v` 's from `esp32-hal-i2c.c` which came in with this PR?
They produce heavy logging loads in debug mode verbose, making verbose hard to use for other purposes.

Or is there a way to suppress the `log_v` 's from this module during runtime?

/cc @stickbreaker

stickbreaker:
@cyberman54 Sure, There hasn't been any I2C problems since this last release, I'll move them behind the `
#define ENABLE_I2C_DEBUG_BUFFER`

Issue #2212, opened 2018-12-18
Occasional OTA crash


poengAlex:
### Hardware:
Board:							ESP32-wroom-32D (custom board)
Core Installation/update date:			1.0.0?
IDE name:							Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                          no
Upload Speed:						115200
Computer OS:                                            Windows 10

### Description:
Hi

I have build a simple test program to check the OTA functionality on ESP32. There are 2 versions of a bin fil I download with on integer value different in the two. Both programs start OTA and download the other bin file. If the same bin fil starts after a reset the OTA must have failed somehow. I also have a timer that restarts the MCU after 5 minutes.
After running the program over the weekend I have 6722 successful OTA and 162 fails. 2.4% fail rate. There is a few instances (<5) where the MCU simply freeze and do the restart failsafe kicks in. But the most of the failures there is a LoadStoreError. The full exception you can find bellow.
Do anyone else have this problem? Any solutions?

### Sketch: 
```cpp

	Update.onProgress(OnProgress); 
	ESPhttpUpdate.rebootOnUpdate(false);
	t_httpUpdate_return ret = ESPhttpUpdate.update(bin);

	switch (ret) {
	case HTTP_UPDATE_FAILED:
		Serial.printf("HTTP_UPDATE_FAILD Error (%d): %s", ESPhttpUpdate.getLastError(), ESPhttpUpdate.getLastErrorString().c_str());
		vTaskDelay(1000);
		ESP.restart();
		break;

	case HTTP_UPDATE_NO_UPDATES:
		Serial.println("HTTP_UPDATE_NO_UPDATES");
		vTaskDelay(1000);
		ESP.restart();
		break;

	case HTTP_UPDATE_OK:
		Serial.println("HTTP_UPDATE_OK");
		vTaskDelay(1000);
		ESP.restart();
		break;
	}
```

### Debug Messages:
```
Guru Meditation Error: Core  1 panic'ed (LoadStoreError). Exception was unhandled.
Core 1 register dump:
PC      : 0x400d8f3c  PS      : 0x00060930  A0      : 0x800d8fdc  A1      : 0x3ffb1d00  
A2      : 0x3ffbc01c  A3      : 0x3ffb1d4f  A4      : 0x00000001  A5      : 0x40000000  
A6      : 0x00043814  A7      : 0x04d8e925  A8      : 0x00000000  A9      : 0xffffffff  
[...]

PR #2259, opened 2019-01-01
Fix WiFiClientRxBuffer::fillBuffer() error, likely solves issue #2212

Jeroen88:
Fix WiFiClientRxBuffer::fillBuffer() error in handing a negative res from the call to recv()

In https://github.com/espressif/arduino-esp32/blob/229d9b7366deee0d8a90fb69c19119f7287c9d1d/libraries/WiFi/src/WiFiClient.cpp#L70 if res is negative and errno equals EWOULDBLOCK then the if() condition is false causing _fill to be decremented. 
The condition res < 0 && errno == EWOULDBLOCK simply means that the socket has no data available (yet), so _fill should not change and, since nothing is read, the return value of fillBuffer should be zero.

This most likely solves the complete #2212 issue, at least it solves the minimum sketch reproducing the error as posted in that issue by @lbernstone. This could also be the root cause of other issues, since on slow connections it is likely that the socket not always has data available.

I also added an extra test on https://github.com/espressif/arduino-esp32/blob/229d9b7366deee0d8a90fb69c19119f7287c9d1d/libraries/WiFi/src/WiFiClient.cpp#L60 by setting _failed to true if the malloc() fails. I am not sure and did not check if the _failed condition is checked in the other functions, but it should not do any harm. Moreover, this did not cause the error.

Happy New Year to everyone!

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -58,6 +58,10 @@ private:
         {
             if(!_buffer){
                 _buffer = (uint8_t *)malloc(_size);
+                if(!_buffer) {
+                    _failed = true;
+                    return 0;
+                }
             }
             if(_fill && _pos == _fill){
                 _fill = 0;
@@ -67,8 +71,10 @@ private:
                 return 0;
             }
             int res = recv(_fd, _buffer + _fill, _size - _fill, MSG_DONTWAIT);
-            if(res < 0 && errno != EWOULDBLOCK) {
-                _failed = true;
+            if(res < 0) {
+                if(errno != EWOULDBLOCK) {
+                    _failed = true;
+                }
                 return 0;
             }
             _fill += res;
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -59,6 +59,7 @@ private:
             if(!_buffer){
                 _buffer = (uint8_t *)malloc(_size);
                 if(!_buffer) {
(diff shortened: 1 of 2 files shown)


libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -58,6 +58,10 @@ class WiFiClientRxBuffer {
         {
             if(!_buffer){
                 _buffer = (uint8_t *)malloc(_size);
+                if(!_buffer) {
```
copercini: Please add an error log like `log_e("Not enough memory to allocate buffer");`

this helps a lot on debug =)

lbernstone:
I'm putting a different fix in for this that seems closer to the root.  Put em both out there and see what shakes out, I figure...

Jeroen88:
@lbernstone, @nikuz confirmed that my PR fixes the error. Also, the if() in WiFiClientRxBuffer::fillBuffer() is obviously wrong...

lbernstone:
Verified.  Looks like you included some BLE stuff in your commit, tho.

Jeroen88:
@lbernstone You are right, I always pull master before creating a branch. So I was surpised to have some BLE stuff in my master

me-no-dev:
I can not merge this with the BLE stuff in there ;)

Jeroen88:
@me-no-dev can you help me get it out? I did a git pull upstream master and then it got in...????

me-no-dev:
delete the BLE folder then call `git submodule update --init --recursive` in the root arduino folder

Jeroen88:
Replaced by #2263

PR #2270, opened 2019-01-03
Added ESP32 FM DevKit board

dragon-engineer:
Added new ESP32 board, FM DevKit. Audio and FM transmitter-based board with OLED, buttons, USB-UART bridge, RF output and I2S DAC with headphones output


atanisoft:
you mention this board has an OLED onboard, if this is the case please add the OLED defines as seen on the heltic variants if applicable.

dragon-engineer:
> you mention this board has an OLED onboard, if this is the case please add the OLED defines as seen on the heltic variants if applicable.

It is a generic I2C OLED. Do you mean to add defines for X and Y size in pixels?

variants/fm-devkit/pins_arduino.h
```diff
@@ -0,0 +1,53 @@
...
+static const uint8_t D8 = 15;
+static const uint8_t D9 = 23;
+static const uint8_t D10 = 0;
+
+// I2C BUS, 2k2 hardware pull-ups
+static const uint8_t SDA = 16;
+static const uint8_t SCL = 17;
+
+// SPI - unused but you can create your own definition in your sketch
+static const uint8_t SCK = 0;
```
me-no-dev: change those to `int8_t` and value `-1`. 0 is a valid pin

PR #2275, opened 2019-01-04
not Validated.  I2C, UART mods to support Cpu freq Changes

stickbreaker:
#  UNTESTED 
## debug coding still active 

**hal-uart modifications**
* changed to support naked receive(operation without FreeRTOS queue)
* changed to improve realtime concurrency
  read(),peek() will access rxFifo if necessary,
  available() will include rxFifo content in the count.
  read(),peek() will return -1 if no data available
* will compensate for cpufrequency changes to maintain configured Baud rate
  Baudrate is NOT validated
* hal-uart supports 3mhz minimum cpu clock rate.

**hal-i2c modifications**
* supports 3mhz minimum cpu clock rate
* reduce i2c bus clock when cpu clock decreases
  Will NOT increase bus clock when cpu freq increases.

**hal-cpu modifications**
* change callback list to double link, to support unwind
* add APB_ABORT_CHANGE ev_type to allow drivers to refuse change request
* change callback to bool

### DIFF
--- a/cores/esp32/esp32-hal-cpu.c
+++ b/cores/esp32/esp32-hal-cpu.c
@@ -28,6 +28,7 @@
 
 typedef struct apb_change_cb_s {
         struct apb_change_cb_s * next;
+        struct apb_change_cb_s * prev;
         void * arg;
         apb_change_cb_t cb;
 } apb_change_t;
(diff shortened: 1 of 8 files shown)


cores/esp32/esp32-hal-cpu.c
```diff
@@ -37,54 +38,110 @@ const uint32_t MHZ = 1000000;
...
+}
+digitalWrite(D_A,LOW);
+digitalWrite(D_B,LOW);
+}  
+
+
 static void initApbChangeCallback(){
     static volatile bool initialized = false;
     if(!initialized){
-        initialized = true;
```
me-no-dev: it was like that on purpose if two threads call it at the same time. please revert :)

cores/esp32/esp32-hal-cpu.c
```diff
@@ -37,54 +38,110 @@ const uint32_t MHZ = 1000000;
...
+        }
+*/
     xSemaphoreGive(apb_change_lock);
+    return success;
+
 }
 
 bool addApbChangeCallback(void * arg, apb_change_cb_t cb){
     initApbChangeCallback();
-    apb_change_t * c = (apb_change_t*)malloc(sizeof(apb_change_t));
```
me-no-dev: why is this whole change necessary? to speed up? how many calls will be made to this function?

stickbreaker: I just hate `goto` and this flow is easier to understand, 
* lock
* check duplicate
 add if new
* unlock

the double linked list may not be necessary,  I just added it while testing to see if execution calling order made a difference,  I was running into deadlocks because of `log_x()` calls during debuging

stickbreaker:
This pr was just an easy way to get your comments.  
The recursiveMutex's were necessary in i2c because in my next version 1.2.0 that supports multithreaded calls I needed a transactional mutex.

The recursiveMutex in UART was an attempt to handle `log_x()` during freq changes.  

Limiting processor speeds in the callback allows minimum speed to be set by the fastest peripheral.
I don't know what minimum speeds are necessary for SPI, i2s, irda(UART), PWM. So this method allows me to got down as far as i2c can handle.  If other hardware has a higher limit, that is ok.

The DEBUG_RX digitalWrites are testing for RX, I am disabling RX interrupt is queue is full.  It is part of my receive validation.
   
This is not a candidate for merge, I just wanted to see what you thought about the additional `APB_ABORT_CHANGE` 

It is just easy to use the diff support of merger.

me-no-dev:
@stickbreaker there are so many changes that it is hard to get a point of what you are doing :) maybe if you want to propose something, keep it separate and easier to follow. Change the least possible code. We can discuss optimisations later. Now I thought that you propose to merge all this and was quite confused by it 😄

cores/esp32/esp32-hal-uart.h
```diff
@@ -56,15 +56,16 @@ void uartEnd(uart_t* uart);
 
 uint32_t uartAvailable(uart_t* uart);
 uint32_t uartAvailableForWrite(uart_t* uart);
-uint8_t uartRead(uart_t* uart);
-uint8_t uartPeek(uart_t* uart);
+int16_t uartRead(uart_t* uart);
```
me-no-dev: better use `int` if size of the value is not important. It was 8bit by default, because uart is 8 bit.

cores/esp32/esp32-hal-uart.c
```diff
@@ -45,73 +45,126 @@ struct uart_struct_t {
...
     uint8_t i, c;
-    BaseType_t xHigherPriorityTaskWoken;
-    uart_t* uart;
+    BaseType_t xHigherPriorityTaskWoken=pdFALSE;
+    uart_t* uart =(uart_t *)arg;
+    if( uart != NULL){
+        UBaseType_t spaces=0,count=0;
+        count = (uart->dev->mem_rx_status.wr_addr - uart->dev->mem_rx_status.rd_addr)%0x80;
+        if(uart->queue != NULL) {
+            spaces=uxQueueSpacesAvailable( uart->queue);
```
me-no-dev: `uxQueueSpacesAvailable` ok to be called from isr?

stickbreaker: I can't find any documentation either way.

me-no-dev:
interesting... on esp32 the uart seems not to have register bits for switching rx or tx on-off... sad

stickbreaker:
See Clock1, this work is done.

Issue #476, opened 2017-06-30
Changing UART Baud rate messes up UART


eroom1966:
Hello,

I think I have a similar issue to https://github.com/espressif/arduino-esp32/issues/441

If I configure UART2 to run at 57600 baud, I can send data messages and all works well.
However, I need to be able to perform configuration of the end device at 9600 baud,
then flip both sides of the interface into 57600 - I cannot get this to work reliably.

I have a fixed message length of 24 bytes (at 57600 baud), but when I send messages, there is some strange buffering occuring, such that on the serial line, 21 bytes of the message(0) are sent, then there is a pause until the next message(1) is sent, then I have the missing 3 bytes from message(0), followed by the next 21 bytes if message(1) - and so on.

This only occurs IF the baud rate is changed from 9600 to 57600, if I start with a baud rate of 57600, the messages are transmitted as expected. so effectively what I have is the following

void setup() {
    Serial2.begin(9600, SERIAL_8N1, gpioX, gpioY);
}

in the functional code, at some point the interface is reconfigured
Serial2.begin(57600, SERIAL_8N1, gpioX, gpioY);

I have tried with various combinations of 
Serial2.end(); // causes the program to stop

and adding delays to no avail.

so can anybody provide a snippett of code to show how to reconfigure the BAUD rate of the interface, without introducing the buffering issue described above.

E.

eroom1966:
I now have a simple loopback testcase on Serial2 that shows this issue, as you will see if you run this, the data sent over the serial is totally corrupted after a change of baud rate
the test simply sends a message with a timestamp
```
write(12    ): 0123456789 BAUD=9600
 read(36    ): 0123456789 BAUD=9600
```
after changing the baud rate, its totally garbled
```
write(20101 ): 0123456789 BAUD=57600
 read(20105 ): ⸮8⸮-E⸮4⸮%l0&⸮⸮⸮⸮?⸮⸮
```

```
#include <HardwareSerial.h>
//
// Note: you need to link IO16 to IO17
//
HardwareSerial Ser2(2);

#define RF_ESP_RX  16
#define RF_ESP_TX  17
#define RF_BAUD_HI 57600
#define RF_BAUD_LO 9600

void setup() {
    Ser2.begin(RF_BAUD_LO, SERIAL_8N1, RF_ESP_RX, RF_ESP_TX);
    Serial.begin(RF_BAUD_LO);
    Serial.println("\r\n** Start **\r\n");
}

[...]

PR #2223, opened 2018-12-20
HardwareSerial - add changeBaudRate method

toomasz:
- This method will be useful for libraries that attempt to find current baud rate of device connected via Serial, for example GSM modems
 - Please be aware that some garbage on Serial may occur(an most of times it does) after baud rate is changed

Fixes:
#441

PR #2276, opened 2019-01-05
Update HardwareSerial.h

terryjmyers:
This should resolve issue #476 and easily allows access to the uartSetBaudRate function in the HAL

### DIFF
--- a/cores/esp32/HardwareSerial.h
+++ b/cores/esp32/HardwareSerial.h
@@ -86,6 +86,10 @@ public:
     {
         return write((uint8_t) n);
     }
+    void setBaudRate(uint32_t baud) 
+    {
+        uartSetBaudRate(_uart, baud);
+    }
     uint32_t baudRate();
     operator bool() const;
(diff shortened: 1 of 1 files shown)


cores/esp32/HardwareSerial.h
```diff
@@ -86,6 +86,10 @@ class HardwareSerial: public Stream
     {
         return write((uint8_t) n);
     }
+    void setBaudRate(uint32_t baud)
```
me-no-dev: don't put the implementation in the header please. just declare here and add to the cpp file

toomasz: Isn't it same as https://github.com/espressif/arduino-esp32/pull/2223 which is already merged?

atanisoft: Yes, it appears to be a duplicate. Instead of calling it "updateBaudRate" this PR is calling it "setBaudRate"

PR #2279, opened 2019-01-05
FIX - AsyncUDP joinMulticastGroup does not work if interface is not specified

forkineye:
If ```joinMulticastGroup`` is used with the default ```tcpip_if``` of ```TCPIP_ADAPTER_IF_MAX```, an ungraceful crash will occur.  This modifies ```joinMulticastGroup``` to utilize ```igmp_joingroup``` instead of ```igmp_joingroup_netif``` when ```TCPIP_ADAPTER_IF_MAX``` is passed in so that all interfaces are iterated over and joined.


me-no-dev:
Thanks @forkineye

PR #2301, opened 2019-01-10
I2C fix READ of zero bytes hardware hang

stickbreaker:
The i2c peripheral will hang if a READ request is issued with a zero data length.  The peripheral
drops into a continuous timeout interrupt response.  The STOP command can not be set out to the I2C
bus. The SLAVE device correctly ACK'd the address byte, with READ bit set, it has control of the SDA
 pin.  The ESP32 send out the next SCL HIGH pulse but, since the SLAVE is in READ Mode, and the First
bit it is sending happened to be a ZERO, the ESP32 cannot send the STOP.  When it releases SDA during
the SCL HIGH, the pin does not change state.  The pin stays low because the SLAVE is outputing a LOW!
The ESP32 drops into a perminent WAIT state waiting for SDA to go HIGH (the STOP).

**esp32-hal-i2c.c**
* add databuff length checks to `i2cRead()` and `i2cWrite()`

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -1609,6 +1609,9 @@ i2c_err_t i2cFlush(i2c_t * i2c)
 }
 
 i2c_err_t i2cWrite(i2c_t * i2c, uint16_t address, uint8_t* buff, uint16_t size, bool sendStop, uint16_t timeOutMillis){
+    if((i2c==NULL)||((size>0)&&(buff==NULL))) { // need to have location to store requested data
+        return I2C_ERROR_DEV;
+    }
     i2c_err_t last_error = i2cAddQueueWrite(i2c, address, buff, size, sendStop, NULL);
 
     if(last_error == I2C_ERROR_OK) { //queued
@@ -1628,6 +1631,9 @@ i2c_err_t i2cWrite(i2c_t * i2c, uint16_t address, uint8_t* buff, uint16_t size, 
 }
 
 i2c_err_t i2cRead(i2c_t * i2c, uint16_t address, uint8_t* buff, uint16_t size, bool sendStop, uint16_t timeOutMillis, uint32_t *readCount){
+    if((size == 0)||(i2c == NULL)||(buff==NULL)){ // hardware will hang if no data requested on READ
+        return I2C_ERROR_DEV;
+    }
     i2c_err_t last_error=i2cAddQueueRead(i2c, address, buff, size, sendStop, NULL);
 
     if(last_error == I2C_ERROR_OK) { //queued


cores/esp32/esp32-hal-i2c.c
```diff
@@ -1628,6 +1631,9 @@ i2c_err_t i2cWrite(i2c_t * i2c, uint16_t address, uint8_t* buff, uint16_t size,
 }
 
 i2c_err_t i2cRead(i2c_t * i2c, uint16_t address, uint8_t* buff, uint16_t size, bool sendStop, uint16_t timeOutMillis, uint32_t *readCount){
```
me-no-dev: if size is 0, why return an error? if size is not zero, but buf is null, then I understand, but size 0 is fine to return OK

stickbreaker: A zero byte read creates an invalid I2C bus state.  Depending on what the first bit the SLAVE responds with, the i2c peripheral may not be able to complete the transmission.  A zero byte write is possible, but a zero byte read is not allowed.  The zero byte write is used as a presence detect event.:
```c++
Wire.beginTransmission(ID);
uin8_t error = Wire.endTransmission();
if(error == 2) { // slave did not answer, is not present
  Serial.printf("NAK, no slave present at 0x%02X\n",ID);
} else if (error == 0) {// Slave ready for communications
  Serial.printf("ACK, slave present at 0x%02X\n",ID);
}
```
The following code is invalid:
```c++
uint8_t requestSize = 0;
uint8_t count = Wire.requestFrom(ID,requestSize);
if( count == requestSize) {// Slave is present and ready for communications
   Serial.printf("ACK, slave present at 0x%02X, received %dbytes\n",ID,count);
   Serial.printf(" requestFrom succeeded =%d(%s)\n",Wire.lastError(),Wire.getErrorText(Wire.lastError()));
   for (uint8_t i = 0 ; i< count; i++){
     Serial.print(Wire.read());
  }
  Serial.println():
}
else { // problem
   Serial.printf(" requestFrom Failed =%d(%s)\n",Wire.lastError(),Wire.getErrorText(Wire.lastError()));
}
```
change `requestSize` to a non zero value and the code works a expected, 
The act of requesting a zero byte read must error out. 

A Zero byte Write will return information, Whether a Slave device is active on the bus.

A Zero byte Read is meaningless. and will only succeed if the correct circumstances exist, else it will hang the i2c bus creating a error cascade that is confusing.

Chuck.

me-no-dev: ok...

valki2:
I cannot stress out how happy I am that you fixed this. i2c was driving me nuts and I had no clue where it came from  :)

asetyde:
I think is my issue .. sht21 not working rc4

PR #2333, opened 2019-01-16
 add missing EEPROM lib to CMakeLists.txt, moved the EEPROM files to proper dir

marcogmaia:
CMake fix for EEPROM lib


me-no-dev:
https://github.com/espressif/arduino-esp32/tree/master/libraries/EEPROM the path was correct ;)

me-no-dev:
ok so the include folder was missing, but why move the files?

PR #2341, opened 2019-01-17
honor CONFIG_TCPIP_TASK_AFFINITY (merged)

qt1:
Honor global preference CONFIG_TCPIP_TASK_AFFINITY in arduino

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -96,7 +96,7 @@ static bool _start_network_event_task(){
         }
     }
     if(!_network_event_task_handle){
-        xTaskCreatePinnedToCore(_network_event_task, "network_event", 4096, NULL, ESP_TASKD_EVENT_PRIO - 1, &_network_event_task_handle, ARDUINO_RUNNING_CORE);
+        xTaskCreatePinnedToCore(_network_event_task, "network_event", 4096, NULL, ESP_TASKD_EVENT_PRIO - 1, &_network_event_task_handle, CONFIG_TCPIP_TASK_AFFINITY);
         if(!_network_event_task_handle){
             log_e("Network Event Task Start Failed!");
             return false;


qt1:
In my case I use this to free the arduino core as I dedicate it to some intensive RMT stuff..

At the time I tried to create this flag but it turned out that the flag already exists and it is used in the ESP IDF code, so it make sense to honor the same flag in the arduino as well.

I use it for a few months now and everything seems OK.

me-no-dev:
I do not say that it's not going to work, but this task is sleeping 99.99% of the time. There are rarely any events coming there.

me-no-dev:
BTW are you using Arduino as IDF component?

qt1:
Yes, Arduino as a component + vscode

This is how my ~/esp looks like:

├── esp-idf
├── openocd-esp32    <-- working some for the time..
├── rmt_sender  <-- my program
│   ├── components
│   │   ├── arduino
│   │   └── ibn_esp32_gen_lib  <-- my library
│   └── main
└── xtensa-esp32-elf

me-no-dev:
I'm thinking maybe we should add this as menuconfig option. Generally I have chosen that task to run on Core 1 because it's an Arduino task and all Arduino things run on core 1.

Will you be satisfied with such option?

qt1:
I think it is already in munuconfig. I'll check again later.

me-no-dev:
i meant to separate the setting for Arduino's loop and the network task :)

qt1:
I found the flag I used here:
components/arduino/tools/sdk/sdkconfig:CONFIG_TCPIP_TASK_AFFINITY_NO_AFFINITY=

Looks like it's already in the config. only needs to be used in the arduino version of the code.

I can change it by running `make menuconfig` then going to  Component config / WiFi / WiFi Task Core ID

me-no-dev:
`CONFIG_TCPIP_TASK_AFFINITY_NO_AFFINITY` is an IDF option and not Arduino :) Also the event task has nothing to do with TCP/IP. I do not agree to make the event task be tied to the tcpip task, but I am willing to add an option so you can select the core that you want that event task to run on :)

qt1:
Yes, the flag is used in lwip layer, not in the arduino code.

A different flag in Arduino will give the control I want (namely - separate some time critical code to the arduino core and keep the TCPIP operational).

Thanks!

me-no-dev:
done :) https://github.com/espressif/arduino-esp32/commit/fa74767b2e4b4808439f81b97cb454e1118621f8

PR #2380, opened 2019-01-25
Board T-Beam: Update pins_arduino.h

cyberman54:
DAC2 is not wired, only DAC1 is -> corrected
LORA pins 27 + 19 are only wired on pcb, but not connected to header -> deleted
Button and LED added


lemmingDev:
static const uint8_t LED_BUILTIN = 14;

I also added the above code to get LED_BUILTIN working on the board - as missing from the pin definition file...

PR #2383, opened 2019-01-26
Add connect timeout to WiFiClient

boarchuz:
Allows a timeout, in milliseconds, to be passed as an optional parameter to WiFiClient.connect().

Currently this is a blocking function which can take up to 20 seconds to return if the server isn't responding. 

- Incidentally, the 'normal' timeout will trigger an error at the usual time so behaviour is unchanged if a timeout isn't provided! (Unfortunately, it will also trigger if the provided timeout is longer; timeout duration is therefore effectively capped at whatever it is normally.) Ideally this would be overriden. Also, a defined default value might be more a better approach (see initial commit for eg.).

- Perhaps using setTimeout would be a better way to set the value. That would leave signatures unchanged and only require setting the value once, instead of tacking it on to every connect().


me-no-dev:
Nice! Thanks for adding this :)

PR #2397, opened 2019-01-29
New version of the Olimex boards ESP32-EVB/Gateway/PoE

Stanimir-Petev:
In Board.txt file added board revision submenu for different revisions of the same board. By selecting different revision in the Arduino menu different value of the predefined macro will be applied which is used inside the pins_arduino.h file to make different defines and values for the specific revision.

Added PartitionScheme submenu to our 3 boards, similar to the other ESP32 boards.

In esp32-gateway pin defintions preprocessor conditions added to distingiush different revisions.

In esp32-poe pin definitions is added ethernet power enable pin and also defined macro: BOARD_HAS_1BIT_SDMMC.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -6,6 +6,7 @@ menu.FlashSize=Flash Size
 menu.PartitionScheme=Partition Scheme
 menu.DebugLevel=Core Debug Level
 menu.PSRAM=PSRAM
+menu.Revision=Board Revision
 
 ##############################################################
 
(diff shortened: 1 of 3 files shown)


me-no-dev:
готово :)

me-no-dev:
Hey we have some issues with the new definitions. The define which changes the board version causes warnings in Arduino. More info can be found here: https://github.com/espressif/arduino-esp32/issues/2463
Let's find a better way to do this :)

Stanimir-Petev:
Thanks for the report! You and stickbreaker are right about it.

I didn't notice that if a board different than ESP32-Gateway is selected there will be no revision submenu and thus no revision selected. As a result - no macro of the board will be defined at all. So it seems the line "#esp32-gateway.build.board=ESP32_GATEWAY" should be uncommented in order to solve this.

This way if no revision is selected (which is in case of any other board chosen) there will be definition of the macro "ARDUINO_ESP32_GATEWAY" WITHOUT value, and if Gateway with revision is selected then we will have the same macro WITH value corresponding to the revision selected and depending on this value in the "pins_arduino.h" other macros will/won't be defined to match the hardware of the revision.

It was just my mistake on testing not noticing this for which I apologize, and thanks for correcting me.

me-no-dev:
@Stanimir-Petev you want to add a PR for this, or should I edit it myself?

Stanimir-Petev:
I don't think another pull request is needed since it is just one line that should be uncommented.
So yes, if possible edit it yourself, but if I have to I will do it.

me-no-dev:
https://github.com/espressif/arduino-esp32/commit/628b8f0c299148b3f90369bc6dd77aa031b2f09a
Done :)

PR #2392, opened 2019-01-28
Update Heltec Automation(TM) boards defintion.

Heltec-Aaron-Lee:
Update Heltec Automation(TM) boards defintion, include WiFi_Kit_32, WiFi_LoRa_32, WiFi_LoRa_32_V2, Wireless_Stick defintions.

Please check.

PR #2407, opened 2019-01-30
Heltech LoRa 32 V2 Update

n9wxu:
Updated the board support for the Heltech ESP32 LoRa V2 hardware.

Changes:
1) Partition updated for the 8MB SPI Flash
2) Added LORA_DIO1 and LORA_DIO2 pins on 35 and 34
3) Added VEXT_CNTRL and PRG_BUTTON on 21 and 0
4) Updated the maximum program size to 3342336

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -2119,6 +2119,50 @@ heltec_wifi_lora_32.menu.UploadSpeed.512000.upload.speed=512000
 
 ##############################################################
 
+heltec_wifi_lora_32_V2.name=Heltec_WIFI_LoRa_32_V2
+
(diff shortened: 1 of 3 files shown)


me-no-dev:
There is another Heltek PR. This overlaps? 
https://github.com/espressif/arduino-esp32/pull/2392

n9wxu:
PR #2392 is good, but the partition map should be updated to take advantage of the 8MB memory.  The partition map in this PR is probably a reasonable choice as it allows OTA with two larger program segments.

me-no-dev:
why not chime in there and collaborate :) seems the requester is actually a Heltek guy, so contact could be good as well :)

PR #2409, opened 2019-01-31
Add default pin mapping for Serial1 and Serial2

xorbit:
This sets the default pin assignments for Serial1 and Serial2 correctly for the wESP32.


me-no-dev:
Merged :) please state the board name in the title next time :) helps when reading the changelog

xorbit:
Thanks!  Sorry I will keep that in mind in the future.

PR #2429, opened 2019-02-06
Added partition menu for OROCA EduBot board.

chcbaram:


me-no-dev:
do you really need that large of partitions?

chcbaram:
> do you really need that large of partitions?

Yes. it uses BLE so takes large memory and there are lots of images and audio data for lcd and audio.

me-no-dev:
how much flash do you have on-board? if more than 4, we can add custom scheme for it.

chcbaram:
> how much flash do you have on-board? if more than 4, we can add custom scheme for it.

Thank you for your kindness. It has 4MB Flash so it's possible to select the partition in the default options. If I need custom partition, I will ask . Thanks again.

Issue #1608, opened 2018-07-06
EEPROM.readString(address) reads string with junk characters at the end


avinashs24:
### Hardware:
Board:	ESP-WROOM-32
Core Installation/update date:	01/JUN/2018
IDE name:	Arduino IDE
Flash Frequency:	80Mhz
Upload Speed:	921600

### Description:
I am trying to write some string to EEPROM and retrieve it later. Whenever I try to read the string using EEPROM.readString(address), one or two junk characters like (`, ?, ") are at the end of the string. Even in the example eeprom_extra, the last two operations that involve writing and reading of strings, the output for the string operations are,

```
.
.
I love ESP32."
Thank You Espressif!`

```

### Sketch:
```cpp

/*
  ESP32 eeprom_extra example with EEPROM library

  This simple example demonstrates using other EEPROM library resources

  Created for arduino-esp32 on 25 Dec, 2017
  by Elochukwu Ifediora (fedy0)
*/

#include "EEPROM.h"

void setup() {
  // put your setup code here, to run once:
  Serial.begin(115200);
  Serial.println("\nTesting EEPROM Library\n");
  if (!EEPROM.begin(64)) {
    Serial.println("Failed to initialise EEPROM");
    Serial.println("Restarting...");
    delay(1000);
    ESP.restart();
  }

  int address = 0;                                  // Same address is used through the example

  EEPROM.writeByte(address, -128);                  // -2^7
  Serial.println(EEPROM.readByte(address));

  EEPROM.writeChar(address, 'A');                   // Same as writyByte and readByte
  Serial.println(char(EEPROM.readChar(address)));

  EEPROM.writeUChar(address, 255);                  // 2^8 - 1
  Serial.println(EEPROM.readUChar(address));

  EEPROM.writeShort(address, -32768);               // -2^15
  Serial.println(EEPROM.readShort(address));

  EEPROM.writeUShort(address, 65535);               // 2^16 - 1
  Serial.println(EEPROM.readUShort(address));

  EEPROM.writeInt(address, -2147483648);            // -2^31
  Serial.println(EEPROM.readInt(address));

  EEPROM.writeUInt(address, 4294967295);            // 2^32 - 1
  Serial.println(EEPROM.readUInt(address));

  EEPROM.writeLong(address, -2147483648);           // Same as writeInt and readInt
  Serial.println(EEPROM.readLong(address));

  EEPROM.writeULong(address, 4294967295);           // Same as writeUInt and readUInt
[...]

PR #2439, opened 2019-02-09
Fixed missing null terminator in EEPROM.readString()

### LINKED ISSUE #1252: EEPROMClass readString Bug
Hi! Is this bug?

String EEPROMClass::readString (int address)
{
  if (address < 0 || address > _size)
    return String(0);

  uint16_t len;
  for (len = 0; len <= _size; len++)
    if (_data[address + len] == 0)
      break;

  if (address + len > _size)
    return String(0);

  char value[len + 1];
  memcpy((uint8_t*) value, _data + address, len);
  value[len + 1] = 0; <- ????
  return String(value);
}

kfitzgerald:
Greetings,

This patch corrects an issue with `EEPROM.readString(address, value, maxLen)` not terminating the read string. The resulting symptoms can be garbage characters when used.

A similar issue, #1608, corrected the overload function `EEPROM.readString(address)`, but did not address this function.

Additionally, this patch should close #1252.

Thanks,
-Kevin


me-no-dev:
arghh :) please rebase :) EEPROM files have moved slightly

me-no-dev:
Thanks!

kfitzgerald:
My pleasure! Thanks @me-no-dev !

PR #2457, opened 2019-02-14
adding uartRxActive to expose the RX state machine status as a boolean value

atanisoft:
In some usages of the UART device it is necessary to check the RX status prior to TX, exposing this method will allow consumers of the UART HAL to not replicate the uart_struct_t content in order to access the status register.


cores/esp32/esp32-hal-uart.c
```diff
@@ -558,3 +558,10 @@ uartDetectBaudrate(uart_t *uart)
 
     return default_rates[i];
 }
+
+/*
+ * Returns the status of the RX state machine, if the value is non-zero the state machine is active.
+ */
+bool uartRxActive(uart_t* uart) {
+    return uart->dev->status.st_urx_out != 0;
```
me-no-dev: You do not care if there are bytes in the fifo? Does the state machine signal busy while RX is waiting for timeout since the last byte?

atanisoft: bytes in the fifo can be detected with uartAvailable, this is specifically checking if the RX state machine is actively processing something or is idle.

atanisoft: the state machine will have a value of zero if it is idle per the docs:
UART_ST_URX_OUT This register stores the value of the receiver’s finite state machine.
0: RX_IDLE;
1: RX_STRT;
2: RX_DAT0;
3: RX_DAT1;
4: RX_DAT2;
5: RX_DAT3;
6: RX_DAT4;
7: RX_DAT5;
8: RX_DAT6;
9: RX_DAT7;
10: RX_PRTY;
11: RX_STP1;
12:RX_STP2;
13: RX_DL1. (RO)

based on the above the assumption is any value other than zero means RX is active in some form.

me-no-dev: what is `RX_DL1` ?

me-no-dev: the thing is... whne you RX data on the uart, after the last stop bit, the uart goes into waiting for timeout state, during which time it expects more bytes. This is normal through a transmission... not all bytes are chained together without small pause. So that is why I am asking :) If you want to use this to switch the mode to TX for example and bus is not shown busy while waiting, you need to watch the RXTO ISR event.

atanisoft: unsure, it is from the esp32 tech ref manual. I haven't seen any details on it outside of that.

atanisoft: hmm, is there a hook point for that ISR?

atanisoft: uart->dev->int_st.rxfifo_tout ?

me-no-dev: `uart->dev->int_st.rxfifo_tout` will be active after the timeout The event can be obtained from the UART ISR :) Here is where it is cleared: https://github.com/espressif/arduino-esp32/blob/5ca809f74db196cd2562bd1375a6ded87a524cfa/cores/esp32/esp32-hal-uart.c#L85

atanisoft: maybe this instead:
```
return uart->dev->status.st_urx_out != 0 || !uart->dev->int_st.rxfifo_tout;
```
?

me-no-dev: it's 1 for a very short time and zeroed in ISR, so you might never get the above to be true. just need to test if `status.st_urx_out` is 0 or 1 while uart waits for tout

atanisoft: ok, so in that case the original check of st_urx_out != 0 would make sense. For the uart waiting on tout, that would depend on the baud rate for how long it might be. is there a good way to wait for rxfifo_tout to flip state? maybe:
```
while(!uart->dev->int_st.rxfifo_tout) {}
return uart->dev->status.st_urx_out != 0;
```

me-no-dev: `while(!uart->dev->int_st.rxfifo_tout) {}` << tis might spin forever. State can be updated in ISR :)

atanisoft: ok maybe a flag in uart_t to indicate "rx_active" that gets set inside the rxfifo_cnt loop and reset outside that loop?

atanisoft: or inside _uart_isr set a flag on uart_t to the value of uart->dev->int_st.rxfifo_tout?

me-no-dev: ughh ... needs more thinking

me-no-dev: Maybe something along those lines :)

```c
struct uart_struct_t {
    uart_dev_t * dev;
#if !CONFIG_DISABLE_HAL_LOCKS
    xSemaphoreHandle lock;
#endif
    uint8_t num;
    xQueueHandle queue;
    intr_handle_t intr_handle;
    bool rx_busy; //new property, set to false by rxfifo_tout in ISR
};

bool uartRxActive(uart_t* uart){
    if(!rxbusy && uart->dev->status.st_urx_out != 0){
        //this can still fail if it happened to ask while waiting for tout
        rxbusy = true;
    }
    return rxbusy;
}
```

atanisoft: @me-no-dev I didn't get a chance to test your update above, did you implement this or is it still TBD?

PR #2472, opened 2019-02-17
Update board file for LOPY4

cyberman54:
removed #define LORA_RST NOT_A_PIN because this runs on an error with LMIC stack (NOT_A_PIN resolves to -1 in Arduino)


me-no-dev:
I think that this should be done in LMIC stack. -1 is a valid value to be takes as "the pin is not defined"

cyberman54:
Your're right. Thus is close this.

PR #2477, opened 2019-02-19
return macaddress

a-c-sreedhar-reddy:


me-no-dev:
why? Please argument your changes :)

a-c-sreedhar-reddy:
`WiFi.macaddress() `returns NULL string when WiFi is not initialized (`WiFi.begin()`). This change returns macaddress irrespective of WiFi mode.

atanisoft:
It will return a value if you call WiFi.mode() prior to calling the Mac address function.

I agree that it isn't ideal though, perhaps in the method it can check the wifi state and if it is not on it can call your method instead?

a-c-sreedhar-reddy:
But what is the use of calling WiFi.mode() when the mac-address is same for any WiFi state.

atanisoft:
> But what is the use of calling WiFi.mode() when the mac-address is same for any WiFi state.
Agreed, it seems to be a bit of a hack to set the mode prior to calling various functions which don't change based on the mode. I'd suggest the following though:
```
    if(WiFiGenericClass::getMode() != WIFI_MODE_NULL){
        esp_wifi_get_mac(WIFI_IF_STA, mac);
    }else {
        esp_read_mac(mac, ESP_MAC_WIFI_STA);
    }
```

This way if the mode has been set it will use the first call and if it hasn't it can use the new call that you have provided. There are a few other places where mode is used and there may be a similar IDF function that will work suitably in the case where mode is null.

the only time I'd expect mode to be null is if WiFi has been turned off.

a-c-sreedhar-reddy:
I updated the changes

PR #2498, opened 2019-02-24
overload Preferences.getBytes similar to nvs so you can get size of t…

lbernstone:
overload Preferences.getBytes similar to nvs so you can get size of the array.

### DIFF
--- a/libraries/Preferences/examples/Prefs2Struct/Prefs2Struct.ino
+++ b/libraries/Preferences/examples/Prefs2Struct/Prefs2Struct.ino
@@ -1,0 +1,43 @@
+/*
+This example shows how to use Preferences (nvs) to store a
+structure.  Note that the maximum size of a putBytes is 496K
+or 97% of the nvs partition size.  nvs has signifcant overhead,
+so should not be used for data that will change often.
+*/ 
+#include <Preferences.h>
+Preferences prefs;
+
+typedef struct {
+  uint8_t hour;
+  uint8_t minute;
+  uint8_t setting1;
+  uint8_t setting2;
+} schedule_t;
+
+void setup() {
+  Serial.begin(115200);
+  prefs.begin("schedule"); // use "schedule" namespace
+  uint8_t content[] = {9, 30, 235, 255, 20, 15, 0, 1}; // two entries
+  prefs.putBytes("schedule", content, sizeof(content));
+  size_t schLen = prefs.getBytesLength("schedule");
+  char buffer[schLen]; // prepare a buffer for the data
+  prefs.getBytes("schedule", buffer, schLen);
+  if (schLen % sizeof(schedule_t)) { // simple check that data fits
+    log_e("Data is not correct size!");
+    return;
+  }
+  schedule_t *schedule = (schedule_t *) buffer; // cast the bytes into a struct ptr
+  Serial.printf("%02d:%02d %d/%d\n", 
+    schedule[1].hour, schedule[1].minute,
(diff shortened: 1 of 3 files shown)


libraries/Preferences/examples/Prefs2Struct/Prefs2Struct.ino
```diff
@@ -0,0 +1,43 @@
+/*
+This example shows how to use Preferences (nvs) to store a
+structure.  Note that the maximum size of a putBytes is 496K
```
stickbreaker: @lbernstone Are you sure the 496K limit is accurate? I thought NVS's limit was much smaller. I though all NVS data blocks were limited to fit inside one 4096 BYTE FLASH page? Base on the standard NVS configs:
```c++
// From IDF nvs_page.hpp
 const uint32_t PSB_INIT = 0x1;
 const uint32_t PSB_FULL = 0x2;
 const uint32_t PSB_FREEING = 0x4;
 const uint32_t PSB_CORRUPT = 0x8;

 const uint32_t ESB_WRITTEN = 0x1;
 const uint32_t ESB_ERASED = 0x2;

 const uint32_t SEC_SIZE = SPI_FLASH_SEC_SIZE;

 const size_t ENTRY_SIZE  = 32;
 const size_t ENTRY_COUNT = 126;
 const uint32_t INVALID_ENTRY = 0xffffffff;
    
 const size_t BLOB_MAX_SIZE = ENTRY_SIZE * (ENTRY_COUNT / 2 - 1);

 const uint8_t NS_INDEX = 0;
 const uint8_t NS_ANY = 255;
```
Using this calculation, max Blob Size is 1984 Bytes?

Chuck.

me-no-dev: pinging @lbernstone :D

lbernstone: I am going by what the docs say: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/nvs_flash.html#keys-and-values.  I didn't test, but I suppose someone (me) should.

stickbreaker:
@lbernstone Sorry, New version.  I last worked on NVS over a year ago,  Looks like they implement a spanning blob sometime within the last year.

from: NVS documentation [Entry and Entry state Bitmap](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/nvs_flash.html#entry-and-entry-state-bitmap)
>For values of primitive types (currently integers from 1 to 8 bytes long), entry holds one key-value pair. For string and blob types, entry holds part of the whole key-value pair. For strings, in case when a key-value pair spans multiple entries, all entries are stored in the same page. _**Blobs are allowed to span over multiple pages by dividing them into smaller chunks.**_ For the purpose tracking these chunks, an additional fixed length metadata entry is stored called “blob index” entry. Earlier format of blobs are still supported (can be read and modified). However, once the blobs are modified, they are stored using the new format.

I just "knew" this value was out of range,  Sorry.

Chuck.

libraries/Preferences/examples/Prefs2Struct/Prefs2Struct.ino
lbernstone: I don't have psram on my test rig here, but I can put as much into putBytes as a I can get into a char array (113792 bytes tested).  Satisfies me enough to not put the energy into writing something to try to stream more :smile:

me-no-dev:
done :) sorry for taking so long @lbernstone

PR #2506, opened 2019-02-25
Fixing temperature

sharkgua:
The temperature function was not sending the proper value. The builtin function was not working (see [espressif/esp-idf#732](https://github.com/espressif/esp-idf/issues/732)), two new files were created (components/arduino-esp32/cores/esp32/esp32-temp.c/h) that implements test_tens.c refered in the previous linked. CmakeLists was modified to take into account the change, as well as censor.c that now calls the proper function.


sharkgua:
Didn't mean to make this, my bad.

PR #2515, opened 2019-02-26
Frog Board ESP32 definition

hosek:
[FrogBoard](https://www.tindie.com/products/fred_IOT/esp32s-esp-wroom32-frogopins-development-board/)

We know its bare ESP32 module development board, but lot of user (mostly newbies) started asking how to work with it and they are little confused what board they should select in Arduino IDE, so having "own" definition would be nice especially for less experienced users.

BTW is there support to add own Example sketches to the definitions? 
Thank you in advance!

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -2849,7 +2849,7 @@ fm-devkit.menu.DebugLevel.verbose.build.code_debug=5
 
 frogboard.name=Frog Board ESP32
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
I use actually a variant of that same board :) the base ESP32 Dev Board is enough. Wrover if that chip is used.

PR #2566, opened 2019-03-10
File with more than 32,767 characters

jimibi:
When the file.bin is more than 32767 characters. 
I change "int" in "long" for contentLength.


me-no-dev:
int on esp32 is 32bit long, not 16bit like on AVR

jimibi:
Yes but there is à problem for file more 32767 bytes, perhaps it's atoi the problems. With this 2 modifications the result is OK.

me-no-dev:
Alright :) let's give it a go

PR #2594, opened 2019-03-18
Add M5Stick-C Board

sakabin:
Added support for M5Stick-C board, Change upload Speed
Project link https://docs.m5stack.com/#/en/core/m5stickc


me-no-dev:
alright... let's see with those weird baudrates :) with many IDEs and apps, such options are not available

PR #2598, opened 2019-03-20
add microseconds

Adrianotiger:
Add microseconds to the ticker library. Usefull for example for DAC/audio projects.

### DIFF
--- a/libraries/Ticker/src/Ticker.cpp
+++ b/libraries/Ticker/src/Ticker.cpp
@@ -1,8 +1,8 @@
-/* 
+/*
   Ticker.cpp - esp32 library that calls functions periodically
 
   Copyright (c) 2017 Bert Melis. All rights reserved.
-  
+
   Based on the original work of:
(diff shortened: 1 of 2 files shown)


lbernstone:
1) You can't remove code that people are already using.  You can add a new function, but must leave the old one.
2) This gets very quickly into dangerous behavior.  If you have a us-process that takes longer to execute than your cycle, it will very quickly blow out your stack, while simultaneously starving the cpu.  It will sneak up on you, since it works when you first write the code, but as you build a larger app and don't have the extra cycles to complete in a timely fashion, you will get stack failures or WDT from unpredictable places.
3) There's a hard limit on how small an interval could even theoretically be completed dependent on the tick rate.  You should probably check that the requested cycle rate is higher than that.

bertmelis:
I'd use a hardware timer (use IDF). Mond that Ticker dispatches callbrack from An auxiliary task. So it comes with the overhead of switching tasks. You might want to avoid this when on a tight schedule.

PR #2605, opened 2019-03-23
Update pins_arduino.h - Adafruit feather ESP32

Gius-8:
I had to make this change to use the Serial UART. Serial1 does not seem to work, while with this change I can use Serial2 correctly.
Do you think it is correct to modify it in this way? Thanks

### DIFF
--- a/variants/feather_esp32/pins_arduino.h
+++ b/variants/feather_esp32/pins_arduino.h
@@ -14,8 +14,8 @@
 static const uint8_t LED_BUILTIN = 13;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
 
-static const uint8_t TX = 17;
-static const uint8_t RX = 16;
+static const uint8_t TX2 = 17;
+static const uint8_t RX2 = 16;
 
 static const uint8_t SDA = 23;
 static const uint8_t SCL = 22;


me-no-dev:
don't know :) let's try to chime in @ladyada and see if she knows what would be the best option. I too wondered why RX and TX are defined to such pins. In any way, you can start any serial on any pins you like:
```Serial1.begin(115200, SERIAL_8N1, rx_pin, tx_pin);``` so predefinitions are not really needed

ladyada:
hihi - here's the schematic
image

and on the Feather we label **IO17** as Serial1 TX and **IO16** as Serial1 RX. Serial is the USB-serial converter and just for debug/upload. Serial1 is Arduino standard naming for the 'external' UART for GPS, cell, etc.
Serial2 is confusing, that's for the second UART, whereas we only have one, so it should be called **Serial1**

me-no-dev:
@ladyada so are those definitions tied anywhere in your code? maybe makes sense to label them RX1/TX1? but will that break anything?

ladyada:
i dont use those literal pin names anywhere - historically those pins have 'just worked' in the definition to set up **Serial1** - if the BSP has changed so that they need to be named differently you can change em. Only requirement is it must be **Serial1** (which is hardcoded everywehere in the universe)

me-no-dev:
Ahh now I am starting to recall what was the whole thing about. I'll have a look and make sure that Serial1 corresponds to those pins by default. Maybe even dig up the old issue for refreshment :)

me-no-dev:
@ladyada @Gius-8 according to [this line](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp#L46), names must be RX1 and TX1 for Serial1 to auto select them.
@Gius-8 will you make the appropriate change? Maybe also replace the pins for RX and TX to 3 and 1 for clarity :)

Gius-8:
Hi everyone, I had to modify otherwise Serial1 didn't work for me.

me-no-dev:
But how did Serial1 worked for you with this definition if you are changing it for Serial2?

me-no-dev:
See here: https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp#L9-L23

You are just defining the same thing, but instead you should be defining RX1 and TX1

me-no-dev:
resolved here: https://github.com/espressif/arduino-esp32/commit/d922557e0183fcf5faf0e59c3e0f6dee52d091d4

Please test :)

Issue #1433, opened 2018-05-24
HTTPClient setTimeout hasn't any effect


fabianoriccardi:
### Hardware:
Board:							 Lolin D32
Core Installation/update date:			24/May/2018
IDE name:							Arduino IDE
Flash Frequency:					40Mhz
Upload Speed:						921600

### Description:
While using the WiFiClient library, the method setTimeout() does't work. I mean, it seems always fixed to certain among (around 18seconds). The code is showed below...You can see that I call setTimeout twice to be sure to set the proper value, but still doesn't work.

The code below is reporting the library example, enriched with setTimeout() method.

### Sketch:
```cpp
void loop(){
// wait for WiFi connection
    if((wifiMulti.run() == WL_CONNECTED)) {

        HTTPClient http;
        http.setTimeout(1000);
        USE_SERIAL.print("[HTTP] begin...\n");
        // configure traged server and url
        
        http.begin("http://192.168.0.101/index.html"); //HTTP
        http.setTimeout(1000);
        USE_SERIAL.print("[HTTP] GET...\n");
        // start connection and send HTTP header
        int start=millis();
        int httpCode = http.GET();
        int end =millis();
        Serial.println(String("Effective timeout: ") + (end-start));
        // httpCode will be negative on error
        if(httpCode > 0) {
            // HTTP header has been send and Server response header has been handled
            USE_SERIAL.printf("[HTTP] GET... code: %d\n", httpCode);

            // file found at server
            if(httpCode == HTTP_CODE_OK) {
                String payload = http.getString();
                USE_SERIAL.println(payload);
            }
        } else {
            USE_SERIAL.printf("[HTTP] GET... failed, error: %s\n", http.errorToString(httpCode).c_str());
        }

        http.end();
    }

    delay(5000);
}
```

### Debug Messages:
```
[HTTP] begin...
[HTTP] GET...
Effective timeout: 18242
[HTTP] GET... failed, error: connection refused
[HTTP] begin...
[HTTP] GET...
Effective timeout: 18498
[HTTP] GET... failed, error: connection refused
```

pcbtborges:
You are using a method that I tried and did not work and it does not seem to be a timeout problem.
Try something like the following:
`//#include <WiFi.h>
#include <HTTPClient.h>

const char* ssid     = "Skynet";
const char* password = "xxx";
const char* host     = "192.168.66.104";
const char* protStr  = "http://";

int    jd    = 1234567;
char  *area  = "Area01";
[...]

PR #2606, opened 2019-03-23
Add server connect timeout to HTTPClient

boarchuz:
The existing setTimeout() function is effective only once the client is already connected to the server. Therefore, if there is a problem establishing a connection to the server in the first place (eg. it is offline), connect() will block for up to ~20 seconds regardless of the value passed to setTimeout. 
See https://github.com/espressif/arduino-esp32/issues/1433

This PR introduces a separate, self-explanatory function - setConnectTimeout(int32_t connectTimeout) - to HTTPClient, which simply takes advantage of the WiFiClient connect timeout functionality recently introduced here: https://github.com/espressif/arduino-esp32/pull/2383

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -429,6 +429,15 @@ void HTTPClient::setAuthorization(const char * auth)
 }
 
 /**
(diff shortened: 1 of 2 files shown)


me-no-dev:
nice!

tonydee2377:
Thank you!
with this instruction you have saved me.

PR #2607, opened 2019-03-24
Fix AsyncUDP receive memory leak

jweigelt:
If _handler is set, pbuf_free is not called. ~AsyncUDPPacket() calls pbuf_free once but only after calling pbuf_ref in it's constructor. The refcount never reaches zero and the memory allocated for pbuf is never released.


me-no-dev:
aahhh!!! nice find! maybe just not ref in the constructor. I think that is a better option?

me-no-dev:
Thought of it, you are correct :) this is a better way. Merging..

PR #2610, opened 2019-03-25
Added PATCH request to HTTPClient

buddemat:
In addition to the existing GET, POST, PUT,... methods, I have added a PATCH call, so that RESTful APIs with the respective functionality can be accessed more easily.


me-no-dev:
thanks! nice :)

PR #2629, opened 2019-04-01
update getLocalTime().

a-c-sreedhar-reddy:


a-c-sreedhar-reddy:
ms should not be given zero right. I think ms here is timeout milliseconds.

lbernstone:
Nope.  If it is 0, it should be able to immediately return a false.  This means time is not synched (maybe the network is not yet connected), and it can be used in synchronous processing.  I don't understand the issue you are fixing with this.

me-no-dev:
It simplifies the method and eliminates repetitive code. Penalty will really be microscopic. I will accept it :)

PR #2641, opened 2019-04-06
Hide log_v behind conditional define, reduce excessive logging

stickbreaker:
@cyberman54 from #2250 recommended reducing unnecessary logging.  There have not been any reported problems with i2c and cpu frequency changes since it was implemented in the 1.0.1 release. This pr hides these verbose log reported behind the `ENABLE_I2C_DEBUG_BUFFER` define.  They will be available if necessary, but normally will not be included in compiled code.

Chuck.


me-no-dev:
do people use verbose a lot?

PR #2623, opened 2019-03-30
Added 2 ffat partition schemes for 4MB modules + improved descriptions.

marcmerlin:
Existing format options were unclear on how much spiffs you were
getting. Also add missing 1MB App/3MB Spiffs split.
This addresses https://github.com/espressif/arduino-esp32/issues/1799

This makes it very clear what each option gives you, and also adds 2 options
to get ffat on 4MB chips..

PR #2588, opened 2019-03-15
Add support for ffat on 4MB ESP32 Dev Module and mhetesp32minikit board.

CelliesProjects:

PR #2649, opened 2019-04-10
Also show how space is allocated for the default partitions

marcmerlin:
Make all the partitions descriptions consistent, they now all show what space allocation you get.


marcmerlin:
followup to https://github.com/espressif/arduino-esp32/pull/2623 and https://github.com/espressif/arduino-esp32/pull/2588 to make it clear what you get from the 2 defaults

me-no-dev:
done :)

PR #2652, opened 2019-04-10
Properly allocate string and remove warning

skrenes:
The former way generates the following warning:
ISO C++ forbids converting a string constant to 'char*'

This change makes a character array the size of the string with null ending. It's clearer and gets rid of the warning.


me-no-dev:
Actually proper fix is to change it to `const char *` :)

skrenes:
> Actually proper fix is to change it to `const char *` :)

Agreed! A lapse in my judgment.

me-no-dev:
Thanks :)

PR #2655, opened 2019-04-11
Improve _uploadReadByte

vicatcu:
This change, without really understanding why, makes WebServer OTA less likely to fail with a broken pipe, and has been empirically demonstrated to recover from a condition that would otherwise prematurely terminate the connection during file transfer.

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -304,10 +304,27 @@ void WebServer::_uploadWriteByte(uint8_t b){
 
 int WebServer::_uploadReadByte(WiFiClient& client){
   int res = client.read();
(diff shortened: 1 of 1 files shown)


vicatcu:
I'm going to test this tomorrow by just trying to do OTA updates and see if the WebServer OTA still fails occasionally. Earlier today, though, I tried a more rushed implementation of this that certainly overcame an apparent problem, as described in the source code comments associated with this PR.

vicatcu:
if we'd rather see a timeout based limitation rather than an retry limitation I could certainly implement such a thing easily enough, but in that case what should be used for the timeout duration?

me-no-dev:
We do not need this PR anymore :P

PR #2656, opened 2019-04-11
Alternative Improve _uploadReadByte 

vicatcu:
This is a timeout based alternative to https://github.com/espressif/arduino-esp32/pull/2655. I haven't tested it yet.

### DIFF
--- a/cores/esp32/Stream.h
+++ b/cores/esp32/Stream.h
@@ -59,10 +59,7 @@ public:
 // parsing methods
 
     void setTimeout(unsigned long timeout);  // sets maximum milliseconds to wait for stream data, default is 1 second
-    unsigned long getTimeout(void)
-    {
-      return _timeout;
-    }
+    unsigned long getTimeout(void);
     bool find(const char *target);   // reads data from the stream until the target string is found
     bool find(uint8_t *target)
     {
--- a/cores/esp32/Stream.h
+++ b/cores/esp32/Stream.h
@@ -59,7 +59,10 @@ public:
 // parsing methods
 
     void setTimeout(unsigned long timeout);  // sets maximum milliseconds to wait for stream data, default is 1 second
-
+    unsigned long getTimeout(void)
+    {
+      return _timeout;
+    }
     bool find(const char *target);   // reads data from the stream until the target string is found
     bool find(uint8_t *target)
     {
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -307,7 +307,7 @@ int WebServer::_uploadReadByte(WiFiClient& client){
   if(res < 0) {
     // keep trying until you either read a valid byte or timeout
     const unsigned long startMillis = millis();
-    const long timeoutIntervalMillis = HTTP_MAX_POST_WAIT;
+    const long timeoutIntervalMillis = client.getTimeout();
     unsigned long currentMillis = startMillis;
     for(;;) {
       // loosely modeled after blinkWithoutDelay pattern
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -304,10 +304,27 @@ void WebServer::_uploadWriteByte(uint8_t b){
 
 int WebServer::_uploadReadByte(WiFiClient& client){
   int res = client.read();
-  if(res == -1){
-    while(!client.available() && client.connected())
+  const uint8_t max_retries = 3;
+  uint8_t numRetries = 0;
+  while(numRetries < max_retries && res == -1){
+    while(!client.available() && client.connected()){
       delay(2);
+    }
+
     res = client.read();
+    // NOTE: it is possible to get here and have all of the following
+    //       assertions hold true
+    //
+    //       -- client.available() > 0
+    //       -- client.connected == true
+    //       -- res == -1
+    //
+    //       a simple retry strategy overcomes this which is to say the
+    //       assertion is not permanent, but the reason that this works
+    //       is ellusive, and possibly indicative of a more subtle underlying
+    //       issue
+
+    numRetries++;
   }
   return res;
 }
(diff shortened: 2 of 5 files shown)


libraries/WebServer/src/Parsing.cpp
```diff
@@ -304,12 +304,39 @@ void WebServer::_uploadWriteByte(uint8_t b){
 
 int WebServer::_uploadReadByte(WiFiClient& client){
   int res = client.read();
-  if(res == -1){
-    while(!client.available() && client.connected())
-      delay(2);
-    res = client.read();
+  if(res < 0) {
+    // keep trying until you either read a valid byte or timeout
+    const unsigned long startMillis = millis();
```
me-no-dev: why const?

libraries/WebServer/src/Parsing.cpp
```diff
@@ -304,12 +304,39 @@ void WebServer::_uploadWriteByte(uint8_t b){
...
+      //       is ellusive, and possibly indicative of a more subtle underlying
+      //       issue
+
+      currentMillis = millis();
+      if(currentMillis - startMillis >= timeoutIntervalMillis) {
+        return -1; // exit on a timeout
+      }
+    }
   }
-  return res;
```
me-no-dev: ALWAYS return as last line ;) compilers complain.

libraries/WebServer/src/Parsing.cpp
vicatcu: > why const?

because it is only written to once in the context of this function... may as well declare it as such

libraries/WebServer/src/Parsing.cpp
vicatcu: ack, I updated this on my branch, it was actually a bug

libraries/WebServer/src/Parsing.cpp
vicatcu: only written to once in this function, so I think it's correct to declare it as such.

me-no-dev: but you set it every time you call this function....

vicatcu: removed the const keywwords

cores/esp32/Stream.h
```diff
@@ -59,7 +59,10 @@ class Stream: public Print
 // parsing methods
 
     void setTimeout(unsigned long timeout);  // sets maximum milliseconds to wait for stream data, default is 1 second
-
+    unsigned long getTimeout(void)
```
me-no-dev: not in the header please :)

libraries/WebServer/src/Parsing.cpp
```diff
@@ -304,11 +304,42 @@ void WebServer::_uploadWriteByte(uint8_t b){
...
+    unsigned long startMillis = millis();
+    long timeoutIntervalMillis = client.getTimeout();
+    unsigned long currentMillis = startMillis;
+    boolean timedOut = false;
+    for(;;) {
+      // loosely modeled after blinkWithoutDelay pattern
+      while(!timedOut && !client.available() && client.connected()){
+        delay(2);
+        currentMillis = millis();
+        timedOut = currentMillis - startMillis >= timeoutIntervalMillis;
```
me-no-dev: I would say, do not use at all `currentMillis` and just replace it with a call to millis(). The compiler will do it underneath anyway, but I think it will make the code slightly more readable

vicatcu: I'd be happy to _not_ use a local variable, but I think we have to at least cast the result to (unsigned long) for the comparison to work and still handle millis() overflow boundary conditions

cores/esp32/Stream.h
vicatcu: addressed in d5c501f7687be64e5995189eb0be44c1fb53c7d8

libraries/WebServer/src/Parsing.cpp
vicatcu: addressed in d5c501f

libraries/WebServer/src/Parsing.cpp
```diff
@@ -304,11 +304,40 @@ void WebServer::_uploadWriteByte(uint8_t b){
...
+  if(res < 0) {
+    // keep trying until you either read a valid byte or timeout
+    unsigned long startMillis = millis();
+    long timeoutIntervalMillis = client.getTimeout();
+    boolean timedOut = false;
+    for(;;) {
+      // loosely modeled after blinkWithoutDelay pattern
+      while(!timedOut && !client.available() && client.connected()){
+        delay(2);
+        timedOut = ((unsigned long) millis()) - startMillis >= timeoutIntervalMillis;
```
atanisoft: unsigned long cast is redundant since millis() returns unsigned long...

vicatcu: ack, removed cast

vicatcu:
Don't merge yet, still conducting testing.

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -289,7 +289,7 @@ int WiFiClient::getOption(int option, int *value)
     size_t size = sizeof(int);
     int res = getsockopt(fd(), IPPROTO_TCP, option, (char *)value, &size);
     if(res < 0) {
-        log_e("%d", errno);
+        log_e("getOption fail on fd %d, errno: %d", fd(), errno);
```
me-no-dev: the name of the function is not necessary. the log facility will print the file name, function name and line number.
AFAIK IDF has a facility to turn the errorno into string :) check the sdk folder. Path should make sense

vicatcu: that's cool... I was just following through on comments from @atanisoft... looks like esp_err_to_name is the function you are thinking of?

vicatcu: addressed by 3e7dd27e0e937203b963c1af3c8eba3f6b267650

vicatcu: addressed by 3e7dd27e0e937203b963c1af3c8eba3f6b267650

me-no-dev: uhmm not sure if this will work for errno though. The method you posted is for esp_err_t

atanisoft: @me-no-dev @vicatcu it won't work for errno, the proper function is strerror(errno)

vicatcu: Thanks @atanisoft, addressed by https://github.com/espressif/arduino-esp32/pull/2663

PR #2668, opened 2019-04-13
Reduced Logging level of Enter and Exit 

Bascy:
Many BLE functions have logging statements on entry and on exit, which is fine when debugging the BLE library but these statement should be on Verbose level


me-no-dev:
Hey! Nice idea! Actually while you are at it, how about you replace all calls to `ESP_LOGx(TAG, ` with `log_x(`? the ESP log is because this code used to be used outside of Arduino, but this is no longer the case. I this a batch find and replace will do :)

Bascy:
> Hey! Nice idea! Actually while you are at it, how about you replace all calls to `ESP_LOGx(TAG, ` with `log_x(`? the ESP log is because this code used to be used outside of Arduino, but this is no longer the case. I this a batch find and replace will do :)

You mean to remove the TAG parameter? I would love to do that, but I was actually hoping on someone _implementing_  this here, because it can be very useful in selective debugging messages

me-no-dev:
Arduino has it's own logging facility. Does not need a tag because it prints the file, function and line number that called the log.
```cpp
//ESP-IDF Log
ESP_LOGE(TAG, "some message");

//Arduino Log
log_e("some message");
```

me-no-dev:
Example Arduino Log call:
```cpp
log_d("Searching for camera address");
```

Result:
`[D][camera.c:980] camera_probe(): Searching for camera address`

me-no-dev:
Ahh if you talk about turning particular log on and off... that is a debatable thing for Arduino. For logs to bu turned on, they need to be also compiled. People will not like their firmwares to increase in size all of a sudden :D

libraries/Preferences/src/Preferences.cpp
```diff
@@ -291,7 +291,7 @@ int8_t Preferences::getChar(const char* key, const int8_t defaultValue){
     }
     esp_err_t err = nvs_get_i8(_handle, key, &value);
     if(err){
-        log_e("nvs_get_i8 fail: %s %s", key, nvs_error(err));
+        log_v("nvs_get_i8 fail: %s %s", key, nvs_error(err));
```
me-no-dev: why would you mute the error? it is an error, so do not mute any of them please. If you get them, then see what you are doing wrong.

Bascy: Because, if you havent saved a preference value yet, the default is assumed, that's why you provide a default. That is not a faulty situation, and not an error, a warning at most

PR #2671, opened 2019-04-14
Removed use of TAG from logging

Bascy:
Arduino implementation of LOG_x doesnt use TAG parameter, so it can be removed from LOG_x statements


me-no-dev:
oh noo :D you must have missread my comments. Arduino's log is NOT `ESP_LOGX(` it is `log_x(...`

me-no-dev:
you know what... I will just accept your previous PR and batch replace those myself :)

Bascy:
Oh sorry, my bad, will try again and await the completing of the build process this time :(

Bascy:
Hmmmm, it will not work anymore if `CONFIG_ARDUHAL_ESP_LOG` is not defined ...
Is that possible?

#if defined(ARDUINO_ARCH_ESP32) && defined(CONFIG_ARDUHAL_ESP_LOG)
#include "esp32-hal-log.h"
#else
#include "esp_log.h"
static const char* LOG_TAG = "BLEAdvertising";
#endif

me-no-dev:
you can remove that whole if/else thing and replace it with only `#include "esp32-hal-log.h"`

me-no-dev:
code was intended to detect if Arduino is installed and forward the log to it

PR #2672, opened 2019-04-14
Refactored use of LOG_X(LOG_TAG, ...) to log_x(...)

Bascy:
LOG_TAG is not used in implementation of logging


libraries/ESPmDNS/src/ESPmDNS.cpp
```diff
@@ -82,7 +82,7 @@ void MDNSResponder::setInstanceName(String name) {
 
 void MDNSResponder::enableArduino(uint16_t port, bool auth){
     mdns_txt_item_t arduTxtData[4] = {
-        {(char*)"board"         ,(char*)ARDUINO_VARIANT},
+        {(char*)"board"         ,(char*)"esp32"},
```
atanisoft: This should be reverted, the ARDUINO_VARIANT variable is injected with the board name typically. The ESPmDNS library checks explicitly for this define and if missing defines it itself as you have done here: https://github.com/espressif/arduino-esp32/blob/master/libraries/ESPmDNS/src/ESPmDNS.h#L48-L51

me-no-dev: yeah, why would you change that?

Bascy: I don't know why all these commits are part of this pull request,  these are from months ago,  and already reverted locally.
It also looks like that code isn't there anymore

robert-alfaro:
Is this refactoring really necessary? It is now broken.  If the kconfig does not select CONFIG_ARDUHAL_ESP_LOG (ESP Log forwarding to Arduino Log) then `log_X` macros are undefined because the includes are missing.

Moreover, previously, if you wanted forwarding then `ESP_LOGX` automatically turned into `log_X`, otherwise `ESP_LOGX` macros would be used if opted out -- this functionality is lost now.

me-no-dev:
this is an Arduino library and will always have CONFIG_ARDUHAL_ESP_LOG defined :) includes will be there and all. What is your use case?

PR #2678, opened 2019-04-16
Converted EEPROM library to use nvs instead of partition.  

lbernstone:
Converted EEPROM library to use nvs instead of partition.  
Removed eeprom partition from all partition table CSV files.


libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
 }
 
 bool EEPROMClass::begin(size_t size) {
-  if (size <= 0) {
-    return false;
+  if (size <= 0) return false;
```
me-no-dev: please try to not use one liners like this one :) new line and brackets are more readable and conform to IDF guidelines as well

libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
...
+      nvs_get_blob(_handle, _name, trunc_key, &size);
+      nvs_set_blob(_handle, _name, trunc_key, size);
+      nvs_commit(_handle);
+      free(trunc_key);
   }
-  //  _mypart = esp_partition_find_first(ESP_PARTITION_TYPE_DATA,ESP_PARTITION_SUBTYPE_ANY, EEPROM_FLASH_PARTITION_NAME);
-  _mypart = esp_partition_find_first(ESP_PARTITION_TYPE_DATA, ESP_PARTITION_SUBTYPE_ANY, _name);
-  if (_mypart == NULL) {
-    return false;
+  if (size > key_size) { // expand or new
```
me-no-dev: `else if(...` here makes sense?

libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
...
+      size_t expand_size = size - key_size;
+      uint8_t* expand_key = (uint8_t*) malloc(expand_size);
```
me-no-dev: no check for NULL?

libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
...
+  res = nvs_get_blob(_handle, _name, NULL, &key_size);
+  if(res != ESP_OK && res != ESP_ERR_NVS_NOT_FOUND) {
+      log_e("Unable to read NVS key: %d", res);
+      return false;
   }
-  if (size > SPI_FLASH_SEC_SIZE) {
-    size = SPI_FLASH_SEC_SIZE;
+  if (size < key_size) { // truncate
+      log_w("truncating EEPROM from %d to %d", key_size, size);
+      uint8_t* trunc_key = (uint8_t*) malloc(size);
```
me-no-dev: no check for NULL?

libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
...
+      size_t expand_size = size - key_size;
+      uint8_t* expand_key = (uint8_t*) malloc(expand_size);
+      if(nvs_set_blob(_handle, "expand", expand_key, expand_size)) {
+        log_e("Not enough space to expand EEPROM from %d to %d", key_size, size);
+        free(expand_key);
+        return false;
+      }
+      free(expand_key);
+      nvs_erase_key(_handle, "expand");
+      uint8_t* hold_data = (uint8_t*) malloc(size);
```
me-no-dev: also missing check for NULL

libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -66,31 +63,59 @@ EEPROMClass::~EEPROMClass() {
...
+      memset(hold_data, 0, size);
+      if(key_size) {
+        log_i("Expanding EEPROM from %d to %d", key_size, size);
+        nvs_get_blob(_handle, _name, hold_data, &key_size);
+        nvs_erase_key(_handle, _name);
+      }
+      nvs_commit(_handle);
+      nvs_set_blob(_handle, _name, hold_data, size);
+      free(hold_data);
+      nvs_commit(_handle);
```
me-no-dev: this whole block of code needs some comments to explain exactly what is going on :) I got lost with all those reads, writes and erases. What data is where... you need 2 blobs to hold the default 4K EEPROM, but reading through your code I got lost.

me-no-dev:
left a few notes :)

libraries/EEPROM/src/EEPROM.cpp
lbernstone: There is no way to expand a key, so in order to do so, it tests for free space, then holds the data while it deletes the key.  If it is a new key, it just fills with zeros.  I think changing the variable name to key_data will make it clearer.

libraries/EEPROM/src/EEPROM.h
```diff
@@ -117,11 +105,10 @@ class EEPROMClass {
     template <class T> T writeAll (int address, const T &);
 
   protected:
-    uint32_t _sector;
+    nvs_handle _handle;
```
me-no-dev: could you maybe forward declare the `nvs_handle` so it is not required to include nvs in the header?

libraries/EEPROM/examples/eeprom_extra/eeprom_extra.ino
```diff
@@ -20,68 +20,117 @@ void setup() {
...
+  address += 21;
+
+  // See also the general purpose writeBytes() and readBytes() for BLOB in EEPROM library
+  EEPROM.commit();
+  address = 0;
+  
+  Serial.println(EEPROM.readByte(address));
+  address += sizeof(byte);
+
+  Serial.println(char(EEPROM.readChar(address)));
```
stickbreaker: Does this actually compile?  Shouldn't it be `printlin((char)EEPROM.readChar(address));`

libraries/EEPROM/examples/eeprom_extra/eeprom_extra.ino
```diff
@@ -20,68 +20,117 @@ void setup() {
...
+  address += sizeof(double);
 
-  EEPROM.writeBool(address, true);
   Serial.println(EEPROM.readBool(address));
+  address += sizeof(bool);
 
-  String sentence = "I love ESP32.";
-  EEPROM.writeString(address, sentence);
   Serial.println(EEPROM.readString(address));
+  address += sentence.length();
```
stickbreaker: Length +1?

libraries/EEPROM/examples/eeprom_extra/eeprom_extra.ino
me-no-dev: travis would have complained :) not sure how Make will like it though

pgrisu:
This is a good feature that save a bit of flash space but it breaks all deployed systems where factory settings are stored into EEPROM flash partition.
Already deployed devices based on previous arduino-esp32 will lose all EEPROM factory/config settings if they move to latest release by FOTA
@me-no-dev Please remember to mark this change as 'breaking changes' in next release notes

me-no-dev:
Will do! @lbernstone any ideas on how that can be detected and moved to NVS?

fedy0:
@me-no-dev
The new move of the EEPROM by this pull request from flexible flash data partition to NVS, in my opinion, is **superfluous**

My reasons:
[1] Already Existing Interface: The [Preference library](https://github.com/espressif/arduino-esp32/tree/master/libraries/Preferences) already provides **a complete and an excellent friendly** access to ESP32's Non-Volatile Storage. As a novice in software, I've always been warned not to repeat myself
[2] Not Backward Compatible: The pull request **breaks changes** not only in deployed systems (which is _fatal_) as mentioned by @pgrisu but also in the classical EEPROM examples.
[3] NVS Usage: Non-Volatile Storage (NVS)  was **primarily** [designed to store key-value pairs]( https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/nvs_flash.html) in flash which the preference library mastered.
[4] Flexibility: The [partition table](https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/partition-tables.html) already encouraged user-defined partitions (with data subtypes starting @0x99) which EEPROMClass can enjoy that this pull request deprecated.
[5] Same storage: All **data** partitions are on one flash memory with **predefined subtypes** (as used in the partition table)  acting as constraints to their method of usage. If an NVS partition is truncated (for example, when the partition table layout is changed), its contents should be erased which is same for the previous EEPROM implementation. Why move the EEPROM?

arduino-esp32 is too matured to start having _breaking changes_ caused by memory partitioning.
Thanks once again for your great contribution in the Preference library @me-no-dev

tomtobback:
Hi, in addition to breaking EEPROM it also breaks any files stored earlier in SPIFFS; the SPIFFS offset moved from 0x291000 to 0x290000. On on my system, any SPIFFS files saved with core 1.0.2 cannot be read by core 1.0.3. I can only read SPIFFS with 1.0.3 when i edit my partitions/default.csv back to the 1.0.2 values (line 6: spiffs). Those 1000 bytes gained are not worth breaking backward compatibility in my humble opinion. I was surprised not to find any documentation on this because i assume lots of people have/will run into this problem.

PR #2681, opened 2019-04-18
Add TTGO T-Watch board definition

lewisxhe:
Add the upcoming TTGO T-Watch board definition


boards.txt
```diff
@@ -3522,4 +3522,72 @@ sparkfun_lora_gateway_1-channel.menu.UploadSpeed.512000.upload.speed=512000
...
+ttgo-t-watch.upload.maximum_size=6553600
+ttgo-t-watch.upload.maximum_data_size=4521984
+ttgo-t-watch.upload.wait_for_upload_port=true
+
+ttgo-t-watch.serial.disableDTR=true
+ttgo-t-watch.serial.disableRTS=true
+
+ttgo-t-watch.build.mcu=esp32
+ttgo-t-watch.build.core=esp32
+ttgo-t-watch.build.variant=t-watch
```
me-no-dev: do not use `-` in the names of the variant and board. those things are later turned into defines and can cause issues.

variants/t-watch/pins_arduino.h
```diff
@@ -0,0 +1,39 @@
...
+
+#define RTC_INT             37
+#define APX20X_INT          35
+#define BMA42X_INT1         39
+#define BMA42X_INT2         4
+
+static const uint8_t KEY_BUILTIN = 36;
+
+static const uint8_t TX = 33;
+static const uint8_t RX = 34;
```
me-no-dev: which serial is this supposed to be?

variants/t-watch/pins_arduino.h
```diff
@@ -0,0 +1,39 @@
...
+
+static const uint8_t SDA = 21;
+static const uint8_t SCL = 22;
+
+static const uint8_t SS    = 13;
+static const uint8_t MOSI  = 15;
+static const uint8_t MISO  = 2;
```
me-no-dev: IO2 is a bootstrap pin. just a warning: if slave pulls it up, the board might not boot :)

variants/t-watch/pins_arduino.h
lewisxhe: Serial1 Already assigned to GPS LORA

variants/t-watch/pins_arduino.h
lewisxhe: SPI has been configured as an SD card slot and must be removed when downloading

variants/twatch/pins_arduino.h
```diff
@@ -0,0 +1,41 @@
...
+// Interrupt IO port
+#define RTC_INT             37
+#define APX20X_INT          35
+#define BMA42X_INT1         39
+#define BMA42X_INT2         4
+
+static const uint8_t KEY_BUILTIN = 36;
+//Serial1 Already assigned to GPS LORA
+static const uint8_t TX = 33;
+static const uint8_t RX = 34;
```
me-no-dev: if this is Serial1, I will suggest that you add:
```cpp
#define TX1 TX
#define RX1 RX
```

So when someone calls `Serial1.begin(baud)` it will run with those pins

lewisxhe: Ok, thank you for your suggestion, it has been modified.

variants/twatch/pins_arduino.h
```diff
@@ -24,8 +24,8 @@
 
 static const uint8_t KEY_BUILTIN = 36;
 //Serial1 Already assigned to GPS LORA
-static const uint8_t TX = 33;
-static const uint8_t RX = 34;
+static const uint8_t TX1 = 33;
+static const uint8_t RX1 = 34;
```
me-no-dev: not rename :) use `#define` as in my example. Reason is that the defines will overwrite the default pins, but a declaration will cause issues:  https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp#L6-L23

lewisxhe: It has been modified. Thanks

PR #2682, opened 2019-04-18
add lolin32 CPU freq choice

jandegr:
With a little bit of copy, paste and rename I added cpu frequency to the arduino IDE menu.
Tested with 240, 160 and 80 MHz.

I am not sure if the plan is to add this to all of the board definitions, so feel free to reject or close this PR.


me-no-dev:
I see no reason to reject this :) give it a world!

PR #2690, opened 2019-04-21
Defined BLE setPrivateAddress()

WhyKickAmooCow:
Implemented the function used to change the advertised BLE address.

As per the specification, there are multiple address types that can be user set (Random, Static, etc.). Currently this library only uses the address issued to the device during manufacture. This pull request adds the functionality of user changeable BLE addresses.

There is currently no implementation of this function that I am aware of, so use in projects should be minimal, if at all. Changing the functions signature should not break any significant number of projects.

### DIFF
--- a/libraries/BLE/src/BLEAdvertising.cpp
+++ b/libraries/BLE/src/BLEAdvertising.cpp
@@ -254,6 +254,27 @@ void BLEAdvertising::stop() {
 } // stop
 
 /**
+ * @brief Set BLE address.
+ * @param [in] Bluetooth address.
+ * @param [in] Bluetooth address type.
+ * Set BLE address.
+ */
+
+void BLEAdvertising::setDeviceAddress(esp_bd_addr_t addr, esp_ble_addr_type_t type)
+{
+	log_v(">> setPrivateAddress")
+
+	m_advParams.own_addr_type = type;
+	esp_err_t errRc = esp_ble_gap_set_rand_addr((uint8_t*)addr);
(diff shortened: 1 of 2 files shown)


me-no-dev:
could you please explain to me and to everyone in this PR's description exactly what this function does and what your change improves. Also, I am not sure that changing the original function's signature is a good thing. Maybe a new function should be added for this? Code that uses it will now break.

me-no-dev:
Any comments on this @WhyKickAmooCow

WhyKickAmooCow:
Given the function's content, maybe you are right in saying that a different function may be better. I would suggest something like "setRandomAddress()".

me-no-dev:
How about `setDeviceAddress`? Since the address will be manually set.

WhyKickAmooCow:
Of course I miss the obvious choice. This will be probably the best option as address types other than random may be set, and that it makes is clear that it is setting the device address. 

As according the BLE spec, as far as I am aware even if you manually set the address, it is still of the random type. So random address would still be appropriate for the most part, but `setDeviceAddress` would still be better.

me-no-dev:
I have the feeling that the meaning of "random" is rather different. More like the different types of MAC addresses for networking. They also have a few different classes. So the address is not really random every time, but is of type "random" (which means something else). Do I get this correctly?

chegewara:
In ble specs there is 4 types of address:
- 1 public, which is burned in esp32 efuse,
- 3 random, esp-idf is using only 1 random type as far as i know (public random).

PR #2694, opened 2019-04-21
Update ALKS board

CornyjK:
Thank you for merging.

https://github.com/RoboticsBrno/ArduinoLearningKitStarter-library/issues/22


me-no-dev:
I see you added the PSRAM option. Do you have PSRAM on-board?

CornyjK:
Mostly not, but there is the possibility (this means board has the ability to have PSRAM, but we do not commonly use it, we might in future). So I would like to have the option.

me-no-dev:
cool :) merged

PR #2701, opened 2019-04-23
Portability from ESP8266, virtual Stream member functions.

dok-net:
Needed for https://github.com/plerup/espsoftwareserial


me-no-dev:
what is the point of using such software serial on esp32? this is a bitbang lib that will have many issues in FreeRTOS and on top of that ESP32 has 3 full UART ports that allow you to assign to any pin.

dok-net:
Don't kill the messenger :-)
1st - leaving aside the use in EspSoftwareSerial, do you agree that the PR is generally useful?
2nd - to answer your question, could you please consider the use case described in https://github.com/plerup/espsoftwareserial/issues/93 ? Personally, no, I haven't checked the capabilities of the HW UARTS in the ESP32 to cover the requirements of the poster.

me-no-dev:
We are a gun-free community :D 
1. Sure that can be agreed!
2. Weird requirement, though recently someone asked me for the same exact thing, but I can not recall the use case.

Is this tested with ESP32? If so, what baudrates and scenarios? << I guess this is more for @plerup to answer :) I am just worried that this would bring more issue reports and will not push people to actually use the hardware serials available on ESP32

dok-net:
1 - thank you, please go ahead :-)
2 - I've done some research, asking myself the same question as you did. Vending machines appear to have a bus protocol using a ninth bit. Whatever.
I am co-admin for https://github.com/plerup/espsoftwareserial :-)
Just a few minutes ago, I added hardware loopback to the loopback.ino example, bitrates and scenarios have been tested and are working just a well as on the ESP8266 for weeks now.
I agree with your concern about people missing the extra HW UARTs in the ESP32, but that point is moot, given that the EspSoftwareSerial is ported and available through the Arduino library manager right now - this PR just concers latest improvements and the ability to go forward without additional porting trouble for me.
FYI, I also have PR pending for backporting the assignInterruptArg() function to ESP8266, so the synergy goes both ways.
https://github.com/espressif/arduino-esp32/pull/2701 and https://github.com/esp8266/Arduino/pull/6003 are currently blocking me from releasing the latest EspSoftwareSerial bits to https://github.com/plerup/espsoftwareserial

me-no-dev:
The current maintainers of ESP8266 Arduino might not be aware of the reason for your PR. I will advise you (as true here) that you approach PR description like you are explaining to someone totally new. This will help on getting reviews, less follow-up questions and a nice note that will accompany the merge.

PR #2706, opened 2019-04-24
Ensure that _size is properly set in begin

lbernstone:


libraries/EEPROM/src/EEPROM.cpp
```diff
@@ -131,6 +131,7 @@ bool EEPROMClass::begin(size_t size) {
   }
 
   _data = new uint8_t[size];
```
stickbreaker: Where is the NULL check? `new` does not always succeed.

Chuck.

me-no-dev: @lbernstone does not like NULL checks :D

lbernstone: Hey, I didn't even write that bit.

projectgus: > new does not always succeed.

`new` doesn't always succeed, but it doesn't return `NULL` unless it's marked with `std::nothrow`.  It throws an exception on failure.

(Arduino is configured with C++ exception handling enabled, but even if exceptions are disabled the firmware will `abort()` if `new` fails without being marked `std::nothrow`)

[gcc will even optimize out a null check on a `new` result if one is added in the code](https://gcc.godbolt.org/z/KLQTGe), as it knows the result can't be NULL. (That's ARM, but Xtensa output will be the same for -Os or higher.)

stickbreaker: @projectgus So, it is a stackdump.  Why is `begin()` a boolean function if it can never return a fail? If  I used this function, I would assume because of the declaration I am responsible for error checking.

projectgus: > Why is begin() a boolean function if it can never return a fail?

I don't know? If I had to guess, I would say `EEPROM:begin()` likely returns bool because of historical Arduino API reasons.

I'm not saying you shouldn't check for a failed allocation. I'm saying that a null check on the result of the `new` operator isn't actually checking anything. There are a number of ways a working out of memory check could be added, the comment hints at some of them.

projectgus: (If it was up to me I'd suggest either adding `std::nothrow` or changing this to malloc().)

me-no-dev: let's go with malloc since that is what is used throughout :)

PR #2710, opened 2019-04-25
Add board "WEMOS D1 MINI ESP32".

dok-net:
The "WEMOS D1 MINI ESP32" category of boards provide pin-out compatibility to the family of shields for the popular WEMOS/LOLIN D1 MINI series of ESP8266 boards. In order to make recompiling sketches as easy as possible, this board description includes definitions for the pins of the D1 MINI.
Among the other names for the same board layout are mini32, HW-665, WEMOS TTGO MINI32.

### DIFF
--- a/variants/d1_mini32/pins_arduino.h
+++ b/variants/d1_mini32/pins_arduino.h
@@ -20,8 +20,8 @@ static const uint8_t D5   = 18;
 static const uint8_t D6   = 19;
 static const uint8_t D7   = 23;
 static const uint8_t D8   = 5;
-static const uint8_t RXD0 = 3;
-static const uint8_t TXD0 = 1;
+static const uint8_t RXD  = 3;
+static const uint8_t TXD  = 1;
 
 #define PIN_SPI_SS   SS   // backward compatibility
 #define PIN_SPI_MOSI MOSI // backward compatibility
--- a/variants/d1_mini32/pins_arduino.h
+++ b/variants/d1_mini32/pins_arduino.h
@@ -20,8 +20,8 @@ static const uint8_t D5   = 18;
 static const uint8_t D6   = 19;
 static const uint8_t D7   = 23;
 static const uint8_t D8   = 5;
-static const uint8_t RXD0 = 9;
-static const uint8_t TXD0 = 10;
+static const uint8_t RXD0 = 3;
+static const uint8_t TXD0 = 1;
 
 #define PIN_SPI_SS   SS   // backward compatibility
 #define PIN_SPI_MOSI MOSI // backward compatibility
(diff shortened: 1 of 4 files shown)


variants/mini32/pins_arduino.h
```diff
@@ -0,0 +1,33 @@
...
+static const uint8_t D1   = 22;
+static const uint8_t D2   = 21;
+static const uint8_t D3   = 17;
+static const uint8_t D4   = 16;
+static const uint8_t D5   = 18;
+static const uint8_t D6   = 19;
+static const uint8_t D7   = 23;
+static const uint8_t D8   = 5;
+static const uint8_t RXD0 = 9;
+static const uint8_t TXD0 = 10;
```
me-no-dev: which serial is this?

variants/d1_mini32/pins_arduino.h
```diff
@@ -0,0 +1,33 @@
...
+static const uint8_t D1   = 22;
+static const uint8_t D2   = 21;
+static const uint8_t D3   = 17;
+static const uint8_t D4   = 16;
+static const uint8_t D5   = 18;
+static const uint8_t D6   = 19;
+static const uint8_t D7   = 23;
+static const uint8_t D8   = 5;
+static const uint8_t RXD  = 3;
+static const uint8_t TXD  = 1;
```
dok-net: I don't remember where I found pin numbers 9 and 10 for RXD0/TXD0 - RX and TX are in d32_core.h already. I am unsuccessful in redirecting Serial to vacate 3 and 1, in order to use EspSoftwareSerial for the USB logging port, and have all three HW UARTs freely available. Is this not supported, to redirect the pin assignment of Serial?

me-no-dev: if you make them to be defines, they will overwrite the default pins for the UARTs:
```cpp
//this will make Serial1.begin(baud) to run on pins 21 and 22
//add RX2/TX2 to overwrite Serial2
#define RX1 21
#define RX2 22
```

dok-net: Yes, thank you for explaining, I understand that from reading the code in HardwareSerial.(h|cpp) and esp32-hal-uart.(h|cpp).
But: the code suggests that Serial (0) can be redirected from 3 (RX) and 1 (TX), for the love of it, I can't that to work. Is the UART0 somehow hardwired to these ports and cannot be redirected, even if the code tries to, and fails silently?
As far as the PR is concerned, I've completed it, would you merge it now as it is?

me-no-dev: if you want to run Serial on other pins, you really need to specify them in Serial.begin
RX and TX are hardcoded to 1 and 3 otherwise as that is the real Serial port (like AVR arduinos and such)

PR #2715, opened 2019-04-26
Copy ESP8266 String w/SSO to ESP32 repo

earlephilhower:
I redid the ESP8266 WString library to enable small string optimization
(SSO) a while back, and think it would be helpful even on the ESP32 with
its higher memory complement.

SSO avoids lots of tiny mallocs() on the heap which cause fragmentation
by using the memory in the class object itself to store the actual
string and only mallocing() for buffers that are larger than what can
fit in thie class object.  Modern C++ std::string implementations have
this optimization as well, but since we're using Arduino strings we had
to roll our own.

### DIFF
--- a/cores/esp32/WString.h
+++ b/cores/esp32/WString.h
@@ -35,295 +35,292 @@ class StringSumHelper;
 // an abstract class used as a means to proide a unique pointer type
 // but really has no body
 class __FlashStringHelper;
-#define F(string_literal) (reinterpret_cast<const __FlashStringHelper *>(PSTR(string_literal)))
+#define FPSTR(pstr_pointer) (reinterpret_cast<const __FlashStringHelper *>(pstr_pointer))
+#define F(string_literal) (FPSTR(PSTR(string_literal)))
 
 // The string class
-class String
-{
-    // use a function pointer to allow for "if (s)" without the
-    // complications of an operator bool(). for more information, see:
-    // http://www.artima.com/cppsource/safebool.html
-    typedef void (String::*StringIfHelperType)() const;
-    void StringIfHelper() const
-    {
-    }
-
-public:
-    // constructors
-    // creates a copy of the initial value.
-    // if the initial value is null or invalid, or if memory allocation
-    // fails, the string will be marked as invalid (i.e. "if (s)" will
-    // be false).
-    String(const char *cstr = "");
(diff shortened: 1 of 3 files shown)


me-no-dev:
Welcome to the ESP32 scene @earlephilhower :) Thanks for the update!

cores/esp32/WString.h
```diff
@@ -35,295 +35,292 @@ class StringSumHelper;
 // an abstract class used as a means to proide a unique pointer type
 // but really has no body
 class __FlashStringHelper;
-#define F(string_literal) (reinterpret_cast<const __FlashStringHelper *>(PSTR(string_literal)))
+#define FPSTR(pstr_pointer) (reinterpret_cast<const __FlashStringHelper *>(pstr_pointer))
```
dok-net: WString.h: 38:0: warning: "FPSTR" redefined
   #define FPSTR(pstr_pointer) (reinterpret_cast<const __FlashStringHelper *>(pstr_pointer))
 
WString.h:29: In file included from
Arduino.h:152: from
[…]
pgmspace.h:35: note  this is the location of the previous definition
   #define FPSTR(p)      ((const char *)(p))

earlephilhower: Looks like I didn't catch that difference that needed patching, sorry.  Weird how it got through CI, though.

Did you want to change the `#define`s back in your own PR, @dok-net, or would you like me to do it?

earlephilhower: PR #2726 undoes the redefinition, returning it to the ESP32 native definition.  Since you have a failing case, can you check that it looks good for you?

dok-net: No more warnings with PR #2726 in place, thank you.

me-no-dev: merged :)

bxparks:
FYI: #2961 re-implements this fix by keeping the definition from ESP8266, but removing the duplicate (but incorrect) definition of `FPSTR()` in ESP32's pgmspace.h. Therefore, the FPSTR() macro is now identical between ESP8266 and ESP32, and future merges from ESP8266 should not cause trouble.

What is different now is the PSTR() macro, defined in <pgmspace.h>. This macro is a no-op in ESP32, but uses PROGMEM in ESP8266.

PR #2728, opened 2019-04-30
Fix semaphores in IDF & std::string assert

h2zero:
Fixes the problem of giving a mutex from a callback with the latest IDF. Also addresses an occasional assert that happens when the btc_task callback gives the semaphore and causes an assert due to both cores potentially writing m_owner concurrently.

### DIFF
--- a/libraries/BLE/src/FreeRTOS.cpp
+++ b/libraries/BLE/src/FreeRTOS.cpp
@@ -61,14 +61,14 @@ uint32_t FreeRTOS::getTimeSinceStart() {
  */
 uint32_t FreeRTOS::Semaphore::wait(std::string owner) {
 	log_v(">> wait: Semaphore waiting: %s for %s", toString().c_str(), owner.c_str());
-
-	m_owner = owner;
 	
 	if (m_usePthreads) {
 		pthread_mutex_lock(&m_pthread_mutex);
 	} else {
 		xSemaphoreTake(m_semaphore, portMAX_DELAY);
 	}
+	
+	m_owner = owner;
 
 	if (m_usePthreads) {
 		pthread_mutex_unlock(&m_pthread_mutex);
@@ -77,6 +77,7 @@ uint32_t FreeRTOS::Semaphore::wait(std::string owner) {
 	}
 
 	log_v("<< wait: Semaphore released: %s", toString().c_str());
+	m_owner = std::string("<N/A>");
 	return m_value;
 } // wait
 
--- a/libraries/BLE/src/FreeRTOS.cpp
+++ b/libraries/BLE/src/FreeRTOS.cpp
@@ -62,13 +62,13 @@ uint32_t FreeRTOS::getTimeSinceStart() {
 uint32_t FreeRTOS::Semaphore::wait(std::string owner) {
 	log_v(">> wait: Semaphore waiting: %s for %s", toString().c_str(), owner.c_str());
 
+	m_owner = owner;
+	
 	if (m_usePthreads) {
 		pthread_mutex_lock(&m_pthread_mutex);
 	} else {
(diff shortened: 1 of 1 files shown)


libraries/BLE/src/FreeRTOS.cpp
```diff
@@ -62,22 +62,21 @@ uint32_t FreeRTOS::getTimeSinceStart() {
 uint32_t FreeRTOS::Semaphore::wait(std::string owner) {
 	log_v(">> wait: Semaphore waiting: %s for %s", toString().c_str(), owner.c_str());
 
+	m_owner = owner;
```
projectgus: Setting the owner before we take the semaphore seems like maybe it can lead to behaviour. Think about this sequence of events:

* Task A calls `s->wait("task_a")`. m_owner is set to "task_a", and then waiting for the semaphore succeeds because noone was holding it.
* Task B calls `s->wait("task_b")`. m_owner is set to "task_b" but then Task B is blocked in xSemaphoreWait() because Task A is still holding the semaphore. 

Now we have a situation where Task A is holding the semaphore, but the owner field indicates Task B is holding it.

h2zero: Good point, hadn't really thought about it as the way the BLE library currently works (on the client side anyway) there is multiple semaphores used and only one task actually takes each of them. Also, as this was a mutex before, the semaphore is usually taken already before we call the wait function to block the task and wait for the callback to give the semaphore back, at which time it sets m_owner to "<N/A>" anyway. 

The reason for this change though is I've encountered many instances of on assert error when giving the semaphore from the callback as both cores sometimes try to write to m_owner at the same time, one in the give() on core 0 and this one after we've unblocked in wait() on core 1.

Anyway, maybe a better solution would be to pin the callback task to core 1, or maybe just remove the setting of m_owner in wait() all together? I'm open to suggestions :).

me-no-dev: What is the point of this `m_owner` anyway? Logging?

h2zero: Yes as far as I can tell it’s only used for logging, but I suppose  someone could call the toString() method on the semaphore to check ownership. 

I’m tempted just to remove it entirely for my use, but others might find it useful.

projectgus: @h2zero

Right, sorry I should have read the full code - I missed that wait() isn't "take", it's "taken then give". Not quite what I expected...

> The reason for this change though is I've encountered many instances of on assert error when giving the semaphore from the callback as both cores sometimes try to write to m_owner at the same time, one in the give() on core 0 and this one after we've unblocked in wait() on core 1.

If this is the problem then I think the best solution is to change all the m_owner assignments so  they only happen when the assignee holding the semaphore, so the semaphore protects against any race condition.  I think this means:

* In `::give()`, move setting of `m_owner` up to before the semaphore is given so it's set while the caller still holds the semaphore.
* In `::wait()`, don't set m_owner at all. As soon as the caller in wat() gets the semaphore it gives it back, and giving a semaphore never blocks, so if you did set m_owner then it gets immediately un-set afterwards which seems like a waste of CPU cycles.

h2zero: Yeah that’s along the line of my thinking as well. Sorry I changed the PR yesterday assuming you wanted me to revert the m_owner move and I didn’t see your reply here first. I’ll make the changes as you’ve mentioned and fix the PR again lol.

h2zero: @projectgus 

Made the changes, tested with my test code that would cause the assert within seconds and has been running over an hour now. Logging works fine as well.

projectgus: Awesome! :)

PR #2734, opened 2019-05-02
FunctionalInterrupt/ScheduledFunctions should be in a library, not in esp32-hal-gpio

dok-net:
After mulling over the situation with the pending PR https://github.com/esp8266/Arduino/pull/6003, I have rewritten the support for std::functional and scheduled interrupts to be less intrusive.
First, the support is revoked. Second, the refactored code is brought back as an example in libraries/ESP32/examples/GPIO/FunctionalInterrupt.
I have unified the code with the ESP8266 version: https://github.com/esp8266/Arduino/pull/6038


dok-net:
I don't know if this redundant, but if integrating the Schedule.h functionality that brings loop-synchronized scheduling, is desired, here's what it takes:

- copy Schedule.(h|cpp) from libraries/ESP32/examples/GPIO/FunctionalInterrupt to cores/esp32.
- add cores/esp32/Schedule.cpp to CMakeLists.txt right before cores/esp32/Stream.cpp.
- add #include "Schedule.h" as last include in Arduino.h.
- add run_scheduled_functions(); right under loop(); in void loopTask(void *pvParameters) which is in cores/esp32/main.cpp.

me-no-dev:
here are a few thoughts :) 
- Do not rename header filenames.
- Do not include ESP8266 Arduino code in examples.
- ESP32 Arduino runs on top of FreeRTOS :) we have threads and other tools to create scheduled commands
- ESP32 Arduino doesn't necessarily have to have 1:1 what ESP8266 Arduino have. Chips are not the same, runtime is not the same... many differences.
- Such PR Should be split in two, where we can in one discuss that Schedule library and another to discuss Functional Interrupt

dok-net:
@me-no-dev :
 
> Do not rename header filenames.
Checked - though they were deleted and resurrected as pure example code, not renamed ;-)

> Do not include ESP8266 Arduino code in examples.
Functional* and Schedule* were previously included from ESP8266. I am working with the WEMOS D1 mini like boards, and I think Arduino is all about portability, please don't restrict this.

> ESP32 Arduino runs on top of FreeRTOS :) we have threads and other tools to create scheduled commands
Again, Functional* and Schedule* did in fact get accepted earlier from a different source, I am removing them due to problems :-)

> ESP32 Arduino doesn't necessarily have to have 1:1 what ESP8266 Arduino have. Chips are not the same, runtime is not the same... many differences.
Naturally, but helping migrate code is not an evil thing, anyway.

> Such PR Should be split in two, where we can in one discuss that Schedule library and another to discuss Functional Interrupt
(Re)moving Schedule is a just a necessary side effect of (re)moving the buggy FunctionalInterrupts.

OK - I hope you can review this PR for it's own merits. The core issue I am addressing is that non-IRAM code will be surreptitiously called from ISRs if FunctionalInterrupt.h gets used.

PR #2737, opened 2019-05-03
StreamString SSO bugfix

mongozmaki:
Bugfix for StreamString/String. Same as https://github.com/esp8266/Arduino/pull/6035

_Note: I did not figure out where the tests are for ESP32, so its just the fix. For the test code see https://github.com/esp8266/Arduino/pull/6041_


me-no-dev:
duplicate :)

Issue #2128, opened 2018-11-30
Arduino.h: error on abs macro definition


KarateBrot:
## Hardware:
Board:							NodeMCU 32-S
Core Installation/update date:			v1.0.0
IDE name:							PlatformIO
Computer OS:                                          Windows 8.1

### Description:
It seems like I have got a similar issue to issue #1734. Somehow the macro definition of `abs` inside `Arduino.h` collides with the `std::abs` function inside the `cmath` library.

### Debug Messages:
```
In file included from src\PowerSupply.h:22:0,
from src\PowerSupply.cpp:11:
src\PowerSupply.cpp: In member function 'void Heater::calibrate()':
C:\Users\Jan Post\.platformio\packages\framework-arduinoespressif32\cores\esp32/Arduino.h:76:16: error: expected unqualified-id before '(' token
#define abs(x) ((x)>0?(x):-(x))
^
src\PowerSupply.cpp:190:15: note: in expansion of macro 'abs'
while (std::abs(current - 10.0) > 1.0 && std::abs(currentLast - 10.0) > 1.0) {
^
C:\Users\Jan Post\.platformio\packages\framework-arduinoespressif32\cores\esp32/Arduino.h:76:16: error: expected unqualified-id before '(' token
#define abs(x) ((x)>0?(x):-(x))
^
src\PowerSupply.cpp:190:49: note: in expansion of macro 'abs'
while (std::abs(current - 10.0) > 1.0 && std::abs(currentLast - 10.0) > 1.0) {
^
*** [.pioenvs\nodemcu-32s\src\PowerSupply.cpp.o] Error 1
```

PR #1783, opened 2018-08-22
cores: replace max and min macros with imports from std

yoursunny:
fixes #1734

PR #2738, opened 2019-05-03
Use std::abs for a float-compatible abs() function

projectgus:
* Other Arduino cores uses a macro to redefine libc abs() to take any  type, meaning `abs(-3.3) == 3.3` not the normal libc result of 3.

* 1e4bf14a3 (#1783) replaced similar `min`, `max` macros with c++ stdlib. However this change includes `<algorithm>` after the line which defines the abs() macro. `<algorithm>` includes `<cstdlib>` which undefines `abs()` and re-defines it.

* As a result, current esp32 Arduino core has a bug compared to other Arduino cores: `abs()` is now the plain libc version which only takes integers, so `abs(-3.3) == 3`. As reported here: https://github.com/espressif/esp-idf/issues/3405

This fix tries to keep in the spirit of #1783 by using libstdc++ rather than a macro, for the same effect (abs can take any compatible type). The other option would be to include `<cstdlib>` before defining the `abs()` macro, so it doesn't get undef-ed again later on.

This will probably also fix https://github.com/espressif/arduino-esp32/issues/2128 as the `abs()` macro is no longer defined.

### DIFF
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -68,12 +68,6 @@
 #define __STRINGIFY(a) #a
 #endif
 
-// undefine stdlib's abs if encountered
-#ifdef abs
-#undef abs
-#endif
-
-#define abs(x) ((x)>0?(x):-(x))
 #define constrain(amt,low,high) ((amt)<(low)?(low):((amt)>(high)?(high):(amt)))
 #define radians(deg) ((deg)*DEG_TO_RAD)
 #define degrees(rad) ((rad)*RAD_TO_DEG)
@@ -160,6 +154,7 @@ void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val);
 #include "HardwareSerial.h"
 #include "Esp.h"
 
+using std::abs;
 using std::isinf;
 using std::isnan;
 using std::max;


MartinL1:
@projectgus I don't agree with this change, as it doesn't maintain backwards compatibility with other Arduino boards. Moving the Arduino abs() macro to the bottom of the "Arduino.h" file to join the min() and max() macros, (after the <algorithm> include):

```
// undefine stdlib's abs if encountered
#ifdef abs
#undef abs
#endif

#define abs(x) ((x)>0?(x):-(x))
```
...allows the Arduino abs() macro to override the standard library's, enabling abs() to operate with floats oncemore and thereby maintaining compatibility with other Arduino boards.

MartinL1:
@projectgus My apologies. I deleted my last comment, as I wrongly assumed that the change to using the standard library wouldn't work with floats.

projectgus:
@MartinL1 No worries, I oversimplified a long story in the PR summary so it's not immediately clear to someone who hasn't followed the whole saga. :) I edited it a bit to try and make it clearer what this change does, vs what was a regression in earlier changes.

me-no-dev:
This has been PITA for a loong time. As far as I remember, we have tried this before and libraries broke.
Fingers crossed 🤞

PR #2743, opened 2019-05-05
Add Schedule.(h|cpp) from ESP8266 to run off loopTask.

dok-net:
This is the non-intrusive part of two existing PRs by @hreintke. It provides source compatibility to ESP8266's Schedule.
An actual benefit that I would use is the ability to easily schedule (periodic) activity to the loop task, without any risk of race conditions, from interrupt service routines. I would likely use this in https://github.com/plerup/espsoftwareserial


me-no-dev:
See my comments in the other PR :)

dok-net:
I currently no longer believe that this or any existing version of Schedule is free of concurrency issues. I am revoking this PR.

PR #2754, opened 2019-05-07
Add Pycom GPy board

devingleeson:
The GPy is an ESP32-based board quite similar to the already-supported Pycom LoPy but the pinout differs somewhat for the integrated LTE modem.


me-no-dev:
Thanks :)

PR #2763, opened 2019-05-09
Fix #2750

### LINKED ISSUE #2750: SPIFFSFS class should not shadow methods in FS class
Please see https://github.com/espressif/esp-idf/issues/3427 for code samples, platform details and discussion. To summarize:

The FS class is not designed to be subclassed insofar as none of its methods are virtual. Re-implementing those methods in SPIFFSFS does not override them, it hides them, so instances of SPIFFSFS can not be passed as references or pointers to the FS class. If they are, the FS methods will be called instead and will not work reliably for a SPIFFS filing system (the existing unit tests are all positive - they test that things work when the files are as they are supposed to be, they don't test negative situations such as files not actually existing).

Instead, an instance of a SPIFFSFS class (such as the instance called _SPIFFS_) should inject a SPIFFS-specific implementation class into the superclass.

judge2005:
This is a fix for https://github.com/espressif/arduino-esp32/issues/2750

### DIFF
--- a/libraries/SPIFFS/src/SPIFFS.cpp
+++ b/libraries/SPIFFS/src/SPIFFS.cpp
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "vfs_api.h"
+
 extern "C" {
 #include <sys/unistd.h>
 #include <sys/stat.h>
@@ -22,6 +24,14 @@ extern "C" {
 #include "SPIFFS.h"
 
 using namespace fs;
+
+class SPIFFSImpl : public VFSImpl
+{
+public:
+    SPIFFSImpl();
+    virtual ~SPIFFSImpl() { }
+    virtual bool exists(const char* path);
+};
 
 SPIFFSImpl::SPIFFSImpl()
 {
--- a/libraries/SPIFFS/src/SPIFFS.cpp
+++ b/libraries/SPIFFS/src/SPIFFS.cpp
@@ -1,3 +1,17 @@
+// Copyright 2015-2016 Espressif Systems (Shanghai) PTE LTD
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 extern "C" {
 #include <sys/unistd.h>
 #include <sys/stat.h>
@@ -13,12 +27,14 @@ SPIFFSImpl::SPIFFSImpl()
 {
 }
 
-bool SPIFFSImpl::exists(const char* path) {
+bool SPIFFSImpl::exists(const char* path)
+{
     File f = open(path, "r");
     return (f == true) && !f.isDirectory();
 }
 
-SPIFFSFS::SPIFFSFS() : FS(FSImplPtr(new SPIFFSImpl())) {
+SPIFFSFS::SPIFFSFS() : FS(FSImplPtr(new SPIFFSImpl()))
+{
 
 }
 
@@ -62,7 +78,8 @@ void SPIFFSFS::end()
     }
 }
 
-bool SPIFFSFS::format() {
+bool SPIFFSFS::format()
+{
     disableCore0WDT();
     esp_err_t err = esp_spiffs_format(NULL);
     enableCore0WDT();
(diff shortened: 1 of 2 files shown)


libraries/SPIFFS/src/SPIFFS.cpp
```diff
@@ -1,32 +1,26 @@
...
 using namespace fs;
 
-SPIFFSFS::SPIFFSFS(FSImplPtr impl)
-    : FS(impl)
-{}
+SPIFFSImpl::SPIFFSImpl()
+{
+}
+
+bool SPIFFSImpl::exists(const char* path) {
```
me-no-dev: please use a new line before the curly braces. I see this a few more times below :)

libraries/SPIFFS/src/SPIFFS.cpp
```diff
@@ -1,32 +1,26 @@
-// Copyright 2015-2016 Espressif Systems (Shanghai) PTE LTD
```
me-no-dev: Do not remove the headers!

libraries/SPIFFS/src/SPIFFS.h
```diff
@@ -1,39 +1,33 @@
-// Copyright 2015-2016 Espressif Systems (Shanghai) PTE LTD
```
me-no-dev: Do not remove the headers!

libraries/SPIFFS/src/SPIFFS.h
```diff
@@ -1,39 +1,33 @@
...
+#include "FSImpl.h"
+#include "vfs_api.h"
 
 namespace fs
 {
 
+class SPIFFSImpl : public VFSImpl
+{
+public:
+	SPIFFSImpl();
```
me-no-dev: too much space?

me-no-dev:
Left a few styling notes. Not sure why you deleted the company headers?

judge2005:
> Left a few styling notes. Not sure why you deleted the company headers?
I implemented these changes in separate files and then overwrote the originals. Anyway - I've added them back...

judge2005:
I merged all of the recent changes into my fork before committing fixes for this. Not sure what effect that will have on attempts by you to merge this specific change to SPIFFS.

libraries/SPIFFS/src/SPIFFS.h
```diff
@@ -14,26 +14,37 @@
 #ifndef _SPIFFS_H_
 #define _SPIFFS_H_
 
+#pragma once
+
 #include "FS.h"
+#include "FSImpl.h"
```
me-no-dev: these two includes should not be in the header, but in the CPP file instead. Nothing in the header depends on them :)

libraries/SPIFFS/src/SPIFFS.h
```diff
@@ -14,26 +14,37 @@
 #ifndef _SPIFFS_H_
 #define _SPIFFS_H_
 
+#pragma once
```
me-no-dev: this is redundant with the above. keep one or the other :)

libraries/SPIFFS/src/SPIFFS.h
judge2005: The header **does** depend on them, but they are also included via vfs_api.h, so at most they are redundant. For the same reason they don't need to be included in the .cpp file (because they are included via vfs_api.h which is included via SPIFFS.h).

Whatever, I have removed them from SPIFFS.h, but I haven't added them to SPIFFS.cpp.

libraries/SPIFFS/src/SPIFFS.h
```diff
@@ -14,26 +14,33 @@
 #ifndef _SPIFFS_H_
 #define _SPIFFS_H_
 
-#include "FS.h"
+#include "vfs_api.h"
```
me-no-dev: FS should be included. vfs_api is a private header.

judge2005: 18 hours ago, you asked me to remove FS.h and FSImpl.h, now you are asking me to add FS.h back in.

If I remove vfs_api.h I will also have to add  FSImpl.h back because SPIFFSImpl is declared in SPIFFS.h.

me-no-dev: No, I asked about `#include "FSImpl.h"` and `#include "vfs_api.h"`.
FS.h is the only one that should be included :) I guess a bit of misunderstanding.

judge2005: This change requires an additional class, SPIFFFSImpl that inherits VFSImpl, which is declared in vfs_api.h, so something is going to need to include vfs_api.h. Currently this class is declared in SPIFFS.h, which is why SPIFFS.h includes vfs_api.h. How would you have me resolve this?

me-no-dev: Does this class need to be public? What happens if the declaration goes into the CPP file?

judge2005: It should be fine if I move it to the CPP file.

me-no-dev:
merged :)

PR #2775, opened 2019-05-13
WifiMulti resilience updates

Daemach:
This update to WiFiMulti is designed to do whatever it takes to get and keep the device online so it can download a configuration file that includes the correct AP credentials.  It adds the following features:

- Failure tracking, to check every entry on the list before resetting and trying them all again.
- AllowOpenAP automatically adds open access points to APList and attempts to connect.
- StrictMode toggle to disable the current behavior which disconnects from an AP if that AP is not on the APList.
- TestConnection connects to a web page on a remote web server and looks for a test phrase to ensure that there is full internet access.  (This is most useful with AllowOpenAP which could add APs that have captive portals, such as hotels, airports, coffee shops, or cable networks like xfinity in the US)
-  Adds an example page for these configuration options.


libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -48,10 +48,19 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout)
...
-            if(!_bStrict || WiFi.SSID()==APlist[x].ssid){
+        if (_bTestConnection && !_bWFMInit){
+            if (testConnection()) {
+                _bWFMInit = true;
                 return status;
             }
+        } else if (!_bStrict) {
+            return status;
+        } else {
+            for(uint32_t x = 0; x < APlist.size(); x++) {
```
atanisoft: since you don't need the index value this can be written as:
```
for(auto ap : APlist) {
  if(WiFi.SSID() == ap.ssid) {
    return status;
  }
}
```

libraries/WiFi/src/WiFiMulti.h
```diff
@@ -59,13 +59,14 @@ class WiFiMulti
     bool _bStrict = true;
     bool _bAllowOpenAP = false;
     bool _bTestConnection = false;
-    String _testPhrase = "This is a very simple HTML file.";
+    String _testPhrase = "simple HTML file.";
     String _testURL = "http://www.brainjar.com/java/host/test.html";
```
atanisoft: It may be better to default this to www.google.com or similar that returns a very small page. Checking the actual content is not super critical to test that the connection was successful to the remote endpoint, just that a connection to the remote was successful.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -261,12 +269,10 @@ bool WiFiMulti::testConnection(int32_t i){
...
     const int httpPort = 80;
     if (!client.connect(host.c_str(), httpPort)) {
-        log_e("connection failed");
-        Serial.println("connection failed");
-        markAsFailed(i);
+        log_e("Connection failed");
         return false;
     } else {
-        log_i("connection to test host successful");
+        log_i("Connected to test host");
```
atanisoft: at this point it should be possible to close the connection without reading the actual data

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -261,12 +269,10 @@ bool WiFiMulti::testConnection(int32_t i){
     WiFiClient client;
     const int httpPort = 80;
```
atanisoft: port needs to be dynamically determined from the test url, it is not guaranteed that all servers will expose port 80 (most are moving to https as a default)

libraries/WiFi/src/WiFiMulti.cpp
Daemach: Much more elegant.  I need to play with vectors more...

libraries/WiFi/src/WiFiMulti.h
Daemach: I was/am concerned that a captive portal would return 200 and fool me.  This test is almost purely to skip captive portals.  I test it with the xfinitywifi AP at the office, but it could be the Starbucks next door.  Thoughts?  I wonder if looking for a 302 header would be better...

libraries/WiFi/src/WiFiMulti.cpp
Daemach: See above.  I would rather skip the text check to make it simpler.  Just need to make sure we don't get trapped.  If this device gets onto xfinitywifi here it doesn't want to get off.

libraries/WiFi/src/WiFiMulti.cpp
Daemach: Hmm.  Can I make an https request directly from wifiClient?  I'm trying to keep it simple by not including more libraries than necessary.  I'm assuming anyone using this function would set the test page manually.  Still, if you feel this is important I'll modify it.

libraries/WiFi/src/WiFiMulti.h
atanisoft: perhaps checking for a 302 would work as well. The ultimate goal would be to avoid downloading the full response when it is not needed in most cases. Only the first few lines of the response will be needed to detect a redirect or if it is a general 200. Having client.find() will result in a lot of heap usage for the duration of the find and it may not even find the string.

libraries/WiFi/src/WiFiMulti.cpp
atanisoft: true, but if a user sets a url with https it would likely fail here as well

libraries/WiFi/src/WiFiMulti.cpp
atanisoft: yeah, ideally this would exit once it receives the initial http header details (all lines until a blank line IIRC). Once the header has been received the connection can be closed. I'd recommend read by line and check the line for known keywords (Location: XXX, HTTP XXX 200|302 etc...) That way you can abort earlier from the line by line checking.

libraries/WiFi/src/WiFiMulti.cpp
Daemach: Yeah.  In my example I specify that it must be http://.  Need to think it through a bit more.

libraries/WiFi/src/WiFiMulti.h
Daemach: will client.find() abort as soon as it finds the target string?  302 Found in this case

Daemach: Make that 200 OK...  I updated the h file.

me-no-dev:
What is the state here @Daemach ? Is this ready for review and merge?

Daemach:
Yes.  Let me know if I need to modify anything.

atanisoft:
@Daemach can you drop the .vscode directories? I don't think we want those polluting the source tree (especially dynamically generated files)

Daemach:
Missed those.  Thanks for the heads-up!

atanisoft:
> Missed those. Thanks for the heads-up!

Looks like a few are still there...

Daemach:
Not seeing them - where?

atanisoft:
file diff shows these directories still:
tools/sdk/include/coap/.vscode/
tools/sdk/include/esp_http_server/.vscode/
tools/sdk/include/log/.vscode/
tools/sdk/include/wpa_supplicant/wpa/.vscode/
variants/airnode/.vscode/

stickbreaker:
@Daemach at top of this page click `Files Changed`  Currently it shows 14 Files Changed.  You should be able to use the triple dots to selected deleted for each of the vscode file files.

Daemach:
Sheesh!  Is there any way to keep vscode from creating those folders?

Daemach:
Thanks for the guidance!  All new to me.

stickbreaker:
@Daemach you might want to review [git ignore file list](https://git-scm.com/docs/gitignore) :grinning:

me-no-dev:
@Daemach please rebase this :)

libraries/WiFi/src/WiFiMulti.h
```diff
@@ -42,10 +43,31 @@ class WiFiMulti
...
+    void APlistClean(void);
+
 private:
     std::vector<WifiAPlist_t> APlist;
+
+    bool _bStrict = true;
+    bool _bAllowOpenAP = false;
+    bool _bTestConnection = false;
+    String _testPhrase = "301 Moved";
+    String _testURL = "http://www.amazon.com";
```
me-no-dev: this will not work in China :)

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -199,6 +266,143 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout)
...
+    _bTestConnection = bTestConnection;
+}
+void WiFiMulti::setTestPhrase(const char* testPhrase){
+    _testPhrase = testPhrase;
+}
+void WiFiMulti::setTestURL(String testURL){
+    _testURL = testURL;
+}
+
+bool WiFiMulti::testConnection(){
```
me-no-dev: why not add the URL as an argument (defaulting to NULL) and then parse it to get the port and so on?

libraries/WiFi/src/WiFiMulti.cpp
me-no-dev: I agree with @atanisoft here :) there are some codes that show all is good.

me-no-dev: actually what is the target? which codes would mean that something went bad?

libraries/WiFi/src/WiFiMulti.h
Daemach: You can set the testURL to whatever you want.  Do you have a better suggestion for a URL that will work everywhere and also redirects with something other than 304?  I'm trying to detect captive portals which normally return 304 or 200.

libraries/WiFi/src/WiFiMulti.cpp
Daemach: I figured most people would just want to turn it on and forget it so I defaulted the test URL and phrase to amazon which I thought would be universal.

libraries/WiFi/src/WiFiMulti.cpp
Daemach: All I am trying to do is ensure that the device can actually reach the internet - captive portals are the problem.  If you're connected to Wifi but behind a captive portal you're not going anywhere.  This functionality is most useful when you have bAllowOpenAP, which automatically connects to any and all open APs in addition to the ones in your list just to get online.  In my case, once it's online it connects to my server, grabs a config file with the correct SSID information then turns bAllowOpenAP off and bStrictMode on.  It's good functionality to put 1000 brand new devices online just by plugging them in and setting your AP to open mode for 15 minutes.

I'm looking for anything but a 304 by default, but if that isn't good enough, you can set your own URL and look for any test phrase that you want, including text in a webpage.  By default I don't want to look for 200's either in case a captive portal happens to return that.  Looking for a permanent redirect seems like the best solution because a captive portal wouldn't do that and a site like google or amazon should be available everywhere.

Daemach:
I would still like to see this added. I’ll rebase on 1.0.3

VojtechBartoska:
Hello, @Daemach please sign CLA, thanks.

libraries/WiFi/examples/WiFIMultiAdvanced/WiFiMultiAdvanced.ino
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+void loop()
+{
+    if( wifiMulti.run() != WL_CONNECTED) {
+        Serial.println("WiFi not connected!");
+        delay(5000);
+    }
+}
```
lucasssvaz, line 47:
```suggestion
}

```

libraries/WiFi/src/WiFiMulti.cpp
lucasssvaz, line 1: Could you please merge the master branch and resolve the conflicts ?

libraries/WiFi/src/WiFiMulti.h
lucasssvaz, line 63: Idk if you need some specific kind of page but google's URL would work for most places, including China.

me-no-dev:
Continued in: https://github.com/espressif/arduino-esp32/pull/9139

PR #2811, opened 2019-05-22
Update HardwareSerial.cpp

IkpeohaGodson:
I changed UART1 rxpin from GPIO9 to GPIO2. I changed UART1 txpin from GPIO10 to GPIO4

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -7,11 +7,11 @@
 #include "HardwareSerial.h"
 
 #ifndef RX1
-#define RX1 9
+#define RX1 2
 #endif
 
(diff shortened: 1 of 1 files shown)


atanisoft:
This should be redefined in pins_arduino.h, the defaults here are from Espressif.

Jeroen88:
Or pass them in the ::begin() function for Serial1:
`
void begin(unsigned long baud, uint32_t config=SERIAL_8N1, int8_t rxPin=-1, int8_t txPin=-1, bool invert=false, unsigned long timeout_ms = 20000UL);
`

me-no-dev:
As @atanisoft said :) different defaults should be added in pins-arduino and not here :)

Issue #2822, opened 2019-05-25
AzureIoT submodule fails to compile due to Werror-maybe-uninitialized


sieren:
### Hardware:
Board:							ESP32 Dev Module / Wroom Kit etc.
Core Installation version:			various (xtensa-esp32-elf-osx-1.22.0-80-g6c4433a-5.2.0)
IDE name:						IDF component, CMake Build system
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Mac OSX & Linux

### Description:
Compiling as an IDF component (Cmake, C++14 enabled) fails due to `-Werror=maybe-uninitialized`.
I submitted a fix to the submodule a while ago: https://github.com/VSChina/ESP32_AzureIoT_Arduino/pull/15 but it seems like it is no longer maintained by @lirenhe who added it originally to this repo.

```
/home/travis/build/sieren/HomeControl/components/arduino-esp32/libraries/AzureIoT/src/az_iot/iothub_client/src/iothubtransport_mqtt_common.c:1058:32: error: 'request_id' may be used uninitialized in this function [-Werror=maybe-uninitialized]
 if (request_id == msg_entry->packet_id)
              ^```

sieren:
Merged and therefore closed :)

PR #2824, opened 2019-05-25
Ignore Werror-maybe-uninitialized from Azure IoT Library

sieren:
Ignore the error about a potentially uninitialized variable
in the Azure library that will occur when compiling as a component
in ESP-IDF.

See https://github.com/espressif/arduino-esp32/issues/2822


me-no-dev:
Thanks @sieren :) Maybe I should ping MS to update the lib

mfk-ableton:
Thanks for merging 👍

Issue #2836, opened 2019-05-28
Bundled AsyncUDP has invalid library.json


andig:
I'm trying a cross-build for esp8266 and esp32 both. I need to use ESPAsyncUDP.

In order to have the esp32 build *NOT* use the ESPAsyncUDP but the built-in AsyncUDP, I've set `lib_compat_mode=strict`.

    [env:esp32]
    platform=espressif32
    # use stage version for added SPIFFS support
    #platform=https://github.com/platformio/platform-espressif32.git
    board=featheresp32
    framework=arduino
    lib_compat_mode=strict
    lib_ldf_mode=deep
    lib_deps=
        ${common.lib_deps}
        ESP Async WebServer@^1.2
    #    https://github.com/me-no-dev/ESPAsyncWebServer
        https://github.com/me-no-dev/AsyncTCP
        AsyncUDP
        SPIFFS
        FS

Problem is that now the `AsyncUDP` is ignored according to the verbose build:

    LDF MODES: FINDER(deep) COMPATIBILITY(strict)
    Platform incompatible library /Users/andig/htdocs/SimpleEVSE-WiFi/.piolibdeps/ESPAsyncTCP_ID305
    Platform incompatible library /Users/andig/htdocs/SimpleEVSE-WiFi/.piolibdeps/ESPAsyncUDP_ID359
    Platform incompatible library /Users/andig/htdocs/SimpleEVSE-WiFi/.piolibdeps/MFRC522_ID63
    Platform incompatible library /Users/andig/.platformio/packages/framework-arduinoespressif32/libraries/AsyncUDP

Looking at the platform directory I see that `libraries/AsyncUDP` contains this `library.json`:

    {
    "name":"AsyncUDP",
    "description":"Asynchronous UDP Library for ESP32",
    "keywords":"async,udp,server,client,multicast,broadcast",
    "authors":
    {
        "name": "Hristo Gochkov",
        "maintainer": true
    },
    "repository":
    {
        "type": "git",
        "url": "https://github.com/me-no-dev/ESPAsyncUDP.git"
    },
    "frameworks": "arduino",
    "platforms":"espressif"
    }

This seems wrong twice- I assume the platform should be "espressif32" _and_ wrapped inside an array or the library.json removed altogether. The latter would limit confusion as it still points to the ESPAsyncUDP repo.

Btw, `frameworks` should also be an array?

Refs https://github.com/platformio/platform-espressif32/issues/186

PR #2837, opened 2019-05-28
Update AsyncUDP library.json

### LINKED ISSUE #186: ESP32
Hi,
    Please anyone tell me how to run BLE and WiFi in ESP32 simultaneously
Thanks

andig:
Fix https://github.com/espressif/arduino-esp32/issues/2836 and https://github.com/platformio/platform-espressif32/issues/186


andig:
ping @me-no-dev does this make sense for you?

libraries/AsyncUDP/library.json
```diff
@@ -1,17 +1,15 @@
...
-  {
-    "type": "git",
-    "url": "https://github.com/me-no-dev/ESPAsyncUDP.git"
-  },
-  "frameworks": "arduino",
-  "platforms":"espressif"
-}
+  "frameworks": [
+    "arduino"
+  ],
```
atanisoft: changing frameworks to a list is not necessary

libraries/AsyncUDP/library.json
```diff
@@ -1,17 +1,15 @@
...
+  "platforms": [
+    "espressif32"
```
atanisoft: by changing this to espressif32 you are desupporting the esp8266, you will need to add "espressif" as a second element in this list.

andig:
Happy to make the requested changes. 

As for the platform I was under the impression that this repo is for esp32 *only* and esp8266 i supported at https://github.com/me-no-dev/ESPAsyncUDP.git? Are you actually suggesting that I should open this PR upstream?

atanisoft:
You are correct that this is only for the esp32, I missed which repo this was on and removed that comment after checking further. This version should only be for espressif32.

andig:
Looking at the other embedded libraries- should we maybe drop the library.json altogether?

atanisoft:
That is an option but it would need to be tested and confirm that PIO won't complain about it being missing.

andig:
> That is an option but it would need to be tested and confirm that PIO won't complain about it being missing.

Just verified that my PIO project compiles without it present. No other included library does have this file either ;)

atanisoft:
Then I'd suggest drop it from here :)

andig:
Pr is already updated- up to you :)

me-no-dev:
Hey @andig :) Long time no see :D

andig:
> Hey @andig :) Long time no see :D

So true- hope you're doing well!

PR #2838, opened 2019-05-29
2 Olimex boards added and 3 slightly modified.

Stanimir-Petev:
Added ESP32-PoE-ISO (similar to ESP32-PoE) and ESP32-DevKit-Lipo (similar to the base ESP32) in the boards.txt
ESP32-EVB, ESP32-Gateway and ESP32-PoE (no ISO) added pins for additional functionality of alternative Serial (Serial1/Serial2) on either UEXT or extention pins.


me-no-dev:
Готово :)

PR #2841, opened 2019-05-29
Added convert method to EEPROM to transfer data from partition to nvs

lbernstone:


me-no-dev:
Uhmm... how will that work if the EEPROM partitions are removed from the schemes? How will you get to execute that code and for it to run on older partition schemes?

lbernstone:
It is just for a one time conversion from partition to nvs.  You run it independently before the begin.  If there is no partition, or if the partition is empty (all 0xFF), then it returns a 0 (no bytes converted).

lbernstone:
It is written such that it does no harm if run multiple times.  So, I could roll it into the begin routine if that makes sense to you.

PR #2848, opened 2019-06-01
Fix map() division by zero

ExcaliburGO:


me-no-dev:
Thanks @ExcaliburGO !

PR #2849, opened 2019-06-02
Ticker updated to match extensions in ESP8266 API

dok-net:
This PR updates the Ticker class to the ESP8266 in corresponding PR there.

Additional info:
This also prepares for an enhancement in another PR that would provide a co-op single-thread, IRQ safe Scheduler.
Even on FreeRTOS a single-thread scheduler in the loop-context is useful: for most sketches OS-level multitasking is not necessary, but synchronization is a difficult effort to get right. Also, the single-thread performance of the ESP32 in measured tests is lower than that of current non-OS SDK ESP8266, so an efficient user-context scheduler for timed and IRQ-based tasks is useful - besides, the async programming pattern this affords is quite popular.


libraries/Ticker/src/Ticker.h
```diff
@@ -26,82 +26,115 @@
...
-    _attach_ms(milliseconds, false, reinterpret_cast<callback_with_arg_t>(callback), arg32);
-  }
-
-  void detach();
-  bool active();
-
-protected:	
-  void _attach_ms(uint32_t milliseconds, bool repeat, callback_with_arg_t callback, uint32_t arg);
-
+	Ticker();
```
me-no-dev: When you change the style of the header like this, it get's hard to track what actual changes you have made. Was the change to spacing necessary?

libraries/Ticker/src/Ticker.h
```diff
@@ -26,82 +26,115 @@
...
+	~Ticker();
+
+	typedef void (*callback_with_arg_t)(void*);
+	typedef std::function<void(void)> callback_function_t;
+
+	void attach(float seconds, callback_function_t callback)
+	{
```
me-no-dev: could all those functions be implemented in the source file instead?

dok-net: @me-no-dev Whatever happened to my replies (did GitHub swallow them?), here's the gist of it:
I am sorry for whitespace causing review difficulties, but it's all about the tooling (GitHub has it's share). I am seriously in favor of keeping the really small and efficient inlineable functions in the header, if you read them verbatim, not in review diff, it's quite tidy - my opinion, of course.
If you absolutely cannot merge this PR without those changes, yes, I'll do them, but I am completely unconvinced as it's rather more difficult making adjustments across multiple files, if there's no compelling ground for that.

dok-net:
@full-stack-ex You've recently made a change to Ticker, could you please review this PR? I've adopted the ULL integer literal fix from your patch.

full-stack-ex:
Well, it looks like the function called from Timer accepts a 64 bit us time (period):
```c++
esp_err_t esp_timer_start_periodic(esp_timer_handle_t timer, uint64_t period);
```
But this one, which calls it, accepts 32 bits only (micros, passed to the above as period):
```c++
void Ticker::_attach_us(uint32_t micros, bool repeat, callback_with_arg_t callback, void* arg)
```
Should it also accept 64 bits, like this, to pass any valid value without truncation?
```c++
void Ticker::_attach_us(uint64_t micros, bool repeat, callback_with_arg_t callback, void* arg)
```

dok-net:
@full-stack-ex Converted API to 64bit time values. Thanks!
@me-no-dev Code is again dissimilar enough to ESP8266, so it doesn't matter, moved most from header into compilation unit as you requested in review comments.

Good to for merge now?

full-stack-ex:
As far as 64 bit conversion is concerned:
> Converted API to 64bit time values.

Good to go.

By the way, there is a side effect, though positive. Just in case, it might affect some documentation, tests, etc. I don't know of any. The old 32 bit millisecond range used to limit the period to UINT32_MAX ms, which is 4294967295/1000/60/60/24 = 49.7 days. The 64 bit millisecond range eliminates that constraint, and so the period may be any practical value. Say, a year, should anybody need that.

hreintke:
@dok-net As the ESP32 timer  functions allow microsecond resolution, should there also be attach_us.. functions ? 
Benefit of that above the alternative timer usage is that the callback in in TASK, not ISR.

hreintke:
@dok-net Sorry for the confusion. Did not check the PR properly.
You are right, it is already included. 
Will delete my original comment.

dok-net:
@me-no-dev Just a heads up :-) All reviewers gave their OK, I've recently followed your request to move as much from the header into the compilation unit as possible.

Domochip:
Thanks a lot Dok-Net for this work. It think this PR is very usefull. I'm just trying to upgrade my first ESP8266 project to ESP32 and encounter this missing feature...

fabiuz7:
Hi all,
is there any possibility to get this PR on master branch anytime soon?
Anyway, thanks for your work!

vortigont:
Any chance to get this simple change merged? Missing <functional> callbacks makes most of 8266 sketches incompatible and forces to use ugly #ifdef hacks. :(

Domochip:
Up

mrengineer7777:
This PR needs to be updated.  See #7664 .

dok-net:
@mrengineer7777 The implementation for ESP8266 has undergone some changes, the ESP32-based as well, or will due to another PR. I lack the need and also the resources to maintain this here PR, unless you have some overriding reason, I shall close it without merging.

mrengineer7777:
@dok-net Congrats on updating your PR.  Unfortunately it deletes one of the examples and the third parameter is wrong on one of the calls.

I'm a fellow developer (not a maintainer), so I can't tell you what to do.  That said, your PR introduces a lot of changes.  It is more likely to get merged with just a few clear changes.  For instance adding new example files.

I like the update on keywords.txt.  I don't know if it is necessary (I don't use Arduino IDE), but seems reasonable.

I don't agree with changing _attach_ms() to _attach_us().

I'm curious about introducing `Ticker::_static_callback(void* arg)`.  How is that used?

Honestly at this point I would likely start with a fresh PR and just introduce minimal changes.  I have successfully written and merged several PRs for this project.  Happy to give feedback.

See example below of how I'm bringing over some changes from ESP8266.  Note I reference prior PRs and code lines.
https://github.com/espressif/arduino-esp32/issues/7766
https://github.com/espressif/arduino-esp32/pull/7774

dok-net:
"Congrats"? You're welcome.
You missed my last comment then? I only forgot to set the PR to draft. I am not expecting this to be of any more interests to the maintainers now that during the past 2.7 years. If you are interested in picking it up, please make a clone soon.

Whenever you spend your time and find issues in code, be as precise as you can when you mention them to the devs, just letting me know that I've got a (third) parameter wrong isn't quite so helpful.

I did just fix a warning that occurs in the TickerParameter.ino sample. I believe my original thinking was that the three examples in this PR are more comprehensible than the previous one, the one I have deleted. Nevermind.

Take care!

mrengineer7777:
> "Congrats"? You're welcome.

Thank you for your efforts.

> Whenever you spend your time and find issues in code, be as precise as you can when you mention them to the devs, just letting me know that I've got a (third) parameter wrong isn't quite so helpful.

Correct.  I didn't go into details because it sounds like you've given up on the PR.  I don't blame you.  It's hard to get a PR merged.  I hear your frustration.  Let's take a step back.  What functionality would you like to see added?

anabolyc:
Really sad to see that such useful change cannot be merged in 4 years. I'm working on few esp32 projects and constantly having issues with inability to use Ticker in classes, due to (honestly speaking) ridiculous type incompatibility. This would nail it, I assume.

mrengineer7777:
@anabolyc I suspect the Ticker class is about to be refactored due to changes in IDF 5.1.  What functionality are you looking for?

anabolyc:
@mrengineer7777 I'm compiling this simple class
```
#pragma once

#include <Ticker.h>

class SampleClass
{
private:
    uint32_t lastUpdate = 0;
    Ticker tick;
public:
    SampleClass(/* args */) {};
    void init(void);
};

void SampleClass::init(void) {
    tick.attach(5, [&](){
        lastUpdate = millis();
    });
}
```
No issues compiling against esp8266, but on the esp32 getting
```
src/SampleClass.h:18:6: error: no matching function for call to 'Ticker::attach(int, SampleClass::init()::<lambda()>)'
```

mrengineer7777:
@anabolyc 
Try using attach_ms() instead of attach().  Check your timing!
Not sure if an inline function (lambda) is supported.  Here's a working example:
https://github.com/espressif/arduino-esp32/blob/master/libraries/Ticker/examples/Blinker/Blinker.ino

anabolyc:
> @anabolyc Try using attach_ms() instead of attach(). Check your timing! Not sure if an inline function (lambda) is supported. Here's a working example: https://github.com/espressif/arduino-esp32/blob/master/libraries/Ticker/examples/Blinker/Blinker.ino

Point is, in the current shape it would only accept free function, then you need workarounds and compromises to access class members. While it just builds against esp8266 version.

lucasssvaz (approved):

P-R-O-C-H-Y (approved):
LGTM.

PR #2850, opened 2019-06-02
Add coop loop() and yield() Scheduler - single-thread efficient scheduling for Timer and IRQs

dok-net:
Even on FreeRTOS a single-thread scheduler in the loop-context is useful: for most sketches OS-level multitasking is not necessary, but synchronization is a difficult effort to get right. Also, the single-thread performance of the ESP32 in measured tests is somewhat lower than that of current non-OS SDK ESP8266, so an efficient user-context scheduler for timed and IRQ-based tasks is useful - besides, the async programming pattern this affords is quite popular.


me-no-dev:
Hey @dok-net :) What is this supposed to be? Another scheduler? We are not turning the ESP32 into ESP8266 here :P We already have a scheduler and all. If anyone finds it hard to spin tasks, maybe they need to do some reading first. I do not agree that we need to add another scheduler to a running OS

me-no-dev:
Also, once you reach 5-6 changes, you need to rebase... we do not need this long list of small changes to one PR :)

dok-net:
@me-no-dev Rebasing certainly isn't for the faint of heart or those that have limited time resources. Squash commits might help, but I wouldn't throw out the work history for anyone's sake.

I tried to provide reasonable motivation for this PR in the original comment - it's certainly useful. I'm beginning to not understand why we have different opinions on what expectation the name "Arduino" is allowed to forster, is the user base supposed to be non-migrating between even similar devices from the same company, and library compatibility not desired? Working in a single-threaded model - that also solves IRQ difficulties just nicely - is a great thing for many users. If this is just for technical pros like us, I must have misunderstood the whole Arduino and maker movement.

I cannot agree with your comment on the commit history, either - GitHub offers a review functionality that shown the resulting diff (even with the option to ignore pure whitespace differences), the number of commits to my repo should not cause you concern. I am eternally sorry for the whitespace changes my work environment creates, but really, that's run of the mill stuff.

OK, what do you think about this:

I'll split out the weak function part by itself - the Scheduler can then be an Arduino portable libraries that you won't have to worry about. I should have done this from the beginning, I guess, I am sorry.

dok-net:
The implementation by the core team on ESP8266 turns out to be quite a work in progress, and this is a competing (unwanted?) competitive development with some advantages and disadvantages in other places, too.
I agree with @me-no-dev, there's also other PRs of mine that are more important to me, and close this, perhaps in favor of #2865. The Ticker update parts are already in another standalone PR, too, so nothing is lost.

PR #2865, opened 2019-06-06
Add hooks for sketches/libraries to perform work on each yield() and completed loop()

dok-net:
For periodic workloads, as in some libraries, being able perform on each yield() and after each loop without modifications to the main sketch code, is very useful, sometimes even necessary. Full multi-tasking / threading may sometimes not be an option due to data atomicity and other concurrency considerations.

### DIFF
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -130,6 +130,9 @@ void init(void);
 void initVariant(void);
 void initArduino(void);
 
+void yield_completed(void);
+void loop_completed(void);
+
 unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout);
 unsigned long pulseInLong(uint8_t pin, uint8_t state, unsigned long timeout);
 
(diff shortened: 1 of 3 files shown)


dok-net:
@me-no-dev @igrr This would be the basic changes required to allow me to do what PR #2850 does, but in a stand-alone library. EspSoftwareSerial comes to mind a as direct user of this - I know you don't find that lib useful, but some serial features (9bit protocols) are requested and not covered by the hardware, AFAI(or anyone else)K.

me-no-dev:
You are on a multithreaded OS. All that code is pointless :) Please understand this! This is not ESP8266.

dok-net:
@me-no-dev I seriously contest that single-threaded cooperative multitasking is pointless, unless it's proven to me that
- all developers can write error-free code in MT environments
- there is no performance hit by OS preemptive context switching over user-space cooperative task switching. According to a lot of research, there is usually more that a niche for the latter.

That is not to say that you have good reason to try and get people to learn FreeRTOS, but I think there's a good deal of sketches and libraries that take advantage of the "Arduino" APIs in esp32-arduino, which is and should remain more than just a way of taking advantage of the build and uploading framework integration.

Those hooks that I am asking for would make this a little easier on the eye, because the stuff in loop() could become part of a tiny "scheduler":

https://github.com/dok-net/CoopTask/releases/tag/0.4.0

Naturally, I am not opposed to any review! If you find a reason why my code doesn't work, I would be more than happy to know. Otherwise, as always, thanks for all the stuff!!!

PR #2867, opened 2019-06-06
Bugfix/http client

Jeroen88:
Made a mistake in parsing the header for HTTP1.0 in #6152. This PR corrects this.


Jeroen88:
Updated the reuse logic to be in line with the ESP8266.

Jeroen88:
Do not merge this PR yet. It has an unsolved problem that I am tracking on the [same PR](https://github.com/esp8266/Arduino/pull/6176) for the ESP8266

me-no-dev:
@Jeroen88 make sure you ping me on gitter once this is OK for merge :) I might miss it here

Jeroen88:
@me-no-dev Sure, I will :) However, I have the same PR for ESP8266 (see [here](https://github.com/esp8266/Arduino/pull/6176)), and I think there are still reuse issues. I think so by looking at the code. However, I miss the protocol knowledge to be sure, nor do I have defined test cases (send request header Connection: Keep-Alive, get response header Connection: Close, I think ::disconnect() should be called, and ::connect() on the next POST, but is that true?; Is it possible that a connection is reset during a Keep-Alive situation and should it be handled by HTTPClient?). 
Could you please have a look at my last post at that PR and give your opinion? It is holiday time here, so I will have other priorities for the next few weeks :)

Jeroen88:
@me-no-dev the HTTPClient library is in line with the ESP8266HTTPClient library, as far as reusing the connection is concerned. I tested the changes on the ESP8266. Also I ran the ESP32 code to test HTTP/1.0 and HTTP/1.1, both with ::setReuse(false) and ::setReuse(true). Everything seems to be fine now...

Issue #2866, opened 2019-06-06
BLE Library Compilation Error with Arduino as ESP-IDF component


coolbreeze413:
### Hardware:
Board:							ANY

### Description:
Compilation Error in BLE_Advertising.cpp at lines 265 and 274:

```
log_v(">> setPrivateAddress")
log_v("<< setPrivateAddress")
```

https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/src/BLEAdvertising.cpp#L265

https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/src/BLEAdvertising.cpp#L274

Both are missing semicolons!

Introduced as part of commit:
https://github.com/espressif/arduino-esp32/commit/bea7bd1852bae767864dbb582ea4d4123bc968c1

atanisoft:
@coolbreeze413 please consider closing this issue as this has been fixed via PR #2876

coolbreeze413:
closing the issue, as fix merged.
apologize for delay, missed the merge notification.

PR #2876, opened 2019-06-09
fix for BLE Library Compile Error #2866

coolbreeze413:
Simple fix for missing ';' in the file libraries/BLE/src/BLEAdvertising.cpp.


elliottkember:
+1 - I just hit this error today out of the blue. Only seemed to pop up when I compiled with `DebugLevel=verbose`.

elliottkember:
Thanks!

PR #2901, opened 2019-06-18
Added support for TinyPICO in the boards list

UnexpectedMaker:
I have added support for the ESP32 PICO-D4 based TinyPICO in the boards list, so users can select that board when using their TinyPICO.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -279,6 +279,64 @@ pico32.menu.DebugLevel.verbose.build.code_debug=5
 
 ##############################################################
 
+tinypico.name=TinyPICO
+
+tinypico.upload.tool=esptool_py
+tinypico.upload.maximum_size=1310720
+tinypico.upload.maximum_data_size=327680
+tinypico.upload.wait_for_upload_port=true
+
+tinypico.serial.disableDTR=true
+tinypico.serial.disableRTS=true
+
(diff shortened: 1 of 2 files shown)


me-no-dev:
I noticed that you have selected DIO as flash transport. Have you thought of giving QIO a go? Coupled with 80MHz and PSRAM, that will make both run QIO 80MHz and give the best performance. Downside is that in this mode HSPI is used for PSRAM and not available as second SPI for Arduino (VSPI is still there)

UnexpectedMaker:
> I noticed that you have selected DIO as flash transport. Have you thought of giving QIO a go? Coupled with 80MHz and PSRAM, that will make both run QIO 80MHz and give the best performance. Downside is that in this mode HSPI is used for PSRAM and not available as second SPI for Arduino (VSPI is still there)

Thanks for the merge @me-no-dev !

I did try using QIO at the very early stages of development, but I was still trying to figure out the PSRAM, and at some stage I just set it back to DIO and left it :( 

I'm going to commit a new update that exposes the flash mode and frequency in the menu, and set it by default to QIO/80, but this way people can change it if they need. 

Thanks for the suggestion!

me-no-dev:
@UnexpectedMaker do you sell the board in europe? What is the best place to grab one?

PR #2903, opened 2019-06-18
 fix empty reply from server error

pmalhaire:
The flush causes an empty response a client side.

see https://github.com/espressif/arduino-esp32/issues/2902


me-no-dev:
thanks :)

Issue #2900, opened 2019-06-18
[Feature Request] It would be useful to know if a received AsyncUDPPacket is part of (and which part of) a fragmented packet


ssilverman:
When I receive large packets, they are often fragmented. It would be useful if `AsyncUDPPacket`, via `AsyncUDP`'s `onPacket` call, contained information about if and how the packet is fragmented, for example, the _fragment offset_ and _total length_ information.

See: https://en.wikipedia.org/wiki/IPv4#Fragmentation_and_reassembly

r1dd1ck:
@ssilverman This "fragmentation" you are experiencing is obviously caused by the application that sends it - most probably because it uses a 1024-byte payload limit/buffer where payloads larger than that get automatically split into multiple packets. I suppose you have to handle this type of "fragmentation" on your own @ sender/receiver ends.

ssilverman:
There should be enough information in the packet to be able to pass along fragmentation information. The fact that it's fragmented should be detectable. It's impossible, currently, to handle the fragmentation on the receiver with the current API.

Fragmentation or the fact that it's happening isn't relevant to this issue, as you're suggesting. I'm only suggesting to enhance the API to provide information that already should be in the fragmented packets, specifically, the _fragmentation offset_ and _total length_ information. See the Wikipedia link in my first post in this issue.

Unless, of course, the ESP32 low-level APIs don't provide access to this information?

ssilverman:
I just went through the `AsyncUDP.cpp` code and having access to `pbuf.tot_len`, in addition to the count inside `_recv`, are things I'm looking for. Let me see about a pull request...

ssilverman:
See the new pull request: #2904.

r1dd1ck:
So you are just looking to extend the functionality of the `AsyncUDP` arduino library - OK.

But have you checked that `pbuf.tot_len` contains the **total payload length** before the data was split at sender side?

ssilverman:
It doesn't matter what my network does; many networks are different. All I'm saying is it's useful to be able to access `pbuf.tot_len`. Are you suggesting this field is never set to something other than the same value as `pbuf.len`?

lbernstone:
[...]

PR #2904, opened 2019-06-18
Adding a way to access the remaining length for fragmented packet data in AsyncUDPPacket

ssilverman:
This adds a way to access the number of remaining bytes for fragmented packet data via a new `AsyncUDPPacket::remainingLength()` method. This simply adds an accessor to data that's already there in the `pbuf` object. This addresses #2900.


me-no-dev:
but this is not the remaining len. It's the length on all pbufs chained after this one.

ssilverman:
Ah, I should have closed this after I closed the related feature request (#2900).

Issue #2899, opened 2019-06-17
Network stack never recovers without a hard reset after receiving closely-spaced larger UDP packets


ssilverman:
### Hardware:
Board: Adafruit ESP32 Feather
Core Installation version: Core v1.0.2 in Arduino 1.8.9, espressif32@1.8.0 and framework-arduinoespressif32@2.10002.190416 in PlatformIO
IDE name: Arduino v1.8.9 and PlatformIO v3.6.7 in VSCode v1.35.1
Flash Frequency: 80MHz
PSRAM enabled: Don't know
Upload Speed: 921600
Computer OS: Mac OSX v10.14.5

### Description:
The network stack never recovers if it receives just tens of larger UDP packets greater than 1024 bytes. This detail got lost in the discussion of #2871, so I'm creating this new bug.

I see this problem almost every time I send somewhere between 20-80 larger UDP packets in non-softAP mode. I see this problem sometimes but rarely when using softAP mode. Sometimes, if I repeatedly run the program, see the network stack crash, and then hard reset the device, I won't see the network stack failure and it will recover from seeing a bunch of large packets. But only sometimes.

To reiterate: The network stack does not recover after receiving a bunch of large packets. Some people see that it recovers just fine if the packets stop, but I never see this. Yes, the device stops seeing packets if it's flooded with them, and yes, some people will see the network stack recover after a short period, but I never do. A hard reset is required.

To continuously send closely-spaced packets, this Bash script is useful:
```bash
while true; do echo -n $(printf '.%.0s' {1..1400}) > /dev/udp/192.168.1.9/8000; sleep 0.05; done
```

The `1400` is the UDP size; I find that 1025 and greater causes the problem for me. Also, change the IP address and sleep (in seconds) to play with different network loads.

The crux is this: I see `onPacket` never called again once a bunch of large packets are received, even if no more packets are sent. I've tried on three different ESP32 Feathers and three different network setups for non-softAP mode.

The effect: The device becomes permanently unusable on the network, without a hard reset, if it sees lots of larger UDP packets in a row.

### Sketch:
```cpp
#include <AsyncUDP.h>
#include <Esp.h>
#include <WiFi.h>

constexpr char kAPName[] = "ChangeMe";
constexpr char kAPPassword[] = "ChangeMe";
constexpr bool isSoftAP = false;  // Change to true for SoftAP mode

AsyncUDP udp;

void setup() {
  Serial.begin(115200);
  while (!Serial && millis() < 4000) {
[...]

PR #2912, opened 2019-06-20
Change some WiFi buffer settings to match esp-idf.  See #2899

lbernstone:
Buffer mismatches appear to be causing WiFi communications to halt.  See #2899 for more details.


stickbreaker:
@lbernstone since you found these relationships between buffer configurations, why don't you add some error checking for the values. added these compiler checks would prohibit this error from happening again.

This is just an example, You will have to figure out the correct rules to test.
```c++
#if (CONFIG_ESP32_WIFI_RX_BA_WIN > CONFIG_ESP32_WIFI_STATIC_RX_BUFFER_NUM )
# error buffer config mismatch
#endif 
```

Do you know the limit values?  Are the current values in your PR "empirical" or calculated?

Chuck.

lbernstone:
They are taken from the default sdkconfig in esp-idf, so not really even empirically derived.  They are compiled into the libraries, so I'm not sure what an error check would do for anybody but travis.  The error check really ought to be upstream in LWiP, where the mismatch is actually causing the problem.

atanisoft:
> They are compiled into the libraries, so I'm not sure what an error check would do for anybody but travis.

These are in fact compiled at the time of compiling arduino-esp32 code and not IDF. This is not always the case though. The WiFi library itself uses these values when it initializes the wifi stack.

r1dd1ck:
These checks will be implemented @ ESP-IDF sooner or later, so I would stay with just fixing the values in the current sdkconfig now & let the checks come in later with an IDF update.

lbernstone:
@atanisoft 
> 
> These are in fact compiled at the time of compiling arduino-esp32 code and not IDF. This is not always the case though. The WiFi library itself uses these values when it initializes the wifi stack.

Doh, you are right.  Only used in headers.

atanisoft:
> Doh, you are right. Only used in headers.

@r1dd1ck is the one he who pointed this out. Almost all settings are precompiled in but there seem to be a handful that are adjustable at the app level.

It would be really nice to be able to override these defaults in our app code without modifying core headers but that isn't possible or maintainable.

lbernstone:
@atanisoft : Copy the header into your sketch folder.  Arduino will use the local copy instead of the include folder version.

atanisoft:
@lbernstone now that is an awesome find!

Daemach:
Does this benefit wifi in general?  If so can you provide a link to the
header that should be copied?

atanisoft:
@Daemach sdkconfig.h I believe

stickbreaker:
@Daemach [tools/sdk/include/config/sdkconfig.h](https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/include/config/sdkconfig.h) The other is file is used for the build config.

Chuck.

me-no-dev:
how does the heap look when you increase the static buffers?

r1dd1ck:
@me-no-dev 
If I remember correctly, it takes about 3-4kB more than with the current arduino defaults (8).
Raising it to the `iperf` example defaults (16) had a 13-14kB impact.

Maybe it could survive even without the static buffer raise, as I suppose the most crucial part was the `RX_BA_WIN` adjustment (16 → 6). Haven't played around with the numbers since then, but the values from this PR have proven to be quite solid, and they are also the defaults for ESP-IDF.

lbernstone:
Those static buffers should be 1K each. In practice, I see a difference of 3120 bytes between the two configs.

me-no-dev:
alright... I'm merging this to include it in the new build

PR #2914, opened 2019-06-21
Fix WebServer hang on client disconnect during upload

romansavrulin:
This tiny PR fixes endless loop inside WebServer library if file upload is interrupted by the browser


me-no-dev:
thanks!

PR #2923, opened 2019-06-24
set connection timeout of a http client.

a-c-sreedhar-reddy:


me-no-dev:
please add some explanation in the box above. Why, what, so on :)

PsuFan:
Is WiFiClient connect default timeout 2 minutes...? Seems like this should be reduced.

https://github.com/espressif/arduino-esp32/blob/ed53b6c8d47d7a0b38c9462dd209c30652268ebf/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L38

PR #2924, opened 2019-06-24
make default connection timeout to 60 seconds

a-c-sreedhar-reddy:


boarchuz:
What does this solve?
Think of -1 as “not overridden” rather than “none”. It won’t block forever. 
The underlying stream class will still return an error when its normal timeout expires. That’s  about 20 seconds by default. If _connectTimeout is <0 (or > the stream timeout), it will have no effect. 

So 60 seconds will do nothing, and is confusing because the effective connection timeout is actually much shorter than that. I think either commit to it and make the default something usefully low (ie. < 5 secs), or leave it.

Also it’s in milliseconds. You want 60000, not 60. Same applies to #2923

a-c-sreedhar-reddy:
When the connection starts and immediately SYSTEM_EVENT_STA_LOST_IP event occurs it does not exit.

me-no-dev:
@a-c-sreedhar-reddy please read what @boarchuz said :)

PR #2948, opened 2019-07-01
ESP8266 => ESP32 tidyup, set correct default port

pfeerick:
Tidied up references to the ESP8266, and set the correct default port for the ESP32.

The latter point is the main reason for this PR - I discovered that when you run the `espota.py` tool from this repo against a ESP32 _without_ overriding the default port it would not work... but it was being overridden when run via the Arduino IDE. 

I'm hoping this is just leftovers, and can be corrected?


me-no-dev:
Thanks :)

PR #2951, opened 2019-07-02
Fix warning -Wempty-body by adding braces to 'if' stmt body

MartyMacGyver:
This resolves the following build warning when using WiFi:

```
[...]\hardware\espressif\esp32\libraries\WiFi\src\WiFiGeneric.cpp: In static member function 'static esp_err_t WiFiGenericClass::_eventCallback(void*, system_event_t*)':

[...]\hardware\espressif\esp32\libraries\WiFi\src\WiFiGeneric.cpp:336:107: warning: suggest braces around empty body in an 'if' statement [-Wempty-body]

     if(event->event_id < 26) log_d("Event: %d - %s", event->event_id, system_event_names[event->event_id]);
```


me-no-dev:
good catch!

PR #2955, opened 2019-07-02
Cleanup WiFiMulti

NathanJPhillips:
First PR to esp32 - testing the waters with a simple clean-up. Four commits, most significantly using macros instead of magic numbers for case labels but also some smaller changes.


me-no-dev:
Hey @NathanJPhillips nice :) merging

Issue #1371, opened 2018-05-01
FPSTR() incorrectly defined in pgmspace.h


bxparks:
### Hardware:
Board:							ESP32 Dev Module
Core Installation/update date:			2018-05-01
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
Upload Speed:					921600

### Description:

The `FPSTR()` macro in `./cores/esp32/pgmspace.h` is incorrectly defined. It is currently
```
#define PSTR(s)       (s)
```
But that macro should do a cast to a `(const __FlashStringHelper*)` instead.
See https://github.com/esp8266/Arduino/blob/master/cores/esp8266/WString.h#L38

The sketch below works on ESP8266 but fails on ESP32.

The solution is to remove `FPSTR()` from `pgmspace.h` and add the following to `WString.h`:
```
#define FPSTR(pstr_pointer) (reinterpret_cast<const __FlashStringHelper *>(pstr_pointer))
#define F(string_literal) (FPSTR(PSTR(string_literal)))
```
(where I have redefined `F()` in terms of `FPSTR()` as done in the ESP8266 version of this file).

I can send you a PR if you agree that this is the correct solution.

### Sketch:
```cpp
#include <Arduino.h>

#if defined(ESP32) || defined(ESP8266)
  #include <pgmspace.h>
#else
  #error Unsupported platform
#endif

const char flashString[] PROGMEM = "abcedf";
const __FlashStringHelper* const flashHelperString = FPSTR(flashString);

void setup() {
}

void loop() {
}
```

### Debug Messages:
```
Build options changed, rebuilding all
In file included from /home/brian/Downloads/Arduino/arduino-1.8.5/hardware/espressif/esp32/cores/esp32/WString.h:29:0,
                 from /home/brian/Downloads/Arduino/arduino-1.8.5/hardware/espressif/esp32/cores/esp32/Arduino.h:150,
                 from /home/brian/dev/arduino/esp32/fpstr_demo/fpstr_demo.ino:1:
/home/brian/Downloads/Arduino/arduino-1.8.5/hardware/espressif/esp32/cores/esp32/pgmspace.h:35:41: error: cannot convert 'const char*' to 'const __FlashStringHelper* const' in initialization
 #define FPSTR(p)      ((const char *)(p))
                                         ^
/home/brian/dev/arduino/esp32/fpstr_demo/fpstr_demo.ino:10:54: note: in expansion of macro 'FPSTR'
[...]

PR #2961, opened 2019-07-04
Fix incorrect definition of FPSTR() macro and move it from pgmspace.h to WString.h (#1371)

bxparks:


me-no-dev:
what have you tested this with?

bxparks:
I'm not sure what kind of testing you are looking for, but the following compiles and prints "Hello PROGMEM" on the ESP8266. It fails to compile on the ESP32:
```C++
#include <pgmspace.h>

// Cannot use F() macro outside of a function, use PROGMEM directly.
static const char HELLO_PROGMEM[] PROGMEM = "Hello PROGMEM";

void printFlashString(const __FlashStringHelper* s) {
  Serial.println(s);
}

void setup() {
  delay(1000);
  Serial.begin(115200);
  while(!Serial); // for Arduino Leonardo/Micro

  printFlashString(FPSTR(HELLO_PROGMEM));
}

void loop() {
}
```

The error is:
```
.../FlashMemory.ino: In function 'void setup()':
FlashMemory:21:35: error: cannot convert 'const char*' to 'const __FlashStringHelper*' for argument '1' to 'void printString(const __FlashStringHelper*)'
   printString(FPSTR(HELLO_PROGMEM));
                                   ^
exit status 1
```

bxparks:
If I apply this PR to my ESP32 core on my Arduino IDE 1.8.9 `.../packages/esp32/hardware/esp32/1.0.2/cores/esp32/{WString.h,prgmspace.h}`, the same program compiles, and prints the correct output on the ESP32.

me-no-dev:
OK it's not a big deal :) let's merge and see if we get any issues before we cut 1.0.3 final

PR #2964, opened 2019-07-05
new board added magicbit

Magicblocks:
new board added magicbit


me-no-dev:
this has way too many options enabled. I imagine the board has a particular hardware. Why all the flash and so on options?

Magicblocks:
Thanks all, modified pull request according to your suggestions and updated.

me-no-dev:
closing as duplicate

PR #2974, opened 2019-07-09
Fix issue with semaphores

chegewara:
Fixes issue that in many cases caused app stuck on semaphores during connection. Im testing it 2 days and it seems to solve many problems.


me-no-dev:
Thanks mate :)

romansavrulin:
@chegewara, that was our bug. Nice catch, thank you!

PR #2976, opened 2019-07-10
Added more menu settings for TinyPICO

UnexpectedMaker:
Added Flash mode and Flash speed menu settings for TinyPICO
Also set Flash mode default to QIO


me-no-dev:
merged :)

Issue #2921, opened 2019-06-23
Off-by-one Error In RMT Example


tedhenry100:
In https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/RMT/RMTWriteNeoPixel/RMTWriteNeoPixel.ino lines 64 and 65 there is an off by one error.

The shift is too much. for the first iteration of the loop `1<<(8-0)` is `100000000` which is not the desired bit mask. The desired bit mask is `1000000`. It should be `1<<(7-0)`

            for (bit=0; bit<8; bit++){
                if ( (color[col] & (1<<(8-bit))) && (led == led_index) ) {

There is a comment from @VerusLogic in the related commit about this problem.

https://github.com/espressif/arduino-esp32/commit/4e96bffe0ef0e3929ba0ecf540c9c6f299b31e14#diff-e59a7abf6f432b9f3d716a15f5f9b715

@david-cermak

david-cermak:
Hi @tedhenry100 

sorry for such a late response.
Yes, there's an off by one error, thanks for noting! Posted a [PR](https://github.com/espressif/arduino-esp32/pull/2986)

PR #2986, opened 2019-07-14
RMT: Fix in bitshift of NeoPixel example project

david-cermak:
Fixed bit shift in demo application of using RMT peripheral. Init data array to be transmitted was off by one, as for the first iteration (i==0) the mask was 1<<8, which resulted in shifted RGB value in `color[]` variable
Closes https://github.com/espressif/arduino-esp32/issues/2921


me-no-dev:
your favourite one char fixes hahahah

Issue #2798, opened 2019-05-20
Very slow connection to WiFi - almost 3.2 seconds, even if the static IP is used.


robertgregor:
Hi,
I have latest 1.0.3 rc1 installed and have simple sketch:

#include <WiFi.h>

boolean connected=false;

IPAddress getIpAddr(String ip) {
            IPAddress addr;
            int iparr[4];
            iparr[0] = ip.toInt();
            for (int i=1; i<4; i++) {
                ip.remove(0, ip.indexOf('.')+1);
                iparr[i] = ip.toInt();               
            }
            return IPAddress(iparr[0], iparr[1], iparr[2], iparr[3]);
}

void connectToWifiNoWait(String ssid, String pwd, String ip, String gw, String mask, String dns1, String dns2, String bssid, String chnl ) {
  if (WiFi.status() == WL_CONNECTED) return;
  Serial.println(PSTR("Connecting wifi..."));
  WiFi.persistent(true); 
  WiFi.setAutoConnect(true);
  WiFi.setAutoReconnect(true);
  if (ip.length() != 0) {
    WiFi.config(getIpAddr(ip), getIpAddr(gw), getIpAddr(mask), getIpAddr(dns1), getIpAddr(dns2));
    Serial.println(PSTR("Config called..."));
  }
  if ((ssid.length() != 0) && (pwd.length() == 0)) {
      WiFi.begin(ssid.c_str());
      Serial.println(PSTR("Begin with ssid called..."));
  } else if ((ssid.length() != 0) && (pwd.length() != 0)) {
    if (bssid.length() == 0) {
      WiFi.begin(ssid.c_str(), pwd.c_str());
      Serial.println(PSTR("Begin with ssid and password called..."));
    } else {
      byte iparr[6];
      for (int i=0; i<6; i++) {
         String part = "0x" + bssid.substring(0,bssid.indexOf(':'));
         bssid.remove(0, bssid.indexOf(':')+1);
         iparr[i] = strtol(part.c_str(),NULL,0);               
      }
      WiFi.begin(ssid.c_str(), pwd.c_str(), chnl.toInt(), iparr);
      Serial.println(PSTR("Begin with ssid, password, channel and bssid called..."));
    }
  }  
}

void setup() {
  Serial.begin(115200);
[...]

Issue #1675, opened 2018-07-25
Connect to Wifi network take almost 3 seconds. I am using static IP and BSSID and channel. On ESP8266, the same code connected within 100ms. What I am doing wrong


robertgregor:
Hi, I am migrating the code of ESP8266 for battery powerred device. But seems, I have a problem to have reasonable time to connect to wifi. It takes at least 3 secods always. Here is the code:

#include <FS.h>
#include <AsyncTCP.h>
#include <ESPAsyncWebServer.h>
#include <SPIFFS.h>

// SKETCH BEGIN
AsyncWebServer server(80);

const char* ssid = "Bob";
const char* pwd = "xxx";
byte ip[] = {192,168,1,115};
byte gw[] = {192,168,1,1};
byte mask[] = {255,255,255,0};
byte dns[] = {194,41,3,193};
byte bssid[] = {0xF8, 0xD1, 0x11, 0x24, 0xB3, 0x84};
byte chnl = 4;

IPAddress getIpAddrFromArr(byte ipAddr[]) {
    return IPAddress(ipAddr[0], ipAddr[1], ipAddr[2], ipAddr[3]);
}

void connectToWifi(String ssid, String pwd, byte ip[], byte gw[], byte mask[], byte dns[], byte bssid[], byte chnl ) {
  if (WiFi.status() == WL_CONNECTED) return;
  Serial.println(F("Connecting wifi..."));
  long st = millis();
  WiFi.enableSTA(true);
  WiFi.mode(WIFI_STA);
  if (ip[0] != 0) WiFi.config(getIpAddrFromArr(ip), getIpAddrFromArr(gw), getIpAddrFromArr(mask),getIpAddrFromArr(dns));
  if ((ssid.length() != 0) && (pwd.length() == 0)) {
      WiFi.begin(ssid.c_str());
  } else if ((ssid.length() != 0) && (pwd.length() != 0)) {
    if (bssid[0] == 0) {
      WiFi.begin(ssid.c_str(), pwd.c_str());
    } else {
      WiFi.begin(ssid.c_str(), pwd.c_str(), chnl, bssid);
    }
    WiFi.setAutoConnect(true);
  }  
  int attempts = 0;
  while (WiFi.status() != WL_CONNECTED) {
      if (++attempts < 300) {
        Serial.print(".");
        delay(50);
      } else break;
  }
  if (WiFi.status() == WL_CONNECTED) {
    long tm = millis() - st;
[...]

PR #2989, opened 2019-07-16
WPA2 connection fix (significant improvement to connection time)

boarchuz:
This seems to finally fix/workaround the bug that has been plaguing WPA2 connections.
In my testing (home network, WPA2 Personal), connection time has improved from **~1200ms to ~300-350ms**.

The issue seems to be that the underlying WiFi API needs esp_wifi_set_config() to be explicitly called. Perhaps some lwIP settings aren't applied. It might even be a clash with the ARP check config specifically.

This is not specific to Arduino, but it is a band-aid for now. Also tested with esp-idf (PlatformIO) , esp-idf + Arduino component (PlatformIO), Arduino (IDE and PlatformIO). 
I'll create an issue on the esp-idf repo soon.

Hopefully helps with #1675, #2798, etc


me-no-dev:
Nice find! Let's merge this and see how it goes :)

PR #2991, opened 2019-07-16
Make yield() overridable

dok-net:
Hi @me-no-dev!
As I've a fully portable co-op scheduler running on ESP8266, ESP32 and Arduino (physically the ATmega 328P - Arduino Pro or Pro Mini) now [https://github.com/dok-net/CoopTask], here is another request to allow nice integration into the loop() with `yield()` (and, later, optionally, `delay()`). AVR Arduino et cetera have this, because `yield()` is overridable and `delay()` cyclically calls `yield()`, but as you know, the scheduling on the Espressif SOC is different.
I just hope that this very small adjustment is agreeable to you, even if you don't think that cooperative MT is useful on FreeRTOS, but users should implement OS tasks instead...

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -144,7 +144,6 @@ unsigned long IRAM_ATTR millis()
 void delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
-    if (yield != __yield) yield();
 }
 
 void IRAM_ATTR delayMicroseconds(uint32_t us)
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -144,7 +144,7 @@ unsigned long IRAM_ATTR millis()
 void delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
-    yield();
+    if (yield != __yield) yield();
 }
 
 void IRAM_ATTR delayMicroseconds(uint32_t us)
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -141,12 +141,10 @@ unsigned long IRAM_ATTR millis()
     return (unsigned long) (esp_timer_get_time() / 1000ULL);
 }
 
-void __delay(uint32_t ms)
+void delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
 }
-
-void delay(uint32_t ms) __attribute__ ((weak, alias("__delay"))); 
 
 void IRAM_ATTR delayMicroseconds(uint32_t us)
 {
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -144,6 +144,7 @@ unsigned long IRAM_ATTR millis()
 void delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
+    yield();
 }
 
 void IRAM_ATTR delayMicroseconds(uint32_t us)
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -44,10 +44,12 @@ float temperatureRead()
     return (temprature_sens_read() - 32) / 1.8;
 }
 
-void yield()
+void __yield()
 {
     vPortYield();
 }
+
+void yield() __attribute__ ((weak, alias("__yield")));
 
 #if CONFIG_AUTOSTART_ARDUINO
 
@@ -139,10 +141,12 @@ unsigned long IRAM_ATTR millis()
     return (unsigned long) (esp_timer_get_time() / 1000ULL);
 }
 
-void delay(uint32_t ms)
+void __delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
 }
+
+void delay(uint32_t ms) __attribute__ ((weak, alias("__delay"))); 
 
 void IRAM_ATTR delayMicroseconds(uint32_t us)
 {
--- a/cores/esp32/HardwareSerial.h
+++ b/cores/esp32/HardwareSerial.h
@@ -97,6 +97,8 @@ protected:
     uart_t* _uart;
 };
 
+extern void serialEventRun(void) __attribute__((weak));
+
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
 extern HardwareSerial Serial;
 extern HardwareSerial Serial1;
--- a/cores/esp32/HardwareSerial.h
+++ b/cores/esp32/HardwareSerial.h
@@ -105,4 +105,4 @@ extern HardwareSerial Serial1;
 extern HardwareSerial Serial2;
 #endif
 
-#endif
+#endif // HardwareSerial_h
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -9,6 +9,13 @@ TaskHandle_t loopTaskHandle = NULL;
 
 bool loopTaskWDTEnabled;
 
+extern "C" void __loop_end(void)
+{
+    /* do nothing by default */
+}
+
+extern "C" void loop_end(void) __attribute__((weak, alias("__loop_end")));
+
 void loopTask(void *pvParameters)
 {
     setup();
@@ -17,6 +24,7 @@ void loopTask(void *pvParameters)
             esp_task_wdt_reset();
         }
         loop();
+        loop_end();
     }
 }
 
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -9,12 +9,10 @@ TaskHandle_t loopTaskHandle = NULL;
 
 bool loopTaskWDTEnabled;
 
-extern "C" void __loop_end(void)
-{
-    /* do nothing by default */
+extern "C" void esp_loop(void) __attribute__((weak));
+extern "C" void esp_loop(void) {
+    loop();
 }
-
-extern "C" void loop_end(void) __attribute__((weak, alias("__loop_end")));
 
 void loopTask(void *pvParameters)
 {
@@ -23,8 +21,7 @@ void loopTask(void *pvParameters)
         if(loopTaskWDTEnabled){
             esp_task_wdt_reset();
         }
-        loop();
-        loop_end();
+        esp_loop();
     }
 }
 
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -9,11 +9,6 @@ TaskHandle_t loopTaskHandle = NULL;
 
 bool loopTaskWDTEnabled;
 
-extern "C" void esp_loop(void) __attribute__((weak));
-extern "C" void esp_loop(void) {
-    loop();
-}
-
 void loopTask(void *pvParameters)
 {
     setup();
@@ -21,7 +16,8 @@ void loopTask(void *pvParameters)
         if(loopTaskWDTEnabled){
             esp_task_wdt_reset();
         }
-        esp_loop();
+        loop();
+        if (serialEventRun) serialEventRun();
     }
 }
 


me-no-dev:
I do not understand why you need those to be overwritten? Those are the proper integrations for ESP32 and FreeRTOS, no other delay is good idea.

dok-net:
Per se, the weak linkage doesn't cause additional delay - if you are separately concerned about the loop_end(), I am willing to drop that from the patch if it helps get the main portion through?

dok-net:
@me-no-dev Oh an juggling three different platforms just writing this library is crazy and full of surprises enough already. I forget that yield() is not a hook on ESP32 despite being so in "regular" Arduino, this patch includes that fix.

me-no-dev:
I fully agree that moving to threading is not easy, but this is where things are going and I find it a bad practice to try and make esp32 be esp8266. Regardless how people see ESP32, it is not an ESP8266 with more pins :) I asked for the wek linkages because they are pointless for the functions you did it for. Delay on ESP32 will always be tied to vTaskDelay for example (RTOS API), yield is mostly pointless, but is there to keep 8266 folks happy.
Arduino for 8266 will be moving to RTOS as well. Once that is done, all those things you do here will be pointless and another set of work will have to happen to bring 8266 to esp32 :)

But my real question is why do you need those weak? what is te benefit?

dok-net:
I'll gladly try to answer your question.
Making the bindings weak allows me to hook into those functions and insert additional behavior. The original code still gets called!
I can't prove it, but I am confident that there a loads of device libraries out there, that are portable in the sense of the Arduino single-thread API. Frankly, FreeRTOS doesn't look for the faint of heart, either.

In `tools/sdk/include/freertos/freertos/FreeRTOSConfig.h,` I find
`#define configUSE_PREEMPTION                    1 `
I imagine it's possible to specify a set of tasks that are non-preemptable among them, basically presenting a cooperative multithreading to each other, and it may be not that hard for someone who learns FreeRTOS to implement a user-friendly wrapper around that. Still. reusing what works on Arduino and ESP8266 seems not such a bad idea, and that's what my CoopTask does. For that, I really need the weak bindings, and I can just hope and very kindly ask you to make that tweak to the core, pleeeaase?

[CoopTask Github repo](https://github.com/dok-net/CoopTask)

dok-net:
@me-no-dev After reading the configuration options documentation and the actual settings in effect here, I find that Arduino ESP32 both preempts for the benefit of higher-priority tasks, and performs round-robin scheduling between same-priority tasks. For any so called bit banging among other things like atomic access and synchronization issues, this requires working with priorities. Please allow the weak bindings such that integration of any additional behavior at yield, delay and the loop() become available to library builders, just like on Arduino and ESP8266. It's absolutely not about making the ESP32 anything ESP8266-like.

dok-net:
@me-no-dev First, only rebased, no changes.
Second, please consider this PR from the point-of-view of Arduino users, not as an ESP8266 emulation initiative.
Third, I may not have been completely clear on this: the weak functions are not about replacing the calls into Free RTOS, but they are there to shim some **additional** code that needs to run periodically. For (Mega AVR) Arduino code, this works for yield() and delay() for nearly 7 years now [[69aead513d](https://github.com/arduino/ArduinoCore-megaavr/commit/69aead513de2f646a186eaacc40fa7c5a3ef5db2)]

me-no-dev:
My issue is that you want to alter the way that those functions will work. And I presume you think that others will use it too. As a user, I do not want those functions to do anything other than what they should do and I do not want that to change because I installed a library. It will mess up my other code.

If you need high prio task, be my guest and spin one ;) there is API for it. Does the official AVR arduino allow overwrite of those functions?

dok-net:
@me-no-dev For the "The Official Arduino AVR core", the answer is yes, it does, see above [[69aead513d](https://github.com/arduino/ArduinoCore-megaavr/commit/69aead513de2f646a186eaacc40fa7c5a3ef5db2)].
Same for [(official) Arduino Core for SAMD21 CPU, commit-ish 973ccc47d8](https://github.com/arduino/ArduinoCore-samd/commit/973ccc47d8fbad115343ea0b74e005509e9bc19b).
Most likely, etc. etc. ;-)
A minor difference, delay() is implemented in terms of a busy-loop over yield there, whereas my ESP32 patch does the appropriate thing on this platform.

About the loop() shim, AVR and SAMD cores have:

```
for (;;) {
	loop();
	if (serialEventRun) serialEventRun();
}
```

I think it's a mistake of [[69aead513d](https://github.com/arduino/ArduinoCore-megaavr/commit/69aead513de2f646a186eaacc40fa7c5a3ef5db2)] not to shim loop in addition to yield() and thus implicitly delay() as well, because of that, the main loop() depends on calls to delay() and yield() for the [(official) Arduino Scheduler library](https://www.arduino.cc/en/Reference/Scheduler) to run at all! Any sketch simply cycling, perhaps just using the synchronous delayMilliseconds(), blocks the scheduler. We can do better than that :-)

me-no-dev:
for yield I can accept. adding serialEventRun like here: https://github.com/arduino/ArduinoCore-samd/search?q=serialEventRun&unscoped_q=serialEventRun I can also accept. delay... not

atanisoft:
Co-operative scheduling doesn't work in a pre-emptive scheduled environment. Trying to shoehorn that together is an act of madness. Embrace the FreeRTOS support that the platform has and let it's scheduler manage the tasks for you!

Keep in mind that the esp32 is a dual core system and is not limited like the other environments your library may work in

dok-net:
All I am asking for is less judgement against, and more freedom for the application and library developers. There is really no need to protect them from making choices, and, well yes, shoehorning FreeRTOS down everyone's throat and calling it Arduino doesn't fit every size.
Can you please explain what I must be missing, where is the detriment in weak binding delay(), which is only relevant to the Arduino API users at all?
While obviously I am trying to make my implementation of CoopTask work better, there can be all sorts of further uses in intercepting `yield()` and `delay()` in (existing) code, the `serialEventRun()` is an indication for that.
Besides, the Arduino world is by default pinned to core 1 on the ESP32.

atanisoft:
By allowing the library to replace delay() it will lead to hard to trace watchdog timeouts for the end user. It will have other effects since the usage of vTaskDelay is critical to FreeRTOS on the esp32 in order to ensure all tasks are executing. It is not all about end user code, there are other housekeeping tasks that run as well.

serialEventRun is an interesting one, I haven't seen it in use so I can't really comment on it too much. But what I can say about it is that it looks like a nice shim for calling a function *after* loop() runs. It doesn't replace the core required functionality of the underlying system.

dok-net:
You are right, there are ways to abuse a delay() shim, but I am sure there also many other ways to implement hard to track errors already :-)
As far as I am concerned, I only need to intercept the very fact that yield() or delay() were called, in order to do the specific state change, only in ADDITION to calling the underlying yield() or delay() implementation eventually. It's not about omitting or replacing that at all.
Reading some sensors periodically, writing to a display every x seconds, handling webserver requests, and responding to button presses, doesn't require multithreading or realtime programming expertise, if it's done from (stackful) coroutines, and as long as `delay()`s don't stop the world. I have seen sketches that are a work of horror, and additionally keep blocking activity like web requests while doing serial IO or I2C, just because delaying one task and doing other activity during that time is so hard on (novice) developers.

atanisoft:
There are a little t of ways to break things, that is certain.

I'd suggest using the yield() hook alone in this case. If they call delay() and you have it in your library readme to avoid that API then it is on the end user. There also are a number of libraries that call delay() internally which could be broken by replacing the underlying delay call (if it is tiny dependent).

And you are right, some of the take you describe don't need to be run as multi-threaded and in some cases would be hindered by it (httpd). A FSM based scheduling system is perfect for some use cases. I use them in my projects running inside a FreeRTOS task with great success, a single task to implement httpd with concurrent connections as an example.

dok-net:
OK, let me see, I am arguing off the assumption, contrary to your's, that there are `delay()` calls in off-the-shelf libraries that would benefit other tasks if transparently yielding to those during the blocking period, much more so than causing trouble. Keep in mind that the Arduino Scheduler library does exactly that, so one could claim that every library is already responsible for explaining to users whether it works with the Scheduler, where it does, where it doesn't, and what to do to get the greatest benefit from including the Scheduler library. No one is forced to use it.

#### Use of delay() in libs
Using delay() in init() or begin() seems common.

##### BMP280 / BME280 libs
A delay to let the hardware catch up in some "forced" mode.

##### BH1750
In multiple places.

##### DHT
Multiple times.

##### SDS011 particulate matter sensor on serial IO
Essential, the specified message rate it 1Hz.

##### E-Paper library
Multiple places in the code.

##### HP303B
Multiple places in measurement code.

These are just the few libs that I have git clones of.

atanisoft:
delay() is definitely used in a number of libraries, which is why it is imperative not to break those libraries by replacing delay(). I'm also not certain that most of the usages of delay() in libraries is correct for all environments and most libraries out there flat out ignore error checking entirely (virtually every Wire usage is missing error checking!)

dok-net:
Yes... let me add that my own shim checks the delay() call's context first, doesn't change the behavior (except the few CPU cycles for the check) at all if not called from a "task", therefore only ever has any noticeable side-effects in code that's in the user's sphere of responsibility anyway. I am really begging now :-)

dok-net:
I've made up my mind, and I can only hope to convince you, that for an Arduino layer, which this is all about, both yield() and delay() must have a way of shimming functionality. Anyone programming against FreeRTOS isn't affected, default behavior isn't affected at all by weak bindings, suppressing this capability is not in the interest of the community, given that plain Arduino has it and has had it for many years now. Please agree.

dok-net:
From [Arduino reference on delay()](https://www.arduino.cc/reference/en/language/functions/time/delay/):

"More knowledgeable programmers usually avoid the use of delay() for timing of events longer than 10’s of milliseconds unless the Arduino sketch is very simple"

The take on this is twofold - one possible reading is that improving delay() by shimming it is a waste, because the "more knowledgeable" avoid it anyway, another reading is that making it more versatile is pretty welcome.

ESP32 specific, it seems that `xTaskAbortDelay` is not available, so one cannot break out of a delay, say, to respond to an interrupt. Is that on purpose?

dok-net:
@me-no-dev @atanisoft I'm proposing #3227 as a way for me to port CoopTask to ESP32 (FreeRTOS in general?) in more palatable way, I hope.

dok-net:
@me-no-dev @atanisoft I hope you'd be disappointed if I'd given up that easily ;-) So - I've read the docs some more, and think I've come up with a less intrusive way to do what I want to do.

If this PR is still OK (for merge) less the delay() shim, I just need one hopefully acceptable modification to delay() that I really can't do without, and that's a nod to AVR Arduino's delay() as well as ESP8266 Arduino master, both of which do something extra on delay:

```
void delay(uint32_t ms) {
    vTaskDelay(ms / portTICK_PERIOD_MS);
    yield(); // <---------------------------
}
```

The yield() after vTaskDelay() allows code to detect that a task was blocked.

Please, this is ESP32 Arduino, not the ESP32 IDF, allow some concessions.

dok-net:
Additional suggestion, I could also do:

```
void delay(uint32_t ms) {
    vTaskDelay(ms / portTICK_PERIOD_MS);
    if (yield != __yield) yield(); // <---------------------------
}
```

This prevents the unnecessary `vPortYield()` after `vTaskDelay()` if `yield()` isn't overridden.

atanisoft:
> I'm obviously not sure if `vPortYield()` in a NOP in Espressif IDF FreeRTOS on ESP32 :-)

vPortYield is not a NOP... https://github.com/espressif/esp-idf/blob/v3.2/components/freertos/portasm.S#L498

dok-net:
Great, so the function pointer comparison is justified. Commited and pushed.

dok-net:
@atanisoft @me-no-dev With what I hope is now a minor-sized PR here, I've also [ported CoopTask](https://github.com/dok-net/CoopTask/releases/tag/2.0.0-rc6) to use FreeRTOS native tasks as substrate. This PR is essential for this to work.

cores/esp32/esp32-hal-misc.c
```diff
@@ -142,6 +144,7 @@ unsigned long IRAM_ATTR millis()
 void delay(uint32_t ms)
 {
     vTaskDelay(ms / portTICK_PERIOD_MS);
+    if (yield != __yield) yield();
```
me-no-dev: uhmmm ... no! no changes to delay ;) in FreeRTOS delay does yield.

dok-net: @me-no-dev I am not even sure that using FreeRTOS tasks, that is, starting AND destroying them, is such a great idea with ESP32 Arduino, as my reading of the docs is that it takes the IDLE task to clean up destroyed tasks completely. Generally I don't see Arduino sketches letting the IDLE task run at all, unless *Yield defers to lower priority tasks?! But no one is required to use *Delay in a sketch, right?

Haven't I proved beyond reasonable doubt that AVR Arduino, as the gold standard, calls the OVERRIDABLE Arduino yield() as part of any Arduino delay()? Everyone is free to use FreeRTOS calls on ESP32, but Arduino libs and Arduino sketches don't and so they can rightfully expect Arduino compatibility, don't you think so, too?

I've learned from discussions here that there is such a thing as using Arduino-esp32 as a so called "component". There's already an incompatibility with that mode in the much lower tick frequency. So maybe, could your propose some #ifdef define for me to use around the above code change, such it will be disabled in the component mode?

I would also be quite willing to research into an even more transparent way of shimming vTaskDelay(), the issue being how to learn the xTicksToDelay value. I haven't found any.

dok-net:
Dear @me-no-dev,
I've finally found a way to do what I want without doing anything you don't want in core - is this PR now up to standards, can you merge, or is there something I'm missing?

me-no-dev:
yup :)

dok-net:
Great - CoopTask 2.2.0 works out of the box on ESP32 now.

PR #2998, opened 2019-07-17
Ble notification/indication status and timeout

romansavrulin:
Sometimes it is useful to see if our BLE indication succeed or not. Current BLE implementation doesn't provide such functionality and has potential deadlock if indication confirmation response is not delivered to ESP by any reason. 

This PR adds:
* TimedWait to FreeRTOS abstraction layer
* Indication timeout (and fix possible deadlock on semaphore)
* Indication/Notification status callback, to give precise control over messages sent to BLE peer.
* Backward codebase compatibility with previous sketches.

### DIFF
--- a/libraries/BLE/src/BLECharacteristic.cpp
+++ b/libraries/BLE/src/BLECharacteristic.cpp
@@ -480,10 +480,20 @@ void BLECharacteristic::notify(bool is_notification) {
 	assert(getService() != nullptr);
 	assert(getService()->getServer() != nullptr);
 
+	auto cb = BLECharacteristicCallbacks();
+
+	BLECharacteristicCallbacks * callback = &cb;
+	if(m_pCallbacks != nullptr){
+		callback = m_pCallbacks;
+	}
+
+	callback->onNotify(this);   // Invoke the notify callback.
+
 	GeneralUtils::hexDump((uint8_t*)m_value.getValue().data(), m_value.getValue().length());
 
 	if (getService()->getServer()->getConnectedCount() == 0) {
 		log_v("<< notify: No connected clients.");
+		callback->onStatus(this, BLECharacteristicCallbacks::Status::ERROR_NO_CLIENT, 0);
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


libraries/BLE/src/BLECharacteristic.cpp
```diff
@@ -480,10 +480,20 @@ void BLECharacteristic::notify(bool is_notification) {
 	assert(getService() != nullptr);
 	assert(getService()->getServer() != nullptr);
 
+	auto cb = BLECharacteristicCallbacks();
```
chegewara: There is no need to create additional callback, which will be created easch time you send notification/indications (memory leak).

Just use the same mechanism like with onRead/onWrite:
```
if(m_pCallback != nullptr){
     callback->onNotify(this);
}
```

romansavrulin: @chegewara I personally prefer to use [Null object pattern](https://en.wikipedia.org/wiki/Null_object_pattern) to maintain [KISS](https://en.wikipedia.org/wiki/KISS_principle) and [DRY](https://ru.wikipedia.org/wiki/Don’t_repeat_yourself)

It's not a big deal for performance or memory considerations with BLE data rates. But much simpler to read, isn't it?

chegewara: If it wont make memory leak, then ok.

romansavrulin: @chegewara It's a scoped variable, placed on stack. It is automatically deleted when we leave its outbound `{}` block ( see [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) )

atanisoft: for consistency it should be using the same approach as onRead/onWrite...

also this shouldn't leak any memory as it is allocated and freed automatically as part of the method execution.

romansavrulin: @atanisoft Did I got you right: you mean use Null object pattern in other callback checks?

atanisoft: The general approach used so far (in this and other libs) is:
```
if(m_pCallback){
     m_pCallback->onNotify(this);
}
```
I'm not certain your approach is any cleaner/readable.

romansavrulin: @atanisoft so, you mean readability of the code with extra

```
if(m_pCallback){
     m_pCallback->onNotify(this);
}
```

in dozen places with other 5-level conditionals is much cleaner than  
```
callback->onNotify(this);
```

and we should use that approach because we used to? This doesn't sound convictive to me, sorry. I'd prefer to replace other usages of that callback to decrease overall codebase and improve maintainability.

atanisoft: I'm fine either way, just be consistent in usage :)

romansavrulin: @atanisoft Thanks, Mike

PR #3003, opened 2019-07-18
Use Uart FreeRTOS timeout instead of busy loop wait

romansavrulin:
Using any timed function to read uart causes warnings in IDLE task. This happens because of busy loop wait for reading uart. We can gracefully utilize FreeRTOS timed read from UART RX queue for same purpose and let CPU use less power/cycles while reading UART


romansavrulin:
Added wrong commit in this PR. Will re-create it from scratch

PR #3004, opened 2019-07-18
HAL read/peek timeout with native freeRTOS blocking mechanism

romansavrulin:
Using any timed function to read uart causes warnings in IDLE task. This happens because of busy loop wait for reading uart. We can gracefully utilize FreeRTOS timed read from UART RX queue for same purpose and let CPU use less power/cycles while reading UART


cores/esp32/Stream.h
```diff
@@ -40,8 +40,8 @@ class Stream: public Print
 protected:
     unsigned long _timeout;      // number of milliseconds to wait for the next char before aborting timed read
     unsigned long _startMillis;  // used for timeout measurement
-    int timedRead();    // private method to read stream with timeout
-    int timedPeek();    // private method to peek stream with timeout
+    virtual int timedRead();    // private method to read stream with timeout
+    virtual int timedPeek();    // private method to peek stream with timeout
```
atanisoft: this may break compatibility with libraries.

romansavrulin: @atanisoft can you be more precise and tell how will it break things? any usage when timedRead should be called directly on Stream.h without virtual inheritance?

cores/esp32/HardwareSerial.h
```diff
@@ -62,6 +62,8 @@ class HardwareSerial: public Stream
     int availableForWrite(void);
     int peek(void);
     int read(void);
+    virtual int timedRead();    // private method to read uart with timeout
+	virtual int timedPeek();    // private method to peek uart with timeout
```
atanisoft: fix the indent

cores/esp32/Stream.h
atanisoft: Stream.h is defined by the Arduino core specification, you can see the declaration of these methods [here](https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Stream.h#L54-L55). By adding virtual to these methods it will potentially break compatibility with any library which overrides these methods without the declaration of virtual. A similar problem has been seen in https://github.com/espressif/arduino-esp32/issues/2755

cores/esp32/esp32-hal-uart.c
```diff
@@ -377,6 +376,10 @@ static void uart_on_apb_change(void * arg, apb_change_ev_t ev_type, uint32_t old
         }
         // wait TX empty
         while(uart->dev->status.txfifo_cnt || uart->dev->status.st_utx_out);
+
+        if (xHigherPriorityTaskWoken) {
+			portYIELD_FROM_ISR();
+		}
```
atanisoft: Fix the indent for this block

cores/esp32/esp32-hal-uart.c
```diff
@@ -277,26 +278,24 @@ uint32_t uartAvailableForWrite(uart_t* uart)
...
+uint8_t uartRead(uart_t* uart, uint8_t *data, size_t to)
 {
     if(uart == NULL || uart->queue == NULL) {
         return 0;
     }
-    uint8_t c;
-    if(xQueueReceive(uart->queue, &c, 0)) {
-        return c;
+    if(xQueueReceive(uart->queue, data, to)) {
+        return 1;
```
atanisoft: this alters the expected behavior of uartRead() which returns the character that was read from the uart device.

cores/esp32/esp32-hal-uart.c
```diff
@@ -277,26 +278,24 @@ uint32_t uartAvailableForWrite(uart_t* uart)
...
+uint8_t uartPeek(uart_t* uart, uint8_t *data, size_t to)
 {
     if(uart == NULL || uart->queue == NULL) {
         return 0;
     }
-    uint8_t c;
-    if(xQueuePeek(uart->queue, &c, 0)) {
-        return c;
+    if(xQueuePeek(uart->queue, data, to)) {
+        return 1;
```
atanisoft: same for this method, the uartPeek() method returns the next character available to read

cores/esp32/esp32-hal-uart.c
romansavrulin: @atanisoft mention 0 timeout in original code. It will return 0 even if there's nothing to read in uart. That's weird behavior and breaks [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) per se and should be fixed.

cores/esp32/esp32-hal-uart.c
romansavrulin: @atanisoft same here

cores/esp32/esp32-hal-uart.c
atanisoft: while it may be weird and break POLA, it is likely going to break a number of existing libraries which rely on the current method signatures and behaviors of the methods. I know I have a library which uses the UART HAL (I'll be moving it over to IDF API instead soon).

romansavrulin: how do you determine case if UART empty or not in your library?

Moving to idf API was the first solution we've considered. But it's not perfect too. If you carefully  look into uart api implementation you'll notice that it's impossible to set RX buffer size, according to example given in documentation. So, we've discarded that solution in favor to fix HAL

atanisoft: uartAvailable() is used to check for presence of data.

As for RX buffer size, that is allocated at time of initialization of the uart via IDF API

romansavrulin: @atanisoft     

So, what do you recommend? add overloaded functions?

atanisoft: You could add a new method with the updated signature with the existing one calling into the new method for compatibility. Since this is C code and not C++ I'm not sure if an overloaded method with the same name is possible.

romansavrulin: Oh, yes. It's C. I'll think how to implement this.

cores/esp32/Stream.h
romansavrulin: I cannot imagine any viable scenario about how virtual can break the things. I've got your points about UART HAL methods, and will workaround them. But it is useless if you discard this PR because of virtual methods. Let's decide if this have chance to be merged?

@me-no-dev, what's your opinion about this ?

atanisoft: I'm not against the addition of virtual to the declaration but it may have unintended breakages in libraries which extend this class

Issue #2981, opened 2019-07-11
delay() function doesn't work for less than 10ms


ducalex:
### Hardware:
Board:							Lolin32
Core Installation version:			d5fdd715efddfe5838996eeee7840223f81f52bd
IDE name:						IDF component
Flash Frequency:					40Mhz
PSRAM enabled:                                     no
Upload Speed:					115200
Computer OS:                                         Windows 8.1

### Description:
delay() doesn't work for periods smaller than one tick(10ms) (it is also rounded down if not divisible by 1 tick). 

It is a small change but I can create a pull request if it is the proper way to do things also, just tell me.

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp
Current code is:
void delay(uint32_t ms)
{
    vTaskDelay(ms / portTICK_PERIOD_MS);
}



//  My suggestion is:
void delay(uint32_t ms)
{
    uint32_t delay_end = millis() + ms;
    vTaskDelay(ms / portTICK_PERIOD_MS);
    while (delay_end > millis()) {
        NOP();
    }
}
```

stickbreaker:
@ducalex 
Why do you think `portTICK_PERIOD_MS` is 10?

Chuck.

stickbreaker:
@ducalex if you are worried the `portTICK_PERIOD_MS` will change in the future, then you could change `delay()` [current Delay coding](https://github.com/espressif/arduino-esp32/blob/7dbda4988b60f83ecfc887c7f92233b2fb3f63dd/cores/esp32/esp32-hal-misc.c#L142)
to:
```c++
void delay(uint32_t ms)
{
    vTaskDelay(ms / portTICK_PERIOD_MS);
    uint32_t remainderUS = (ms % portTICK_PERIOD_MS)*1000;
    if(remainderUS) delayMicroseconds(remainderUS);
}
```

ducalex:
> @ducalex
> Why do you think `portTICK_PERIOD_MS` is 10?
> 
> Chuck.

````
#define CONFIG_FREERTOS_HZ 100
#define configTICK_RATE_HZ				( CONFIG_FREERTOS_HZ )
#define portTICK_PERIOD_MS			( ( TickType_t ) 1000 / configTICK_RATE_HZ )\
````

vTaskDelay's argument is a tick number. Therefore it works correctly only on multiples of 10ms.

Your code would also resolve the problem indeed.

[...]

PR #3014, opened 2019-07-19
Added documentation regarding delay() resolution in esp-idf component

ducalex:
This is a documentation improvement regarding #2981

### DIFF
--- a/docs/esp-idf_component.md
+++ b/docs/esp-idf_component.md
@@ -68,6 +68,11 @@ If you are writing code that does not require Arduino to compile and you want yo
 #endif
 ```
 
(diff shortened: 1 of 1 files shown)


me-no-dev:
worth saying that Arduino is actually compiled with 1kHz setting ;)

ducalex:
But how standalone arduino-esp32 is compiled doesn't really matter to the arduino-esp32 as a component method, this is the component installation README.

I don't mind rewording the text, I'm just trying to see what improvement are you suggesting here!

ducalex:
Something like "Arduino-esp32 standalone is configured with 1khz and as such you need to configure your local project 1khz as well" ?

PR #3024, opened 2019-07-22
The progress callback is now correctly invoked with Update.write

ducalex:
The progress callback was only invoked with Update.writeStream before.


vicatcu:
@me-no-dev I can confirm this fixes the problem, would be great if it made it into 1.0.3 release.

lbernstone:
:+1: https://github.com/lbernstone/asyncUpdate

PR #3034, opened 2019-07-26
Fix BLE stop advertising not working

beegee-tokyo:
BLEAdvertising::handleGAPEvent -> ESP_GAP_BLE_ADV_STOP_COMPLETE_EVT should NOT call start()!


beegee-tokyo:
Thanks.

Issue #3006, opened 2019-07-18
Incoherent WiFiEvent() calls


jjmonteiro:
### Hardware:
```
Board: ESP32 Dev Module
Core Installation version: ?1.0.0
IDE name: Arduino IDE
--------------------------
Chip Rev: 1
Sdk Ver : v3.2-111-g524308c98
Cpu Freq: 240
Cycles  : 11074815
--------------------------
Heap Size  : 342420
Free Heap  : 316340
SPIRam Size: 0
SPIRam Free: 0
Flash Mode : 2
Flash Size : 4194304
Flash Speed: 80000000
--------------------------
Computer OS: Windows 10
```

### Description:
Calls to WiFiEvent(WiFiEvent_t event) are not making sense

### Sketch:  (leave the backquotes for [code formatting]
```
#include <WiFi.h>


void WiFiEvent(WiFiEvent_t event) {

    switch (event) {

    case SYSTEM_EVENT_WIFI_READY:
        Serial.println("SYSTEM_EVENT_WIFI_READY");
        break;

    case SYSTEM_EVENT_SCAN_DONE:
        Serial.println("SYSTEM_EVENT_SCAN_DONE");
        break;

    case SYSTEM_EVENT_STA_START:
        Serial.println("SYSTEM_EVENT_STA_START");
        break;

    case SYSTEM_EVENT_STA_STOP:
        Serial.println("SYSTEM_EVENT_STA_STOP");
        break;

    case SYSTEM_EVENT_STA_CONNECTED://or STARTED ?
        Serial.println("SYSTEM_EVENT_STA_CONNECTED");
        break;

    case SYSTEM_EVENT_STA_DISCONNECTED:
        Serial.println("SYSTEM_EVENT_STA_DISCONNECTED");
        break;

    case SYSTEM_EVENT_STA_AUTHMODE_CHANGE:
        Serial.println("SYSTEM_EVENT_STA_AUTHMODE_CHANGE");
        break;

    case SYSTEM_EVENT_STA_GOT_IP:
        Serial.println("SYSTEM_EVENT_STA_GOT_IP");
        break;

    case SYSTEM_EVENT_STA_LOST_IP:
        Serial.println("SYSTEM_EVENT_STA_LOST_IP");
        break;

    case SYSTEM_EVENT_AP_START:
        Serial.println("SYSTEM_EVENT_AP_START");
        break;

    case SYSTEM_EVENT_AP_STOP:
[...]

PR #3085, opened 2019-08-12
 Fix WiFi disconnect event not being propagated. 

edoardoo:
Relative to #3006.
A short recap:
I've just removed the true from the disconnect event callback in order to prevent the WiFi module to be stopped. This somehow prevented the disconnect event to be propagated.  
Not sure about other side effects.


me-no-dev:
I think it kinda makes sense not to restart the radio, but maybe it was there for a reason that is no longer true :)

edoardoo:
😄 yeah, I also thought that it is probably some kind of inheritance of the earliest stages of esp specifications. 
It's not much, but at the moment I'm using it without the `true` on 6 devices and they are all successfully propagating the disconnect event.

simap:
I find myself here because I'm getting an AUTH_EXPIRE every so often (from minutes to a day or so), and the esp32 is not reconnecting. Reconnect is set:
```
WiFi.setAutoReconnect(true);
```

I suspect that maybe the code that is supposed to reconnect here:
```
            WiFi.disconnect();
            WiFi.begin();
```
which was modified in this PR, isn't reconnecting because mode still has STA set. the `begin()` call calls to `WiFiGenericClass::enableSTA()`, which checks that before trying. 

I wrote a small program that would let me trigger this pair of calls, and it is unreliable. Sometimes wifi reconnects, but often it doesn't. I added some logging in `WiFiGenericClass::enableSTA()` and indeed, it is triggering the guard that checks the current mode.

When I put the `true` back in `WiFi.disconnect(true);` the pair reliably restarts WiFi. 

I am using version 1.0.4. 

I think this change should be reverted, perhaps there is some other way to allow disconnect propagation. I don't know why my WiFi is causing AUTH_EXPIRE in the first place, but it seems that in this state the code cannot be resilient to such things. Without this the ESP32 will just drop off WiFi and never come back until reset. FWIW, I have dozens of ESP8266 devices that do not have this problem on the same WiFi network.

PR #3104, opened 2019-08-17
Magicbit variant

Magicblocks:
Changed magicbit variant 
Updated BUILTIN_LED

PR #3106, opened 2019-08-17
Analog write

Magicblocks:


atanisoft:
This contains the same content as https://github.com/espressif/arduino-esp32/pull/3104. You will need to create a branch for the analogWrite pieces *PRIOR* to your boards.txt updates and submit a clean PR.

PR #3108, opened 2019-08-17
Analog write

Magicblocks:
Added analogWrite based on LedC


lbernstone:
Failing on Cmakelists.txt, but in any case, this is not using a "pwm" device, so should be a part of esp32-hal-ledc

PR #3110, opened 2019-08-18
Analog write 1

Magicblocks:
Added analogWrite functions to ledc library


me-no-dev:
for this to really work, a resource tracking is necessary for ledc, sidgmadelta and pwm. Available channels to be given based on the requirements and so on. With hardcoding channels and timers, some libs that use them will stop to work.

PR #3119, opened 2019-08-20
bump CONFIG_ESP32_WIFI_STATIC_RX_BUFFER_NUM to improve RX performance

me-no-dev:


marcmerlin:
This is the first of 2 or more CLs that took over 55KB of RAM on my chip when using ESP32 and PSRAM.
Because of this CL and subsequent ones, I am unable to further update the ESP32 tree until it reverts back to its original use from May 2019.

More options and performance are great, but they should not be default, they should be options like ESP8266 which lets you pick performance vs memory.
Filed bug: https://github.com/espressif/arduino-esp32/issues/4008

PR #3139, opened 2019-08-26
Added the pragma line to the top

jubueche:
This creates a warning instead of an error and is not ignored by the compiler.


me-no-dev:
Thanks :) Merged!

PR #3147, opened 2019-08-27
Add config, menu partition Wrover

chio7896:
Hi, we  are working with ESP-32 CAM and our aplication needs configuration for use more flash memory en with this configuration we can edit this size of memory, 

We will work a lot of teams with this project is the reason that we need this actialitation.


me-no-dev:
I agree with @atanisoft fully. I can accept you to add the missing partition schemes to Wrover, but not to alter and remove boards and other definitions. You can use the ESP32 Dev Board, enable PSRAM and have the same thing as WROVER and all partitions.

chio7896:
Sorry, I delete the boards for accident, only need the partition from Wrover.

PR #3165, opened 2019-09-01
Inline ESP::getCycleCount() to make it safe to call from ISRs

dok-net:
As the subject line says, it has been found by users (of EspSoftwareSerial), that the function is not ISR-safe right now. Inlining should efficiently fix this, IMHO.


me-no-dev:
to make it safe, mark it as IRAM ;) now it is not safe, because it has to be loaded from flash. ESP32 != ESP8266

dok-net:
Is that an observation you can explain based on generated code? What part of inline, __attribute__((always_inline)), and inline assembly still allows for flash access here, when invoked from a code that resided in IRAM? I don't get it - what's the difference to ESP8266 then?

dok-net:
@me-no-dev I've looked at the output from objdump, seems like the code in question is always inlined. Whenever it's not, I get errors from the compiler from calling getCycleCount() in ISRs. To make a long story short, this PR expresses what the compiler optimizations affect right now anyway, and that's a good thing for the human reader, isn't it?

me-no-dev:
And how about you mark it as IRAM and make it easily understandable that it can be run in ISR?

dok-net:
Done.

PR #3175, opened 2019-09-04
Fix long Ticker period conversion: cast ms to uint64_t when calculating us

full-stack-ex:
Long duration was truncated when calculating milliseconds * 1000. ESP32 only: functions esp_timer_start_* require microseconds as uint64_t. ESP8266 is not affected as it uses a different API.

### DIFF
--- a/libraries/Ticker/src/Ticker.cpp
+++ b/libraries/Ticker/src/Ticker.cpp
@@ -43,9 +43,9 @@ void Ticker::_attach_ms(uint32_t milliseconds, bool repeat, callback_with_arg_t 
   }
   esp_timer_create(&_timerConfig, &_timer);
   if (repeat) {
-    esp_timer_start_periodic(_timer, milliseconds * 1000);
+    esp_timer_start_periodic(_timer, ((uint64_t)milliseconds) * 1000);
   } else {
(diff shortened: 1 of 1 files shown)


libraries/Ticker/src/Ticker.cpp
```diff
@@ -43,9 +43,9 @@ void Ticker::_attach_ms(uint32_t milliseconds, bool repeat, callback_with_arg_t
   }
   esp_timer_create(&_timerConfig, &_timer);
   if (repeat) {
-    esp_timer_start_periodic(_timer, milliseconds * 1000);
+    esp_timer_start_periodic(_timer, ((uint64_t)milliseconds) * 1000);
```
atanisoft: Add "ULL" to the 1000 as well to also force cast there. This is also done in the millis() function as seen [here](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-misc.c#L137-L140).

full-stack-ex: The 1000ULL added. Prior change (the uint64_t cast) reverted as redundant. Tested: Ticker used to fire after 900+ seconds, now it hasn't fired for 1342s, and counting.

Issue #3181, opened 2019-09-05
Print.printf() always runs vsnprintf twice 


knifter:
It seems to me that Print.printf(fmt, ...) is allocating a local buffer but never actually uses it, always running vsnprintf() twice. A buffer of size 64 is allocated but then vsnprintf is called with (NULL, 0, ...) parameters just to calculate the length. Then it always allocates a new buffer to vsnprintf into. I've modified the code in such a way that loc_buf[64] is used at first and if it proofs insufficient, only then allocate another. I think this was the original idea.

Also: vsnprintf() returns int, and negative int for an error. I guess this may wreck havoc in various ways when that gets assigned to size_t (uint) and becomes some (large) positive number.

Have a look at this diff for the problem and the fix:

    diff --git a/cores/esp32/Print.cpp b/cores/esp32/Print.cpp
    index 8c29534..57d725f 100644
    --- a/cores/esp32/Print.cpp
    +++ b/cores/esp32/Print.cpp
    @@ -52,15 +52,19 @@ size_t Print::printf(const char *format, ...)
         va_list copy;
         va_start(arg, format);
         va_copy(copy, arg);
    -    size_t len = vsnprintf(NULL, 0, format, copy);
    +    int len = vsnprintf(temp, sizeof(loc_buf), format, copy);
         va_end(copy);
    +    if(len < 0) {
    +        va_end(arg);
    +        return 0;
    +    };
         if(len >= sizeof(loc_buf)){
             temp = new char[len+1];
             if(temp == NULL) {
                 return 0;
             }
    +        len = vsnprintf(temp, len+1, format, arg);
         }
    -    len = vsnprintf(temp, len+1, format, arg);
         write((uint8_t*)temp, len);
         va_end(arg);
         if(len >= sizeof(loc_buf)){

atanisoft:
@knifter Please send a PR with this change.

stickbreaker:
@knifter I think you also need to call `va_end(arg)` in the allocation failure exit.
```c++
     if(len >= sizeof(loc_buf)){
         temp = new char[len+1];
         if(temp == NULL) {
+++         va_end(arg);
            return 0;
         }
```

Chuck.

stickbreaker:
one more stupid question:  I am questioning the use of `delete[]` in the original code.  I think the array delete `delete[]` should not be used.  Just the standard `delete`;

```c++
  temp = new char[len+1];
...
  delete[] temp;
```
[...]

PR #3184, opened 2019-09-06
#3181 printf double vsnprintf() fix, malloc, va_end

knifter:
These are the changes mentioned in #3181


cores/esp32/Print.cpp
```diff
@@ -52,19 +52,24 @@ size_t Print::printf(const char *format, ...)
...
-        temp = new char[len+1];
+        temp = (char*) malloc(len+1);
         if(temp == NULL) {
+            va_end(arg);
             return 0;
         }
+        len = vsnprintf(temp, len+1, format, arg);
     }
-    len = vsnprintf(temp, len+1, format, arg);
     write((uint8_t*)temp, len);
```
stickbreaker: I've just thought of something, This print function is used for many different subsystems. I think some of them may actually have cases where `write()` can fail. instead of returning the number of characters to be written, actually return the true number.
```c++
int actualLen = write((uint8_t*)temp,len);
...
return actualLen;
```

knifter: Ill add/fix that

cores/esp32/Print.cpp
```diff
@@ -66,12 +66,11 @@ size_t Print::printf(const char *format, ...)
...
         len = vsnprintf(temp, len+1, format, arg);
     }
-    write((uint8_t*)temp, len);
     va_end(arg);
-    if(len >= sizeof(loc_buf)) {
+    if(temp != loc_buf){
         free(temp);
     }
-    return len;
+    return write((uint8_t*)temp, len);
```
stickbreaker: don't do it this way, the buffer has already been freed, `free()` will overwrite the content of `temp` with the delete chain links.

knifter: oops... absolutely right, a sec

stickbreaker: you could:
```c++
if(temp != loc_buf){
     len = write((uint8_t*)temp,len);  
     free(temp);
}
return len;
 ```

Comment (knifter) [reply cores/esp32/Print.cpp]: I also want to write(temp) if temp == loc_buf

**Comment by stickbreaker** [comment]
@knifter Found one more at line 162:
```c++
size_t Print::print(struct tm * timeinfo, const char * format)
{
    const char * f = format;
    if(!f){
        f = "%c";
    }
    char buf[64];
    size_t written = strftime(buf, 64, f, timeinfo);
    print(buf);
    return written;
}
```
Instead of `return written;` change it to `return print(buf);`  
That is the only other function that was not returning actual characters written.

Thanks for doing this. Every little bit helps!

me-no-dev:
@stickbreaker that is unrelated though :)

PR #3188, opened 2019-09-08
Bugfix/detect baudrate

Jeroen88:
Expose uartStartDetectBaudrate(uart_t *) in esp32-hal-uart.h and call it from HardwareSerial::begin() if baudrate detection is requested (by passing a baudrate of 0) to solve baudrate detection problems.


cores/esp32/esp32-hal-uart.c
```diff
@@ -401,7 +401,12 @@ uint32_t uartGetBaudRate(uart_t* uart)
     if(uart == NULL) {
         return 0;
     }
+
     uint32_t clk_div = (uart->dev->clk_div.div_int << 4) | (uart->dev->clk_div.div_frag & 0x0F);
+    if(!clk_div) {
```
me-no-dev: in what case is `clk_div` == 0?

Jeroen88: I do not know, but I got a division by zero error and stack trace pointed to uartGetBaudRate(). It was a 'random' error, not reproducible...

me-no-dev: Interesting... ok :)

me-no-dev:
After I merge one request, will update the branch and merge :)

Jeroen88:
@me-no-dev Thnx :)

PR #3190, opened 2019-09-08
Add support for WiFi long range mode

me-no-dev:


asetyde:
This feature can work only between Espressif devices ? or also other router ? we are interested in more range ( no problem on lower bandwidth )

beegee-tokyo:
I played with Espressifs Longe-Range WiFi before. None of my phones or my laptop could see the ESP32 AP I build. But I could connect from another ESP32 in STA mode. ESP8266 didn't see the AP either. 

So I guess it's ESP32 to ESP32 only.

asetyde:
it can be useful a ESP8266 ROUTER project, porting to ESP32U , using a big antenna, it can be an idea ... we ve many problems with router SO, slow, many difference between models, bug , user manumission

asetyde:
Hi, about this using mesh on this arduino esp32 library need to activate long range supporto on WIFI ?

PR #3197, opened 2019-09-11
Added timeout to WiFiScan.cpp to prevent getting stuck at WIFI_SCAN_RUNNING

EricMc1289:
If you start a scan for networks if the esp is connecting to an ap the scan request is aborted by the sdk without returning an error. See here espressif/esp-idf#3044 for reference.

In my case the scan routine is running in its own thread so i can't assume when the scanNetworks is triggered. If it gets triggered while the esp is connecting the connect has priority and the SYSTEM_EVENT_SCAN_DONE event is never triggered. In that particular case the WIFI_SCANNING_BIT is never cleared and you're not able to scan for networks without a reset.

To prevent that case i modified the WiFiScan.ccp to include a timeout mechanism to reset a failed scan. So I added a define SCAN_ERROR_DELAY with a value of 10000 for 10s as timeout and the parameter _scanStarted.

If the SYSTEM_EVENT_SCAN_DONE event is not fired within the SCAN_ERROR_DELAY the scanComplete method returns WIFI_SCAN_FAILED to indicate that the scan failed.

### DIFF
--- a/libraries/WiFi/src/WiFiScan.cpp
+++ b/libraries/WiFi/src/WiFiScan.cpp
@@ -22,7 +22,6 @@
 
  */
 
-#define SCAN_ERROR_DELAY 10000 //See scan as failed after 10s of scan time
 
 #include "WiFi.h"
 #include "WiFiGeneric.h"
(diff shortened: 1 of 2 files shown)


me-no-dev:
please fix tabs and spaces and also make the timeout be connected to `max_ms_per_chan * 20`.

PR #3210, opened 2019-09-15
Fix for issue_3209.

william-ferguson-au:
#include "esp32-hal-log.h" is mandatory is order to build BLEDevice.cpp.
It can't be left up to a compiler variable.

### DIFF
--- a/libraries/BLE/src/BLEDevice.cpp
+++ b/libraries/BLE/src/BLEDevice.cpp
@@ -32,13 +32,7 @@
 #include "esp32-hal-bt.h"
 #endif
 
-#if defined(CONFIG_ARDUHAL_ESP_LOG)
 #include "esp32-hal-log.h"
-
-#else
-#include "esp_log.h"
-static const char* LOG_TAG = "BLEDevice";
-#endif
 
 
 /**


me-no-dev:
leftovers from when this library was hosted elsewhere :) it was used so it can compile in IDF without Arduino. Just removing it can cause other issues, like if you use Arduino as IDF component and have not selected to forward IDF errors to Arduino's logging. Then you will get compile error for undefined ESP_LOGx

william-ferguson-au:
So what's the solution then?

me-no-dev:
do as you have done for all files in the library and change `ESP_LOGx(TAG, ` to `log_x(`

me-no-dev:
that will force the library to use Arduino's logging and since it's a part of it, that is fine :)

me-no-dev:
what arduino libraries use `ESP_LOGx(Tag,` ? arduino libs use `log_x(`

me-no-dev:
ESP_LOGx is IDF API that is not controllable through the Arduino IDE and API. Once included IDF libs are precompiled, that is what it is for them. Therefore Arduino uses log_x as logging API and that is controllable through the IDE.

william-ferguson-au:
My mistake.

william-ferguson-au:
OK, reopening because I believe I was correct and have run into this again trying to upgrade from 1.0.3 to master. Our conversation became derailed at this comment https://github.com/espressif/arduino-esp32/pull/3210#issuecomment-532091661.

BLEDevice.cpp is failing to compile because it **can't** find the log_x() functions.

It can't find them because they are declared in esp32-hal-log.h and unlike all the other BLE sources which always include esp32-hal-log.h, BLEDevice.cpp only includes esp32-hal-log.h when the compiler var CONFIG_ARDUHAL_ESP_LOG is defined.

BLEDevice.spp should always include esp32-hal-log.h instead of omitting it if CONFIG_ARDUHAL_ESP_LOG is not defined.

me-no-dev:
so... you made one change to one file, because you had issues with that file. I asked you to make the same change to all files and to replace any occurrence of ESP_LOGx with log_x. If you are not willing to do it, just say so and I'll do the whole thing myself. As-is, I will not merge just that change.

william-ferguson-au:
Sorry, there seems to be some confusion here.
I changed BLEDevice.cpp because it was the only BLE file whose header had not been changed when Bascy refactored all the BLE files to switch from ESP_LOGX to log_x (16-APR-2019).

As far as I am aware there aren't any arduino-esp32 files that are using ESP_LOGX.
So the only change that needs to be made is the one that I have changed.

me-no-dev:
yup confusion :)

PR #3216, opened 2019-09-16
Fix HTTP Client with SSL

me-no-dev:


Jeroen88:
@me-no-dev Why so difficult? Why where the ::connect() methods with timeout introduced in the first place? Wouldn't it be far better to call the parent class WiFiClient::setTimeout() method?

I did not have time to check it, but I think there is a nasty error in start_ssl_client() that might be the root cause of the problems. In [this](https://github.com/espressif/arduino-esp32/blob/06a399b84aaeb99a80588582c621e946aba58b11/libraries/WiFiClientSecure/src/ssl_client.cpp#L82) line and the next, not a pointer and a sizeof() an _int_ should be passed, but a pointer and a sizeof a _struct timeval_. I think because the first element of the struct is a timeout in seconds, the error was never discovered...

``
    struct timeval tv;
    tv.tv_sec = 0;
    tv.tv_usec = timeout * 1000;
    if(lwip_setsockopt(SO_RCVTIMEO, &tv, sizeof(struct timeval)) < 0) {
        return -1;
    }
    if(lwip_setsockopt(SO_SNDTIMEO, &tv, sizeof(struct timeval)) < 0) {
        return -1;
    }
I
``

In the above code a timeout in milliseconds is set, but, since WiFiClient::setTimeout() is in seconds, I strongly believe (just because this method existed earlier than ::setHandshakeTimeout(), which is in milliseconds) that the connect() timeout parameter should be in seconds. But, even better: remove these methods and call the parent class setTimeout()

Also, because the timeouts are messy between WiFiClient, HTTPClient and the socket timeout that the timeout parameters should be renamed to show in what unit they are, so not _int32_t timeout_ but _int32_t timeout_milliseconds_ or something like that.

Also I think that [this](https://github.com/espressif/arduino-esp32/blob/06a399b84aaeb99a80588582c621e946aba58b11/libraries/WiFiClientSecure/src/ssl_client.cpp#L84) line could also better be replaced by the parent class WiFiClient::setNoDelay()

I did not test any of this, I did not even try to code it, so maybe I am completely wrong...

me-no-dev:
this will do for now :) rewrite of both clients is in order ;) Also is integrating with the official Arduino API. Timeouts are good ;) we need them but they can be implemented better

PR #3220, opened 2019-09-17
Fix F_Fat::format return false when succeed

luc-github:
this is due to usage of `esp_err_t result;` instead of boolean, ESP_OK = 0, so it is false


me-no-dev:
@lbernstone want to have a look?

lbernstone:
Looks good to me.

luc-github:
@lbernstone thank you ^_^

PR #3227, opened 2019-09-18
Abortable Arduino delay()

dok-net:
Currently, there is (seems to be?) no way to break out of the Arduino `delay()` equivalent implementation on ESP32 Arduino core.
This PR proposes a different FreeRTOS method of implementing task delay, that allows resuming the loop() task (or any other task using Arduino-style sketches with `delay()`).
Measured timings for very short delay values, that is 0, 1, 2µs, are identical to the replaced code.

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -54,6 +54,12 @@ void yield()
 extern TaskHandle_t loopTaskHandle;
 extern bool loopTaskWDTEnabled;
 
+void IRAM_ATTR abortLoopDelay(){
+    if(loopTaskHandle != NULL){
+        vTaskNotifyGiveFromISR(loopTaskHandle, NULL);
+    }
+}
+
 void enableLoopWDT(){
     if(loopTaskHandle != NULL){
         if(esp_task_wdt_add(loopTaskHandle) != ESP_OK){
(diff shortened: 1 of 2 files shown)


cores/esp32/esp32-hal-misc.c
```diff
@@ -54,6 +54,12 @@ void yield()
 extern TaskHandle_t loopTaskHandle;
 extern bool loopTaskWDTEnabled;
 
+void IRAM_ATTR abortLoopDelay(){
+    if(loopTaskHandle != NULL){
+        vTaskNotifyGiveFromISR(loopTaskHandle, NULL);
```
atanisoft: This is an ISR function and is not usable outside the ISR context, you should check the caller context to see if it needs ISR calling or can use the non-ISR call. You can do this via:
```
if (xPortInIsrContext()) {
  vTaskNotifyGiveFromISR(....);
} else {
  vTaskNotifyGive(....);
}
```

dok-net: The API documentation isn't specific on this, thanks for the hint, done.

me-no-dev:
Why do you want to abort delay that you called yourself? If you need to cancel a delay, then you should not use delay to begin with. Will not agree with this PR. Sorry

dok-net:
Verdict heard :-)

PR #3240, opened 2019-09-21
Changed WeMosBat.name

FraserLee:
The old version would crash the Stino plugin, which appears not to support quotes in board names


me-no-dev:
thanks :) nice find!

PR #3245, opened 2019-09-22
Fixed FFat::end. Fixes #3244

### LINKED ISSUE #3244: After FFat.end(), subsequent calls to FFat.begin() fail
### Hardware:
Board:							Wemos Lolin32
Core Installation version:			1.0.3
IDE name:						VSCode / Arduino IDE
Flash Frequency:					80MHz
PSRAM enabled:                                     n/a
Upload Speed:					921600
Computer OS:                                         Windows 10

### Description:
After calling FFat.begin(), doing some file work and then FFat.end(), subsequent calls to FFat.begin() always fail.

### Way to reproduce:
Open the included FFat_Test example. Confirm it works (don't forget to use a partition scheme with a FAT partition). Then, on setup() around line 163, replace the following block: 
```cpp
    if(!FFat.begin()){
        Serial.println("FFat Mount Failed");
        return;
    }
    Serial.printf("Total space: %10lu\n", FFat.totalBytes());
    Serial.printf("Free space: %10lu\n", FFat.freeBytes());
    listDir(FFat, "/", 0);
    writeFile(FFat, "/hello.txt", "Hello ");
```
with this one:
```cpp
    if(!FFat.begin()){
        Serial.println("FFat Mount Failed");
        return;
    }
    Serial.printf("Total space: %10lu\n", FFat.totalBytes());
    Serial.printf("Free space: %10lu\n", FFat.freeBytes());
    listDir(FFat, "/", 0);
    FFat.end();
    delay(5000);
    if(!FFat.begin()){
        Serial.println("FFat Mount Failed");
        return;
    }
    writeFile(FFat, "/hello.txt", "Hello ");
```
The second FFat.begin() returns false. and "FFAT Mount Failed" prints on serial. If debug is enabled, the following message appears.

### Debug Messages:
```
[E][FFat.cpp:56] begin(): Mounting FFat partition failed! Error: -1
```

lbernstone:
_wl_handle was not being cleared properly

### DIFF
--- a/libraries/FFat/src/FFat.cpp
+++ b/libraries/FFat/src/FFat.cpp
@@ -81,7 +81,7 @@ bool F_Fat::format(bool full_wipe, char* partitionLabel)
 {
     esp_err_t result;
     bool res = true;
-    if(_wl_handle){
+    if(_wl_handle != WL_INVALID_HANDLE){
         log_w("Already Mounted!");
         return false;
     }
--- a/libraries/FFat/src/FFat.cpp
+++ b/libraries/FFat/src/FFat.cpp
@@ -40,16 +40,20 @@ const esp_partition_t *check_ffat_partition(const char* label)
 
 bool F_Fat::begin(bool formatOnFail, const char * basePath, uint8_t maxOpenFiles, const char * partitionLabel)
 {
-    if(_wl_handle){
+    if(_wl_handle != WL_INVALID_HANDLE){
         log_w("Already Mounted!");
         return true;
     }     
 
-    if (!check_ffat_partition(partitionLabel)) return false;
+    if (!check_ffat_partition(partitionLabel)){
+        log_e("No fat partition found on flash");
+        return false;
+    }
 
     esp_vfs_fat_mount_config_t conf = {
       .format_if_mount_failed = formatOnFail,
-      .max_files = maxOpenFiles
+      .max_files = maxOpenFiles,
+      .allocation_unit_size = CONFIG_WL_SECTOR_SIZE
     };
     esp_err_t err = esp_vfs_fat_spiflash_mount(basePath, partitionLabel, &conf, &_wl_handle);
     if(err){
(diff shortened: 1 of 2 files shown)


libraries/FFat/src/FFat.cpp
```diff
@@ -40,16 +40,20 @@ const esp_partition_t *check_ffat_partition(const char* label)
 
 bool F_Fat::begin(bool formatOnFail, const char * basePath, uint8_t maxOpenFiles, const char * partitionLabel)
 {
-    if(_wl_handle){
+    if(_wl_handle != WL_INVALID_HANDLE){
         log_w("Already Mounted!");
         return true;
```
stickbreaker: Ok, this case is if the Object has already been opened, Is there any way to verify that the `basePath` points to the previous open?  Also, does it point to the same `partitionLabel`?

I can see where someone would use `begin()` to change the current mount point/partition.  

Would you actually force a dismount before processing the mount?  

What happens to a fileObject when you dismount the file system?

Chuck.

lbernstone: Forcing an end before another begin certainly takes care of closing any opened files, and seems pretty intuitive.  Unmount is handled by the fatfs unmount.  I expect that this would do unpleasant things to dependent objects. Should clearing the mountpoint signal FS(impl) to close all files?

stickbreaker: I can see a couple of paths:
* Flush/close all open files.  Cause every access to orphaned fileObject to return an error.  Make it painful for someone to make this mistake.
* Fail the `begin()` if the mount point is already created.

I can see people going to light sleep first dismounting (to reduce current consumption..) then re-mounting after wakeup.  How does deep Sleep affect it?  It seems to me the Deep Sleep is just a program controlled power off with a power on reset on wakeup?  If so, then Deep Sleep is equivalent to uncontrolled shutdown with a potential dirty volume on wakeup?

How much effort are you willing to make it "foolproof"?  I think you should use the second option.  Make a call to `begin()` with an active mount a fatal error.  Unmount(protect filesystem) then return fail.

me-no-dev: one of you please ping me after you have agreed :)

mouridis: btw @lbernstone, it seems this fix somehow breaks FFat.format(). I downloaded the revised FFat library files from your fork to replace the files in my local repo. The fix indeed solves #3244. But it seems to render FFat.format useless. It returns no error, but it doesn't erase the FFAT partition. Please check. You can use the library example.

lbernstone:
I'll work on a way to do a second begin & cleaning up the file objects this weekend in a separate PR.  Go ahead and merge this if there are no more issues.

mouridis:
@lbernstone Did you check my comment above about this PR breaking FFat.format()?

lbernstone:
@mouridis Yes, I put in another [commit](https://github.com/espressif/arduino-esp32/pull/3245/commits/1311f94e116add8491660c165b1d9753258afabe) that fixed format.

PR #3252, opened 2019-09-23
Added board definition for VintLabs boards

pvint:
I've had a lot of users confused as to which board to select for VintLabs devkits, so I'd like to add the board definition (which also allows using PWMx variables to reference PWM driver outputs on the PWM driver kit)

Thanks!


me-no-dev:
done :)

PR #3255, opened 2019-09-23
Add partition schemes to adafruit featheresp32

tpitman:


boards.txt
```diff
@@ -1654,6 +1654,17 @@ featheresp32.menu.DebugLevel.debug.build.code_debug=4
 featheresp32.menu.DebugLevel.verbose=Verbose
 featheresp32.menu.DebugLevel.verbose.build.code_debug=5
 
+featheresp32.menu.PartitionScheme.default=Default
+featheresp32.menu.PartitionScheme.default.build.partitions=default
+featheresp32.menu.PartitionScheme.minimal=Minimal (2MB FLASH)
```
me-no-dev: is this one necessary? feathers are all 4MB or above, right?

tpitman: Not necessary. I wonder if there is still value in keeping it to simulate other devices? I don't feel strongly either way.

me-no-dev: i would suggest to remove it :) it's kinda pointless

tpitman: I am new to this. How do I make that change?

atanisoft: > I am new to this. How do I make that change?

go here: https://github.com/tpitman/arduino-esp32/tree/patch-1 and modify the file directly (no need to create a PR against that branch as it is local to your own tree). Once you modify that file on that tree it will reflect in this PR.

tpitman: Thanks. Figured it out just after posting comment.

You are very patient with noobs. Thanks.

me-no-dev: do not worry that build fails. cause is external problem :)

tpitman:
Interesting that it passed yesterday. Did something change over night?

boards.txt
tpitman: Interesting that it passed yesterday. Did something change over night?

atanisoft: > Interesting that it passed yesterday. Did something change over night?

SSL cert expiration for dl.espressif.com. https://github.com/espressif/arduino-esp32/issues/3259

tpitman: Timing everything.

PR #3267, opened 2019-09-25
Non-functional deinit/init in BLEDevice.cpp

billprozac:
Pretty sure this was a typo as the deinit doesn't actually allow for reinit if that is an ifndef.  Changed in my local copy and can now deinit and reinit just fine.  Also, not sure why we are checking for the architecture here.  Just curious.


me-no-dev:
yup :) typo!

billprozac:
thanks, boss!

PR #3276, opened 2019-09-26
Fix UUID length should be 36

tanakamasayuki:
1.0.2 > 0000180d-0000-1000-8000-00805f9b34fb
1.0.3 > 180d-0000-1000-8000-00805f9b34fb


me-no-dev:
uhmmm it was wrong in 1.0.2 ;) now it' ok

PR #3277, opened 2019-09-26
Add Added unimplemented getCharacteristicsByHandle function

tanakamasayuki:
Defined in BLERemoteService.h
But, Not included in BLERemoteService.cpp


libraries/BLE/src/BLERemoteService.cpp
```diff
@@ -226,6 +226,22 @@ std::map<std::string, BLERemoteCharacteristic*>* BLERemoteService::getCharacteri
 	return &m_characteristicMap;
 } // getCharacteristics
 
+/**
+ * @brief Retrieve a map of all the characteristics of this service.
+ * @return A map of all the characteristics of this service.
+ */
+std::map<uint16_t, BLERemoteCharacteristic*>* BLERemoteService::getCharacteristicsByHandle() {
+	log_v(">> getCharacteristicsByHandle() for service: %s", getUUID().toString().c_str());
```
me-no-dev: log already prints the method name.

tanakamasayuki:
# 1.0.3\libraries\BLE\src\BLERemoteService.h
```c
	std::map<std::string, BLERemoteCharacteristic*>* getCharacteristics();
	std::map<uint16_t, BLERemoteCharacteristic*>* getCharacteristicsByHandle();  // Get the characteristics map.
```
getCharacteristicsByHandle is defined.
But, Not available.

```c
sketch\BLE_clientEx.ino.cpp.o:(.literal._Z15connectToServerv+0x24): undefined reference to `BLERemoteService::getCharacteristicsByHandle()'
```

Therefore, it was created with reference to the getCharacteristics function.

```c
/**
 * @brief Retrieve a map of all the characteristics of this service.
 * @return A map of all the characteristics of this service.
 */
std::map<std::string, BLERemoteCharacteristic*>* BLERemoteService::getCharacteristics() {
	log_v(">> getCharacteristics() for service: %s", getUUID().toString().c_str());
	// If is possible that we have not read the characteristics associated with the service so do that
	// now.  The request to retrieve the characteristics by calling "retrieveCharacteristics" is a blocking
	// call and does not return until all the characteristics are available.
	if (!m_haveCharacteristics) {
		retrieveCharacteristics();
	}
	log_v("<< getCharacteristics() for service: %s", getUUID().toString().c_str());
	return &m_characteristicMap;
} // getCharacteristics
```

getCharacteristics to getCharacteristicsByHandle
m_characteristicMap to m_characteristicMapByHandle

libraries/BLE/src/BLERemoteService.cpp
me-no-dev: please remove `">> getCharacteristicsByHandle() for ` from the log lines

Issue #3288, opened 2019-09-27
Proposed fix for AsyncUDP and AsyncUDPPacket


ssilverman:
Referencing https://github.com/espressif/arduino-esp32/issues/3287#issuecomment-535727588.

I'll get a pull request done at some point soon, but what do people think of my proposed fix in that comment? It covers both the pass-as-reference and pass-as-copy cases for implementing a `AuPacketHandlerFunction`. If there is agreement, I'll create a pull request for it.

To summarize the fix:
1. Add a call to `pbuf_ref` in the `AsyncUDPPacket` copy constructor, and
2. Call `pbuf_free` regardless (in `AsyncUDP::_recv`), outside the if and not in an else.

Here are the steps to show why I think my fix works:
1. The reference count starts at 1, so `pbuf_free` needs to be called at least once.
2. Case 1: Pass-as-reference version.
   a. `packet` gets created and the ref count is now 2.
   b. The packet is handled, then the `packet` destructor gets called and the ref count is back to 1.
   c. `pbuf_free` is called regardless. The ref count is zero.
3. Case 2: Pass-as-copy version.
   a. `packet` gets created and the ref count is now 2.
   b. The copy is created and the ref count is now 3.
   c. When the handler ends, both the copy and `packet` destructors get called and the ref count goes back to 1.
   d. `pbuf_free` is called regardless. The ref count is zero.

ssilverman:
See #3290

ssilverman:
Another thought: we could also delete the copy constructor and force use of the reference form of the callback.

Issue #3287, opened 2019-09-26
v1.0.3: Network stack never recovers without a hard reset after receiving a fixed number of UDP packets


ssilverman:
This is the same bug as #2899, but for v1.0.3.

### Hardware:
Board: Adafruit ESP32 Feather
Core Installation version: Core v1.0.3 in Arduino 1.8.10
IDE name: Arduino v1.8.10
Flash Frequency: 80MHz
PSRAM enabled: Don't know
Upload Speed: 921600
Computer OS: Mac OSX v10.14.6

### Description:
I refer the reader to the main description in #2899, but in summary, here are two programs for testing.

To continuously send closely-spaced packets (in bash):
```bash
while true; do echo -n $(printf '.%.0s' {1..1400}) > /dev/udp/192.168.1.9/8000; sleep 0.05; done
```
[Change the IP address to your board's IP address.]

You'll notice that after about 32 packets, no more packets are received.

### Sketch:
```cpp
#include <AsyncUDP.h>
#include <Esp.h>
#include <WiFi.h>

constexpr char kAPName[] = "ChangeMe";
constexpr char kAPPassword[] = "ChangeMe";
constexpr bool isSoftAP = false;  // Change to true for SoftAP mode

AsyncUDP udp;

void setup() {
  Serial.begin(115200);
  while (!Serial && millis() < 4000) {
    // Wait for Serial
  }
  Serial.println("Starting.");

  if (isSoftAP) {
    Serial.println("Starting SoftAP...");
    if (WiFi.softAP(kAPName, kAPPassword)) {
      Serial.print("    IP: ");
      Serial.println(WiFi.softAPIP());
    } else {
      Serial.println("ERROR: Starting SoftAP!");
    }
  } else {
    if (WiFi.begin(kAPName, kAPPassword)) {
      while (!WiFi.isConnected()) {
        delay(500);
      }
      Serial.print("    IP: ");
      Serial.println(WiFi.localIP());
      Serial.print("    Subnet: ");
      Serial.println(WiFi.subnetMask());
      Serial.print("    Gateway: ");
      Serial.println(WiFi.gatewayIP());
    } else {
      Serial.println("    ERROR: Connecting to AP!");
    }
  }

  if (!udp.listen(8000)) {
    Serial.println("ERROR: Starting UDP server!");
  }
  udp.onPacket(onPacket);
}

int counter = 0;

void onPacket(AsyncUDPPacket &packet) {
[...]

PR #3290, opened 2019-09-27
Fix for AsyncUDP's packet handling callback

ssilverman:
This is a fix for both pass-as-reference and pass-as-copy versions. See https://github.com/espressif/arduino-esp32/issues/3287 (original bug) and https://github.com/espressif/arduino-esp32/issues/3288 (reasoning).


ssilverman:
The copy constructor needs to call a function, hence its long-form version. :/

libraries/AsyncUDP/src/AsyncUDP.cpp
```diff
@@ -682,9 +698,8 @@ void AsyncUDP::_recv(udp_pcb *upcb, pbuf *pb, const ip_addr_t *addr, uint16_t po
         if(_handler) {
             AsyncUDPPacket packet(this, this_pb, addr, port, netif);
             _handler(packet);
-        } else {
-            pbuf_free(this_pb);
```
me-no-dev: on destruct the packet will free the pbuf. if it does not, that should be fixed there, or where the packet is copied

libraries/AsyncUDP/src/AsyncUDP.cpp
```diff
@@ -324,6 +324,22 @@ AsyncUDPPacket::AsyncUDPPacket(AsyncUDP *udp, pbuf *pb, const ip_addr_t *raddr,
...
+      _len(packet._len),
+      _index(packet._index)
+{
+    memcpy(_remoteMac, packet._remoteMac, sizeof(_remoteMac));
+    pbuf_ref(_pb);
+}
+
 AsyncUDPPacket::~AsyncUDPPacket()
 {
     pbuf_free(_pb);
```
me-no-dev: here the pbuf should be freed. should take care of `pbuf_ref` as on that depends if `pbuf_free` will work

libraries/AsyncUDP/src/AsyncUDP.cpp
ssilverman: `pbuf_free`, according to the code, needs to be called at least once, even if the handler isn't used. As soon as the handler is used with the new `AsyncUDPPacket` object, its constructor and destructor will `pbuf_ref` and `pbuf_free`, respectively. That's why the copy constructor needs to add a `pbuf_ref`, to balance things out if the pass-as-copy is used.

What are your thoughts on the logic in #3288? I tested the code with my example from #3287 using both pass-as-reference and pass-as-copy forms.

libraries/AsyncUDP/src/AsyncUDP.cpp
ssilverman: It is freed, as you point out, in the destructor. This balances the call to `pbuf_ref` in the constructor. That's why my change also adds a call to `pbuf_ref` in the copy constructor. This solves the imbalance.

ssilverman: As written, the code won't successfully work for _both_ pass-as-reference and pass-as-copy forms, no matter where the `pbuf_free` call is used in `AsyncUDP::_recv`, either outside or inside an `else`. That's why adding to the copy constructor solves the problem for both cases, if the `pbuf_free` is moved outside the if/else.

libraries/AsyncUDP/src/AsyncUDP.cpp
ssilverman: Another possibility for a fix: Don't call `pbuf_ref` in the constructor or copy constructor, and don't call `pbuf_free`in the destructor. Have all management outside the `AsyncUDPPacket` object. However, code that currently exists and assumes the class will manage this will break.

libraries/AsyncUDP/src/AsyncUDP.cpp
me-no-dev: I have tried your fix before and it did not work in many situations. so I reverted it back. you could look in the history... should be there :) need to figure out better that pbuf ref/unref count and not leak or double free

ssilverman: Interesting. Could you show me some situations in which my fix fails? I’d like to see if I can resolve them. All I know is that my fix doesn’t fail in any of my tests.

me-no-dev: unfortunately I can not recall...

ssilverman: So you’re saying that in the past you’ve already tried adding this fix _with_ changing the copy constructor and it didn’t work? What do you think of the logic in #3288? I don’t see what’s wrong with it.

Could you at least show which commits you undid where the fix didn’t work?

thbl:
@me-no-dev any news on this issue ? :)

libraries/AsyncUDP/src/AsyncUDP.cpp
thbl: @me-no-dev ?

ssilverman:
@thbl I just use my own fork (I haven't looked at this in over a year, so I'm not sure what the current status is). I'm still not sure what the problem is with my fix.

ssilverman:
@thbl Looks like @me-no-dev took this fix about 2 months ago but didn't use my PR: https://github.com/espressif/arduino-esp32/commit/8134a42162f11cb01155038a2465848d6b2b84bc
Closing because the PR is not needed anymore. Turns out I was right; don't know why my fix was rejected out of hand: https://github.com/espressif/arduino-esp32/pull/3290#discussion_r329368293

PR #3314, opened 2019-10-02
Base64::encode : const correctness / String by reference passing

dirkmueller:
Avoid passing String by-value, which is slightly less efficient
as it involves a full copy-constructor/tempstring creation.


me-no-dev:
Thanks :) totally agree

PR #3329, opened 2019-10-07
fix authrization problem

lightzl:
I compiled the sample "WebServer -> HttpAdvancedAuth". But the right username and password can not login in. I found the file "WebServer.c" may be have some problem. So I fix it. Please check it.


me-no-dev:
thanks :)

PR #3345, opened 2019-10-11
Add an aditional (void *) arg to the RMT callback (much like Ticker() et.al.).

dirkx:
Add an aditional (void *) arg to the RMT callback - to allow more flexible handling of the callback (e.g. by passing a private struct or a class pointer).

Same pattern as used by the Ticker() and many others. Example updated & new example with a trapoline added.

This lets you handle multiple RMT handlers with ease & without many global variables or tricks (as the current API does not provide access to the channel).

### DIFF
--- a/cores/esp32/esp32-hal-rmt.c
+++ b/cores/esp32/esp32-hal-rmt.c
@@ -94,6 +94,7 @@ struct rmt_obj_s
     transaction_state_t tx_state;
     rmt_rx_data_cb_t cb;
     bool data_alloc;
+    void * arg;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


me-no-dev:
nice :) will hold it for a bit, just to make sure that 1.0.4 is all good before I change API

suhlig:
If I understand the situation correctly, there is no way to use a classes member function as callback before this PR is merged. The example that is part of the PR illustrates this nicely by using a trampoline that passes `this`. 

Today this _is_ possible with the ESP8266 Ticker lib, but _not_ with the one for ESP32. 

Please merge 🙏

PR #3356, opened 2019-10-14
SPI: Ensure all read-only data pointers are marked as const

davefiddes:
This changes all SPI functions that take data pointers which are
not modified so that the declaration is const. This allows them
to be used with const data (i.e. held in flash). No functional
changes are required.

The defnitions of spiWrite() and spiTransferBytes()  in
esp-hal-spi.h/c have been updated to be consistent.

Tests:
 - Build a simple sketch using SPI.writePattern() and
   SPI.transferBytes()  which uses const data and verify that the
   attached device functions as expected.


me-no-dev:
Agreed :) good change indeed

Issue #3215, opened 2019-09-16
Build failed on ESP_IDF


boonkey:
### Hardware: (irrelevant for this issue)
Board: M5 ESP32CAM
Core Installation/update date: latest
IDE name: IDF-with Arduino
Flash Frequency: 40
Flash size: 4 MB
PSRAM enabled: ?no?
Upload Speed: ?115200?
Computer OS: Ubuntu 18.04.3 LTS

I'm trying to add the HTTPClient header file (and all associated files, libraries, methods, etc.) to M5's WIFI_STA project [(link)](https://github.com/m5stack/m5stack-cam-psram/tree/master/wifi/wifi_sta) in order to be able to send GET requests.

I followed the steps in the docs [(doc)](https://github.com/espressif/arduino-esp32/blob/master/docs/esp-idf_component.md):
1. cloned the repo into components/ as arduino/
2. tried to make
I had the exact same issue as [WiFiClientSecure make issue #2154](https://github.com/espressif/arduino-esp32/issues/2154), so I followed the instructions there to solve. 
3. `Component Config -> mbedTLS -> TLS Key Exchange Methods -> 
  [*] Enable pre-shared-key ciphersuits
    [*] Enable PSK based ciphersuite modes`
I now still have building issues, appearing unrelated to the previous one. 
### LOG: 
`In file included from /home/user/esp/m5stack-cam-psram/wifi/wifi_sta/main/main.c:18:0:
/home/user/esp/m5stack-cam-psram/wifi/wifi_sta/components/arduino/libraries/HTTPClient/src/HTTPClient.h:32:18: fatal error: memory: No such file or directory
compilation terminated.
/home/user/esp/esp-idf/make/component_wrapper.mk:285: recipe for target 'main.o' failed
make[1]: *** [main.o] Error 1
/home/user/esp/esp-idf/make/project.mk:530: recipe for target 'component-main-build' failed
make: *** [component-main-build] Error 2
`
please help me to solve this issue.
Thanks

me-no-dev:
you forgot to mention the ESP-IDF version, but I will guess it's not 3.2 :)
Arduino tracks IDF branch release/3.2 currently

boonkey:
I changed ESP-IDF version to release/3.2
It did not solve the issue 
`In file included from /home/user/esp/m5stack-cam-psram/wifi/wifi_sta/main/main.c:18:0:
[...]

Issue #2154, opened 2018-12-05
WiFiClientSecure make issue


Goldenstarc:
### Hardware: (Doesn't matter for this issue !)
Board:							node32
Core Installation/update date:			latest
IDE name:							IDF-with Arduino
Flash Frequency:					40
PSRAM enabled:                                          ?no?
Upload Speed:						?115200?
Computer OS:                                          MacOS

### Description:
error in make command !
USING LATEST LIBRARIES without any change !!

`CXX build/arduino-esp32/libraries/WiFiClientSecure/src//ssl_client.o
/Users/hamid/Documents/Projects/ESP-IDF/IDF-Arduino/components/arduino-esp32/libraries/WiFiClientSecure/src/ssl_client.cpp: In function 'int start_ssl_client(sslclient_context*, const char*, uint32_t, const char*, const char*, const char*, const char*, const char*)':
/Users/hamid/Documents/Projects/ESP-IDF/IDF-Arduino/components/arduino-esp32/libraries/WiFiClientSecure/src/ssl_client.cpp:144:67: error: 'mbedtls_ssl_conf_psk' was not declared in this scope
                  (const unsigned char *)pskIdent, strlen(pskIdent));
                                                                   ^
make[1]: *** [libraries/WiFiClientSecure/src//ssl_client.o] Error 1
make: *** [component-arduino-esp32-build] Error 2`

atanisoft:
you need to enable MBED_TLS_PSK options in your sdkconfig, please refer to the arduino sdkconfg for an example, there are four or five lines you need to enable.

Goldenstarc:
ok dude, I'm surrender ! :)
I just get a new version (TODAY) and it fixed
I have a simple function to update my esp repos in bash profile. it may help  :)

`function esp_update()
{
currentDir=$PWD && \
cd ~/esp && \
cd arduino-esp32 && \
git pull origin master && \
git submodule update --init --recursive && \
cd .. && \
cd esp-idf && \
git pull origin master && \
pip install esptool --upgrade && \
cd $currentDir
}`

henryjw:
For anyone else that runs into this issue, another solution is to disable the WifiClientSecure component altogether. This can be done by following the steps below:

1) Run `make menuconfig` from the project's directory
2) Select "Arduino Configuration"
3) Select "Include only specific Arduino libraries"
4) Deselect Enable WifiClientSecure" and any other libraries you don't need

Goldenstarc:
[...]

PR #3364, opened 2019-10-15
Add an error message in case of invalid configured dependency mbedTLS.

Boldie:
Especially if the user wants to use the library as component in IDF,
there are some pitfalls while doing make menuconfig. One is this missing
dependency which will now fail with a better error message with a hint to
the user how to fix it.

refs #2154 #3215

### DIFF
--- a/libraries/WiFiClientSecure/src/ssl_client.cpp
+++ b/libraries/WiFiClientSecure/src/ssl_client.cpp
@@ -19,6 +19,9 @@
 #include "ssl_client.h"
 #include "WiFi.h"
 
+#ifndef MBEDTLS_KEY_EXCHANGE__SOME__PSK_ENABLED
+#  error "Please configure IDF framework to include mbedTLS -> Enable pre-shared-key ciphersuites and activate at least one cipher"
+#endif
 
 const char *pers = "esp32-tls";
 


libraries/WiFiClientSecure/src/ssl_client.cpp
```diff
@@ -19,6 +19,9 @@
 #include "ssl_client.h"
 #include "WiFi.h"
 
+#ifndef MBEDTLS_KEY_EXCHANGE__SOME__PSK_ENABLED
```
atanisoft: This would need to be prefixed with CONFIG_... But I don't see this key on IDF v3.2 or IDF v3.3.

atanisoft: CONFIG_MBEDTLS_PSK_MODES and CONFIG_MBEDTLS_TLS_ENABLED would be better to check for here.

Boldie: This is not quite easy to handle with the CONFIG_* definitions because internally the library will use a lot of things together (the basic config must be set + at least one cipher must be selected) to enable this function. So I choose to use the library definition which really enables the function to not have duplicate logic in there.

atanisoft: The problem is the define you have used will never exist, regardless of sdkconfig entries selected. If you switch to the CONFIG_XXX entries I posted above it will at least compile for test cases.

But you are right, there is still the requirement to have at least one key exchange format enabled which will require explicit checking for the various keys used by the code.

me-no-dev:
done :)

PR #3425, opened 2019-10-29
Fixed wrong relative path for lwip/arpa/inet.h

mathiasbredholt:
The path given in the original lwip header from esp-idf refers to a path that doesn't exist in arduino-esp32.

### DIFF
--- a/tools/sdk/include/lwip/arpa/inet.h
+++ b/tools/sdk/include/lwip/arpa/inet.h
@@ -15,6 +15,6 @@
 #ifndef INET_H_
 #define INET_H_
 
-#include "../../../lwip/src/include/lwip/inet.h"
+#include "lwip/inet.h"
 
 #endif /* INET_H_ */


atanisoft:
This will need to be submitted to esp-idf and not here since this header is copied as-is from there

mathiasbredholt:
@atanisoft But the problem is that the path structure of the header files in the esp-idf is different than the one in arduino-esp32. When I'm including the header file in an Arduino project, I'm getting this error https://github.com/mathiasbredholt/libmapper_arduino/issues/3.

atanisoft:
Be that as it may, I don't think this will be allowed to merge without being fixed in https://github.com/espressif/esp-idf.  @me-no-dev what are your thoughts?

me-no-dev:
this is a bug in ESP-IDF. Please post the issue there and state that you are using IDF3.2 of Arduino :) I can not accept any changes to `tools/sdk`. All those files come from IDF and should be fixed there.

PR #3427, opened 2019-10-30
Added MGBOT IOTIK 32A and IOTIK 32B boards

vrxfile:
Hi!

I added new Russian education boards IOTIK 32A and IOTIK 32B:
http://www.iotik.ru/iotik32a/
http://www.iotik.ru/en/iotik32b/

Thanks!

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -4226,3 +4226,278 @@ vintlabs-devkit-v1.menu.DebugLevel.debug.build.code_debug=4
 
 ##############################################################
 
+mgbot-iotik32a.name=MGBOT IOTIK 32A
+
+mgbot-iotik32a.upload.tool=esptool_py
+mgbot-iotik32a.upload.maximum_size=1310720
+mgbot-iotik32a.upload.maximum_data_size=327680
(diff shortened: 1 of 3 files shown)


me-no-dev:
HI! There are a lot of options added for those boards. I am sure that most are useless at least. Please selct partition schemes that make sense for the device. Also you do not have PSRA, why do you have the option enabled? If you do not have 16MB of flash, remove that as well. If the board does not use custom pins that differ from the regular ESP32 based modules, you do not need a special variant header and can use the default one.

vrxfile:
Hi! We are going to produce boards with different flash memory sizes (also with ESP32-WROOM-32 16 Mb flash) and with ESP32-WROVER with PSRAM. May we leave all these settings for now?

me-no-dev:
how about the variant headers?

vrxfile:
Please help me. What is the difference between special and default variant header? What do I need to fix?

me-no-dev:
if your board does not need any special pin assignment, you can give it the default esp32 variant and delete the header

vrxfile:
We have LED and IR receiver and different names for some pins, because we use it in our educational documents and instructions.

PR #3435, opened 2019-10-30
Sht post

Mjohncecil:
Enables esp8266 to make post requests to a prepared Raspberry Pi.

Test:
Install any missing libraries and prepare environment to run code for esp8266

Save network username, network password, and post destination to the appropriate variables in the .ino file.

run postRequest.py in rPi.

Upload .ino file to esp8266 and open serial monitor.

If the measurements taken by the sht31 match the numbers recieved by pi the post request is functioning correctly.

Disconnect sht and check if DS18B20 measurements are also being sent.


ESP_SHT3x/ESP_SHT3x.ino
```diff
@@ -0,0 +1,232 @@
...
+    Serial.println("Failed to read temperature");
+  }
+ 
+  if (! isnan(hum)) {
+    Serial.print("Hum. % = "); Serial.println(hum);
+    postProtocol("H: " + (String) hum);
+  } else { 
+    Serial.println("Failed to read humidity");
+  }
+}
```
donnyhyun: Seems like this is sending temperature and humidity in two separate POST requests. They should be sent in a single request in JSON format.

PR #3441, opened 2019-11-01
Add Uri with support for regexUri and globUri

Bmooij:
Same as for [esp8266](https://github.com/esp8266/Arduino/pull/6696)


joysfera:
FYI, this broke backward compatibility and it was not mentioned in the release notes.

PR #3442, opened 2019-11-01
Corrections of Stream.Find, FindUntil and added FindMulti - like AVR-Core Libraries

ale-trevizoli:
Find has some bug that is not working with Ethernet.find() so, I copied code from Stream.h and Stream.cpp in AVR-CORE library and now it's working perfectly.
I don't know where was the error, but an Ethernet.find compiled to MEGA2560 was working but not working when compiled to esp32, after corrections of code (copy of AVR-Core libraries) it's working perfect.
So probably has some error on original ESP32-Core library.

Below is part of code that was working with MEGA2560 and not with ESP32 libraries.
client.find never return TRUE with ESP32 original library and with AVR it's works.

boolean esp32_fw_update(EthernetClient &client, DecodedHeader &header, const String &field_filename, const String &field_crc) {

  char bound[header.boundary.length()+3];
  char term[]="\r\n";
  
  strcpy(bound,header.boundary.c_str());
  strcat(bound,term);
  while (client.find(bound)) { 
    String line=client.readStringUntil('\r');


cores/esp32/Stream.cpp
```diff
@@ -85,7 +85,7 @@ void Stream::setTimeout(unsigned long timeout)  // sets the maximum number of mi
 unsigned long Stream::getTimeout(void) {
   return _timeout;
 }
-
+/*
```
atanisoft: instead of leaving a lot of dead code behind, just delete it.

cores/esp32/Stream.cpp
```diff
@@ -141,6 +141,114 @@ bool Stream::findUntil(const char *target, size_t targetLen, const char *termina
     return false;
 }
 
+*/
+
+// ******************** ALEXANDRE
+
+ // find returns true if the target string is found
+bool  Stream::find(char *target)
```
atanisoft: you should restore the const signatures for these methods.

cores/esp32/Stream.h
```diff
@@ -60,6 +60,24 @@ class Stream: public Print
...
+
+    bool find(char target) { return find (&target, 1); }
+
+    bool findUntil(char *target, char *terminator);   // as find but search ends if the terminator string is found
+    bool findUntil(uint8_t *target, char *terminator) { return findUntil((char *)target, terminator); }
+
+    bool findUntil(char *target, size_t targetLen, char *terminate, size_t termLen);   // as above but search ends if the terminate string is found
+    bool findUntil(uint8_t *target, size_t targetLen, char *terminate, size_t termLen) {return findUntil((char *)target, targetLen, terminate, termLen); }
+
+/*
```
atanisoft: same here about dead code and retaining the const signature updates.

ale-trevizoli: I leave this code to Espressif decide what to do. About const it was not there on original AVR Core code.

cores/esp32/Stream.cpp
ale-trevizoli: Const signature was not there on original AVR Core code.

cores/esp32/Stream.h
me-no-dev: unless the code modifies the chars, they should be const. Regardless if that is on AVR or not :)

PR #3219, opened 2019-09-17
Support for Master mode, Pin and SSP

IonicEV:
I witnessed a lot of folks struggling with master mode on ESP32 (initiating connection to other slave devices). Some examples were available(example_spp_initiator_demo.c), but they lacked support for IO, congestions and other aspects covered nicely in original BluetoothSerial class. I took a step forward and combined code provided in example_spp_initiator_demo.c and BluetoothSerial class to enable Master mode, Pin and SSP for legacy BT in BluetoothSerial class. Hopefully, you would find it useful.

PR #3446, opened 2019-11-03
Fixed incorrect usage of xEventGroupWaitBits

WoLpH:
Related discussion in #3219

### DIFF
--- a/libraries/BluetoothSerial/src/BluetoothSerial.cpp
+++ b/libraries/BluetoothSerial/src/BluetoothSerial.cpp
@@ -154,7 +154,7 @@ static uint8_t _spp_tx_buffer[SPP_TX_MAX];
 static uint16_t _spp_tx_buffer_len = 0;
 
 static bool _spp_send_buffer(){
-    if((xEventGroupWaitBits(_spp_event_group, SPP_CONGESTED, pdFALSE, pdTRUE, portMAX_DELAY) & SPP_CONGESTED)){
+    if(xEventGroupWaitBits(_spp_event_group, SPP_CONGESTED, pdFALSE, pdTRUE, portMAX_DELAY) & SPP_CONGESTED){
         esp_err_t err = esp_spp_write(_spp_client, _spp_tx_buffer_len, _spp_tx_buffer);
         if(err != ESP_OK){
             log_e("SPP Write Failed! [0x%X]", err);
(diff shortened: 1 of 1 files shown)


IonicEV:
You may get note on bool vs. wait bitwise value. Originally, those changes resulted in logic errors by mistake. It may fail equal operator logic: (true != bit), as 1 != 2. It does not affect the existing code though.

me-no-dev:
please bring back the braces and check if it's different than 0. While non-null value is technically TRUE, I would like the code to show that properly :)

WoLpH:
I've added the `!= 0` and I've added parentheses (I assume you meant those instead of braces? Braces are `{` and `}`)

me-no-dev:
surely :D excuse my "English". braces, brackets, parentheses... it's a bit confusing

WoLpH:
I know how it can be, English is not my native language either :)

me-no-dev:
to me:
- `(` braces
- `{` curly braces
- `[` square braces / brackets
- parentheses sounds like it should be `'` or `"`
- my own language sounds strange to me... go figure

Shunyavic:
I'm having some trouble connecting my ESP32-WROOM-32 to any other device via bluetooth. I'm using the SerialtoSerialBTM example with the address and name modified only. The log I'm getting from the debug is the following:

```
[E][esp32-hal-cpu.c:159] setCpuFrequencyMhz(): Can not switch to 240 MH⸮f⸮⸮⸮⸮CPU frequency rated for 160MHz.
[D][esp32-hal-cpu.c:178] setCpuFrequencyMhz(): PLL: 320 / 2 = 160 Mhz, APB: 80000000 Hz
[W][esp32-hal-psram.c:30] psramInit(): PSRAM init failed!
[I][BluetoothSerial.cpp:514] _init_bt(): device name set
[I][BluetoothSerial.cpp:227] esp_spp_cb(): ESP_SPP_INIT_EVT
[I][BluetoothSerial.cpp:120] btSetPin(): pin set
The device started in master mode, make sure remote BT device is on!
[I][BluetoothSerial.cpp:726] connect(): master : remoteAddress
[I][BluetoothSerial.cpp:292] esp_spp_cb(): ESP_SPP_DISCOVERY_COMP_EVT

```

I've seen that it sets the address right, and when doing the _connect()_ function it returns _false_ when doing _waitForConnect(SCAN_TIMEOUT)_ 
I believe that problem comes from the following line that was modified in this merge.
`return (xEventGroupWaitBits(_spp_event_group, SPP_CONNECTED, pdFALSE, pdTRUE, xTicksToWait) & SPP_CONNECTED) != 0;`

I've tried to connect to multiple deviced and it always returns false, any way to fix this?

atanisoft:
@Shunyavic open an issue for your issue rather than hijack a merged PR.

WoLpH:
While this code fixes the bug that it was erroneously telling you it worked, it doesn't fix anything else.

I haven't been able to get the master mode working well either

PR #3454, opened 2019-11-06
Add single quotes to fix bash error when there is a space in either paths {build.source.path} or {build.path}

Jeroen88:
Using Ubuntu as operating system, and the latest Arduino IDE (1.8.10), I have a space in the path to the source directory resulting in the following error during compilation:

```
bash: line 0: [: too many arguments
cp: target 'partitions.csv' is not a directory
exit status 1
Error compiling for board Node32s.
```

Adding single quotes to the paths {build.source.path} and {build.path} fixes the error.

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -63,8 +63,8 @@ compiler.elf2hex.extra_flags=
 
 # Build Dir: {build.path}
 # Sketch Dir: {build.source.path}
-recipe.hooks.prebuild.1.pattern=bash -c "[ ! -f {build.source.path}/partitions.csv ] || cp -f {build.source.path}/partitions.csv {build.path}/partitions.csv"
-recipe.hooks.prebuild.2.pattern=bash -c "[ -f {build.path}/partitions.csv ] || cp {runtime.platform.path}/tools/partitions/{build.partitions}.csv {build.path}/partitions.csv"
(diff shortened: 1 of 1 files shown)


me-no-dev:
shouldn't quotes be on the whole string and not just the variable part? Or is this some Arduino builder hackery ?

Jeroen88:
@me-no-dev I am no Arduino builder expert, but I tried putting the quotes on the quotes on the whole  string, like this:

`'{build.source.path}/partitions.csv'`

and then 2 times, but that doesn't work...

me-no-dev:
did you try sloeber? it's an eclipse plugin that turns the configs into a sort of makefile and I have had tons of headaches with it and quotes... if you have not teted it, please give it a shot and make sure that you can compile ok

Jeroen88:
I do not use eclipse nor schloeber, just the Arduino IDE. Could you test it?

me-no-dev:
maybe you should give it a try ;) you might like it! http://eclipse.baeyens.it/index.shtml

Jeroen88:
I might give it a try once, thanks! I already installed pio a few months ago, compiled my sketches, but no time to try it further... :/

me-no-dev:
i will give it a go, but I recommend it... cuz it'd better than the IDE ;)

Jeroen88:
Okay okay :)

PR #3458, opened 2019-11-07
SPI: Fix discarded-qalifiers warning when compiling with all warnings

davefiddes:
This fixes an error introduced with changeset b847f41 which
tightened the use of const for read-only data. The helper
function __transferBytes also requires the const qualifier on
outgoing data. Without this change a warning is displayed
when compiling with the Arduino IDE set to display "All"
compiler warnings.

Local functions __transferBytes() and __spiTranslate32() are now marked as static.

Unused local function __spiTranslate24() has been removed.

Tests:
 - Build an ESP32 SPI sketch that uses static const data to send
   to an SPI device using the SPI.transferBytes() API

### DIFF
--- a/cores/esp32/esp32-hal-spi.c
+++ b/cores/esp32/esp32-hal-spi.c
@@ -532,16 +532,6 @@ uint8_t spiTransferByte(spi_t * spi, uint8_t data)
     return data;
 }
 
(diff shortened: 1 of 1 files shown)


me-no-dev:
please mark the method as static as well :)

me-no-dev:
thanks :)

PR #3463, opened 2019-11-08
fix typo BLE_client.ino

takanotume24:
fix typo

### DIFF
--- a/libraries/BLE/examples/BLE_client/BLE_client.ino
+++ b/libraries/BLE/examples/BLE_client/BLE_client.ino
@@ -154,7 +154,7 @@ void loop() {
     // Set the characteristic's value to be the array of bytes that is actually a string.
     pRemoteCharacteristic->writeValue(newValue.c_str(), newValue.length());
   }else if(doScan){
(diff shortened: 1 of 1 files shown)


fzesp32:
Hi everybody,  
I am doing a project using esp32 in order to receive the values ​​of three Bluetooth Low Energy sensors. Then he sends them to the database using WiFi.
The result is that I receive the sensor values ​​of bluetooth low energy in the Serieller Monitor in Arduino IDE, but when I want to send them with the same esp32 using wifi to the database, only one sensor values ​​are sent.
The program code is your BLE Client, with modifications added to receive more than one sensor. But as I mentioned, the problem is that only one sensor's values ​​are sent to the database using WiFi.
Can I use the timer in that code? Where can I use it? Because I am very beginner, can you send me where to put the timer?

Do you have another idea for solving the issue in case the timer doesn't work?
Thanks a lot for your help

chegewara:
This is not the place to ask questions about your code, neither its my private email. For such questions is esp32.com forum, stackoverflow etc

PR #3469, opened 2019-11-10
configTime

lbernstone:
Fixed configTime so it doesn't crash if tcpip is not initialized.


cores/esp32/esp32-hal-time.c
```diff
@@ -45,6 +45,11 @@ static void setTimeZone(long offset, int daylight)
  * */
 void configTime(long gmtOffset_sec, int daylightOffset_sec, const char* server1, const char* server2, const char* server3)
 {
+#ifndef _TCPIP_ADAPTER_H_
```
atanisoft: This doesn't seem to be sufficient, it just checks that tcpip_adapter.h has been included. Perhaps there should be a check against ```WiFi.getMode() != WIFI_MODE_NULL)``` or some variant there of via a generic function call (in core vs in wifi lib)

PR #3485, opened 2019-11-15
ESP32 I2S with INMP441 DEMO

q383186554:
HI Dear Author:
    I am trying to write a program to get data from INMP441 , the module is a DAC output module.
    There is only one HiFreq_ADC.ino demo in the arduino-esp32 example and the old version demo could not be compiled because some of methods have been removed.
    I tried for a week but the voice and wave graph always being strange, could you please write a demo,save the voice in computer *.wav and show wave graph on arduino IDE.
    Thank you very much.


me-no-dev:
why is the PR here? you are pulling to your repo, not to this one.

PR #3502, opened 2019-11-20
BLE: reception support multiple service data

egrekov:
There are devices that issue more than one service data. For example: xiaomi mi thermometer. 

This code solves this problem and does not break compatibility. You can read more about this pull request [here (pull request for nkolban)](https://github.com/nkolban/ESP32_BLE_Arduino/pull/24) Adapted the request for the current master.


eos1d3:
This pull request is important for multiple service data. The problem exists for months.

PR #3512, opened 2019-11-25
Fix RMT buffer allocation

collin80:
As the code sits, you can't use the last RMT buffer, basically because of an off by one style error. The comparison I changed should be "greater than" only not "greater than or equal to." It's OK if i+j = MAX_CHANNELS. In the case of having four inputs, all using 2 buffers each, this leads to i being 0, 2, 4, 6 and each takes up two buffers so the allocation is 0-1, 2-3, 4-5, 6-7. The 6-7 is the problem because i+j will equal 8 but really that still means only 6-7 are being used as j is always one larger than the actual offset from the starting buffer.


me-no-dev:
nice find :) just love those type of bugs

PR #3515, opened 2019-11-25
Add TTGO T Journal Camera module & improve commenting

WizardTim:
Added pinout for TTGO T Journal device and improved the intent of comments.


scharrin:
Seems like need a board option for the t-journal also.

WizardTim:
@scharrin I just use the `ESP32 Dev Module` board to compile this camera web server sketch. If you're doing more than that you might need a custom board config.

PR #3521, opened 2019-11-27
BLE: removed double delete of characteristics

crnchbng:
Calling the destructor or disconnect of BLEClient causes a heap corruption. 

This is due of the destructor of the BLERemoteService: 
The removeCharacteristics() method tries to delete the same characteristic twice, one time per map where its pointer is stored in.


TheNitek:
Any chance to get this merged? It's currently impossible to "delete" a BLEClient, so memory leaks are unavoidable

PR #3527, opened 2019-11-28
Cleanup SPI_Multiple_Buses

muman613:
The existing example contains some issues which could cause confusion. 

Looking at line #35 it is clear that the commented line should be calling the `begin` method of `vspi` and not `hspi`. By using preprocessor symbols it makes the code more readable, and being able to control whether the SPI bus standard pins are used or the alternate pins by setting a symbol makes it easier.

Also the setting of the SS pins to output now is easier, and also the raising and lowering of the SS lines during sending is a lot easier and clearer.

Please consider this PR for merging.


me-no-dev:
merged :)

PR #3529, opened 2019-11-29
Add WIFI_REASON_CONNECTION_FAIL in WiFiGeneric.cpp

felixstorm:
`WIFI_REASON_CONNECTION_FAIL` has been added to `wifi_err_reason_t` some time ago:  
https://github.com/espressif/arduino-esp32/blob/cec3fca4ad4a39feb463f9298ab3238819732d50/tools/sdk/include/esp32/esp_wifi_types.h#L95

But it was still missing from `system_event_reasons` in `WiFiGeneric.cpp` sometimes causing panics here:  
https://github.com/espressif/arduino-esp32/blob/cec3fca4ad4a39feb463f9298ab3238819732d50/libraries/WiFi/src/WiFiGeneric.cpp#L348


me-no-dev:
good one!

PR #3531, opened 2019-11-29
Updated BLERemoteCharacteristic to exposre esp_gatt_auth_req_t parame…

cipherz:
For a project, I needed to be able to specify the esp_gatt_auth_req_t parameter for read and write value. I thought this might be valuable to others and I am therefore submitting a pull request for consideration.


cipherz:
Makes sense, which class should be setup with it? BLEDevice, BLERemoteService or BLERemoteCharacteristic? I guess the question here is would the auth likely be the same for connections with client, all services or just for a single characteristic.

Thank you - and I'd be more than happy to update the PR based on your feedback.

chegewara:
BLERemoteCharacteristic and probably BLERemoteDescriptor too to make it consistent.

This of course will require to call function to set required auth for each characteristic, but according to bluetooth specs some characteristics cant have auth set to other than NONE.

cipherz:
Thank you for the quick response and guidance. I will update the PR and resubmit for reconsideration later today.

cipherz:
I updated the PR to not have auth on each read/write as suggested but rather as a private variable for BLERemoteCharacteristic/Descriptor, which is defaulted to none when created and can be modified via setAuth.

PR #3535, opened 2019-12-03
Defined BLE setAdvertisementType

syonbori:
Currently we have no way to set BLE Advertisement Type.
We have an initial value of `ADV_TYPE_IND` and the member is private.

iBeacon should advertise with `ADV_TYPE_NONCONN_IND`.
esp-idf's example code specifies this value.

https://github.com/espressif/esp-idf/blob/93a8603c545fb8e54741d6685146e2f3b874378d/examples/bluetooth/bluedroid/ble/ble_ibeacon/main/ibeacon_demo.c#L57


Legion2:
@chegewara  Shouldn't this be added to https://github.com/nkolban/esp32-snippets/tree/master/cpp_utils, because the library is maintained in the nkolban/esp32-snippets repo? And then sync the library as described in https://github.com/nkolban/esp32-snippets/blob/master/cpp_utils/ArduinoBLE.md#replacing-the-version-that-comes-with-arduino-esp32 back to this repo.

Currently the libraries are out of sync. For example the `setStaticPIN` feature is supported in nkolban/esp32-snippets but not here.

chegewara:
Currently arduino BLE library and cpp_utils are separate libraries. If there is something that is missing in arduino, but exists in cpp_utils feel free to create PR here.

PR #3536, opened 2019-12-03
Changed init from if-statement to while-loop

JoshuaPerk:
Due to the ESP32 boot process, calling SD.begin() too soon can result in inconsistent "Card Mount Failed" issues. Placing the init within the while loop forces the sketch to retry the critical loading of the card vs. failing and halting the entire sketch.


me-no-dev:
and if the card is not in, it will loop forever. You can add a delay and retry if it fails, but not a while loop.

PR #3579, opened 2019-12-17
Fix SerialBT.flush

mishafarms:
flush should not flush the read buffer, it should wait till all output is drained from the write buffer.


lbernstone:
https://github.com/espressif/arduino-esp32/pull/3433 was recently merged to fix this.

mishafarms:
#3433 is for the uart serial port, This is for the Standard Bluetooth SerialBT.flush. I don't think #3433 fixes it.

libraries/BluetoothSerial/src/BluetoothSerial.cpp
```diff
@@ -652,7 +652,11 @@ size_t BluetoothSerial::write(const uint8_t *buffer, size_t size)
 
 void BluetoothSerial::flush()
 {
-    while(read() >= 0){}
+    if (_spp_tx_queue != NULL){
+        while(uxQueueMessagesWaiting(_spp_tx_queue) > 0){
+	    delay(50);
```
me-no-dev: 50 is a bit too much :) there ought to be a better way.

PR #3602, opened 2019-12-27
Added partition and flash size menus for VintLabs boards

pvint:
As the title says, I would like to add the partition and flash size menus for the VintLabs board definitions.

Thanks!


pvint:
Thanks!

PR #3606, opened 2019-12-31
Move GH CI scripts to Python3

earlephilhower:
Python2 is EOL in 12 hours, so make sure we call python3 explicitly (and
pip3) in the CI scripts.


earlephilhower:
Closing as this is turning into a can of worms.  The core platform.txt also needs to be updated to Python3 as well as the installation docs and tools in order to make this work.  Sorry for the noise!

PR #3607, opened 2019-12-31
Update Server.h for arduino core compatibility

orvisevans:
https://github.com/arduino/ArduinoCore-API/blob/master/api/Server.h
The ArduinoCore uses virtual void begin() =0;  Changing this allowed me to compile a project using the Ethernet library.  Otherwise I would get the following compile error:
```
src\main.cpp:35:16: error: cannot declare variable 'server' to be of abstract type 'EthernetServer'
 EthernetServer server(80);
                ^
In file included from src\main.cpp:21:0:
.pio\libdeps\featheresp32\Ethernet_ID872\src/Ethernet.h:253:7: note:   because the following virtual functions are pure within 'EthernetServer':
 class EthernetServer : public Server {
       ^
In file included from C:\users\orvis\.platformio\packages\framework-arduinoespressif32\cores\esp32/Arduino.h:152:0,
                 from .pio\libdeps\featheresp32\Ethernet_ID872\src/Ethernet.h:51,
                 from src\main.cpp:21:
C:\users\orvis\.platformio\packages\framework-arduinoespressif32\cores\esp32/Server.h:28:18: note:      virtual void Server::begin(uint16_t)
     virtual void begin(uint16_t port = 0) = 0;
```

### DIFF
--- a/cores/esp32/Server.h
+++ b/cores/esp32/Server.h
@@ -25,7 +25,7 @@
 class Server: public Print
 {
 public:
-    virtual void begin(uint16_t port=0) =0;
+    virtual void begin() =0;
 };
 
 #endif


me-no-dev:
it does break other things though. We will move to the official API soon enough :)

ivoh95:
@me-no-dev > it does break other things though. We will move to the official API soon enough :)

For us to move to the official API, what all needs to be changed? I believe this goes deeper into the way wifi as works well and not just Ethernet? What files would need to be changed to move to the official API?

CircuitSetup:
Was any progress ever made with this?

orvisevans:
Not by me, unfortunately.

jaap-smd:
I have the same problem. 
I want to use Socket.IO with this websockets library: links2004/WebSockets@^2.3.2. 
After making some minor changes in the websockets library to make it work, I got this error:
`lib/WebSockets/src/WebSocketsServer.cpp: In constructor 'WebSocketsServer::WebSocketsServer(uint16_t, const String&, const String&)':
lib/WebSockets/src/WebSocketsServer.cpp:49:55: error: invalid new-expression of abstract class type 'EthernetServer'
_server = new WEBSOCKETS_NETWORK_SERVER_CLASS(port);`
With the comment: WEBSOCKETS_NETWORK_SERVER_CLASS (= ETHERNET) is a pure virtual function.
ETHERNET is derived from server. So when I change in server.h:
`virtual void begin(uint16_t port=0) =0;`
to
`virtual void begin(uint16_t port=0);`
the code is working like expected. Well for me at least.
Any chance there will be a fix?

thujer:
Hi

I had the same problem a year ago.

I think that this problem was fixed in a newer version of the platform -
try to look if using latest.

We fixed this problem provisionally by downgrading ESP32 platform to 1.7.0.
in platformio.ini

[env:esp32dev]
; platform = espressif32
platform = espressif32@1.7.0  ; If some problem with library compatibility
board = esp32dev
framework = arduino

Another thing - try to look what version of WiFi or Ethernet library do You
using - there are one from Arduino libraries and one from ESP32 libraries -
I had similar problem with this


út 29. 12. 2020 v 15:59 odesílatel jaap-smd <notifications@github.com>
napsal:

thujer:
https://esp8266hints.wordpress.com/2019/04/05/esp32-w5500-simple-working-example/


st 30. 12. 2020 v 11:12 odesílatel Tomáš Hujer <thujer@gmail.com> napsal:

PR #3608, opened 2020-01-02
Update WiFiClient.cpp

Vigeant:
fixed the connected() function so that it only checks errno if recv returns a value of -1.

"in the even of an error, errno is set to indicate the error" --manpage

This fixes the ESP32 Webserver when dealing with a modern webserver with a slow SD card.


me-no-dev:
thanks :)

NathanSweet:
Why is `_connected = true;` here? IMO an unexpected error should disconnect.
```
              default:
                  log_i("Unexpected: RES: %d, ERR: %d", res, errno);
                  _connected = true;
                  break;
```
I call `client.connected()` every loop. When my client disconnects, I'm getting flooded with:
```
14:29:00.846 -> [615859][I][WiFiClient.cpp:531] connected(): Unexpected: RES: -1, ERR: 9
```
What is error 9 and why isn't it handled?

PR #3611, opened 2020-01-04
Added file.close to examples

### LINKED ISSUE #3609: BUG? ESP32 FFAT partition not hold second file(with format false!)
### Hardware:
Board:							ESP32-WROOM-32U IPEX 16MB flash
Core Installation/update date:			?arduino->boards manager -> install esp 1.0.4
IDE name:							Arduino IDE
Flash Frequency:					40Mhz
PSRAM enabled:                                          no
Upload Speed:						115200
Computer OS:                                            Windows 10 ,Ubuntu

### Description:
hi i was excited about speed ffat against spiffs it's really fast but in tests i found a problem when writing from code more file than just one and reset the board so it is readable only the first file the next one .. i see their names but the content is nothing

TEST Windows :
//Read the flash file system (previously formatted) and copy as local disc image
esptool.py --chip esp32 --port COM3 --baud 921600 read_flash 0x300000 0x100000 fatfs.img

OSF mount fatfs.img as disk(read the file same as arduino code .. the first file is OK and the others have 0 bytes in size)

-> write NEW file to mounted disk 

//Write a disc image
esptool.py --chip esp32 --port COM3 --baud 921600 --before esp32r0 --after hard_reset write_flash -z --flash_mode dio --flash_freq 40m --flash_size detect 0x300000 fatfs.img

result : same .. empty file (some files contain garbage)

sketch : example -> FFAT_TEST (FORMAT : first format FFAT and then change to false)

writeFile(FFat, "/First.txt", "Hello First");
writeFile(FFat, "/Second.txt", "Hello Second"); //after reset board this is empty file

### Debug Messages:
```
Enable Core debug level: Debug on tools menu of Arduino IDE, then put the serial output here 
```

lbernstone:
Fixes #3609

### DIFF
--- a/libraries/FFat/examples/FFat_Test/FFat_Test.ino
+++ b/libraries/FFat/examples/FFat_Test/FFat_Test.ino
@@ -52,6 +52,7 @@ void readFile(fs::FS &fs, const char * path){
     while(file.available()){
         Serial.write(file.read());
     }
+    file.close();
 }
 
 void writeFile(fs::FS &fs, const char * path, const char * message){
@@ -67,6 +68,7 @@ void writeFile(fs::FS &fs, const char * path, const char * message){
     } else {
         Serial.println("- write failed");
     }
+    file.close();
 }
 
 void appendFile(fs::FS &fs, const char * path, const char * message){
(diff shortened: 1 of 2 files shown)


me-no-dev:
IMO close is called on the exit of the function anyway as part of the destruct. What am I missing?

lbernstone:
In practice, I would agree.  But, in an example, I think it is more appropriate to demonstrate good hygiene on the file handles.  The user with the issue was somehow able to write the file and it did not close/persist.

me-no-dev:
sure

Issue #3584, opened 2019-12-19
When using TOUCH 8 and TOUCH 9, the corresponding GPIO should be exchanged


zztiger123:
### Hardware:
Board:	                           ESP32 Dev Module(Devkitc_v4)
Core Installation version:	   I copy the github website to the arduino yesterday
IDE name:	                   Arduino IDE
Flash Frequency:	           40Mhz
PSRAM enabled:               no
Upload Speed:	                   115200
Computer OS:                   Ubuntu

### Description:
Problem: When using TOUCH 8 and TOUCH 9, the corresponding GPIO should be exchanged.
Details: In the arduino esp32 library, TOUCHPAD T8 corresponds to GPIO33, TOUCHPAD T9 corresponds to GPIO32, but in my actual test, TOUCHPAD T8 corresponds to GPIO32, and TOUCHPAD T9 corresponds to GPIO33. Is this a problem in the arduino library?

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp

//Change the code below by your sketch
//Change the code below by your sketch
void setup() {
 
  // put your setup code here, to run once:
  Serial.begin(9600);
  Serial.printf("idf_version: %s\n", esp_get_idf_version());//idf_version: v3.2.3-14-gd3e562907
//  pinMode(32, INPUT);
//  pinMode(33, INPUT);
//    pinMode(27, INPUT);

}

void loop() {
  // put your main code here, to run repeatedly:
  int val_14 = touchRead(T6);  //14
  int val_27 = touchRead(T7);  //27
  int val_33 = touchRead(T8);  //33
  int val_32 = touchRead(T9);  //32
  Serial.printf("%d ,%d ,%d ,%d\n", val_14, val_27, val_33, val_32);

}
```
### Debug Messages:
I tested four case:touch GPIO 14(T6), touch GPIO 27(T7), touch GPIO 33(T8), touch GPIO 32(T9),
The log in case 1(GPIO 14(T6)):right
`0 ,110 ,122 ,126`
The log in case 2(GPIO 27(T7)):right
`106 ,0 ,122 ,125`
The log in case 3(GPIO 33(T8)):maybe wrong,I think the third value should be zero.
`107 ,110 ,120 ,0`
The log in case 4(GPIO 32(T9)):maybe wrong,I think the fourth value should be zero.
`107 ,110 ,0 ,126`

PR #3620, opened 2020-01-08
bugfix(touchpad):swap touch8 and touch9 #3584

TreeNewbeer:


UnexpectedMaker:
I dont know if this is the cause of the **current** issue where touch 8 and 9 are reversed, but as it stands as of Oct 2020, 8 and 9 are reversed and this has been tested on ESP32 Arduino 1.0.4 on multiple board types.

@me-no-dev Can this please be looked into? The issue has been confirmed by me and at least one other person and it 100% reproducible.

Thanks :)

PR #3628, opened 2020-01-09
Update CaptivePortal.ino

DanielLester83:
Small change to insure wifi starts off and avoid crashing


DanielLester83:
Also if you change the ip to 8.8.4.4 it fixes the lack of a 2nd dns option because it makes both dns options on android 8.8.4.4. This ensures all DNS traffic goes through right place and causes a login page to automatically popup.

`IPAddress apIP(8, 8, 4, 4); //solution for android & lack of 2nd DNS option'

me-no-dev:
cool! merged :)

PR #3638, opened 2020-01-13
Added the channel parameter to the scanNetworks() function of WiFiScan class

ingframin:
Added the channel parameter to the scan function in order to select which channel to scan.
The default value is 0 (= all channels) to preserve backward compatibility.


me-no-dev:
thanks!

PR #3377, opened 2019-10-17
Use IDF's ADC Driver and Add analogReadMilliVolts

me-no-dev:

PR #3648, opened 2020-01-20
Fix commit d8b2098 (PR #3377) which includes esp_adc_cal

robert-alfaro:
Fix recent commit d8b2098 (PR #3377) which includes `esp_adc_cal` component but does not update `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,7 +205,7 @@ set(COMPONENT_ADD_INCLUDEDIRS
 
 set(COMPONENT_PRIV_INCLUDEDIRS cores/esp32/libb64)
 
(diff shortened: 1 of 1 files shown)


robert-alfaro:
I am using arduino-esp32 as a component in an IDF project..I'm not sure if Arduino or PlatformIO supports CMake, yet, however having CI would be great to catch issues using other (future default) build environments.

me-no-dev:
as soon as we move to 4.x and we can use Cmake to build the Arduino libs, the CI will do that. We have even now a check that makes sure that all files are listed in cmake, but we can not test against the libs without cmake to begin with :)

PR #3655, opened 2020-01-21
Greatly reduced error rate of WMath map() roundtrip

dok-net:
Half, or zero errors, depending on in/out…… ranges, for round-trip mapping at the same performance.
(Based on "improved_map" from ESP8266's Servo.cpp)

### DIFF
--- a/cores/esp32/WMath.cpp
+++ b/cores/esp32/WMath.cpp
@@ -66,11 +66,11 @@ long random(long howsmall, long howbig)
 }
 
 long map(long x, long in_min, long in_max, long out_min, long out_max) {
-    long divisor = (in_max - in_min);
-    if(divisor == 0){
-        return -1; //AVR returns -1, SAM returns 0
-    }
-    return (x - in_min) * (out_max - out_min) / divisor + out_min;
+    const long dividend = out_max - out_min;
+    const long divisor = in_max - in_min;
+    const long delta = x - in_min;
+
+    return (delta * dividend + (divisor / 2)) / divisor + out_min;
 }
 
 unsigned int makeWord(unsigned int w)


dok-net:
Dropped divide-by-zero kludge emulation of AVR CPUs.
Here, too, the sketch to prove the change:
```c++
uint32_t expired;
uint32_t iterations;

void setup()
{
	Serial.begin(115200);
	while (!Serial);
	Serial.println("check map");
	expired = 0;
	iterations = 0;
}

constexpr int imin = 0;
constexpr int imax = 180;
constexpr int omin = 1000;
constexpr int omax = 2000;

int i = 0;
int fails = 0;

void loop()
{
	++iterations;
	i = (i + 1) % (imax + 1);

#if defined(ESP8266) || defined(ESP32)
	cli();
	volatile auto start = ESP.getCycleCount();
#else
	// Set Timer 1 to normal mode at F_CPU.
	TCCR1A = 0;
	TCCR1B = 1;
	cli();
	volatile uint16_t start = TCNT1;
#endif

	volatile auto o = map(i, imin, imax, omin, omax);

#if defined(ESP8266) || defined(ESP32)
	sei();
	volatile auto cycles = ESP.getCycleCount() - start;
#else
	sei();
	volatile uint16_t finish = TCNT1;
	uint16_t cycles = finish - start;
#endif

	expired += cycles;
	
	if (i != map(o, omin, omax, imin, imax))
	{
		++fails;
	}

	if (iterations > 100000)
	{
		Serial.print("Cycles per map(...) = ");	Serial.print(expired / iterations);
		Serial.print(", fails = ");	Serial.print(fails);
		Serial.println();
		expired = 0;
		iterations = 0;
		fails = 0;
	}
}
```

dok-net:
https://github.com/arduino/ArduinoCore-avr/pull/309
https://github.com/esp8266/Arduino/pull/7027

Issue #3657, opened 2020-01-23
Potential deadlock in BluetoothSerial


jcdevel:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			current git master
IDE name:						Arduino IDE
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Ubuntu 18.04 LTS

### Description:
I am sending timestamps every 100 ms from my ESP32 node via BluetoothSerial (do not ask why, just taking baby steps with Bluetooth and ESP32). I am using an Android Bluetooth Terminal app to print the timestamps on my phone. I have noticed that if I close the connection from the phone and open it again, after a couple times the ESP32 node freezes. A hard reset is needed to bring it back to life.

I have performed some analysis and debugging and I have traced the root cause to the blocking calls in the transmission task of the BluetoothSerial component. I have prepared a couple of patches that fix the issue. I will do a pull-request for revision.

### Sketch:
```cpp

#include <stdlib.h>
#include <string.h>
#include <BluetoothSerial.h>

BluetoothSerial SerialBT;

void setup() {
  Serial.begin(115200);
  SerialBT.begin("ESP32test"); //Bluetooth device name
  Serial.println("Start!");
}

void loop() {
  static unsigned long prev;
  unsigned long now;
  char out_buf[32];
  unsigned int len;

  now = millis();
  if ((now - prev) >= 100) {
    //periodic message
    itoa(now, out_buf, 10);
    len = strlen(out_buf);
    out_buf[len] = '\n';
    out_buf[len+1] = '0';
    SerialBT.write((uint8_t *)out_buf, len+1);
    prev = now;
  }
}

```

emmabeebe:
I am also having this issue. I am using serialBT to send data between two esp32 devices and when I manually pull the power to one of the devices the remaining device just hangs and freezes.

jcdevel:
@emmabeebe Could you verify if my PR fixes the issue on your side? Thanks!

Benik3:
I tested it with my GPS Serial to BT, where I encountered this problem pretty often and it looks, that it's fixed :)

jcdevel:
Could somebody review this issue? (PR #3658)

emmabeebe:
I have also reviewed it and the issue has been fixed.

max5555:
Make it faster

```
#include <Arduino.h>
#include "BluetoothSerial.h"

[...]

PR #3658, opened 2020-01-23
Fix potential deadlock in BtSerial if connection is closed

jcdevel:
This PR fixes issue #3657. Root cause is traced down to blocking calls in the transmission task within the BluetoothSerial component. The fix is explained in the commit message.

### DIFF
--- a/libraries/BluetoothSerial/src/BluetoothSerial.cpp
+++ b/libraries/BluetoothSerial/src/BluetoothSerial.cpp
@@ -240,7 +240,7 @@ static void esp_spp_cb(esp_spp_cb_event_t event, esp_spp_cb_param_t *param)
             esp_spp_disconnect(param->open.handle);
         }
         xEventGroupClearBits(_spp_event_group, SPP_DISCONNECTED);
-        xEventGroupSetBits(_spp_event_group, SPP_CONNECTED);
+        xEventGroupSetBits(_spp_event_group, SPP_CONNECTED | SPP_CONGESTED);
         break;
 
     case ESP_SPP_CLOSE_EVT://Client connection closed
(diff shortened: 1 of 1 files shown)


libraries/BluetoothSerial/src/BluetoothSerial.cpp
```diff
@@ -241,6 +243,7 @@ static void esp_spp_cb(esp_spp_cb_event_t event, esp_spp_cb_param_t *param)
         }
         xEventGroupClearBits(_spp_event_group, SPP_DISCONNECTED);
         xEventGroupSetBits(_spp_event_group, SPP_CONNECTED);
+        xEventGroupSetBits(_spp_event_group, SPP_CONGESTED);
```
stickbreaker: Why don't you roll this up into one `xEventGroupSetBits()`
```c++
 xEventGroupSetBits(_spp_event_group, SPP_CONNECTED | SPP_CONGESTED );
```

jcdevel: Yes, you are right. To be honest, I did not give it much thought. I do not think it makes a great difference in performance nor in code readability, but I can change it if necessary for merging.

jcdevel:
Could this PR be considered for merging by an authoritative reviewer? Thanks.

madprunes:
Thank you for this, unfortunately I have spent hours over the last few days tracking down this same issue, only to find a google search of the line which I found causing the issue resulted in a fix from 8 months ago which still hasn't been merged. I had done work arounds which stopped the deadlock occurring in most situations apart from a disconnect because of bad reception.

jcdevel:
There are 152 un-merged PRs up to today, most of them still pending from review. I understand everybody is busy and there is little time for official reviewers to make the extra effort of reviewing and merging PRs from the community. However, I created mine 9 months ago. All this makes me think there is no real interest for public contributions.

Anyway, it is not my project and it is not my call to decide what is is merged and what it is not. I will upload my fixes and improvements in my fork, so anyone interested can grab them from there.

Issue #2004, opened 2018-10-28
HardwareSerial incorrect timing with non standard baudrate


aster94:
### Hardware:
Board:							ESP32 Lolin 32 mini
Core Installation/update date:			last from package manager (1.0.0?)
IDE name:						vscode
Flash Frequency:					tried both 40 and 80MHz
Upload Speed:					460800
Computer OS:                                         xubuntu 18.04

### Description:
Using a baudrate different form the standard (10400) brings to incorrect timing of the serial communication
I made a few samples with a salea logic analizer (clone) with the above settings:
![config](https://user-images.githubusercontent.com/21184718/47618372-6e1a3a00-dad2-11e8-9b92-e220e11cf5e1.png)

Correct behaviour from an arduino mega:
![mega 10400 o](https://user-images.githubusercontent.com/21184718/47618274-78880400-dad1-11e8-8973-0e07c4b8e68a.png)
[mega 10400 o.txt](https://github.com/espressif/arduino-esp32/files/2522911/mega.10400.o.txt)

while esp plays a few jokes
![esp 10400 o](https://user-images.githubusercontent.com/21184718/47618287-93f30f00-dad1-11e8-9582-b7434fa1391c.png)
[esp 10400 o.txt](https://github.com/espressif/arduino-esp32/files/2522912/esp.10400.o.txt)

same at 80mhz
![esp 80 10400 o](https://user-images.githubusercontent.com/21184718/47618292-a10ffe00-dad1-11e8-97ae-cb63f4ef4bcb.png)
[esp 80 10400 o.txt](https://github.com/espressif/arduino-esp32/files/2522913/esp.80.10400.o.txt)

same happens if i remove the odd parity bit
![esp 10400 n](https://user-images.githubusercontent.com/21184718/47618322-db799b00-dad1-11e8-8e5a-facb39ba4624.png)
[esp 10400 n.txt](https://github.com/espressif/arduino-esp32/files/2522916/esp.10400.n.txt)

while at a standard baud rate (115200) all is ok
![esp 115200 n](https://user-images.githubusercontent.com/21184718/47618304-b9801880-dad1-11e8-837a-bd2d4a804a2f.png)
[esp 115200 n.txt](https://github.com/espressif/arduino-esp32/files/2522914/esp.115200.n.txt)

### Sketch:
for esp
```
HardwareSerial ser(2);

void setup()
{
    ser.begin(10400,SERIAL_8O1);
}

void loop()
{
[...]

PR #3664, opened 2020-01-24
Minimize HardwareSerial Receive and Transmit delays

hreintke:
In current HardwareSerial/esp-uart-hal 
- on Transmit
In the uart configutarion setting there is a parameter tx_idle_num.
tx_idle_num is : idle interval after tx FIFO is empty(unit: the time it takes to send one bit under current baudrate)
This tx_idle_num is set to 256.
Resulting in a forced transmit idle time after txFifo empty to next.

This PR sets the tx_idle_num value to zero at uartBegin(). 
Result is an immediate start of transmission when a char is send.

- On Receive
The chars are first received in rxFifo buffer which is transfered to HardwareSerial Queue on either fifo filled with 112 chars or 2 bytes idle time. 
Resulting in a delay at application level processing the incoming messages.

This PR takes the characters in rxFifo into account when calling available(), read() and peek().
Result is an immediate transfer from queue and/or fifo when application requests data.

To demonstrate/test I have an application doing message transfer at 19200 baud
- send a 6 byte message
- receives a 6 bytes message
- responds to that with a 6 byte message.

Running with Master : duration 19.3 ms
Running with PR : duration 9.3 ms

SerialMaster

SerialPR

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -208,6 +208,11 @@ uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rx
         uart->dev->conf0.stop_bit_num = ONE_STOP_BITS_CONF;
         uart->dev->rs485_conf.dl1_en = 1;
     }
+
+    // tx_idle_num : idle interval after tx FIFO is empty(unit: the time it takes to send one bit under current baudrate)
+    // Setting it to 0 prevents line idle time/delays when sending messages with small intervals
+    uart->dev->idle_conf.tx_idle_num = 0;  //
+
     UART_MUTEX_UNLOCK();
 
     if(rxPin != -1) {
@@ -265,7 +270,7 @@ uint32_t uartAvailable(uart_t* uart)
     if(uart == NULL || uart->queue == NULL) {
         return 0;
     }
-    return uxQueueMessagesWaiting(uart->queue);
+    return (uxQueueMessagesWaiting(uart->queue) + uart->dev->status.rxfifo_cnt) ;
 }
 
 uint32_t uartAvailableForWrite(uart_t* uart)
@@ -276,12 +281,27 @@ uint32_t uartAvailableForWrite(uart_t* uart)
     return 0x7f - uart->dev->status.txfifo_cnt;
 }
 
+void uartRxFifoToQueue(uart_t* uart)
+{
+	uint8_t c;
+    UART_MUTEX_LOCK();
+    while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
+        c = uart->dev->fifo.rw_byte;
+        xQueueSend(uart->queue, &c, 0);
+    }
+    UART_MUTEX_UNLOCK();
+}
+
 uint8_t uartRead(uart_t* uart)
 {
     if(uart == NULL || uart->queue == NULL) {
         return 0;
     }
     uint8_t c;
+    if ((uxQueueMessagesWaiting(uart->queue) == 0) && (uart->dev->status.rxfifo_cnt > 0))
+    {
+    	uartRxFifoToQueue(uart);
+    }
     if(xQueueReceive(uart->queue, &c, 0)) {
         return c;
     }
@@ -294,6 +314,10 @@ uint8_t uartPeek(uart_t* uart)
         return 0;
     }
     uint8_t c;
+    if ((uxQueueMessagesWaiting(uart->queue) == 0) && (uart->dev->status.rxfifo_cnt > 0))
+    {
+    	uartRxFifoToQueue(uart);
+    }
     if(xQueuePeek(uart->queue, &c, 0)) {
         return c;
     }
--- a/cores/esp32/esp32-hal-uart.h
+++ b/cores/esp32/esp32-hal-uart.h
@@ -59,8 +59,6 @@ uint32_t uartAvailableForWrite(uart_t* uart);
 uint8_t uartRead(uart_t* uart);
 uint8_t uartPeek(uart_t* uart);
 
-void uartRxFifoToQueue(uart_t* uart);
-
 void uartWrite(uart_t* uart, uint8_t c);
 void uartWriteBuf(uart_t* uart, const uint8_t * data, size_t len);
 
--- a/cores/esp32/esp32-hal-uart.h
+++ b/cores/esp32/esp32-hal-uart.h
@@ -59,6 +59,8 @@ uint32_t uartAvailableForWrite(uart_t* uart);
 uint8_t uartRead(uart_t* uart);
 uint8_t uartPeek(uart_t* uart);
 
+void uartRxFifoToQueue(uart_t* uart);
+
 void uartWrite(uart_t* uart, uint8_t c);
 void uartWriteBuf(uart_t* uart, const uint8_t * data, size_t len);
 


cores/esp32/esp32-hal-uart.c
```diff
@@ -294,6 +314,10 @@ uint8_t uartPeek(uart_t* uart)
         return 0;
     }
     uint8_t c;
+    if ((uxQueueMessagesWaiting(uart->queue) == 0) && (uart->dev->status.rxfifo_cnt > 0))
```
stickbreaker: If queue is not zero, then you ignore the data in the fifo?

I would think you would always add the `rxfifo_cnt` to the `uartPeek()` result.
```c++
if (uart->dev->status.rxfifo_cnt >0)
```

cores/esp32/esp32-hal-uart.c
```diff
@@ -208,6 +208,11 @@ uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rx
         uart->dev->conf0.stop_bit_num = ONE_STOP_BITS_CONF;
         uart->dev->rs485_conf.dl1_en = 1;
     }
+
+    // tx_idle_num : idle interval after tx FIFO is empty(unit: the time it takes to send one bit under current baudrate)
+    // Setting it to 0 prevents line idle time/delays when sending messages with small intervals
+    uart->dev->idle_conf.tx_idle_num = 0;  //
```
stickbreaker: Good Catch !!!!!!!!!!!!!!! :1st_place_medal:

cores/esp32/esp32-hal-uart.h
```diff
@@ -59,6 +59,8 @@ uint32_t uartAvailableForWrite(uart_t* uart);
 uint8_t uartRead(uart_t* uart);
 uint8_t uartPeek(uart_t* uart);
 
+void uartRxFifoToQueue(uart_t* uart);
```
stickbreaker: I would not export the this function.  It should only be call from inside `esp32-hal-uart.c`

cores/esp32/esp32-hal-uart.c
hreintke: If queue is not zero, the peek result is the first in the queue -> no need to transfer from fifo to queue now. 
Alternative is in both read and peek and possible even in available to always transfer all fifo to queue. 
It's a design decision, not functional, I'll change if decided to do the other way.
Just let me know

cores/esp32/esp32-hal-uart.h
hreintke: Agree, I just wasn't sure whether all .c functions needed to be in .h because of unexpected usage.
I'll remove from .h

cores/esp32/esp32-hal-uart.c
stickbreaker: The code sequence where this could matter is:
```c++
int packet_length = 10;
while (Serial.peek() < packet_length){
  delay(5);
}
processPacket();
```
So, I'm waiting for a complete packet before I process it. But this loop would not pass until rxTimeout or rxFifioFull.

You could decide not to empty the fifo, but just add the rxFifoCount to the returned.

What do you think? 

Also, there is something weird with the rxFifoCount, do you see in the original ISR coded where the "empty" test is rxFifoCount OR  `(uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)`?  
It looks like rxFifoCount can be zero and the FiFo is FULL? @me-no-dev ?
perhaps rxFifoCount is not always correct?

Chuck.

hreintke: Serial.peek() doesn't return the size of the queue but the fist char in the queue, Serial.available() does return the number of chars available.
So your example should be
```
int packet_length = 10;
while (Serial.available() < packet_length){
  delay(5);
}
processPacket();
```
And that is providing the correct answer. 

On the rxFifocount. 
I agree with you I expect(ed) that the loop 
while(uart->dev->status.rxfifo_cnt) {... } 
would do the job..
But because of the current code I kept it identical.

dflogeras:
I am seeing an issue when using Serial1.readBytes() that I have bisected down to this commit.

I'm using the serial port to send a query to a sensor device, which then responds with a fixed size (1024 byte) packet.  I do this over and over.  After this commit, it sometimes returns from readBytes() claiming it has read 1024 bytes, but it could not have (see attached pictures).

So first pic is a good transaction.  Channel 5 is a GPIO that goes high while I enter my readBytes() call, then is set low again when I have successfully read 1024 bytes.  As expected it does not go low until after all bytes have been transmitted.  (Note Channel 6 is labelled 'tx' from the point of view of the sensor, not the ESP32).
good

Here's what happens fairly frequently after this commit.  As you can see, Channel 5 goes back low again prior to the data even arriving at the pin.  The software is getting data from somewhere, but it cannot possibly be the uart as the data hasn't arrived yet.
bad

Reverting this commit, and it works as expected.

me-no-dev:
@dflogeras have you looked at the bytes that it claims it have received?
Can you share what the differences are and from which byte it happens?

me-no-dev:
@dflogeras also if you can play with `uart->dev->idle_conf.tx_idle_num = 0;` and try a positive value like 1 or 10?

hreintke:
@dflogeras 
I am trying to simulate your environment so see if I can replicate the issue.
At what baudrate are you sending ? 

PS Are you able to do local patches to esp-hal-uart.c to verify solutions ?

dflogeras:
I am running it at 460800 baud during that phase.

I can certainly patch esp-hal-uart.c if it is built as part of the Arduino build process.  I am working on another job today, but will get back to this tomorrow and can try tx_idle_num changes.

I didn't diagnose down to the byte yet, for I will need to change my test setup for triggering on when it happens, which I can't do until at least tomorrow.

hreintke:
@dflogeras 
Can you update `uint8_t uartRead(uart_t* uart)`, line 295 of esp-hal-uart.c to :  
```
uint8_t uartRead(uart_t* uart)
{
    if(uart == NULL || uart->queue == NULL) {
        return 0;
    }
    uint8_t c;
    UART_MUTEX_LOCK();
    if ((uxQueueMessagesWaiting(uart->queue) == 0) && (uart->dev->status.rxfifo_cnt > 0))
    {
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            c = uart->dev->fifo.rw_byte;
            xQueueSend(uart->queue, &c, 0);
        }
    }
    UART_MUTEX_UNLOCK();
    if(xQueueReceive(uart->queue, &c, 0)) {
        return c;
    }
    return 0;
}
```
And see if there is different behavior ?

@me-no-dev : 
This way I have the ` (uart->dev->status.rxfifo_cnt > 0))` check within the `UART_MUTEX_LOCK();`

dflogeras:
Patching esp32-hal-uart.c as above: now instead of uartRead() returning early with incorrect data, now it always times out with a short count (even though I can see the sensor is sending 1024 bytes on the logic analyzer).  Seemed to be always in the 700s, but not always the same number of bytes.

Trying values 1 and 10 for tx_idle_num both still exhibited the original behaviour (returning a full read, but clearly early and with incorrect data).

hreintke:
@dflogeras 
I am running the test using two ESP32's.
Think I reproduced the issue, using both the original as the patched version.
Needed to go to 921600 baud to get it faulting.

First observation : The first byte being wrong in a message is always nr 256. 
Is there a possibility that you can verify if that is at your side too ?

hreintke:
@dflogeras 
With this uartRead() I have no fails. Can you test in your environment ?
```
uint8_t uartRead(uart_t* uart)
{
    if(uart == NULL || uart->queue == NULL) {
        return 0;
    }
    uint8_t c;
    UART_MUTEX_LOCK();
    //disable interrupts
    uart->dev->int_ena.val = 0;
    if ((uxQueueMessagesWaiting(uart->queue) == 0) && (uart->dev->status.rxfifo_cnt > 0))
    {
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            c = uart->dev->fifo.rw_byte;
            xQueueSend(uart->queue, &c, 0);
        }
    }
    //enable interrupts
    UART_MUTEX_UNLOCK();
    if(xQueueReceive(uart->queue, &c, 0)) {
        return c;
    }
    return 0;
}
```

dflogeras:
I will check when I am home later in the day.

In the meantime, can you verify that the problem goes away for you by reverting this commit?

hreintke:
Ok, will verify and report. Probably tomorrow.

dflogeras:
OK.  Your last patch on top of this commit had the same behaviour as your previous patch: timeouts after the default 1s with short byte count (in the 730-750 bytes range when I asked for 1024)

Your previous comment of needing to go to 800kBaud to reproduce made me think of something else though.  I forgot to mention, that I am running my CPU at 80MHz (for power consumption).  If I switch to 160 or 240, I get the full data (both this commit unpatched, as well as with your latest patch).

What I don't know is what that means.  Is running the CPU faster just hiding a race condition?  I am hard pressed to believe that a 80MHz CPU cannot keep up with a 400kBaud (45 kBytes/s), while doing almost nothing else.

hreintke:
@dflogeras Good to mention.
With 80Mhz and 460800 baud I get rxFifo overflow errors.
I enabled the interrupt for this : 
```
void uartEnableInterrupt(uart_t* uart)
{
    UART_MUTEX_LOCK();
    uart->dev->conf1.rxfifo_full_thrhd = 112;
    uart->dev->conf1.rx_tout_thrhd = 2;
    uart->dev->int_ena.rxfifo_ovf = 1; // add for ovf test

    esp_intr_alloc(UART_INTR_SOURCE(uart->num), (int)ESP_INTR_FLAG_IRAM, _uart_isr, NULL, &uart->intr_handle);
    UART_MUTEX_UNLOCK();
}
```
Add check in isr
```
static void IRAM_ATTR _uart_isr(void *arg)
{
    uint8_t i, c;
    BaseType_t xHigherPriorityTaskWoken;
    uart_t* uart;
    
    uart = &_uart_bus_array[1];        // 0 for serial, 1 for serial1, 2 for serial2
    if (uart->dev->int_st.rxfifo_ovf);  // overflow occurred
    {
    	uart->dev->int_clr.rxfifo_ovf = 1; // clear overflow interrupt bit
    	digitalWrite(27, !digitalRead(27));
    }

    for(i=0;i<3;i++){
```
And i see the GPIO 27 toggle -> loosing chars because of fifo overflow.
Changing to 160Mhz -> no ovf

dflogeras:
Ok good we're on the same page now.  So, the hardware literally cannot keep up @ 80MHz?

dflogeras:
Also, to put it another way, why was it not losing characters prior to this commit?

hreintke:
It's the software which can't keep up at 80Mhz, that should empty the fifo.

dflogeras:
Sorry, yes that's what I mean, the hardware being the CPU @ 80Mhz

That still doesn't explain why it was (seemingly) working prior to this.

hreintke:
Will check on that in the coming days.

stickbreaker:
@hreintke @dflogeras you need a minimum number of cpu cycle to service an interrupt. With the I2C subsystem, I adjust the FIFO thresholds inversely with cpu clock and limit maximum I2C bus clock at lower cpu speeds.  It seems to work at 800k Baud and 112 as Fifo Full Threshold, with Fifo Threshold at 112 that leaves 16 byte periods to start emptying the fifo before overflows happen. That calculates to 200us with a 80MHz clock that is 16k clock cycles.  The interrupt would trigger about every 1.4ms.  It looks like that is not enough.

Try reducing the Fifo threshold inversely with cpu clock and Baud,  If 112 works at 240mhz and 800KHz and you divide cpu by 3, divide threshold by 3, So, 112 becomes 38.  It still uses the full Fifo, but it gives you additional time to react.

Chuck.

dflogeras:
@stickbreaker I agree with your assessment of reducing the 'high water alarm' on the fifo.

However in practice it didn't fix the problem I am seeing.  I reduced the threshold to 32 (in addition to @hreintke last patch from the above comment), and in received even less short-count data after timing out.

hreintke:
@dflogeras : 
I cleaned and rechecked my development environment and am checking exact timings.

Just to be sure : Are you emptying the uart buffers with `while Serial.available() { Serial.read()} ` before requesting the new data ?

hreintke:
@stickbreaker @me-no-dev 
I have been measuring interrupt service times on uart .
This is the full ISR : 
```
static void IRAM_ATTR _uart_isr(void *arg)
{
    uint8_t i, c;
    BaseType_t xHigherPriorityTaskWoken;
    uart_t* uart;

    for(i=0;i<3;i++){
        uart = &_uart_bus_array[i];
            continue;
        }
        uart->dev->int_clr.rxfifo_full = 1;
        uart->dev->int_clr.frm_err = 1;
        uart->dev->int_clr.rxfifo_tout = 1;
        if (i==1) digitalWrite(26,HIGH);
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            c = uart->dev->fifo.rw_byte;
            }
        }
        if (i==1) digitalWrite(26,LOW);
    }

    if (xHigherPriorityTaskWoken) {
        portYIELD_FROM_ISR();
    }
}
```
 The detail : 
```
            }
```
has a check on full queue before sending to queue.
But xQueueSendFromISR returns false when trying to write to a full queue.
When updated to : 
```
```
            if(uart->queue != NULL) {
            }
```
It does have the same functionality.

ISR Timing 
80 Mhz, 112 fifo, current version : 1.4 msec
80 Mhz, 112 fifo, updated version : 0.7 msec
160 Mhz, 112 fifo, current version : 0.55 msec
160 Mhz, 112 fifo, updated version : 0.32 msec

What is your opinion on this update ?

**Comment by stickbreaker** [comment]
@hreintke 
```c++
static void IRAM_ATTR _uart_isr(void *arg)
{
    uint8_t i, c;
    BaseType_t xHigherPriorityTaskWoken = FALSE;
    uart_t* uart;

    for(i=0;i<3;i++){
        uart = &_uart_bus_array[i];
            continue;
        }
        uart->dev->int_clr.rxfifo_full = 1;
        uart->dev->int_clr.frm_err = 1;
        uart->dev->int_clr.rxfifo_tout = 1;
        if (i==1) digitalWrite(26,HIGH);
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            c = uart->dev->fifo.rw_byte;
            if(uart->queue != NULL ) {
            }
        }
        if (i==1) digitalWrite(26,LOW);
    }

    if (xHigherPriorityTaskWoken) {
        portYIELD_FROM_ISR();
    }
}
```
I don't see any problem with your suggested changes. My only question relates back to the choice of servicing all three UARTs whenever one of them triggers an interrupt.  The `for()` loop. I haven't convinced Me-No-Dev yet :grinning: 

Remember to remove your debug (`digitalWrite()`) before you post the merge request.
Chuck.

stickbreaker:
I just had a stupid idea, the current code drops characters when the queue becomes full.  What do you think about a condition to stop emptying the fifo if the fifo is lower than the interrupt trigger and the queue is full, allow foreground a chance to process and save the chars in the fifo?

This code is UNTESTED and Just off-the-cuff, to BBW(Buyer Be Ware!)
```c++
static void IRAM_ATTR _uart_isr(void *arg)
{
    uint8_t i, c;
    BaseType_t xHigherPriorityTaskWoken = FALSE;
    uart_t* uart;

    for(i=0;i<3;i++){
        uart = &_uart_bus_array[i];
            continue;
        }

  // adjust rxfifo empty point based on latency between interrupt triggering and ISR firing
  // minimum free space of 8 characters
        if (uart->dev->int_st.rxfifo_full ) { //only adjust if this UART triggered fifoFull
            if( uart->dev->status.rxfifo_cnt > uart->dev->conf1.rxfifo_full_thrd){ // delayed service
                if(  (128 - uart->dev->status.rxfifo_cnt ) < 8 ) { // almost overflowed Fifo
                    if (uart->dev->conf1.rxfifo_full_thrhd > 32 ) { // min fifo level
                        uart->dev->conf1.rxfifo_full_thrhd -= 1; // trigger earlier 
                    }
                }    
            }
            else if ( uart->dev->conf1.rxfifo_full_thrd <120){// under utilizing Fifo
                uart->dev->confg1.rxfifo_full_thrd += 1;
            }
        }
// end of fifo trigger adjust

        uart->dev->int_clr.rxfifo_full = 1;
        uart->dev->int_clr.frm_err = 1;
        uart->dev->int_clr.rxfifo_tout = 1;
        if (i==1) digitalWrite(26,HIGH);
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            c = uart->dev->fifo.rw_byte;
            if(uart->queue != NULL ) {
                if ( !xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken)){
                // queue is full, exit if fifo level is below fifo trigger
                    if( uart->dev->status.rxfifo_cnt <   uart->dev->conf1.rxfifo_full_thrhd )
                       continue;
                }
              }
        }
        if (i==1) digitalWrite(26,LOW);
    }

    if (xHigherPriorityTaskWoken) {
        portYIELD_FROM_ISR();
    }
}
```
@dflogeras might be seeing the character loss because the queue is overflowing, not because the fifo is overflowing.  If the interrupt is triggering more often then his fore ground process does not have a chance to process before the `Serial()` queue is overflowed.

Chuck.

hreintke:
@stickbreaker 
Will go in more detail through your suggestion but the problems is that you have to get the char from the uart `c = uart->dev->fifo.rw_byte;` before you know that the buffer is full.
That character is always lost.

stickbreaker:
You are correct. The only way not to loose that char would be to check for QueueFull before attempting the add. But, based on your results.  That check before call is expensive in time.

>ISR Timing 
80 Mhz, 112 fifo, current version : 1.4 msec
80 Mhz, 112 fifo, updated version : 0.7 msec
160 Mhz, 112 fifo, current version : 0.55 msec
160 Mhz, 112 fifo, updated version : 0.32 msec

One way around this would be to store the queue size in `UART` record and use
`UBaseType_t uxQueueMessagesWaitingFromISR(const QueueHandle_t xQueue)` to see the available space
```c++
//cut
        uart->dev->int_clr.frm_err = 1;
        uart->dev->int_clr.rxfifo_tout = 1;
        if (i==1) digitalWrite(26,HIGH);
        size_t freeCount = 0;
        if(uart->queue ) {
            freeCount = uart->queueSize - uxQueueMessagesWaitingFromISR(uart->queue );
        }
        while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
            if(freeCount-- > 0) {
                c = uart->dev->fifo.rw_byte;
                !xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
            )
            else {
                // queue is full, exit if fifo level is below fifo trigger
                    if( uart->dev->status.rxfifo_cnt <   uart->dev->conf1.rxfifo_full_thrhd ){
                        break; // stop emptying fifo, Queue is full, but fifo has room
                    }
                    else { // empty one char from FIFO and discard
                         c = uart->dev->fifo.rw_byte;
                    }
            }
        }
        if (i==1) digitalWrite(26,LOW);
    }
```

Oops, used `continue` when I meant `break` :grimacing: 

For this code to work, the `size_t queueSize;` needs to be added to the `uart` structure.

Chuck.

stickbreaker:
This is better inner loop code:'
```c++
    while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
        c = uart->dev->fifo.rw_byte;
        if(freeCount-- > 0) {
          !xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
        }
        else {
                // queue is full, exit if fifo level is below fifo trigger
            if( uart->dev->status.rxfifo_cnt <   uart->dev->conf1.rxfifo_full_thrhd ) {
                break; // stop emptying fifo, Queue is full, but fifo has room
            }
        }
    }
    if (i==1) digitalWrite(26,LOW);
```
Always have to empty at least one char from fifo. So, it will only loose the char if rxFifo_Full interrupt triggered and the Queue is already full.

Chuck.

dflogeras:
I can confirm that just by removing the xQueueIsQueueFullFromISR from the isr code, I can now keep my head above water @ 80MHz.

hreintke:
@stickbreaker 
I was thinking on saving the already taken char in the uart struct. 
When a new interrupt comes, first try the saved char, if that succeeds, take next char and give that a try.
That prevents the need of keeping track of queue size.
I'll do some testing on that, probably tomorrow.

What also might help is moving to Ringbuffer
 [https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/freertos_additions.html](url) 
That has the possibility to add multiple chars in one call, further limiting the number of queue calls within the isr.

stickbreaker:
@dflogeras Have you verified where the loss of data was happening in your case?  Was  the loss because of rxfifo overflow or because the Queue was full?

Chuck.

stickbreaker:
@hreintke saving the "extra" character inside the `uart` structure would create other requirement, you would have to test if the "extra" character is valid and move it during every `read()` and `available()`, `isr()`.  

Another possible problem with leaving the data in the rxfifo is when does `rxfifo_tout` trigger?  Would it continuously re-trigger every `tout` period until the rxfifo is empty?

All of my coding is untested theoretical. So, there are probably many hidden pitfalls.

Chuck.

hreintke:
Submitted PR  #3713 with bugfix and isr improvement.

@stickbreaker 
Another day, new thoughts on not discarding chars when queue is full.
In principle this is an application issue : Not emptying the buffer quick enough as chars are coming in.
It can be improved somewhat by additional checks but ...
- It is no full solution, only delaying the issue
- it makes the isr more complex
- The application developer has easy solutions available
   - Lower the baudrate
   - Increase the buffer size `HardwareSerial::setRxBufferSize`
   - Increase cpu speed

My suggestion would be to leave it as is.

On fifo size.
I would set the fifo size at `Serial.begin()' depending on baudrate and cpu speed and not change it during isr processing.
Maybe add a method to Serial to set it different from the default.

What is your opinion ?

atanisoft:
> I would set the fifo size at `Serial.begin()' depending on baudrate and cpu speed and not change it during isr processing.

It should change as part of the CPU frequency callback which adjusts the baud rate parameters as well. It should not be adjusted inside the ISR.

stickbreaker:
@hreintke @atanisoft  I agree with both of you, my code would add complexity and uncertainty.  Neither of which is good in a foundational driver.  

@hreintke maybe have some persistent flag "queueOverfow", or a `log_e("Overflow uart(%d)",i);` so the user gets some notification that their foreground task is not processing the queue quick enough?

@atanisoft any idea on a formula to generate the `rxfifo_full` value? the 112 was me-no-dev's empirical solution.  I was hoping that my auto adjuster code would generate a value that would match the users application.

Chuck.

aster94:
solved https://github.com/espressif/arduino-esp32/issues/2004

PR #3667, opened 2020-01-25
IDF release/v4.0

me-no-dev:
esp-face: master 9fc6796
esp32-camera: master b1c9712


platform.txt
```diff
@@ -25,7 +25,7 @@ compiler.sdk.path={runtime.platform.path}/tools/sdk
...
 
 compiler.c.cmd=xtensa-esp32-elf-gcc
-compiler.c.flags=-std=gnu99 -Os -g3 -Wno-frame-address -ffunction-sections -fdata-sections -fstrict-volatile-bitfields -mlongcalls -nostdlib -fstack-protector {compiler.warning_flags} -Wno-error=unused-function -Wno-error=unused-but-set-variable -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wextra -Wno-unused-parameter -Wno-sign-compare -Wwrite-strings -Wno-old-style-declaration -mfix-esp32-psram-cache-issue -MMD -c
+compiler.c.flags=-std=gnu99 -Os -g3 -Wno-frame-address -ffunction-sections -fdata-sections -fstrict-volatile-bitfields -mlongcalls -nostdlib -fstack-protector {compiler.warning_flags} -Wno-error=unused-function -Wno-error=unused-but-set-variable -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wextra -Wno-unused-parameter -Wno-sign-compare -Wwrite-strings -Wno-old-style-declaration -Wno-implicit-fallthrough -mfix-esp32-psram-cache-issue -MMD -c
```
liebman: Why ignore more warnings?  Why not add some annotation to the location the warnings are in and leave it enabled to catch accidental fall-thru cases?  This blog entry has a thorough discussion of the annotation options: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

platform.txt
```diff
@@ -22,20 +22,20 @@ compiler.warning_flags.all=-Wall -Werror=all -Wextra
...
 
 compiler.c.cmd=xtensa-esp32-elf-gcc
-compiler.c.flags=-std=gnu99 -Os -g3 -fstack-protector -ffunction-sections -fdata-sections -fstrict-volatile-bitfields -mlongcalls -nostdlib -Wpointer-arith {compiler.warning_flags} -Wno-maybe-uninitialized -Wno-unused-function -Wno-unused-but-set-variable -Wno-unused-variable -Wno-deprecated-declarations -Wno-unused-parameter -Wno-sign-compare -Wno-old-style-declaration -MMD -c
+compiler.c.flags=-std=gnu99 -Os -g3 -Wno-frame-address -ffunction-sections -fdata-sections -fstrict-volatile-bitfields -mlongcalls -nostdlib -fstack-protector {compiler.warning_flags} -Wno-error=unused-function -Wno-error=unused-but-set-variable -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wextra -Wno-unused-parameter -Wno-sign-compare -Wwrite-strings -Wno-old-style-declaration -Wno-implicit-fallthrough -mfix-esp32-psram-cache-issue -MMD -c
```
liebman: Why ignore more warnings? (-Wno-implicit-fallthrough) Why not add some annotation to the location the warnings are in and leave it enabled to catch accidental fall-thru cases? This blog entry has a thorough discussion of the annotation options: https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/

atanisoft: many of these are coming from IDF's build system likely.

liebman: If they are via an include file then the include can be wrapped in something like
```
#pragma GCC diagnostic push 
#pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
#include <freertos/FreeRTOS.h>   // or whatever include file that gives warnings
#pragma GCC diagnostic pop
```

atanisoft: Adding those pragma statements would be a bit better than masking a lot of warnings/errors in the code but it will become a maintenance nightmare when you start having to modify files under tools/sdk/ since these are maintained outside this repo.

atanisoft: but in general I agree, clean up the warnings and drop the flags :)

liebman: You don't have to modify any files in  `tools/sdk`.  Just file file that includes one that causes warnings.

me-no-dev: I left them because some of the warnings will come from IDF's headers. We can not mess with those. In order to provide best compatibility, I opted to use the same exact flags as IDF. Most of the things, if you note, just turn errors into warnings, so you could tstill compile and deal with it later

PR #3673, opened 2020-01-25
Conditionally include libraries based on sdkconfig selectives (cmake)

robert-alfaro:
Conditionally include/compile Azure_Iot library. Not only is it wasteful to compile if unused, it breaks on release build configuration.

This change begins to pave the way for honoring the SDK configuration Library Selections

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,47 +37,6 @@ set(CORE_SRCS
   cores/esp32/WString.cpp
   )
 
-set(LIBRARY_SRCS
-  libraries/ArduinoOTA/src/ArduinoOTA.cpp
-  libraries/AsyncUDP/src/AsyncUDP.cpp
-  libraries/BluetoothSerial/src/BluetoothSerial.cpp
-  libraries/DNSServer/src/DNSServer.cpp
-  libraries/EEPROM/src/EEPROM.cpp
-  libraries/ESPmDNS/src/ESPmDNS.cpp
-  libraries/FFat/src/FFat.cpp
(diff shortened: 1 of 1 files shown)


robert-alfaro:
@me-no-dev looks like the eventual goal for this PR relates to work that was done for selective compilation already..only that work was GNU Make specific (Issue #1655), correct? For a long while now I've bee wondering why these selections have zero effect, now I see.  Anyways, I can make similar changes to the CMakeLists.txt, however, can you shed light on what may need to change in the check-cmakelists.sh github script?

me-no-dev:
4.0 is on it's way and Azure lib is not inside :) will that do?

robert-alfaro:
That's great, although I'm not sure we're switching to + just yet. In any case selective compilation of other parts is still desired. I specified azure here because it causes problems and is large.

me-no-dev:
so by default it should include it. else you see the error in the builder.

robert-alfaro:
What's odd is if CI sdkconfig has it enabled and it's failing. cmakelist script tests against COMPONENT_SRCS which does include azure library.

robert-alfaro:
@me-no-dev I believe the test is presenting a false-negative. Line https://github.com/espressif/arduino-esp32/blob/9b75c65fc76ea67b7b8420dbc2c8ccc917182765/.github/scripts/check-cmakelists.sh#L18 only captures values hardcoded in COMPONENT_SRCS.  It does not know that the list is built up and _is actually complete_.

h2zero:
I hope this gets merged, I need to turn off BLE to use NimBLE without compile errors.

PR #3675, opened 2020-01-26
fix removeApbChangeCallback() error in spiStopBus()

stickbreaker:
spiStartBus() was using spiStopBus() to init the hardware, one of spiStopBus() functions is to unregister the runtime CPU clock speed change callback. But, spiStartBus() only wanted to init the hardware.  This patch separates the hardware init into a standalone function spiInitBus() that both spiStartBus() and spiStopBus() call.


cyberman54:
Tested, works, thanks for fixing the bug!

PR #3683, opened 2020-01-29
Add Senses IoT board name "Senses's WEIZEN"

Isaranu:
Please add our product "Senses's WEIZEN" (ESP32 series) into boards.txt
Thank you very much.


me-no-dev:
I think that you have started from a very old boards.txt file. Please start again from a fresh one.

PR #3689, opened 2020-01-30
Add 'ESP32 EDU KITS' in board.txt

kitsrnd:
'ESP32 EDU KITS' board based on ESP32 PICO was added.


kitsrnd:
I'm sorry I will request later since we decided to change the board name.

PR #3691, opened 2020-01-30
Fix failing Updater with https (WifiClientSecure)

nuclearcat:
In arduino code

>        size_t written = Update.writeStream(client2);

Current code often fails with error "Error Occurred. Error #: 6", with this patch this error doesn't appear.
I tested it on my side, but it should be unit tested on other scenarios, if possible.


nuclearcat:
Anybody alive?

nuclearcat:
Added additional patch to reset timeouts counter, this bug appeared on very slow ppp link (GSM)

nuclearcat:
I will make better version of patch now, for latest master

PR #3693, opened 2020-01-30
add board TTGO T-Display

joerglemke:
This adds a new board TTGO T-Display.
This also requires related changes in platform-espressif32. See my pull request there.

This board is like a base ESP32 dev board with integrated display.
Hence, the additions are modified copies from the ESP32 DEV board.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -4838,12 +4838,12 @@ ttgo-t-display.build.flash_freq=40m
 ttgo-t-display.build.flash_mode=dio
 ttgo-t-display.build.boot=dio
 ttgo-t-display.build.partitions=default
-ttgo-t-display.build.defines=
+ttgo-t-display.build.commondefs=-DARDUINO_TTGO_Display -DUSER_SETUP_LOADED=1 -DST7789_DRIVER=1 -DTFT_WIDTH=135 -DTFT_HEIGHT=240 -DCGRAM_OFFSET=1 -DTFT_MOSI=19 -DTFT_SCLK=18 -DTFT_CS=5 -DTFT_DC=16 -DTFT_RST=23 -DTFT_BL=4 -DTFT_BACKLIGHT_ON=HIGH -DLOAD_GLCD=1 -DLOAD_FONT2=1 -DLOAD_FONT4=1 -DLOAD_FONT6=1 -DLOAD_FONT7=1 -DLOAD_FONT8=1 -DLOAD_GFXFF=1 -DSMOOTH_FONT=1 -DSPI_FREQUENCY=40000000 -DSPI_READ_FREQUENCY=6000000
 
 ttgo-t-display.menu.PSRAM.disabled=Disabled
-ttgo-t-display.menu.PSRAM.disabled.build.defines=
+ttgo-t-display.menu.PSRAM.disabled.build.defines={build.commondefs}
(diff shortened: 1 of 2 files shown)


boards.txt
```diff
@@ -4838,12 +4838,12 @@ ttgo-t-display.build.flash_freq=40m
 ttgo-t-display.build.flash_mode=dio
 ttgo-t-display.build.boot=dio
 ttgo-t-display.build.partitions=default
-ttgo-t-display.build.defines=
+ttgo-t-display.build.commondefs=-DARDUINO_TTGO_Display -DUSER_SETUP_LOADED=1 -DST7789_DRIVER=1 -DTFT_WIDTH=135 -DTFT_HEIGHT=240 -DCGRAM_OFFSET=1 -DTFT_MOSI=19 -DTFT_SCLK=18 -DTFT_CS=5 -DTFT_DC=16 -DTFT_RST=23 -DTFT_BL=4 -DTFT_BACKLIGHT_ON=HIGH -DLOAD_GLCD=1 -DLOAD_FONT2=1 -DLOAD_FONT4=1 -DLOAD_FONT6=1 -DLOAD_FONT7=1 -DLOAD_FONT8=1 -DLOAD_GFXFF=1 -DSMOOTH_FONT=1 -DSPI_FREQUENCY=40000000 -DSPI_READ_FREQUENCY=6000000
```
atanisoft: pin definitions (TFT_MOSI, etc) should be in pins_arduino.h, the rest of these should likely go there as well.

Overriding SPI_FREQUENCY here also doesn't seem like a good idea.

joerglemke: Thanks for commenting! I almost fully agree with you.

That board has built-in display. According to their documentation they have an adaption of the TFT_eSPI display library. The official documentation of this library says that you should either edit a  .H file of the library or alternatively for PlatformIO, that you could supply these parameters on the compiler command line.
So I was looking for a way to supply these defines also using Arduino IDE. Thats why I defined them here.

I would also prefer to define these in pins.arduino.h .
Problem with TFT_eSPI is, it includes its own definitions BEFORE including Arduino.h. And so pins_arduino.h would come too late in the process.
So your finding is good, but it would require changes in that library, or project specific editing a library file, or supplying all those board specific params in each sketch.

What do you think?

atanisoft: TFT_eSPI already requires project specific edits for the "user config" file so it doesn't seem like a blocker, there is also a Setup25_TTGO_T_Display.h file that defines all of these for TFT_eSPI in the library.

Teslafly:
Why was this closed without merging? This board is only becoming more popular and more available.

BlueAndi:
@joerglemke Could you reopen please? The board is not merged yet.

pascalpfeil:
@atanisoft any idea why this was not merged?

pascalpfeil:
Ok thank you - I will get this board in the mail in a few week, I might have a look myself

FilipParyz:
Hi, what up with this board? Is it merged yet? :)

atanisoft:
> Is it merged yet? :)

Nope, this PR was closed by the filer rather than leaving open for merge. Feel free to submit a new PR.

DocBrown101:
Has it still not been merged?

PR #3694, opened 2020-01-30
Fixes esp32_out.ld error

chopmann:


tools/platformio-build.py
```diff
@@ -71,7 +71,7 @@
     CXXFLAGS=[
         "-fno-rtti",
         "-fno-exceptions",
-        "-std=gnu++11"
+        "-std=gnu++17"
```
atanisoft: IDF defaults to gnu++11 so this should not be updated to gnu++17.

chopmann:
I have a working Demo with c++17 and the newer toolchain
https://github.com/HamburgChimps/ibis_wagenbus

atanisoft:
@chopmann it isn't about having support for a newer version or not. The esp-idf build scripts default the gnu++11. I have also successfully overridden this to use c++14 without issues, even on esp-idf v3.x.

chopmann:
> @chopmann it isn't about having support for a newer version or not. The esp-idf build scripts default the gnu++11. I have also successfully overridden this to use c++14 without issues, even on esp-idf v3.x.

I know that. Of course it works with c++14 with platformio. What im saying is:
the current package `toolchain-xtensa32` wont work with this PR or https://github.com/espressif/arduino-esp32/pull/3670 
As far as I understood it needs the newer toolchain with gcc8.2 and not the older one

atanisoft:
That is correct, PlatformIO can not use IDF v4 without a lot of major reworking in PlatformIO core.

I no longer use PlatformIO, but your example of using gnu++17 linked above is the only approach usable with PlatformIO. It will not work outside of that env and won't work for those using cmake based esp-idf builds that include arduino-esp32. For cmake there is a way to do it but it would be better if it was configurable from the esp-idf side instead.

PR #3695, opened 2020-01-31
increase stack size

NickFisherAUT:
Avoid stack overflow

### DIFF
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -25,7 +25,7 @@ extern "C" void app_main()
 {
     loopTaskWDTEnabled = false;
     initArduino();
-    xTaskCreateUniversal(loopTask, "loopTask", 8192, NULL, 1, &loopTaskHandle, CONFIG_ARDUINO_RUNNING_CORE);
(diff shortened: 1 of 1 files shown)


NickFisherAUT:
avoid running in overflow of rtos stack

stickbreaker:
I don't expect this patch to be accepted. If you are overflowing 8k of stack your local variable usage is excessive, or you are using excessive recursion.  The esp32 is a MICRO controller, Not a PC.  If this patch was accepted all other users would loose 8k of RAM for no purpose.

If you have large structures on stack(local variables) investigate Dynamic allocation (`new,malloc,calloc`).

Chuck.

liebman:
You can also run things that require more stack in a separate Task that you create and provide with a larger stack.

a-c-sreedhar-reddy:
@stickbreaker  Will tail recursion help for recursion?

stickbreaker:
@a-c-sreedhar-reddy sure, if the compiler can optimize for it, The basis for "tail recursion" optimization is that there is no additional stack growth for each looping call. But, it is a special case. Most recursive function use the stack for temp storage, The "tail recursion" function is structured such that the stack is clean before the recursive call.  The simple example of walking a binary tree cannot be "tail recursion" structured. 

Chuck.

PR #3703, opened 2020-02-03
Add a new board of KITS for IoT education

kitsrnd:
Please add our product 'KITS ESP32 EDU' which is a variant of ESP32 Pico into boards.txt
Thank you in advance.


me-no-dev:
Sorry for the delay. We will be reworking how boards are listed and constructed, so there will be changes in the near future.

PR #3713, opened 2020-02-05
HardwareSerial bugfix & improvement

hreintke:
This PR includes : 

Bugfix : Disable/enable uart interrupts during transfer from fifo to queue

Improvement : Remove unnecessary check for full queue in uart isr

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -85,7 +85,7 @@ static void IRAM_ATTR _uart_isr(void *arg)
         uart->dev->int_clr.rxfifo_tout = 1;
         while(uart->dev->status.rxfifo_cnt || (uart->dev->mem_rx_status.wr_addr != uart->dev->mem_rx_status.rd_addr)) {
             c = uart->dev->fifo.rw_byte;
-            if(uart->queue != NULL && !xQueueIsQueueFullFromISR(uart->queue)) {
+            if(uart->queue != NULL)  {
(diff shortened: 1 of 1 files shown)


bdring:
[The previous commit](https://github.com/espressif/arduino-esp32/commit/ed220bd0420236e9ff95888f7459f0eb9c97a570) seems to have broken [my code](https://github.com/bdring/Grbl_Esp32). This PR appears to have fixed it.

The problem appeared as random, but regular corruption of the data. The corruption usually looks like a swap of the order of 2 characters.

My firmware is a CNC controller. In the UART data are move commands that are software flow controlled and "real time" command like status requests and pause, etc., that can be sent at any time. Therefore, the UART data needs to be rapidly read to check for real time commands. Anything that is not a real time is placed into another buffer. This is done using RTOS task. Typically the baud rate is 115200, but users are free to change that.

hreintke:
@me-no-dev : 
Can you check this PR and merge if OK.
If you see issues with it, please let me know so I can work on those.

bdring:
@hreintke @me-no-dev 

Can this be expedited? This appears to fix a [previous commit](https://github.com/espressif/arduino-esp32/commit/ed220bd0420236e9ff95888f7459f0eb9c97a570) that broke serial communications.

maudette01432:
Waiting anxiously - would be great to have this fix in the main arduino board manager version as well.

bdring:
Please expedite this. Serial comms is currently broken.

PR #3720, opened 2020-02-06
Move _STREAM_BOUNDARY before _STREAM_PART

abergmann:
The boundary delimiter (_STREAM_BOUNDARY) needs to be send before the body part (_STREAM_PART) too follow RFC2046. This caused `ffplay`/`ffmpeg` to fail to open the MJPEG stream.

[RFC2046: 5.1. Multipart Media Type](https://tools.ietf.org/html/rfc2046#section-5.1)
"... The body must then contain one or more body parts, **each preceded by a boundary delimiter line**, and the last one followed by a closing boundary delimiter line. ..."

After the change `ffplay`, `ffmpeg` and `mplayer` worked fine with the MJPEG stream.

### DIFF
--- a/libraries/ESP32/examples/Camera/CameraWebServer/app_httpd.cpp
+++ b/libraries/ESP32/examples/Camera/CameraWebServer/app_httpd.cpp
@@ -404,14 +404,14 @@ static esp_err_t stream_handler(httpd_req_t *req){
             }
         }
         if(res == ESP_OK){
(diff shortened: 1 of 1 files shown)


me-no-dev:
nice find! I was wondering why it does not work, given that regular requests do not follow this order (boundary comes after the content)

me-no-dev:
could I bother you to also add it here: https://github.com/espressif/esp-who/blob/master/examples/single_chip/camera_web_server/main/app_httpd.c#L590

PR #3728, opened 2020-02-09
prefer custom mac address over efuse mac address for wifi interface

ben-mkiv:
if no custom mac is set it'll read the mac from efuse storage as before

im not sure if this might also be required to be changed for other interfaces like ethernet and bluetooth (a quick search didnt find any other reference for esp_efuse_mac_get_default())


ben-mkiv:
as alternative `esp_read_mac()` could be used, which does handle custom and efuse mac address

Issue #3576, opened 2019-12-16
SPI_HAS_TRANSACTION has issues with being defined


Avamander:
I'm one of the maintainers of github.com/nRF24/RF24 and for some reason our previously working code for detecting SPI transaction support does not work for ESP32:
```c++
#if defined (SPI_HAS_TRANSACTION) && !defined (SPI_UART) && !defined (SOFTSPI)
  #define RF24_SPI_TRANSACTIONS
#endif // defined (SPI_HAS_TRANSACTION) && !defined (SPI_UART) && !defined (SOFTSPI)
```

and based on https://github.com/espressif/arduino-esp32/blob/b847f41e24c1080d1e29aceb92b6a0199b05047f/cores/esp32/esp32-hal-spi.h#L25 I'd expect it to work, but it doesn't.

For some reason ESP32 users are [saying that transactions don't get enabled](https://github.com/nRF24/RF24/issues/508) and they need frankly weird [patches](https://github.com/nRF24/RF24/pull/549/files) to fix those issues. There's a bit of debugging done in the linked issue, it's... weird.

It would be very appreciated if someone on the core team could take a look at what's missing in ESP32 core that it doesn't act like an Arduino board for those people. 

### Hardware:
Core Installation version:			1.0.3
IDE name:						Arduino IDE and Platform.io
Computer OS:                                         Unspecified

Avamander:
Ping

Avamander:
Ping

Avamander:
Ping

atanisoft:
@Avamander if you know of a fix for this, feel free to submit a PR with it for review.

Avamander:
@atanisoft https://github.com/espressif/arduino-esp32/pull/3730

Though by now the suspicion is that platform.io "eats" the flag somewhere, but technically the PR should fix it.

atanisoft:
@Avamander there is a build-platformio.py file under tools that will need to add that default flag. That should fix it for PIO as well.

Avamander:
@atanisoft 
> there is a build-platformio.py file under tools that will need to add that default flag. That should fix it for PIO as well.

I made that change too.

PR #3730, opened 2020-02-10
Made sure that libraries get told that SPI transactions are supported

Avamander:
These changes make sure that Arduino libraries could accurately detect if SPI transactions are enabled, this will very likely improve the performance of a lot of different Arduino libraries. 

Fixes https://github.com/espressif/arduino-esp32/issues/3576

Fixes https://github.com/nRF24/RF24/issues/508

(Please squash merge. Also if other files have to be changed, maintainers can push to this PR's branch)


me-no-dev:
It should be done like this: https://github.com/espressif/arduino-esp32/commit/7af4490fcc2b9f549414921d9a8ad1392303f4d8

PR #3735, opened 2020-02-12
Allow using custom linker scripts in PlatformIO build script

valeros:
This will allow users to set any linker script from `platformio.ini` file using the following syntax:
```ini
[env:esp32dev]
platform = espressif32
framework = arduino
board = esp32dev
board_build.ldscript = custom_ldscript.ld
```
This is a more correct way of specifying it in comparison with using the `-Wl,-T` flag.


ivankravets:
@me-no-dev Could you merge it?

szechyjs:
@me-no-dev any updates on this? the `platformio/platform-espressif32` issue above requires this fix.

PR #3764, opened 2020-02-24
Fix crash when deleting BLEClient

TheNitek:
Currently the ESP crashes when delete is called on a BLEClient instance because the BLERemoteService tries to delete its characteristics twice.

This has already been fixed here https://github.com/nkolban/esp32-snippets/blob/fe3d318acddf87c6918944f24e8b899d63c816dd/cpp_utils/BLERemoteService.cpp#L267 but somehow didn't make it into the master/current releases.


TheNitek:
Any chance to get this merged? Without this the BLEClient is next to unusable in some scenarios.

Obrekr:
Same changes/fix as https://github.com/espressif/arduino-esp32/pull/3521.

PR #3811, opened 2020-03-12
Readme edits

Sagar73594:
to make changes in Master/Readme.md  files.


me-no-dev:
I do not agree with the change. You should instead explain what could be causing the issue and how adding capacitance to the EN pin can help in resolving.

You are talking about boards that have faulty reset circuits which cause the EN pin to go HIGH before IO0 goes HIGH, thus booting back into firmware.

PR #3853, opened 2020-03-26
Exposed the lastError.

n9wxu:
Sorry about the clang_format restructuring the entire file.

The purpose of lastError is to determine if a GET actually returned the default or the stored value.  If it returns the default I will want to issue a PUT to create the key.

Likely a GET with no key should automatically store the default at that key and return the default.

This is appropriate for all GET functions except the handful of functions that accept a buffer and do not use defaults.

### DIFF
--- a/libraries/Preferences/src/Preferences.cpp
+++ b/libraries/Preferences/src/Preferences.cpp
@@ -15,35 +15,40 @@
 
 #include "nvs.h"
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
I have no idea what the change is now that the whole file is reformatted. Please add only the change (you can do this from github.com directly: edit the file and submit pr)

PR #3860, opened 2020-03-28
Support OTA updates with Flash encryption

UCIS:
This patch allows the OTA updater to be used with Flash encryption. Using the ESP partition API instead of the direct Flash access API transparently encrypts the data to be written if Flash encryption is enabled.

### DIFF
--- a/libraries/Update/src/Updater.cpp
+++ b/libraries/Update/src/Updater.cpp
@@ -40,7 +40,7 @@ static bool _partitionIsBootable(const esp_partition_t* partition){
     if(!partition){
         return false;
     }
-    if(!ESP.flashRead(partition->address, (uint32_t*)buf, 4)) {
+    if (esp_partition_read(partition, 0, (uint32_t*)buf, 4) != 0) {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


UCIS:
@robertpoll 
Actually I think it might not be necessary at all to update the first/signature byte. It appears the updater changes the first byte to keep the chip from accidentally booting an incomplete image - I did not notice that part before.

1. If the otadata partition is valid, the chip should continue to boot the old firmware until the update has completed and the otadata partition has been updated. The otadata might be invalid only after the initial flash, if it was not part of the initial flash. In this case, the bootloader will prefer the factory or first ota partition, which will most likely be the original version and not the incomplete update.
2. The [ESP32 bootloader](https://github.com/espressif/esp-idf/blob/f7d1ed5f329c14c356a710c4c10ba50cea277b55/components/bootloader_support/src/esp_image_format.c#L100) uses a checksum byte and an optional sha256 hash (or the secure boot signature) to validate the image.

Other than that, your solution is better. I don't know the Arduino system well enough to tell you how to call the ESP flash functions. Personally I think the ESP wrapper class is redundant since you can call the ESP-IDF methods just as easy. I ended up using the ESP-IDF OTA update functions directly instead since that works without patching.

robertpoll:
@UCIS 
I did have the same thought but took the lazy route and just re-implemented it assuming it was there for a reason. It's possible it's just a hang-over from the ESP8266 implementation but without picking over the bootloader code to verify I'm inclined to leave it in. If anyone else has an opinion on whether it's safe to remove that would be useful. On the ESP wrapper, again not sure of the logic for it but it's straightforward to add the functions so best to remain consistent I think.

How would you like to proceed? Incorporate into this PR, or close and I'll create a new one? Happy either way...

UCIS:
@robertpoll feel free to take over and create a new PR. This one is mostly here to help others who may run into the same problem, but I don't plan on testing or updating this as I am no longer using the Updater class in my own application.

PR #3880, opened 2020-04-08
added new board Logsens V1p1

satyanraj:
Added new board Logsens V1.1 from Imbrios Systems


satyanraj:
Any issue in adding new boards to the list ?

Issue #3205, opened 2019-09-13
Serial.end() not reattaching pin to GPIO matrix


luketanti:
After trying to switch back two UART pins to GPIO by using the Serial.end() function, the pins do not respond to digitalWrite commands.

I have found out that by using `gpio_matrix_out(gpio_number, SIG_GPIO_OUT_IDX, false, false);` I am able to use the pin again after ending the Serial.

stickbreaker:
@luketanti after using `Serial.end()` did you do a `pinMode(pin,OUTPUT)` before doing a `digitalWrite(pin,value)`?

Chuck.

luketanti:
> @luketanti after using `Serial.end()` did you do a `pinMode(pin,OUTPUT)` before doing a `digitalWrite(pin,value)`?
> 
> Chuck.

Yes, offcourse.

Here are two snippets. The first one is when the issue occurs and the second is how I managed to get the pin to respond to digitalWrite commands.

```
  Serial2.end();
  pinMode(K_TX, OUTPUT);
  pinMode(K_RX, INPUT);
  digitalWrite(K_TX, HIGH);
  delay(298);
  digitalWrite(K_TX, LOW);
  delay(23);
  digitalWrite(K_TX, HIGH);
  delay(23);

  Serial2.begin(10417, SERIAL_8N1, K_RX, K_TX);  // K-LINE
```

```
  Serial2.end();
  gpio_matrix_out(K_TX, SIG_GPIO_OUT_IDX, false, false);
  pinMode(K_TX, OUTPUT);
  pinMode(K_RX, INPUT);
  digitalWrite(K_TX, HIGH);
  delay(298);
  digitalWrite(K_TX, LOW);
  delay(23);
  digitalWrite(K_TX, HIGH);
  delay(23);

  Serial2.begin(10417, SERIAL_8N1, K_RX, K_TX);  // K-LINE
```

atanisoft:
Did you try gpio_reset(K_TX); instead of gpio_matrix_out?

luketanti:
I have tried to compile with gpio_reset(K_TX); but it failed.

> exit status 1
> 'gpio_reset' was not declared in this scope

Shouldn't Serial.end() automatically release the pins allocated to other use?

atanisoft:
Sorry that should have been gpio_reset_pin. And yes I'd think that Serial.end() should release the pin to normal/default state. I believe there was an older issue for the same.

luketanti:
[...]

PR #3894, opened 2020-04-11
Fixes UART detach.  Fixes #3878

### LINKED ISSUE #3878: serial.end() doesn't work
### Description:
`Serial.end()` should free the serial pins and leave the user the ability to do `digitalWrite()` on them, but it doesn't work. Instead to drive the TX pin to a LOW state i need to reset the board.

I tested the code below with the following cores:
- last stable release from platformio
- master branch
- master + #3713

I also made three logic analyzer captures: [la.zip](https://github.com/espressif/arduino-esp32/files/4441212/la.zip)
An example from one of these, as you can see it shows that the pin is driven low only once (after upload of the code or reset)

image

### Sketch: 
```cpp

#include <Arduino.h>

HardwareSerial ser(2);
#define TX_PIN 17

void setup(){}

void loop()
{
    pinMode(TX_PIN, OUTPUT);
    delay(10);

    digitalWrite(TX_PIN, HIGH);
    delay(50);
    digitalWrite(TX_PIN, LOW);
    delay(300);
    digitalWrite(TX_PIN, HIGH);
    delay(50);

    ser.begin(9600);
    delay(10);

    for (uint8_t n = 1; n < 5; n++)
    {
        ser.write(n);
        delay(10);
    }
    ser.end();
    delay(300);
}

```

NOTE: looks like that the problem was here before #3713 and #3664

lbernstone:
From what I can tell, there used to be a hidden mechanism to detach the uart pins in gpio_matrix_out.  This no longer works, so we need to keep track of what pins we assign.


aster94:
any news on this?

drojf:
Are you sure that `pinMatrixInDetach()` takes the same first argument as `pinMatrixOutDetach()`? If you look at the definition/SPI code for the same thing, you'll find it takes a different argument to the ...OutDetach() function: 

https://github.com/espressif/arduino-esp32/blob/37a7fb3d6ada8a9230c96e2df999c95ef10fd4d3/cores/esp32/esp32-hal-matrix.c#L28-L41

https://github.com/espressif/arduino-esp32/blob/37a7fb3d6ada8a9230c96e2df999c95ef10fd4d3/cores/esp32/esp32-hal-spi.c#L147-L181

I guess there could also be an error in the `esp32-hal-matrix.c` file definition for `pinMatrixInDetach()`.

lbernstone:
I didn't change the header definitions for either of the pinMatrix detach functions, so I'm pretty sure the arguments are correct.  This PR is specifically to deal with the fact that the UART_TXD_IDX pins (which are in the 200s, IIRC) no longer work in the IDF functions.  The SPI versions of those pins must still work, since nobody is complaining about SPI::end failing.

dka23:
See also #3205

PR #3910, opened 2020-04-16
Add Delay to _runUpdate() in ArduinoOTA.cpp to Avoid Watchdog Resets

nixmeer:
In some cases, the while-loop in `ArduinoOTA.cpp` (line 267) causes a watchdog reset. Adding a `delay(1)` might not be the best solution but works as a quick fix. `yield()` didn't do the job.

Any better solution appreciated!


libraries/ArduinoOTA/src/ArduinoOTA.cpp
```diff
@@ -320,6 +320,7 @@ void ArduinoOTAClass::_runUpdate() {
         } else {
             log_e("Write ERROR: %s", Update.errorString());
         }
+        delay(1);
```
nixmeer, line 323: frees the core for a moment to avoid watchdog resets

me-no-dev:
the problem with this approach is that it delays every packet, which causes the transfer to take more time. More appropriate approach is to do it every few hundred ms and delay for 2ms or more.

nixmeer:
Hello again :) 

I am no software developer, so my fix works for me but is for sure not the best way to go. Isn't there a way to just reset the watchdog since the function itself is doing what it is supposed to do? Or maybe do a `vTaskDelay(0)` which will result in no delay if no other task needs to run?

igrr:
FYI, there is a configuration option in the underlying SPI flash driver that allows yielding during erase operations:

https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig.html#config-spi-flash-yield-during-erase (if CONFIG_SPI_FLASH_YIELD_DURING_ERASE is enabled, every CONFIG_SPI_FLASH_ERASE_YIELD_DURATION_MS the driver will yield for CONFIG_SPI_FLASH_ERASE_YIELD_TICKS).

nixmeer:
So in order to get the Arduino OTA work reliably, should this be activated by default? Or a `yield()` be added instead of a delay? (I did not test this yet)

lucasssvaz:
IMO we should properly address this issue by add the config @igrr mentioned to the sdkconfig and closing this PR.

lucasssvaz:
Closing in favor of espressif/esp32-arduino-lib-builder#154

PR #3915, opened 2020-04-19
Rmt improvements

dirkx:
Add an aditional (void *) arg to the RMT callback - to allow more flexible handling of the callback (e.g. by passing a private struct or a class pointer). 
a
Same pattern as used by the Ticker() and many others. 

Example updated & new example with a trapoline added.


me-no-dev:
this will also go into 2.0.0 because it changes API. I otherwise agree :)

dirkx:
Yes - I tried hard to not break the API - but you would then always be able to work around things with the 'child lock' sort of current shared secret. So I sort of could not figure it out. Sorry !

me-no-dev:
Functionality already present in master. Closing :)

PR #3918, opened 2020-04-19
Update CaptivePortal.ino

acontini:
Row 30 WiFiClient client = server.available();   // listen for incoming clients
should read WiFiClient client = client.available();   // listen for incoming clients

ref https://www.arduino.cc/en/Reference/WiFiClientAvailable

void loop() {
  // if there are incoming bytes available
  // from the server, read them and print them:
  if (client.available()) {
    char c = client.read();
    Serial.print(c);
  }

### DIFF
--- a/libraries/DNSServer/examples/CaptivePortal/CaptivePortal.ino
+++ b/libraries/DNSServer/examples/CaptivePortal/CaptivePortal.ino
@@ -27,7 +27,7 @@ void setup() {
 
 void loop() {
   dnsServer.processNextRequest();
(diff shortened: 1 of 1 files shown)


lbernstone:
Nope.  server.available returns an object that is the client connection.  Code works as written.

acontini:
In my system platformio doesn't compile it. I'll investigate further

acontini:
I've tested the example as a stand alone code and it works.
Sorry, my mistake. 
Andrea

PR #3926, opened 2020-04-22
Add support of unified provisioning to Arduino v4.0

sweetymhaiske:
Here 1. WiFiProv.ino sketch is added that allows arduino users to do provisioning via SoftAP or BLE. WiFi.beginProvision( ) API is designed for provisioning in Arduino.
2. In WiFiProv.h provisioning class is defined.
3. WiFiProv.cpp contains implementation for provisioning class.
4. README.md file is added which contains detail information for working.


me-no-dev:
i'll close this as 4.0 will not be released

PR #3934, opened 2020-04-23
Create and change some placeholder/compatibility include files.  

ewpa:
The request comes from the desire to compile some GNU/Linux source code.  However, the existing include file caused any program that included it to FTBFS.


atanisoft:
this can not be accepted here, it needs to be submitted upstream to esp-idf.

PR #3940, opened 2020-04-25
Create pins_arduino.h

MaksimKlyu:


atanisoft:
If this variant is new it is also needed in boards.txt

PR #3995, opened 2020-05-13
improve & fix BLEScan when wantDuplicates

saknarak:
- when wantDuplicates, no need to check duplicate and no more insert into vector
- delete advertisedDevice when not insert into vector, fix memory leak
- add showParse when you just want raw advertised data


libraries/BLE/src/BLEScan.cpp
```diff
@@ -107,19 +111,25 @@ void BLEScan::handleGAPEvent(
...
+					if (m_shouldParse) {
+						advertisedDevice->parseAdvertisement((uint8_t*)param->scan_rst.ble_adv, param->scan_rst.adv_data_len + param->scan_rst.scan_rsp_len);
+					} else {
+						advertisedDevice->setPayload((uint8_t*)param->scan_rst.ble_adv, param->scan_rst.adv_data_len + param->scan_rst.scan_rsp_len);
+					}
 					advertisedDevice->setScan(this);
 					advertisedDevice->setAddressType(param->scan_rst.ble_addr_type);
 
-					if (!found) {   // If we have previously seen this device, don't record it again.
+					if (!m_wantDuplicates && !found) {   // If we have previously seen this device, don't record it again.
```
chegewara, line 122: What will happen if you want duplicates?
`m_wantDuplicates = true`
In that case this will always return false and device will never be added to vector:
`if (!m_wantDuplicates && !found)`

saknarak: when start continuous scan with callback. vector add allocate without de-allocatted that cause ESP restart.

when not to keep in vector
1. callback was defined.
or 2. wantDuplicates = true
or 3. already in vector (found = true)

Ben79543:
on #4627 I explain why this commit is an issue for me: in a croweded BLE environment (100s devices), a 5sec scan would crash the ESP even in the first scan. So I needed to limit the number of detected devices and I did it with a counter in the callback. 

But with this commit, when a callback is defined there is no record to vector no matter what, and the scan returns no result. 

Easiest is to re-allow to call the callback and populate the vector. Not sure why this case was removed.

chegewara:
I agree, we need this to be fixed or reverted.

Ben79543:
I have created a PR for this : #5241
My first PR hopefully done well...

PR #3990, opened 2020-05-11
Esp32s2 Support Test

me-no-dev:

PR #3998, opened 2020-05-13
fix flash mode file locations

meltdown03:
Fixes unable to flash error: `bootloader_***_**m.bin` not found due to it being in a specific directory, "./tools/sdk/esp32/bin" or "./tools/sdk/esp32s2/bin"


meltdown03:
Fixed in #3990

Issue #3971, opened 2020-05-05
Build as component fails


PaulMartinsen:
How does the build system know which paths to add as includes? In particular `wifi_provisioning/manager.h` is missing from the include list (although it is present in `components/wifi_provisioning/include`)

I have been following the instructions at [esp-idf_component.md](https://github.com/espressif/arduino-esp32/blob/master/docs/esp-idf_component.md) to build the library but the build fails at files that include `wifi_provisioning/manager.h`. This seems to be a [recent addition](https://github.com/espressif/arduino-esp32/commit/2c9b648502f1cfdbe2e9a1ef9c26582718633830).

It looks like the include for `components\wifi_provisioning\include` is missing from the build command. Here's the offending build step, reformated for clarity. 
```FAILED: esp-idf/components/CMakeFiles/idf_component_components.dir/libraries/WiFiClientSecure/src/ssl_client.cpp.obj
ccache C:\Programming\Languages\ESP-IDF\Tools\tools\xtensa-esp32-elf\1.22.0-80-g6c4433a5-5.2.0\xtensa-esp32-elf\bin\xtensa-esp32-elf-g++.exe  -DESP_PLATFORM -DGCC_NOT_5_2_0=0 -DHAVE_CONFIG_H -DIDF_VER=\"v3.3.2\" -DMBEDTLS_CONFIG_FILE=\"mbedtls/esp_config.h\" 
-I../components/variants/esp32 
-I../components/cores/esp32 
-I../components/libraries/ArduinoOTA/src 
-I../components/libraries/AsyncUDP/src 
-I../components/libraries/AzureIoT/src 
-I../components/libraries/BLE/src 
-I../components/libraries/BluetoothSerial/src 
-I../components/libraries/DNSServer/src 
-I../components/libraries/EEPROM/src 
-I../components/libraries/ESP32/src 
-I../components/libraries/ESPmDNS/src 
-I../components/libraries/FFat/src 
-I../components/libraries/FS/src 
-I../components/libraries/HTTPClient/src 
-I../components/libraries/HTTPUpdate/src 
-I../components/libraries/NetBIOS/src 
-I../components/libraries/Preferences/src 
-I../components/libraries/SD_MMC/src 
-I../components/libraries/SD/src 
-I../components/libraries/SimpleBLE/src 
-I../components/libraries/SPIFFS/src 
-I../components/libraries/SPI/src 
-I../components/libraries/Ticker/src 
-I../components/libraries/Update/src 
[...]

PR #4003, opened 2020-05-16
Added missing wifi_provisioning dependency (#3971).

PaulMartinsen:
This addresses issue #3971 where build failed when building as a component because the include directory for wifi_provisioning was missing.


me-no-dev:
thanks :)

Issue #3615, opened 2020-01-06
BasicOTA and "ESP32 sketch data upload" tool


JPMJPM:
Hi,

Board: ESP32 Dev module
Core Installation: esp32 v1.0.2
IDE name: Arduino IDE v1.8.10
CPU Frequency: 240MHz
Flash Frequency: 80MHz
PSRAM enabled: Disabled
Upload Speed: 115200
Computer OS: Windows 10

I am testing BasicOTA and uploading the sketch works fine using
the port "hostname at 192.168.1.28 (ESP32 Dev Module)" but the
"ESP32 sketch data upload" tool does no work.

Regards.

lbernstone:
It works as designed, which is that it doesn't do that.  The sketch upload tool in Arduino IDE is for directly connected devices.  However, the [espota.py](https://github.com/esp8266/Arduino/blob/master/tools/espota.py) script can upload spiffs partition images, and the [update library](https://github.com/espressif/arduino-esp32/tree/master/libraries/Update) can be used to write flexible updaters to meet your needs (push/pull/http/etc).
There are lots of examples that google knows about. 
https://github.com/lbernstone/asyncUpdate/blob/master/AsyncUpdate.ino

JPMJPM:
I forgot to say that I use OTA password "ArduinoOTA.setPassword("password");".
With "ESP32 sketch data upload" there is not any dialog box pops up asking for
password that's way  the data upload is not possible.

Timmmy001:
Hello together.
I can confirm the problem:
Without a OTA-Password (Hash in my case) it works.

With a password I get this error message in Arduino IDE after password-windows

> Sending invitation to 192.168.188.36 
Traceback (most recent call last):
  File "espota.py", line 345, in <module>
  File "espota.py", line 340, in main
  File "espota.py", line 127, in serve
UnicodeDecodeError: 'ascii' codec can't decode byte 0xdc in position 38: ordinal not in range(128)
Failed to execute script espota

(Win 10 - German // ArduinoIDE 1.8.13 hourly)

Timmmy001:
Hello again.
I found the problem. It is a incompatible with special character like "ü" "ö" "ß" (German).
Normal characters like "abc 123 ?!$"  are ok.

Timmmy001:
I'am not really a python expert.
But it works - need a compile to .exe (#4005)

Timmmy001:
The espota.pl file also works without modifications. :-|
[...]

PR #4005, opened 2020-05-16
Update espota.py with utf-8

Timmmy001:
Add 'utf-8' option for correct password-handling with special character ä,ü,ö,ß, etc.
See [Python info](https://coderwall.com/p/mvwzcw/python-hashlib-and-unicode)
Fix: #3615

### DIFF
--- a/tools/espota.py
+++ b/tools/espota.py
@@ -128,10 +128,10 @@ def serve(remoteAddr, localAddr, remotePort, localPort, password, filename, comm
     if(data.startswith('AUTH')):
       nonce = data.split()[1]
       cnonce_text = '%s%u%s%s' % (filename, content_size, file_md5, remoteAddr)
-      cnonce = hashlib.md5(cnonce_text.encode('utf-8')).hexdigest()
-      passmd5 = hashlib.md5(password.encode('utf-8')).hexdigest()
+      cnonce = hashlib.md5(cnonce_text.encode('utf-8-sig')).hexdigest()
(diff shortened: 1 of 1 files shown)


Wolftronics-SBC:
hi please modified uft-8 to utf-8-sig working "uft16 languages"

Timmmy001:
@Wolftronics-SBC : I do it. But beware: This is now a breaking change!
I think this is not a very good idea:
Password | code : hash
admin |               :  21232f297a57a5a743894a0e4a801fc3
admin | utf-8       :  21232f297a57a5a743894a0e4a801fc3
admin | utf-8-sig :  dad6faebab54e34a71faee9455e78a61
admin | utf-16     :  2e05d1660c002f82c6100cac1eda1235

Wolftronics-SBC:
string.encode() python default option utf-8 what you sent in modification is practically doing nothing... but we need to analyze the code better... returns to utf-8 in this row "sock2.sendto(message.encode(), remote_address)" in principle, encode() should not be here, because the variables have already been assigned the character encoding...

Timmmy001:
Yeah, you're right.
The espota.pl file also works without modifications. :-|
But the compiled exe differs from the behaviour.
I have compiled the file myself as .exe and it now works with special characters.

Can someone trigger the recompilation in GitHub?

lbernstone:
This should be submitted and vetted at [esp8266/arduino](https://github.com/esp8266/Arduino/blob/master/tools/espota.py) first.

PR #4012, opened 2020-05-19
Ota update

sweetymhaiske:
This sketch provide functionality for OTA firmware upgrade.


libraries/Update/examples/OTA_Update/OTA_Update.ino
```diff
@@ -0,0 +1,83 @@
...
+    "BQUHAgEWHGh0dHBzOi8vd3d3LmRpZ2ljZXJ0LmNvbS9DUFMwDQYJKoZIhvcNAQEL\n"\
+    "BQADggEBAC/iN2bDGs+RVe4pFPpQEL6ZjeIo8XQWB2k7RDA99blJ9Wg2/rcwjang\n"\
+    "B0lCY0ZStWnGm0nyGg9Xxva3vqt1jQ2iqzPkYoVDVKtjlAyjU6DqHeSmpqyVDmV4\n"\
+    "7DOMvpQ+2HCr6sfheM4zlbv7LFjgikCmbUHY2Nmz+S8CxRtwa+I6hXsdGLDRS5rB\n"\
+    "bxcQKegOw+FUllSlkZUIII1pLJ4vP1C0LuVXH6+kc9KhJLsNkP5FEx2noSnYZgvD\n"\
+    "0WyzT7QrhExHkOyL4kGJE7YHRndC/bseF/r/JUuOUFfrjsxOFT+xJd1BDKCcYm1v\n"\
+    "upcHi9nzBhDFKdT3uhaQqNBU4UtJx5g=\n"\
+    "-----END CERTIFICATE-----";
+
+const char * url = "https://s3.amazonaws.com/fwimageupload-customer2/customers/bokne05bbses72ho07lg/firmwareimages/im45HvgMR.bin";
```
chegewara, line 61: You shouldnt post real link to your S3 buckets and files.

libraries/Update/src/OTA_Update.cpp
```diff
@@ -0,0 +1,70 @@
...
+    config.cert_pem = server_cert;
+    config.skip_cert_common_name_check = skip_cert_common_name_check;
+    config.event_handler = _http_event_handler;
+    esp_err_t ret = esp_https_ota(&config);
+    if (ret == ESP_OK) {
+        ESP.restart();
+    } else {
+        log_e("Firmware upgrade failed");
+    }
+    while (1) {
```
chegewara, line 49: If update failed, maybe its better to delete task instead?

libraries/Update/src/OTA_Update.cpp
```diff
@@ -0,0 +1,70 @@
...
+    log_i("Attempting to connect to the server.........");
+
+    esp_http_client_config_t config;
+    config.url = ota_url;
+    config.cert_pem = server_cert;
+    config.skip_cert_common_name_check = skip_cert_common_name_check;
+    config.event_handler = _http_event_handler;
+    esp_err_t ret = esp_https_ota(&config);
+    if (ret == ESP_OK) {
+        ESP.restart();
```
chegewara, line 45: Unconditional reset on update success is not always good thing.

me-no-dev: agreed, you have an event handler. use it to tell the user that all went good, so that they can decide if they want to reboot now or later (maybe save some state)

libraries/Update/src/OTA_Update.cpp
me-no-dev: yup. delete the task if not needed

libraries/Update/src/OTA_Update.h
```diff
@@ -0,0 +1,10 @@
+#include "esp_http_client.h"
+#define HttpEvent_t esp_http_client_event_t
+class OTA_UpdateClass {
+
+    public:
+    void start_firmware_upgrade(const char * cert, const char * url, bool skip_cert_cn_check);
+    void onHttpEvent(void (*ptr)(HttpEvent_t *));
+};
+
+extern OTA_UpdateClass OTA;
```
me-no-dev, line 10: use AWS_OTA and not OTA. You are reserving a high level namespace for just one specific implementation. If AWS is not fitting, maybe HTTPS or something. not just OTA. this goes for the file names, class names and the exposed object.

libraries/Update/examples/OTA_Update/OTA_Update.ino
```diff
@@ -0,0 +1,83 @@
...
+    while (WiFi.status() != WL_CONNECTED) {
+        Serial.print(".");
+        delay(1000);
+    }
+
+    Serial.print("Connected to ");
+    Serial.println(ssid);
+    
+    OTA.onHttpEvent(HttpEvent);
+    OTA.start_firmware_upgrade(server_certificate, url, true); //Pass server certificate, URL, skip_cert_common_name_check
```
me-no-dev, line 79: what if no certificate? can I update from my own server?

sweetymhaiske: it depends upon config option CONFIG_OTA_ALLOW_HTTP

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,96 @@
...
+        delay(1000);
+    }
+
+    Serial.print("Connected to ");
+    Serial.println(ssid);
+    
+    HttpsOTA.onHttpEvent(HttpEvent);
+    HttpsOTA.start_firmware_upgrade(server_certificate, url, true); //Pass server certificate, URL, skip_cert_common_name_check
+
+    while(1)
```
me-no-dev, line 82: this is the purpose of the loop function. move the code inside this `while` to the loop

me-no-dev: ah wait... so you are waiting for status true. what will happen if it does not come? This will loop forever?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+esp_err_t _http_event_handler(esp_http_client_event_t *evt)
+{
+    http_func_ptr(evt); 
+    return ESP_OK;
+}
+
+void aws_iot_task(void *param)
+{
+    log_i("Starting OTA example");
```
me-no-dev, line 37: this is library code, not example. maybe remove the line. The log below should make more sense too. 3 dots at the end are the norm, more dots do not clarify the situation better.

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,80 @@
...
+        ota_successful = true;
+    } else {
+        log_e("Firmware upgrade failed");
+        if(ota_task_handle != NULL) {
+            vTaskDelete(ota_task_handle);
+        }
+    }
+    while (1) {
+        vTaskDelay(1000 / portTICK_PERIOD_MS);
+    }
```
me-no-dev, line 56: what happened to deleting the task?

libraries/Update/src/HttpsOTAUpdate.h
```diff
@@ -0,0 +1,12 @@
+#include "esp_http_client.h"
+#define HttpEvent_t esp_http_client_event_t
+class HttpsOTAUpdateClass {
+
+    public:
+    void start_firmware_upgrade(const char * cert, const char * url, bool skip_cert_cn_check);
+    void onHttpEvent(void (*ptr)(HttpEvent_t *));
+    bool ota_successful();
+    bool ota_fail();
```
dhrishi, line 9: Some function names in Camel case and some in Snake case? Can you please maintain consistency as per the ESP32 Arduino standards?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "esp_https_ota.h"
+
+#include"HttpsOTAUpdate.h"
+#include"Esp.h"
+
+static TaskHandle_t ota_task_handle = NULL;
+const char * server_cert;
+const char * ota_url;
+bool skip_cert_common_name_check;
+void (*http_func_ptr)(HttpEvent_t*);
```
dhrishi, line 26: Can you please use a better name? Something like `http_event_cb`?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+void aws_iot_task(void *param)
+{
+    esp_http_client_config_t config;
+    config.url = ota_url;
+    config.cert_pem = server_cert;
+    config.skip_cert_common_name_check = skip_cert_common_name_check;
+    config.event_handler = _http_event_handler;
+    esp_err_t ret = esp_https_ota(&config);
+    if(ret == ESP_OK){
```
dhrishi, line 44: Please see the formatting, here and at other places

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "esp_http_client.h"
+#include "esp_https_ota.h"
+
+#include"HttpsOTAUpdate.h"
+#include"Esp.h"
+
+static TaskHandle_t ota_task_handle = NULL;
+const char * server_cert;
+const char * ota_url;
+bool skip_cert_common_name_check;
```
dhrishi, line 25: Can all the above variables be static?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,85 @@
...
+    http_func_ptr = ptr;
+}
+
+void HttpsOTAUpdateClass::start_firmware_upgrade(const char *cert, const char *url, bool skip_cert_cn_check)
+{
+    server_cert = cert;
+    ota_url = url;
+    skip_cert_common_name_check = skip_cert_cn_check;
+
+    if (xTaskCreate(&aws_iot_task, "aws_iot_task", 9216, NULL, 5, &ota_task_handle) != pdPASS) {
```
dhrishi, line 80: Can you please use a macro for stack size, for better readability?

libraries/Update/src/HttpsOTAUpdate.cpp
dhrishi: Or, instead can you directly pass them (through a structure or something) through `xTaskCreate` in `start_firmware_upgrade`

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include"HttpsOTAUpdate.h"
+#include"Esp.h"
+
+static TaskHandle_t ota_task_handle = NULL;
+const char * server_cert;
+const char * ota_url;
+bool skip_cert_common_name_check;
+void (*http_func_ptr)(HttpEvent_t*);
+bool success = false;
+bool fail = false;
```
dhrishi, line 28: A single int variable like `status` would suffice?

libraries/Update/src/HttpsOTAUpdate.cpp
dhrishi: Also, where is `ota_task_handle` assigned? Am I missing anything here?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,80 @@
...
+    config.cert_pem = server_cert;
+    config.skip_cert_common_name_check = skip_cert_common_name_check;
+    config.event_handler = http_event_handler;
+    esp_err_t ret = esp_https_ota(&config);
+    if(ret == ESP_OK){
+        status = 1;
+    } else {
+        status = 0;
+        if(ota_task_handle != NULL) {
+            vTaskDelete(ota_task_handle);
```
dhrishi, line 49: You can just pass `NULL` here if you are executing in the context of the same task that needs to be deleted

libraries/Update/src/HttpsOTAUpdate.cpp
dhrishi: Oops, my bad. I confused the last parameter for the `void *pvParameters` (3rd last parameter)

dhrishi:
@sweetymhaiske Why are you adding a new commit everytime the changes are done? Are the commits squashed during merge? cc @me-no-dev

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+
+#include "esp32-hal-log.h"
+#include "esp_http_client.h"
+#include "esp_https_ota.h"
+
+#include"HttpsOTAUpdate.h"
+#include"Esp.h"
+#define stackSize 9216
```
dhrishi, line 21: Macros should be in all caps

me-no-dev:
> @sweetymhaiske Why are you adding a new commit everytime the changes are done? Are the commits squashed during merge? cc @me-no-dev

Yup. All is squashed and then merged. Surely a rebased branch looks better, but there is no need for it before all is finalised.

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,101 @@
...
+            Serial.println("Http Event On Connected");
+            break;
+        case HTTP_EVENT_HEADER_SENT:
+            Serial.println("Http Event Header Sent");
+            break;
+        case HTTP_EVENT_ON_HEADER:
+            Serial.printf("Http Event On Header, key=%s, value=%s\n", event->header_key, event->header_value);
+            break;
+        case HTTP_EVENT_ON_DATA:
+            //Serial.printf("Http Event On Data, len=%d\n", event->data_len);
```
dhrishi, line 23: Either uncomment or remove the print statement?

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,101 @@
...
+        case HTTP_EVENT_ON_FINISH:
+            Serial.println("Http Event On Finish");
+            break;
+        case HTTP_EVENT_DISCONNECTED:
+            Serial.println("Http Event Disconnected");
+            break;
+    }
+}
+
+const char * server_certificate = "-----BEGIN CERTIFICATE-----\n" \
```
dhrishi, line 34: static?

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,101 @@
...
+     "Y3JsMB0GA1UdDgQWBBSoSmpjBH3duubRObemRWXv86jsoTANBgkqhkiG9w0BAQsF\n" \
+     "AAOCAQEA3TPXEfNjWDjdGBX7CVW+dla5cEilaUcne8IkCJLxWh9KEik3JHRRHGJo\n" \
+     "uM2VcGfl96S8TihRzZvoroed6ti6WqEBmtzw3Wodatg+VyOeph4EYpr/1wXKtx8/\n" \
+     "wApIvJSwtmVi4MFU5aMqrSDE6ea73Mj2tcMyo5jMd6jmeWUHK8so/joWUoHOUgwu\n" \
+     "X4Po1QYz+3dszkDqMp4fklxBwXRsW10KXzPMTZ+sOPAveyxindmjkW8lGy+QsRlG\n" \
+     "PfZ+G6Z6h7mjem0Y+iWlkYcV4PIWL1iwBi8saCbGS5jN2p8M+X+Q7UNKEkROb3N6\n" \
+     "KOqkqm57TH2H3eDJAkSnh6/DNFu0Qg==\n" \
+     "-----END CERTIFICATE-----\n";
+
+const char * url = ""; //state url of your firmware image
```
dhrishi, line 62: static?

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,101 @@
...
+    HttpsOTA.startFirmwareUpgrade(server_certificate, url, true); //Pass server certificate, URL, skip_cert_common_name_check
+
+    while(1)
+    {
+        Serial.print(".");
+        delay(10000);
+        if(HttpsOTA.otaStatus() == 1){ //OTA firmware successfully written
+            Serial.println("Firmware written successfully. To reboot device call API ESP.restart()");
+            //Serial.println("Rebooting ESP");
+            //ESP.restart();
```
dhrishi, line 90: You can remove the commented code, as the same message is conveyed through the print statement

libraries/Update/examples/HTTPS_OTA_Update/Readme.md
```diff
@@ -0,0 +1,30 @@
+# OTA Firmware Upgrade for Arduino
+This sketch allows Arduino user to perform Over The Air (OTA) firmware upgrade. It uses https.
```
dhrishi, line 2: https -> HTTPS at all the places

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
sweetymhaiske: It will monitor both cases for success and failure and accordingly it will break

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+
+#include "esp32-hal-log.h"
+#include "esp_http_client.h"
+#include "esp_https_ota.h"
+
+#include"HttpsOTAUpdate.h"
+#include"Esp.h"
```
dhrishi, line 20: A white space missing after include in 2 statements above

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
me-no-dev: I actually wonder wether that is not pointless all together. Given that we have the events above, why not do the prints and reboots there? Having such while loop is not best IMO. If you have to use anything, use a semaphore and wait for it to be given by the event handler above. You could do that in the class too I guess and expose `HttpsOTA::waitForResult(timeout_ms)` which would do that exact thing and will save the Semaphore from the users.

libraries/Update/src/HttpsOTAUpdate.cpp
me-no-dev: I'm thinking of how to automate this and use IDF's formatter.

libraries/Update/src/HttpsOTAUpdate.cpp
me-no-dev: I think that the following is enough for this, though I am not sure why you need a task for that at all. Arduinos are used to blocking API, so running the function directly is totally acceptable. It's even fine if the callback will run in the user task.

sweetymhaiske: Without creating the task the code fails to initialise HTTP connection.

sweetymhaiske: The main task stack size may not sufficient to establish HTTPs connection.

me-no-dev: then lets increase the main task size. Holding the memory for this one in no way makes it better. Plus you do not exit the task, but instead loop forever. If you want to use your own task, exit it when done.

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,99 @@
...
+        case HTTP_EVENT_ON_FINISH:
+            Serial.println("Http Event On Finish");
+            break;
+        case HTTP_EVENT_DISCONNECTED:
+            Serial.println("Http Event Disconnected");
+            break;
+    }
+}
+
+static const char * server_certificate = "-----BEGIN CERTIFICATE-----\n" \
```
me-no-dev, line 33: move this to the top of the sketch.

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
me-no-dev: Since you have callbacks, you do not need this block at all. Please remove it.

libraries/Update/src/HttpsOTAUpdate.cpp
me-no-dev: here is how I see this function:
```cpp
void https_ota_task(void *param)
 {
     esp_err_t ret = esp_https_ota((const esp_http_client_config_t *)param);
     if(ret == ESP_OK){
         status = HTTPS_OTA_SUCCESS;
     } else {
         status = HTTPS_OTA_FAIL;
     }
     vTaskDelete(NULL);
}
```

me-no-dev: You should use `xEventGroupSetBits` and `xEventGroupGetBits` to handle the status change and getter. One could also wait (with timeout) for a result to come.

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+
+#include "esp32-hal-log.h"
+#include "esp_http_client.h"
+#include "esp_https_ota.h"
+
+#include "HttpsOTAUpdate.h"
+#include "Esp.h"
+#define STACKSIZE 9216
```
me-no-dev, line 21: this is a bit too generic. Maybe use `OTA_TASK_STACK_SIZE` instead

libraries/Update/src/HttpsOTAUpdate.h
```diff
@@ -0,0 +1,14 @@
+#include "esp_http_client.h"
+#define HttpEvent_t esp_http_client_event_t
+#define HTTPS_OTA_SUCCESS 1
+#define HTTPS_OTA_FAIL 0
+
+class HttpsOTAUpdateClass {
+
+    public:
+    void startFirmwareUpgrade(const char * cert_pem, const char * url, bool skip_cert_common_name_check);
```
me-no-dev, line 9: change the method name to `begin`

me-no-dev: does it not make more sense url to be first argument? maybe add default value to `skip_cert_common_name_check `? Again, what if I do not have the certificate? I am lame and have web server on my RaspberryPi that I want to update from. Have Letsencrypt for SSL and that is all I know. Can I update without the certificate? Like I can connect to an SSL server without certificate otherwise. Sure it's not the most secure thing, but I am just testing and not deploying commercially.

sweetymhaiske: it depends upon config option CONFIG_OTA_ALLOW_HTTP
RIght now this option is not set , Hence certificates are mandatory.

me-no-dev: I do not want HTTP, it's still HTTPS, but just like you browse the web and do not provide yourself a certificate for every SSL site you visit. Get it? Skip certificate verification. Is it possible?

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,92 @@
+#include "WiFi.h"
```
dhrishi, line 1: A comment mentioning the details/purpose of the example?

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,92 @@
...
+#include "HttpsOTAUpdate.h"
+
+static const char* ssid     = "your-ssid";  // your network SSID (name of wifi network)
+static const char* password = "your-password"; // your network password
+
+static const char * url = ""; //state url of your firmware image
```
dhrishi, line 7: Please maintain consistency in the syntax and make changes as per ESP32 Arduino guidelines

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+        xEventGroupSetBits(ota_status, OTA_FAILURE_BIT);
+        vTaskDelete(NULL);
+    }
+    while (1) {
+        vTaskDelay(1000 / portTICK_PERIOD_MS);
+    }
+}
+
+int HttpsOTAUpdateClass::otaStatus()
```
dhrishi, line 52: Why do you need ota in the name? The class has the name, so you can rename this to status?

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,83 @@
...
+        status = HTTPS_OTA_SUCCESS;
+    }
+    if(set_bit & OTA_FAILURE_BIT) {
+        status = HTTPS_OTA_FAIL;
+    }
+    xEventGroupClearBits(ota_status, OTA_SUCCESSFUL_BIT | OTA_FAILURE_BIT);    
+    return status;
+}
+
+void HttpsOTAUpdateClass::onHttpEvent(void (*http_event_cb_t)(HttpEvent_t*))
```
dhrishi, line 65: Why not use a typedef here for better readability?

libraries/Update/src/HttpsOTAUpdate.cpp
sweetymhaiske: > You should use `xEventGroupSetBits` and `xEventGroupGetBits` to handle the status change and getter. One could also wait (with timeout) for a result to come.

Yes Required changes are made in new commit

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -0,0 +1,81 @@
...
+int HttpsOTAUpdateClass::status()
+{
+    EventBits_t set_bit =  xEventGroupWaitBits(ota_status, OTA_SUCCESSFUL_BIT | OTA_FAILURE_BIT, true, false, portMAX_DELAY);
+    if(set_bit & OTA_SUCCESSFUL_BIT) {
+        check_status = HTTPS_OTA_SUCCESS;
+    }
+    if(set_bit & OTA_FAILURE_BIT) {
+        check_status = HTTPS_OTA_FAIL;
+    }
+    xEventGroupClearBits(ota_status, OTA_SUCCESSFUL_BIT | OTA_FAILURE_BIT);
```
me-no-dev, line 59: do you need to clear the bits here? maybe on `begin`? Also I see that you do not check if the event group is already created in begin. maybe you should check that and clear the bits if the group exists. Clearing here would mean that status can be asked only once, and that should not be a limitation IMO.

me-no-dev: Also I see that you are waiting for the bits. Is that necessary? Maybe status `UPDATING` makes sense?

me-no-dev:
```cpp
typedef enum {
  HTTPS_OTA_IDLE,
  HTTPS_OTA_UPDATING,
  HTTPS_OTA_SUCCESS,
  HTTPS_OTA_FAIL
} HttpsOTAStatus_t;
```

libraries/Update/src/HttpsOTAUpdate.h
```diff
@@ -0,0 +1,20 @@
...
+    HTTPS_OTA_SUCCESS,
+    HTTPS_OTA_FAIL
+}HttpsOTAStatus_t;
+
+class HttpsOTAUpdateClass {
+
+    public:
+    void begin(const char *url, const char *cert_pem, bool skip_cert_common_name_check = true);
+    void onHttpEvent(void (*http_event_cb_t)(HttpEvent_t *));
+    int status();
```
me-no-dev, line 17: return type should be `HttpsOTAStatus_t`

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
```diff
@@ -0,0 +1,96 @@
+//This sketch provide the functionality of OTA Firmware Upgrade
+#include "WiFi.h"
+#include "HttpsOTAUpdate.h"
+
+static const char *ssid     = "your-ssid";  // your network SSID (name of wifi network)
+static const char *password = "your-password"; // your network password
+
+static const char *url = ""; //state url of your firmware image
```
me-no-dev, line 8:
```cpp
static const char *url = "https://example.com/firmware.bin"; //state url of your firmware image
```

libraries/Update/examples/HTTPS_OTA_Update/HTTPS_OTA_Update.ino
me-no-dev: will you take care of adding a bit more info, like:
```cpp
// This sketch shows how to implement HTTPS firmware update Over The Air.
// Please provide your WiFi credentials, https URL to the firmware image and the server certificate.

me-no-dev (approved):

Issue #3903, opened 2020-04-13
SD_MMC library issue with BOARD_HAS_1BIT_SDMMC


Stanimir-Petev:
### Hardware:
Board:							ESP32-Gateway
Core Installation version:			1.0.4
IDE name:						Arduino IDE 1.8.10
Flash Frequency:					80Mhz
PSRAM enabled:                                     no
Upload Speed:					115200
Computer OS:                                         Windows 10

### Description:
I tried to use the example sketch from SD_MMC library: SDMMC_Test in 1Bit mode. SD card example itself works without problems. But the other 3 pins that are not used in this mode behave like they have been initialized as alternative function so they can't be used as GPIO (I have attached LED on them and tried to blink it). After some research I have found a way to solve the problem but it involved editing the library itself.
In the file SD_MMC.cpp lines 65-67 by default are:
```cpp
    if(mode1bit) {
        host.flags = SDMMC_HOST_FLAG_1BIT; //use 1-line SD mode
    }
```
That changes the mode from 4 to 1 bit, but based on what I saw it doesn't change slot_config structure and thus the pins to be initialized. So I added one line within the body of the if operator and is now:
```cpp
    if(mode1bit) {
        host.flags = SDMMC_HOST_FLAG_1BIT; //use 1-line SD mode
        slot_config.width = 1;
    }
```
So now it works but I am not sure if this is the best approach to the problem since I have to change library file. Is there another way to be done? If no - is there any chance to see fix for that in one of the future official releases of the Espressif package?

### Sketch:
Basically the sketch is the default example SDMMC_Test + added blinking LED on pin 4 and the setup - attached LED on pin 4.
[SDMMC_Test_toggle_4.txt](https://github.com/espressif/arduino-esp32/files/4469677/SDMMC_Test_toggle_4.txt)

DanKoloff:
I can confirm this problem exists for both SD card libraries from the examples ("SD_MMC" and "SD"). Defining 1-bit SD card mode still bugs the pins that are used by default for 4-bit SD card mode.

lbernstone:
If you fork the repository, edit the file (right in your browser), and submit a PR, then this can be fixed.

Stanimir-Petev:
I tried that now. I committed that change into my fork repository. But since I already had a pending pull request (from ~3 weeks ago) it got affected too. Before the commit (on the SD library) the previous pull request had 21/21 checks passed but now is 0/0 and has a yellow dot on it.
[...]

PR #4020, opened 2020-05-21
Make 1bit SDMMC interface work

cyberman54:
According to ESP-IDF API
https://github.com/espressif/esp-idf/blob/dc14d027cec49260b78efb1c3724b1710f2f3ec0/docs/en/api-reference/peripherals/sdmmc_host.rst :

"To configure the bus width, set the width field of :cpp:class:`sdmmc_slot_config_t`. For example, to set 1-line mode:

sdmmc_slot_config_t slot = SDMMC_SLOT_CONFIG_DEFAULT();
slot.width = 1;
"

E.g. necessary on Olimex ESP32-POE, because this board is using GPIO12 for other purpose (switching PoE power). GPIO12 ususally is used in 4 bit SDMMC interface. Thus, should be disabled in 1bit mode to avoid conflicts if GPIO12 is used for other purpose.

### DIFF
--- a/libraries/SD_MMC/src/SD_MMC.cpp
+++ b/libraries/SD_MMC/src/SD_MMC.cpp
@@ -64,6 +64,7 @@ bool SDMMCFS::begin(const char * mountpoint, bool mode1bit)
 #endif
     if(mode1bit) {
         host.flags = SDMMC_HOST_FLAG_1BIT; //use 1-line SD mode
+	slot_config.width = 1;
     }
 
     esp_vfs_fat_sdmmc_mount_config_t mount_config = {


DanKoloff:
https://github.com/OLIMEX/ESP32-POE/issues/14#issuecomment-632505613

DanKoloff:
One of our employees tried to do it here:

https://github.com/espressif/arduino-esp32/issues/3903

me-no-dev:
take a look here :) https://github.com/espressif/arduino-esp32/blob/master/variants/esp32-evb/pins_arduino.h#L31 this is olimex EVB. Also 1bit SDMMC

DanKoloff:
> take a look here :) https://github.com/espressif/arduino-esp32/blob/master/variants/esp32-evb/pins_arduino.h#L31 this is olimex EVB. Also 1bit SDMMC

I believe that the issue is not properly and clearly explained in the post. I believe that the issue is: 

- The pins used for 4-bit mode are not freed to use WHEN and IF you decide to set the 1-bit mode. Meaning that 1-bit mode works but the rest of the pins are still not free to use.

me-no-dev:
@DanKoloff killing me with this nickname :D alright, clear! Merging..

DanKoloff:
@cyberman54 Thanks for the contribution

PR #4021, opened 2020-05-22
update M5Camera pins

Gitshaoxiang:
update the pins number:
product docs:
https://docs.m5stack.com/#/en/unit/m5camera


me-no-dev:
what happens with the old pins? my current camera?

Gitshaoxiang:
actually we just produce few "old pins" version  camera before.  it already stop produce more than one year. we must user are use the "new pins" version

me-no-dev:
maybe we should add it as V2 instead. You will be breaking cameras this way, regradles when they are made. As you see, I added support for all your cameras.

libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
```diff
@@ -11,6 +11,7 @@
 #define CAMERA_MODEL_WROVER_KIT // Has PSRAM
 //#define CAMERA_MODEL_ESP_EYE // Has PSRAM
 //#define CAMERA_MODEL_M5STACK_PSRAM // Has PSRAM
```
me-no-dev, line 13: feel free to add better explanation to both boards, so it is easier for users to distinguish. Like "manufactured before ..." or something else.

Gitshaoxiang: i think it is enough. i add the "version B" note. our user will know it.  actually , we just saled few old version. most is version B.  that is why i hope change it not add a new version.   but add a version B define is a good solution. could you meger the PR.?

me-no-dev:
done :)

PR #4023, opened 2020-05-23
Expose setTimeZone to Arduino IDE

dearmash:
Allow to set time zone to get time from RTC without network connectivity.


dearmash:
The basic idea is to be able to set the timezone separately from configTime, that way if network fails, but RTC is still OK, the time zone is correct.  

This is useful for sketches where you put the esp32 into deep sleep between calls where the RTC might retain "close enough" time, and the NTP refresh is just to update accuracy.

PR #4030, opened 2020-05-27
Camera pin mappings for the M5Stack ESP32CAM device

nigel-mellor:
M5Stack's product offering includes various ESP32-based camera devices. They are all broadly similar, all being based on the ESP32-D0WD core, but have some subtle differences between variants around PSRAM inclusion and camera pin mappings.

The ESP32CAM "unit" product (SKU: U007) has no PSRAM and maps the "Pixel Data 0" (Y2 GPIO/OV2640 D2) camera pin to IO17 rather than IO32 as is the case with most other M5Stack ESP32 camera devices.

The ESP32CAM also requires the "hmirror" and "vflip" camera configuration bits setting in order to capture a true image.

The relevant M5Stack pin configuration documentation can be found [here](https://docs.m5stack.com/#/en/unit/esp32cam?id=pinmap).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ tools/mkspiffs/mkspiffs.exe
 .vs/
 __vm/
 *.vcxproj*
+.vscode/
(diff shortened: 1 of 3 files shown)


me-no-dev:
I was not sure if those were even sold, since it makes little sense to have the camera without PSRAM

nigel-mellor:
Still showing as "in stock" on the M5Stack website: https://m5stack.com/collections/m5-unit/products/esp32-camera

They can manage SVGA capture (800 x 600) without PSRAM but now seem to be one of the very few ESP32 camera boards sold without PSRAM. The PCB seems to have solder pads and provision for adding a PSRAM device.

me-no-dev:
I actually have a couple of those boards, so I am aware, but was not aware that they are sold. As you said, JPEG and SVGA is the limit.

Issue #4046, opened 2020-05-31
Writing >20 Bytes to Characteristic causes all registered Write Callbacks to fire


emitategh:
As explained in title: When sending more than 20 bytes to any Characteristic causes all registered Write Callbacks to fire.

**Board**: ESP32 DevKit v1
**Chip**: ESP32D0WDQ6 (revision 
**Features**: WiFi, BT, Dual Core, 240MHz
**Auto-detected Flash size**: 4MB
**Core Instalation date**: 01/05/2020
**IDE**: Arduino 1.8.12
**Upload Speed**: 115200
**Computer OS**: Windows 10 Pro

**Description**:
Same as title

**Example Sketch**

```cpp
#include <Arduino.h>
#include <BLEDevice.h>
#include <BLEUtils.h>
#include <BLEServer.h>
#include <BLECharacteristic.h>
// See the following for generating UUIDs:
// https://www.uuidgenerator.net/

#define SERVICE_UUID         "4fafc201-1fb5-459e-8fcc-c5c9c331914b"
#define CHARACTERISTIC1_UUID "aeb5483e-36e1-4688-b7f5-ea07361b26a8"
#define CHARACTERISTIC2_UUID "beb5483e-36e1-4688-b7f5-ea07361b26a8"

class Characteristic1_Callbacks: public BLECharacteristicCallbacks {
  void onRead(BLECharacteristic *pCharacteristic){
    Serial.println("Characteristic1 Read");
  }
  void onWrite(BLECharacteristic *pCharacteristic){
    Serial.print("Characteristic1 Written: ");
    Serial.println((char *) pCharacteristic->getData());
  }
};

class Characteristic2_Callbacks: public BLECharacteristicCallbacks {
  void onRead(BLECharacteristic *pCharacteristic){
    Serial.println("Characteristic2 Read");
  }
  void onWrite(BLECharacteristic *pCharacteristic){
    Serial.print("Characteristic2 Written: ");
    Serial.println((char *) pCharacteristic->getData());
  }
};

void setup() {
  Serial.begin(115200);
  Serial.println("Starting BLE work!");

  BLEDevice::init("Long name works now");
  BLEServer *pServer = BLEDevice::createServer();
  BLEService *pService = pServer->createService(SERVICE_UUID);
  BLECharacteristic *pCharacteristic = pService->createCharacteristic(
                                         CHARACTERISTIC1_UUID,
                                         BLECharacteristic::PROPERTY_READ |
[...]

PR #4050, opened 2020-06-01
esp32-snippets BLE replacement

vicatcu:
This PR was created by simply simply running this NodeJS script from a clean checkout of the latest / master branch of https://github.com/nkolban/esp32-snippets

```javascript
const fs = require('fs');
const dest_dir = '/home/me/arduino-esp32/libraries/BLE/src';
const src_dir =  './cpp_utils';
const dest_files = fs.readdirSync(dest_dir);
const src_files = fs.readdirSync(src_dir);

src_files.forEach(file => {
  if (dest_files.indexOf(file) >= 0) {
    console.log(`copying file "${file}"`);
    fs.copyFileSync(`${src_dir}/${file}`, `${dest_dir}/${file}`);
  }
});

console.log('done');
```

It does not add new files, it merely replaces existing files in the arduino-esp32 library folder one-for-one.

This is in light of https://github.com/nkolban/esp32-snippets/issues/796 and by proxy aims to address   https://github.com/espressif/arduino-esp32/issues/4046#issuecomment-636528614


vicatcu:
This is never going to be merged, and maybe it shouldn't be

h2zero:
For the most part this should be merged with a few changes, but like you said "never going to be merged".

I maintain a library [here](https://github.com/h2zero/NimBLE-Arduino) based on this API that is updated a **much** lighter weight if you want to check it out, I'd appreciate some more testers before release 😄.

PR #4055, opened 2020-06-03
Fix BLEClient handling of app registration results

buxtronix:
GATTC events are propagated to all clients, however clients should only act on events directed at their connection (conn_id). This fixes this so that clients ignore events not for them.


buxtronix:
On closer consideration, this may not be the root problem, I think the issue is that the gattc_if is not being set properly at initial client registration. I'll update this PR after verifying.

buxtronix:
Hm no, it seems the first commit works, not the second, but still doesnt seem to be the root cause. I'll close this and dig a bit more.

chegewara:
IIRC esp_ble_server is using gatt_if to distinguish connected peers, but esp_ble_client is using conn_id.

PR #4056, opened 2020-06-03
Fix to allow more than one certificate to be loaded (espressif#3248).

pulquero:
Static ptr meant any load would overwrite the first. Also ensured the buffer is null-terminated.


libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -274,18 +274,15 @@ bool WiFiClientSecure::verify(const char* fp, const char* domain_name)
...
-      free(dest);
-  }
-  dest = (char*)malloc(size);
+  char *dest = (char*)malloc(size+1);
   if (!dest) {
     return nullptr;
   }
   if (size != stream.readBytes(dest, size)) {
     free(dest);
     dest = nullptr;
```
basemotive, line 283: This should return nullptr. Because dest is modified further on in the code, this would lead to problems.

PR #4064, opened 2020-06-06
Have BLECLient gattc event handlers verify conn_id

buxtronix:
esp32 BLE core seems to send events to all clients, even those for others. This PR ensures that events are only handled when the connection is correct.


buxtronix:
I note that Espresso's ESP-IDF also acknowledges the event being broadcast to all clients:

https://github.com/espressif/esp-idf/blob/master/examples/bluetooth/bluedroid/ble/gattc_multi_connect/main/gattc_multi_connect.c#L150

PR #4070, opened 2020-06-07
rewrite GetChipID.ino

cweinhofer:
The previous GetChipID.ino example sketch was confusing and misleading -- and didn't provide an example that would actually be useful to a new user. Not only did it simply output the MAC address (which is already easily available) but it outputted it in reverse order.

The code has been rewritten so that it outputs the exact same thing as the ESP.getChipId() function on ESP8266 (a 32-bit integer matching the last 3 bytes of the MAC address). This is both a useful value and likely what a new user would expect given the name of this sketch.


cweinhofer:
Any idea when this pull request might get approved? Or even just get some feedback from someone?

PR #4078, opened 2020-06-09
Add static pin support

chegewara:


chegewara:
Add static pin support to ble

PR #4085, opened 2020-06-13
Extra PR (on top of #4050) for #4046: BLEDevice::deinit(false) should clear initialized.

geeksville:
Thanks to @vicatcu for the base PR.  @chegewara would you mind checking this change - see my comment in this commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ tools/mkspiffs/mkspiffs.exe
 .vs/
 __vm/
 *.vcxproj*
+.vscode/
--- a/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
+++ b/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
@@ -12,7 +12,8 @@
 //#define CAMERA_MODEL_ESP_EYE // Has PSRAM
 //#define CAMERA_MODEL_M5STACK_PSRAM // Has PSRAM
 //#define CAMERA_MODEL_M5STACK_V2_PSRAM // M5Camera version B Has PSRAM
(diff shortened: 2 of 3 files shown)


chegewara:
There is too many unnecessary changes. I doubt it will be accepted the way it is by @me-no-dev

geeksville:
@chegewara My commit is only one line.  The PR above it is just a massive copy of your tree I think?

chegewara:
Thats the problem. Mentioned PR makes too many changes, by copying from Kolban's repository, which will discard all PRs made here for some time.
Kolban's repository has few updates that could be used here, but some changes in this library are not compatible.

geeksville:
ok understood - sorry I thought you were working on that branch!  I agree then, I didn't realize the two trees had diverged.

geeksville:
Though it is a good datapoint, because that branch definitely fixed nasty BLE corruptions was was seeing.

vicatcu:
to be clear,  my PR was extremely naive transfer of diffs. I would also not be shocked if it were not accepted, but something needs to happen...

geeksville:
@chegewara and @me-no-dev:

Okay now that I know it isn't a matter of the other branch being the only one that is updated.  And not knowing the history - I'd like to ask some advice:

* I just looked into the root cause of this bug and I can make a nice clean small commit that fixes just this one issue.  I will submit it shortly.
* What is the history of these two 'branches' - it looks like there was more recent engineering in the other one?  Any idea of if there are valuable fixes missing from this one?

chegewara:
Its hard to say, because i dont follow code in any of them for some time. I have my own code that i am using and if something is not working i am fixing it.
In Kolban's repo there is few fixes that get merged after arduino started using its own code, and in arduino code there is few fixes since then.
Currently i am providing mostly support with my knowledge, answering questions etc.

PR #4088, opened 2020-06-14
WebServer: Fix OOB write

Tapped:
Successful exploitation could lead to arbitrary code execution.

The bug can be reproduced by running the following in a browser
```js
const formData = new FormData();
for (let i = 0;i < 33;++i) { formData.append("foo", i.toString()); }
await fetch("http://esp.local", { method: 'POST', body: formData });
```


Tapped:
@me-no-dev would be nice with a review

Tapped:
@me-no-dev I think this one should be merged.

PR #4089, opened 2020-06-14
Webserver: Add content length check

Tapped:
Without having a Content-Length check, sending a payload bigger than 512kB would crash the application and force a restart, In other words it prevents this type of DoS attack.

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -167,6 +167,10 @@ bool WebServer::_parseRequest(WiFiClient& client) {
 
     if (!isForm){
       size_t plainLength;
+      if (contentLength > HTTP_MAX_CONTENT_SIZE) {
+        log_e("'Content-Length' was larger than max content size (%d bytes)", HTTP_MAX_CONTENT_SIZE);
+        return false;
+      }
       char* plainBuf = readBytesWithTimeout(client, contentLength, plainLength, HTTP_MAX_POST_WAIT);
       if (plainLength < contentLength) {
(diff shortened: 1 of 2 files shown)


me-no-dev:
hmmm... that should not be the case. TCP should handle how much data is currently in queue regardless of the content length

Tapped:
I think I tested exploting this, however I'm not sure. Though by looking here: https://github.com/espressif/arduino-esp32/blob/34553f71e5272874f7b7eba64a3cb7cf472026e5/libraries/WebServer/src/Parsing.cpp#L54

We'll do an out of memory allocation if a too big payload is sent, so I think it's still an issue. Also by looking at the code there it seems like you can make the server hang, since it waits for all the data up to the content length without timeout, but that's another issue.

me-no-dev:
it should never try to allocate for the whole thing, it allocates based on available data, but that amount depends on TCP window and speed of consumption.

Tapped:
Yes, but I mean my computer can easily send a 10 MiB HTTP body without problems, which will crash the ESP.

Tapped:
Actually nevermind I didn't read what came after the `realloc`. It is actually checking if the allocation failed, and if so it'll we just opt-out.

Tapped:
We still have the issue of saturating this read loop, by sending small TCP packets in a very slow rate, which will prevent other clients to do requests while that happens. So the read loop should also have a timeout check I think if we want this server implementation to be more solid.

me-no-dev:
this server is not means as high performance, but rather as a very simple server. There are other more robust solutions out there.

PR #4090, opened 2020-06-14
Webserver: Add warning about library being unsafe

Tapped:
The Webserver code:
* Lacks tests
* Low code quality
* It's handling client requests in serial, which means by design it's easy to DoS
* Lacks an error reporting interface for users of the library

In my opinion, we should rather recommend and ship a battle tested embedded HTTP library than writing one ourself. For example https://github.com/cesanta/mongoose might be an option.


me-no-dev:
this is a bit unnecessary :) no sane developer will use it in production

PR #4103, opened 2020-06-19
ESP32S2: Static IP and DHCP stop fixes

atanisoft:
Fixing static IP configuration so it doesn't get overwritten by DHCP as part of WiFi.begin().

Fixing DHCP client stop if WiFi.config() is called before WiFi.begin() (as done in WiFiClientStaticIP.ino)


hasenradball:
What were happend if I had first called ‚WiFi.begin()‘ and then WiFi.config()?

atanisoft:
> What were happend if I had first called ‚WiFi.begin()‘ and then WiFi.config()?

You will have a DHCP address once connected to the AP and after calling WiFi.config() it should have your static IP.

Issue #3522, opened 2019-11-27
Change of WiFi behavior after recent commit: disconnect(true) seems to *permanently* close WiFi


mouridis:
### Hardware:
Board:							LOLIN32
Core Installation version:			Latest commit (4846d9c1862f38b8a7c50e5590dbcf70d21f77f8)
IDE name:						Arduino IDE / VSCode
Flash Frequency:					80Mhz
PSRAM enabled:                                    no
Upload Speed:					921600
Computer OS:                                        Win10-x64

### Description:
An old piece of code that worked for some time stopped working after updating core to latest commit (4846d9c1862f38b8a7c50e5590dbcf70d21f77f8). More specifically calls to WiFi.scanNetworks() failed with messages like this in the serial:
 ```
E (8376) wifi: esp_wifi_scan_start 1245 wifi not start
-2
E (9563) wifi: esp_wifi_scan_start 1245 wifi not start
-2

```

By git bisecting, I found the last core version that my code works to be commit e50613622e3d9dc3f7204b2c24f53ab026cc756f. Code stops working in commit 0cdfb0b193d9bf04adc50bf0d6020e95a28f7d21.

By slimming down my code, this is a simple sketch to reproduce the issue:
```cpp
#include <WiFi.h>
#include <HTTPClient.h>

void setup() {
  Serial.begin(115200);
  WiFi.persistent(false);
  WiFi.setAutoConnect(false);
  WiFiCheck();
}

void loop() {
  boolean trigger = false;
  while (Serial.available() > 0) {
    String temp = String(Serial.read());
    trigger = true;
  }
  if (trigger) {
    Serial.println(WiFi.scanNetworks());
  }
}


void WiFiCheck() {
  WiFi.mode(WIFI_STA);
  if (!connectToAP()) return;
  getPayloadFromServer();
  WiFi.disconnect(true);
}


boolean connectToAP() {
  Serial.printf("Attempting AP connection (10 sec timeout).\n");
  WiFi.begin("YourSSID", "YourPass");
  for (uint8_t i = 0; i < 10; i++) {
    if (WiFi.status() == WL_CONNECTED) break;
    delay(1000);
  }
  return (WiFi.status() == WL_CONNECTED);
}


void getPayloadFromServer() {
  HTTPClient http;
  Serial.printf("Sending HTTP GET request to remote server.\n");
  http.begin("http://ip-api.com/json/");
  if (http.GET() == HTTP_CODE_OK) {
[...]

PR #4114, opened 2020-06-24
Fix issue #3522 (WiFi does not restart after stopped)

mouridis:
This commit fixes issue https://github.com/espressif/arduino-esp32/issues/3522 where WiFi service fails to start after a WiFi.disconnect(true) or a WiFi.mode(WIFI_OFF).

I'm just doing the PR. Initial fix found by @rob58329 and modified by @lbernstone.


mc-hamster:
What are the next steps to get this merged in?

Rob58329:
Perhaps ask @me-no-dev to commit/merge this, as he committed the original "Add support for WiFi long range mode (#3190)".

I'm not sure how one would ask @me-no-dev though?

Issue #3677, opened 2020-01-27
BLEScan first scan ok, next no name, no service UUID


Nagymadar:
I am using BLE_scan.ino example. Unfortunately only the first scan gives valid device serviceUUID and Name. From the second scan onward SENSOR_PRO device gives no serviceUUID the HMSoft device gives empty string as Name. It seems it is an old bug, but still no solution? Serial data from boot:

Scanning...
Advertised Device: Name: SENSOR_PRO, Address: ed:3d:9a:96:86:a3, serviceUUID: 6a800001-b5a3-f393-e0a9-e50e24dcca9e 
Advertised Device: Name: HMSoft, Address: 90:9a:77:27:65:15, manufacturer data: 484d909a77276515, serviceUUID: 0000ffe0-0000-1000-8000-00805f9b34fb, txPower: 0 
Devices found: 2
Scan done!

Advertised Device: Name: SENSOR_PRO, Address: ed:3d:9a:96:86:a3 
Advertised Device: Name: , Address: 90:9a:77:27:65:15, manufacturer data: 484d909a77276515, serviceUUID: 0000ffe0-0000-1000-8000-00805f9b34fb, txPower: 0 
Devices found: 2
Scan done!

Advertised Device: Name: SENSOR_PRO, Address: ed:3d:9a:96:86:a3 
Advertised Device: Name: , Address: 90:9a:77:27:65:15, manufacturer data: 484d909a77276515, serviceUUID: 0000ffe0-0000-1000-8000-00805f9b34fb, txPower: 0 
Devices found: 2
Scan done!

etc...

chegewara:
Its hard to do anything if i cant reproduce issue:
```
03:27:33.927 -> Device: 
03:27:33.927 -> TX power => -7
03:27:33.927 -> RSSI => -96
03:27:33.927 -> Name: Chege
03:27:34.226 -> Device: 
03:27:34.226 -> TX power => -7
03:27:34.226 -> RSSI => -85
03:27:34.226 -> Name: Chege
03:27:34.493 -> Device: 
03:27:34.493 -> TX power => -7
03:27:34.493 -> RSSI => -85
03:27:34.493 -> Name: Chege
03:27:34.792 -> Device: 
03:27:34.792 -> TX power => -7
03:27:34.792 -> RSSI => -86
03:27:34.792 -> Name: Chege
03:27:35.092 -> Device: 
03:27:35.092 -> TX power => -7
03:27:35.092 -> RSSI => -84
[...]

PR #4126, opened 2020-07-01
Set scan_duplicate in BLE scan params

### LINKED ISSUE #3770: Scanning for bluetooth-devices works only once
I have quite a complex application which is low on memory. From time to time this application scans for surrounding bluetooth devices. As the whole BLEScan-stuff consumes plenty of RAM, after such a scan everything is deinitialised to save the related memory:

```
   BLEDevice::init("");
   pBLEScan = BLEDevice::getScan(); //create new scan
   if (pBLEScan)
   {
      pBLEScan->setAdvertisedDeviceCallbacks(new MyAdvertisedDeviceCallbacks());
      pBLEScan->setActiveScan(true); //set passive scan; active scan uses more power, but get results faster
      pBLEScan->setInterval(100);
      pBLEScan->setWindow(99);  // less or equal setInterval value
      BLEScanResults foundDevices = pBLEScan->start(5,false);
      pBLEScan->stop();
      pBLEScan->clearResults();   // delete results fromBLEScan buffer to release memory
   }
   BLEDevice::deinit(true);
```
My problem: this works well for the very first call. On the second and all following calls, no more bluetooth-devices are found, means BLEDevice::deinit() or BLEDevice::init() seems to be incomplete and probably leaves some variables back in a state, which avoids further successful scanning.

So my question: how can I reinitialise the whole BLE-stuff correctly in order to successfully scan the devices more than once?

Thanks!

buxtronix:
This value is uninitialised and as such can be a random (and invalid) value. It's needs to be set per the espressif documentation here:

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/bluetooth/esp_gap_ble.html#_CPPv4N21esp_ble_scan_params_t14scan_duplicateE

This PR sets it to DUPLICATE_DISABLE. Chosen as this is needed to ensure all scan data is populated in the scan callback, per this comment in the IDF:

https://github.com/espressif/esp-idf/blob/master/components/bt/host/bluedroid/stack/btm/btm_ble_gap.c#L3591

"//if scan duplicate is enabled, the adv packet without scan response is allowed to report to higher layer"

We **don't** want it to report to the higher layer (ie BLEScan.cpp) **unless** it has the active scan response.

Seems to resolve #3770 #3677 and possibly others.

### DIFF
--- a/libraries/BLE/src/BLEScan.cpp
+++ b/libraries/BLE/src/BLEScan.cpp
@@ -25,6 +25,7 @@ BLEScan::BLEScan() {
 	m_scan_params.scan_type          = BLE_SCAN_TYPE_PASSIVE; // Default is a passive scan.
 	m_scan_params.own_addr_type      = BLE_ADDR_TYPE_PUBLIC;
 	m_scan_params.scan_filter_policy = BLE_SCAN_FILTER_ALLOW_ALL;
+  m_scan_params.scan_duplicate     = BLE_SCAN_DUPLICATE_DISABLE;
 	m_pAdvertisedDeviceCallbacks     = nullptr;
 	m_stopped                        = true;
 	m_wantDuplicates                 = false;
--- a/libraries/BLE/src/BLEScan.cpp
+++ b/libraries/BLE/src/BLEScan.cpp
@@ -25,7 +25,7 @@ BLEScan::BLEScan() {
 	m_scan_params.scan_type          = BLE_SCAN_TYPE_PASSIVE; // Default is a passive scan.
 	m_scan_params.own_addr_type      = BLE_ADDR_TYPE_PUBLIC;
 	m_scan_params.scan_filter_policy = BLE_SCAN_FILTER_ALLOW_ALL;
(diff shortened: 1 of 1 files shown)


chegewara:
Strange, i never needed that to get duplicates in v1.0.4, but maybe in master it is set to true in idf.

h2zero:
From my testing in NimBLE this will you get a lot of spam packets (and callbacks) if disabled (0x0), better make sure `wantDuplicates` in `BLEScan::setAdvertisedDeviceCallbacks` is false. 

Although it may work different in bluedroid, but I doubt it, the same command is sent to the controller.

You should consider adding a method to BLEScan in this PR to allow the app to control this.
Here is what I have in the NimBLE library:
```
void NimBLEScan::setDuplicateFilter(bool active) {
    m_scan_params.filter_duplicates = active;
} // setDuplicateFilter
```

buxtronix:
setAdvertisedDeviceCallbacks already contains a parameter wantDuplicates, which is whether to filter dupes in BLEScan.

PR #4131, opened 2020-07-02
Make UDP task priority configurable.

Ivorforce:
UDP task priority ought to be configurable, especially with TCP task priority defaulting to the same value. A user may want to take a choice to priorize one over the other, or to have more room for his own vtask priorities below UDP beyond 0-2.


thbl:
@me-no-dev please merge this to master :D

me-no-dev:
this is not enough :) it also needs to be added here: https://github.com/espressif/arduino-esp32/blob/master/Kconfig.projbuild#L84

Ivorforce:
Done

PR #4133, opened 2020-07-03
Fix parameter to BLEDevice::updatePeerDevice

buxtronix:
::addPeerDevice and ::removePeerDevice are called with m_appId, so should ::updatePeerDevice as all use the same parameter for the underlying map's key.


chegewara:
Hi,
thanks for pointing it. Actually the idea was to update peer device map to use gatt if after connecting, which is not available when peer device is created, so it was bug in remove peer device. But now i see your fix may be better, not sure now.

Thanks

chegewara (approved):

PR #4143, opened 2020-07-07
Rename eep -> partitions.bin & hex -> bin

mjsir911:
This is in line with the [arduino platform specification](https://arduino.github.io/arduino-cli/platform-specification/#recipes-for-extraction-of-executable-files-and-other-binary-data).

specifies that the file extension after `recipe.objcopy` is arbitrary. The AVR platform uses `eep` & `hex`, while the esp32 platform seems to have file extensions `partition.bin` & `bin`.

Note: I ran into this using the [arduino platform specification to generate makefiles](https://gitlab.com/mjsir911/makeduino), this PR lets other software know about the dependency tree better.

Tested on Arduino 1.8.7

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -83,14 +83,14 @@ recipe.ar.pattern="{compiler.path}{compiler.ar.cmd}" {compiler.ar.flags} {compil
 ## Combine gc-sections, archives, and objects
 recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} {compiler.c.elf.extra_flags} -Wl,--start-group {object_files} "{archive_file_path}" {compiler.c.elf.libs} -Wl,--end-group -Wl,-EL -o "{build.path}/{build.project_name}.elf"
 
-## Create eeprom
-recipe.objcopy.eep.pattern={tools.gen_esp32part.cmd} -q "{build.path}/partitions.csv" "{build.path}/{build.project_name}.partitions.bin"
+## Create partitions.bin
+recipe.objcopy.partitions.bin.pattern={tools.gen_esp32part.cmd} -q "{build.path}/partitions.csv" "{build.path}/{build.project_name}.partitions.bin"
 
-## Create hex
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks @mjsir911 :) Any chance you rename `partitions.bin` to just `partitions`?

mjsir911:
@me-no-dev done

me-no-dev:
I wanted only `recipe.objcopy.partitions.bin.pattern` -> `recipe.objcopy.partitions.pattern` :)

mjsir911:
> I wanted only `recipe.objcopy.partitions.bin.pattern` -> `recipe.objcopy.partitions.pattern` :)

This defeats the purpose of this PR, the file extension of the generated file is `partition.bin`. If the file extension portion of the recipe is incorrect then tools cannot infer the filename of the generated file.

me-no-dev:
> tools cannot infer the filename of the generated file
>

I'm not sure I understand what you mean? what does the recipe name have to do with anything?

p.s. you do need to revert your last commit

mjsir911:
> > tools cannot infer the filename of the generated file
> 
> I'm not sure I understand what you mean? what does the recipe name have to do with anything?

As specified in the [arduino platform specification](https://arduino.github.io/arduino-cli/platform-specification/#recipes-for-extraction-of-executable-files-and-other-binary-data):

> `recipe.objcopy.FILE_EXTENSION_1.pattern=[.....]`
> FILE_EXTENSION_x **must** be replaced with the extension of the extracted file

which is saying that the file extension between `objcopy.` & `.pattern` needs to be the file extension of the resulting file. This is useful for tools (like [mine](https://gitlab.com/mjsir911/makeduino)) to better infer the build steps needed & the intermediate files.

It looks like [the Arduino app](https://github.com/arduino/Arduino/blob/master/arduino-core/src/cc/arduino/Compiler.java#L393) [allows for](https://github.com/arduino/arduino-cli/blob/8252a8e3e3fc489f4df958344fa5f0b70f5d1333/legacy/builder/recipe_runner.go#L60) file extensions with `.`s in it

Issue #1327, opened 2018-04-17
Include md5.h throws compile error #include expects "FILENAME" or <FILENAME>


hreintke:
### Hardware:
Board:							Nano32
Core Installation/update date:		Latest git
IDE name:						Sloeber Arduino plugin V4.2

### Description:

When compiling I get an error on md5.h when compiling WebAuthentication.cpp
Error is: md5.h:29:10: error: #include expects "FILENAME" or \<FILENAME\> ->  #include MBEDTLS_CONFIG_FILE
The compile command contains : -DMBEDTLS_CONFIG_FILE="mbedtls/esp_config.h"

Updating within md5.h to #include "mbedtls/esp_config.h" then it works.

Google revealed that it is a combination of the compiler command and the include statement.

### Sketch:
```cpp

#include "esp2.h"
#include "ESPAsyncWebServer.h"


//The setup function is called once at startup of the sketch
void setup()
{
// Add your initialization code here
}

// The loop function is called in an endless loop
void loop()
{
//Add your repeated code here
}
```

william-ferguson-au:
The mbedtls files need better handling of the MBEDTLS_CONFIG_FILE define.

In Windows, ```#include MBEDTLS_CONFIG_FILE``` will fail with 
```Error is: md5.h:29:10: error: #include expects "FILENAME" or <FILENAME> -> #include MBEDTLS_CONFIG_FILE```


One solution is to modify each of the mbdetls files with:

```
#if !defined(MBEDTLS_CONFIG_FILE)
#include "config.h"
#else
// From https://github.com/Sloeber/arduino-eclipse-plugin/issues/1028
#define VAL(str) #str
#define TOSTRING(str) VAL(str)
#include TOSTRING(MBEDTLS_CONFIG_FILE)
#endif
```

But I'm sure a more refined solution exists.

See https://github.com/Sloeber/arduino-eclipse-plugin/issues/1028

mikemoretti3:
Even with the latest set of files that seem to all contain the above fix, it still happens.  I had to edit the Environment and change the A.COMPILER.CPREPROCESSOR.FLAGS where it says -DMBEDTLS_CONFIG_FILE\="mbedtls/esp_config.h" to -DMBEDTLS_CONFIG_FILE\="""mbedtls/esp_config.h"""   note the triple quotes.

smuellener:
Is this issue still open or resolved by now? esp 1.0.3 seems to have this bug for Windows still

[...]

PR #4159, opened 2020-07-12
Parse the command line import of mbedtls/esp_config.h based on source

madhephaestus:
The arduino version of the define string looks like this
```
"-DMBEDTLS_CONFIG_FILE=\"mbedtls/esp_config.h\""
```
but the eclipse version looks like this:
```
-DMBEDTLS_CONFIG_FILE="mbedtls/esp_config.h"
```


lbernstone:
This file comes from [upstream](https://github.com/espressif/esp-idf/tree/master/components/mbedtls).  It cannot be changed here.

atanisoft:
This could also be considered a bug in how eclipse is generating the define content.

madhephaestus:
This patch resolves the error when compiling the md5 stack in Sloeber.

```
Arduino/hardware/espressif/esp32/tools/sdk/include/mbedtls/mbedtls/md5.h:42:10: error: #include expects "FILENAME" or <FILENAME>
 #include MBEDTLS_CONFIG_FILE
```
While i understand this can not be changed here, it would be helpful for folks who get this error to have a known work around. This PR stands as documentation of a workaround then.

lbernstone:
Either close it or move it to issues, since it will never be merged.

madhephaestus:
This PR works around #1327

PR #4197, opened 2020-07-23
Change check for BT_ENABLE to really be a check for BLUEDROID_ENABLED

geeksville:
Hi,

My project has migrated away from the bluedroid based Bluetooth API in this project, but I'd still like to be able to use the rest of the arduino api.  It turns out, the change needed to make that happen is quite minor (though it does need to be applied many places).  The existing code was checking for the CONFIG_BT_ENABLED flag, when really it would be better to check for CONFIG_BLUEDROID_ENABLED. Because the BLE classes in here are Bluedroid specific.

With this change, projects that have turned off BLUEDROID_ENABLED (but left on BT_ENABLED) will build correctly.  Since existing projects (including the prebuilt esp32-arduino-lib-builder) binaries define BLUEDROID_ENABLED any time BT_ENABLED is set - it shouldn't break existing usage.

Btw - I'm now using Nimble as my BLE stack and it works great (and is super small).  It is also a nicer (and open source) API than bluedroid, so it doesn't need the same amount of helpful wrapper classes.


me-no-dev:
could you please fix the conflicts with WiFiProv?

h2zero:
Would it be possible to change [this line](https://github.com/espressif/arduino-esp32/blob/c2b3f2d6afc1911db1974a324782730dd642d524/cores/esp32/esp32-hal-misc.c#L220) also? 

It likes to release the memory here if one of the `btxxxxxx()` functions isn't called and also creates compilation complications when using NimBLE and Ardruino as an IDF component with cmake.

geeksville:
@h2zero sure - I can do that at the same time.  I think that code was added after I submitted this PR.

h2zero:
@geeksville Perfect, thanks!

geeksville:
@me-no-dev Ok - merged now.

@h2zero I just looked more carefully at your comment.  I'm not sure we want to conditionalize that line with something different.  In my usage of Nimble I just define btInUse() (overriding the weak reference in this library) to return true.  Which seemed to be a clean way of doing things (because bt is in use - just not using the classic BLE stack).

h2zero:
@geeksville The issue I have with that line is with my library that refactored the original to used NimBLE instead. In this scenario it forces me to require Arduino as a dependency to be able to override that function, which doesn't succeed for some reason when using IDF 4.0+.

Besides that I don't see how the change would do any harm. @me-no-dev might see something I don't though. Either way I'll deal with it.

me-no-dev:
example is failing :) please fix

```
/home/runner/Arduino/hardware/espressif/esp32/libraries/WiFi/examples/WiFiProv/WiFiProv.ino: In function 'void setup()':
/home/runner/Arduino/hardware/espressif/esp32/libraries/WiFi/examples/WiFiProv/WiFiProv.ino:59:38: error: 'WIFI_PROV_SCHEME_BLE_EVENT_HANDLER_FREE_BTDM' was not declared in this scope
   WiFi.beginProvision(provSchemeBLE, WIFI_PROV_SCHEME_BLE_EVENT_HANDLER_FREE_BTDM, WIFI_PROV_SECURITY_1, "abcd1234", "PROV_XXX", NULL, NULL);
                                      ^
```

me-no-dev:
fixed and merged into 2.0.0 release branch

PR #4209, opened 2020-07-27
Enable precompiled libraries

cpq:
The latest versions of Arduino IDE shifted the responsibility for precompiled libraries support to the core developers, which breaks precompiled library support in esp32 Arduino core. See https://github.com/arduino/ArduinoCore-avr/pull/52 for more details:

```
In this new version of the builder we are not doing any heuristics to find the right spot where the ldflags should be inserted (this was causing many bugs on its own); instead, we fully trust the core makers to add explicit support to precompiled libs.
```

This chage re-enables precompiled library support in the esp32 Arduino core.


tanakamasayuki:
Hi.
I think it's a different place to put the options.

https://github.com/arduino/ArduinoCore-avr/pull/52

recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} {compiler.c.elf.extra_flags} -Wl,--start-group {object_files} "{archive_file_path}" {compiler.c.elf.libs} **{compiler.libraries.ldflags}** -Wl,--end-group -Wl,-EL -o "{build.path}/{build.project_name}.elf"


but, This PR

recipe.c.combine.pattern="{compiler.path}{compiler.c.elf.cmd}" {compiler.c.elf.flags} {compiler.c.elf.extra_flags} **{compiler.libraries.ldflags}** -Wl,--start-group {object_files} "{archive_file_path}" {compiler.c.elf.libs} -Wl,--end-group -Wl,-EL -o "{build.path}/{build.project_name}.elf"

I got an error and could not compile it in my environment.

gizero:
Hi! @tanakamasayuki's suggestion seems correct. Neither arduino-esp32 1.0.5 nor 1.0.6 releases solved my original issue with using [BSEC-Arduino-library](https://github.com/BoschSensortec/BSEC-Arduino-library). I still need to "override" the combine recipe as suggested [here](https://github.com/BoschSensortec/BSEC-Arduino-library#3-modify-the-platformtxt-file) to make the linker happy.

Having `{compiler.libraries.ldflags}` listed before the list of the object files (`object_files `) and outside the linker's archive list sounds weird to me, but I may be wrong. @cpq can you please help understand why with this PR you choose to leave `{compiler.libraries.ldflags}` outside the linker's archive list (delimited by the `--start-group` and `--end-group` linker options)?

gizero:
After a better look at the repo history, I noticed this PR intentionally moved `{compiler.libraries.ldflags}` outside the archive list after https://github.com/espressif/arduino-esp32/pull/3783 (merged just the day before) had it where @tanakamasayuki and my tests would suggest... I'm becoming even more confused...

PR #4212, opened 2020-07-29
Fix ttgo twatch & tbeam board definition error

lewisxhe:
This is the wrong definition submitted before, which caused the compiler to issue a warning. Please pass this request


boards.txt
```diff
@@ -3544,8 +3544,8 @@ t-beam.serial.disableRTS=true
 
 t-beam.build.mcu=esp32
 t-beam.build.core=esp32
-t-beam.build.variant=t-beam
-t-beam.build.board=T-Beam
+t-beam.build.variant=tbeam
```
matthijskooijman, line 3547: Doesn't this break building for this board? AFAIK `build.variant` is used as a directory name under the `variants` directory, but there is only `variants/t-beam`, not `variants/tbeam`. So IIUC, this change must either be reverted, or the variant directory should be renamed?

Issue #3374, opened 2019-10-17
HttpClient follow redirects


Clickau:
Would it be possible to add an option to automaticallly follow http redirects, like there is for the esp8266? ( https://github.com/esp8266/Arduino/commit/e46ccae9d5294753f3b31d0e88d39e98fb9ca558 )

wga22:
Bump

PR #4240, opened 2020-08-10
Add support for following redirects in HTTPClient

Clickau:
Ported from ESP8266 . Fixes issue #3374 
The user can use HTTPClient::setFollowRedirects with HTTPC_DISABLE_FOLLOW_REDIRECTS, HTTPC_STRICT_FOLLOW_REDIRECTS or HTTPC_FORCE_FOLLOW_REDIRECTS to enable and disable this feature. Responds to codes 301, 302, 303, 307.

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -548,29 +548,106 @@ int HTTPClient::sendRequest(const char * type, String payload)
  */
 int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
 {
-    // connect to server
-    if(!connect()) {
-        return returnError(HTTPC_ERROR_CONNECTION_REFUSED);
-    }
-
-    if(payload && size > 0) {
-        addHeader(F("Content-Length"), String(size));
-    }
-
-    // send Header
-    if(!sendHeader(type)) {
-        return returnError(HTTPC_ERROR_SEND_HEADER_FAILED);
-    }
-
-    // send Payload if needed
-    if(payload && size > 0) {
-        if(_client->write(&payload[0], size) != size) {
-            return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+    int code;
+    bool redirect = false;
+    uint16_t redirectCount = 0;
+    do {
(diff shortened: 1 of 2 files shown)


simonjamain:
@Clickau 
I don't know if this is the right place for this, but the new (much wanted) feature doesn't quite work for me. I use it for getting an asset link from the github API (https://github.com/simonjamain/openmosfet/releases/download/0.1.0-alpha/openmosfet.esp32.bin for example) which give a 302 redirect to a much larger url (540 characters).

By setting `setFollowRedirects(HTTPC_DISABLE_FOLLOW_REDIRECTS)` I got the 302 code but with `HTTPC_STRICT_FOLLOW_REDIRECTS` or `HTTPC_FORCE_FOLLOW_REDIRECTS` i get a 404.

You can test this issue by compiling and running my (platformio) project on this specific commit : https://github.com/simonjamain/openmosfet/tree/f326aa19b7b86527de9099b5a1cf89c7045eb2a6

The code starting from line 58 of [autoUpdater.cpp](https://github.com/simonjamain/openmosfet/blob/f326aa19b7b86527de9099b5a1cf89c7045eb2a6/src/autoUpdater.cpp) will be what you're interested in.

The sequence is started by calling `OMAutoUpdater::requestUpdate()` which you can do from your network by calling a **FETCH** request on `http://openmosfet.local/api/core` or clicking the "Mise à jour" button on the interface at `http://openmosfet.local` from your browser.
 
Hope this is of interest for you, and I didn't do something wrong and wasted you time.

Thank for the much appreciated work (which is way out of my league)!

thegreatdane6:
I ran into the same issue as @simonjamain and after some investigation I found the issue. In case of a redirection to an url with a new host, the host does not get replaced. So if one tries to download the latest release off of github, at some point the host swaps from github.com to some amazon storages. In the logs i can see, the location header returns the correct url with the host pointing to the aws storage. From the code linked in the PR i cannot see tough why this happens, the issue must be somewhere within the "beginInternal" function. For now i just implemented a the rediction myself and it works just fine. This should only be temporary solution tough.

lbernstone:
This is not the place to report issues.  Please open an issue if you are not planning on fixing the problem.

PR #4246, opened 2020-08-11
shallow clone to make installation faster

dimitre:
shallow clone (board and submodules) to make installation faster


lbernstone:
There will be no submodules in the next version, and it is unlikely there will be any in the future.

Issue #4257, opened 2020-08-14
 MD5 Verification fails on HTTPUpdate after unsuccessful OTAUpdate


CriPstian:
### Hardware:
Board:							esp32doit-devkit-v1
Core Installation version:			1.0.4
IDE name:						CLion
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Ubuntu

### Description:
MD5 Verification fails on HTTP Update (with no x-MD5 header response) after unsuccessful OTA Update because the _target_md5 string in Updater class is not restored to default value.

### Sketch:
No need for a sketch, This is a reproducible issue by following two basic steps:

1. Fail an OTAUpdate by cancelling the update in the middle of the process
2. Trigger an HTTPUpdate with a server that has no x-MD5 header set on the response.
    => The update will fail with MD5 Check fail

### Debug Messages:
```
MD5 Failed: expected:<_target_md5>, calculated:<md5>
```

CriPstian:
I've created a small PR #4260  for cleaning up the MD5 related fields since every use of the Updater class would leave garbage values at the site

CriPstian:
This issue continues to create issues on my side. Is anybody up for review?

CriPstian:
This issue was fixed by #4260

PR #4260, opened 2020-08-15
Reinit updater md5 related fields for cleanup

CriPstian:
MD5 cleanup on begin
Typos


lbernstone:
Can this not be part of _reset() so they are cleared on any abnormal result?

CriPstian:
I thought about it, but maybe it's better to have a clean start than cleaning up at the end.
There may be a case where the value is needed after a failure for a print maybe

CriPstian:
Fixes #4257

PR #4263, opened 2020-08-17
fix hwSerial tx only flush

bertmelis:
standard HardwareSerial `flush` should only flush TX.


bertmelis:
Argh... Given the change, CI failing is probably not due to this PR.

PR #4289, opened 2020-08-24
Change define "ttgo-t-watch.build.board=T-Watch" to "ttgo-t-watch.build.board=T_Watch" prevent "warning: ISO C99 requires whitespace after the macro name"

sweetlilmre:
The "-" in the board name causes a compiler warning when this is expanded to -DARDUINO_BOARD="T-Watch".
This pull request fixes those warnings by replacing the "-" with "_" and aligns with other board definitions.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -4100,7 +4100,7 @@ ttgo-t-watch.serial.disableRTS=true
 ttgo-t-watch.build.mcu=esp32
 ttgo-t-watch.build.core=esp32
 ttgo-t-watch.build.variant=twatch
-ttgo-t-watch.build.board=T-Watch
+ttgo-t-watch.build.board=T_Watch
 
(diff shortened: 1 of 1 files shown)


datacute:
@me-no-dev 
Is it possible that the platformio builds are failing because platformio also needs to change the board definition, but platformio are waiting till this commit is merged? https://github.com/platformio/platform-espressif32/pull/378

The build.board turns into a compile time define in the recipes in platform.txt e.g. in recipe.c.o.pattern
-DARDUINO_{build.board}
https://github.com/espressif/arduino-esp32/blob/master/platform.txt#L72

Defines are not allowed to have hyphens in their name.

T-Watch isn't the only board with an invalid build.board value

me-no-dev:
already fixed in another commit

PR #4296, opened 2020-08-25
Added board - microS2

microDev1:
Added esp32s2 based microS2 dev board.


atanisoft:
@microDev1 who produces this board and why are they using the VID from @UnexpectedMaker?

microDev1:
> @microDev1 who produces this board and why are they using the VID from @UnexpectedMaker?

Hi @atanisoft, the microS2 is my creation. This VID belongs to Adafruit, checkout [this](https://github.com/adafruit/circuitpython/issues/3319) conversation. :)

PR #4298, opened 2020-08-26
Add ArduinoOTE = ArduinoOTA over Ethernet

Miq1:
I needed the ability to do OTA via Ethernet, so I modified ArduinoOTA.cpp and ArduinoOTA.h - just a few lines to be swapped in when the ``-DARDUINOOTE`` compiler flag is set. The default is still using WiFi.


me-no-dev:
this is not a clean solution because it requires a define switch for the project (not possible in ArduinoIDE). My suggestion is to copy the library and remove mDNS as well.

I am closing this PR

Issue #3960, opened 2020-05-03
WiFiServer doesn't accept start/stop/start


etrinh:
Hi,

WiFiServer (/WebServer) doesn't allow to stop the service and start it again in a row.
The SO_REUSEADDR is forgotten in WiFiServer to allow to reuse a socket when it was just closed (else we have a long to death timeout...)

Issue seems to have been first noticed in #1267.

Adding:
  int enable = 1;
  setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(int));

Right after:
  sockfd = socket(AF_INET , SOCK_STREAM, 0);
  if (sockfd < 0)
    return;

Make the wifiserver (/webserver).begin() works just after close().
It is more important to make this work as the WiFiServer (/WebServer) API is damn poor and doesn't provide notification of failure (neither of success/running)... :((((

Thank you

everslick:
#1267

Issue #1267, opened 2018-03-27
Should we have SO_REUSEADDR on listening sockets?


everslick:
Closing a listening socket that has open connections leaves the socket in a TIME_WAIT state for approximately 2 minutes. This prevents the socket from being reused immediately. IMHO on an embedded device this is not very useful because we will most likely start the same server on the same port anyway.

Setting the SO_REUSEADDR on the listening socket would make the port available immediately again.

everslick:
Still valid.

everslick:
Keep

PR #4306, opened 2020-08-28
Allow faster reuse of socket, to be able to restart WifiServer.

ahorn42:
See #3960 for more details of the problem and the solution. I only implemented what was proposed in this ticket, as it solves my problem, which was the same as in this ticket. Credits for the code going to @etrinh ;-)

This also is a more consistence behaviour compared to esp8266, where it also is possible to restart the wifiserver immediately on the same port.


me-no-dev:
this should be optional, though I do not mind it being default

ahorn42:
@me-no-dev what would be your suggestion to make it configurable? Would an additional overload for `begin` with a flag be an idea?  

like:
```c
void WiFiServer::begin(uint16_t port, bool socketReuse)
```

everslick:
#1267

everslick:
> ```void WiFiServer::begin(uint16_t port, bool socketReuse)```

how about using the same method, but with default parameter:

```void WiFiServer::begin(uint16_t port, bool socketReuse = true)```

me-no-dev:
i agree with @everslick

ahorn42:
@me-no-dev thanks for implementing/merging it! - you were faster than me, wanted to do this on the weekend ;D

marioelectron:
setTimeout in ESP32 not work

PR #4322, opened 2020-09-03
Update AzureIoT submodule to include the SNTP fix

photomoose:
This PR brings the AzureIoT library submodule up to date, so that the "Operating mode must not be set while SNTP client is running" SNTP bug fix is included in the Arduino ESP32 core.

See https://github.com/VSChina/ESP32_AzureIoT_Arduino/issues/5 for a description of the issue (occurs on my Adafruit Feather ESP32 when a timeout occurs when sending messages to Azure IOT Hub).

The fix for the above issue was merged into master (see https://github.com/VSChina/ESP32_AzureIoT_Arduino/pull/9).

Tested by explicitly fetching the platform package from my fork, as follows:

```
[env:featheresp32]
platform = espressif32
board = featheresp32
framework = arduino
platform_packages =
	framework-arduinoespressif32 @ https://github.com/photomoose/arduino-esp32.git
```


lbernstone:
It's ok to leave this for people looking for a fix, but AzureIoT will be removed from the next version, so the PR will never be merged.

photomoose:
@lbernstone What are the reasons for removing it? Also, once removed, is there an alternative?

lbernstone:
It is being removed as a submodule from this repository.  The project is not going away.  You will just need to download it separately like any other arduino library.

me-no-dev:
nothing wrong to merge this since it's still a part of the repo :) but will be removed as submodule because it's hard to sync fixes and updates.

PR #4342, opened 2020-09-15
add I2C_RX_FIFO_OVF_INT_ST handling to i2c_isr_handler_default

lienbacher:
I have run into a Problem where my ESP32 would crash with this error message: `[E][esp32-hal-i2c.c:1013] i2c_isr_handler_default(): unknown int=4`
It turns out `i2c_isr_handler_default()`does not handle `I2C_RXFIFO_OVF_INT_ST` at all which ultimately leads to a crash. Handling it the same way as a I2C_RXFIFO_FULL_INT_ST fixes the crashing and only leads to gracefully handled errors in the I2C transaction.

### DIFF
--- a/cores/esp32/esp32-hal-i2c.c
+++ b/cores/esp32/esp32-hal-i2c.c
@@ -945,6 +945,14 @@ static void IRAM_ATTR i2c_isr_handler_default(void* arg)
             activeInt &=~I2C_RXFIFO_FULL_INT_ST;
         }
 
+        if(activeInt & I2C_RXFIFO_OVF_INT_ST) {
+            emptyRxFifo(p_i2c);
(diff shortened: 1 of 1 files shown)


lienbacher:
> I have run into a Problem where my ESP32 would crash with this error message: `[E][esp32-hal-i2c.c:1013] i2c_isr_handler_default(): unknown int=4`
> It turns out `i2c_isr_handler_default()`does not handle `I2C_RXFIFO_OVF_INT_ST` at all which ultimately leads to a crash. Handling it the same way as a I2C_RXFIFO_FULL_INT_ST fixes the crashing and only leads to gracefully handled errors in the I2C transaction. 
I have been kindly made aware of this fix by u/azureice on reddit.

PR #4370, opened 2020-09-29
Create S.ODI_Ultra_v1.0/pins_arduino.h for new board s.odi Ultra v1

SPELECautomations:
S.ODI Ultra v1 @MySPELEC, a new dev board based on esp32 with multiple onboard sensors.

PR #4369, opened 2020-09-29
update board.txt added new dev board s.odi Ultra v1

SPELECautomations:
S.ODI Ultra v1 @MySPELEC, a new dev board based on esp32 with multiple sensors onboard.

PR #4372, opened 2020-09-29
Add support for S.ODI_Ultra_v1.0

SPELECautomations:
Add a new Dev Board S.ODI Ultra v1 in esp32 library


SPELECautomations:
only consider the last two commits

me-no-dev:
there are 3 PRs open. Please delete them all and create one clean PR instead.

atanisoft:
@me-no-dev I believe this is the only PR that should be merged and PR #4370 and PR #4369 should simply be rejected (they are both included in this PR).

PR #4384, opened 2020-10-02
Update boards.txt

bonnom:
I got sick of searching for my ESP32 board in the Arduino board list. It was a giant mess. I rearranged the list. I put the official devboards/modules first. The rest of the boards are all in alphabetical order.


bonnom:
Some errors were still in it.

PR #4385, opened 2020-10-03
Fix support for following redirects added by previous merge

trandi:
ee88c42c3b5e87a2fc09303f9e82dd8172e99990 added support for following redirects  (#4240)

However I couldn't make it work because of 2 issues in HTTPClient.cpp:

1) in the newly added setURL() we do "_canReuse = true" which means we preserve the client but kill the connection. The latter invalidates some of the memory which means that later _client.available() or similar blow up completely

2) getString() doesn't deal with when _size == -1  which can be a valid state (see existing comment re "Server not sending Content-Length header"). I've changed that condition to how it was in one of the previous versions and removed the optimisation around reserving memory upfront, as this seems to be done later on just fine..

It's my first time I contribute to this project so please bear with me if I got things wrong :+1: 
Dan

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -919,21 +919,19 @@ int HTTPClient::writeToStream(Stream * stream)
  */
 String HTTPClient::getString(void)
 {
-    StreamString sstring;
-
-    if(_size > 0) {
-        // try to reserve needed memmory
-        if(!sstring.reserve((_size + 1))) {
+    // _size can be -1 when Server sends no Content-Length header
+    if(_size > 0 || _size == -1) {
+        StreamString sstring;
+        // try to reserve needed memory (noop if _size == -1)
+        if(sstring.reserve((_size + 1))) {
+            writeToStream(&sstring);
+            return sstring;            
+        } else {
             log_d("not enough memory to reserve a string! need: %d", (_size + 1));
-            return "";
         }
     }
-    else {
-        return "";
-    }
 
-    writeToStream(&sstring);
-    return sstring;
+    return "";
 }
 
 /**
@@ -1441,8 +1439,10 @@ bool HTTPClient::setURL(const String& url)
         _port = (_protocol == "https" ? 443 : 80);
     }
 
-    // disconnect but preserve _client (clear _canReuse so disconnect will close the connection)
-    _canReuse = false;
+    // disconnect but preserve _client. 
+    // Also have to keep the connection otherwise it will free some of the memory used by _client 
+    // and will blow up later when trying to do _client->available() or similar
+    _canReuse = true;
     disconnect(true);
     return beginInternal(url, _protocol.c_str());
 }


libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -919,21 +919,14 @@ int HTTPClient::writeToStream(Stream * stream)
...
-        // try to reserve needed memmory
-        if(!sstring.reserve((_size + 1))) {
-            log_d("not enough memory to reserve a string! need: %d", (_size + 1));
-            return "";
-        }
-    }
-    else {
-        return "";
+    // _size can be -1 when Server sends no Content-Length header
+    if(_size != 0) {
```
atanisoft, line 923: if size can be -1 then this block may be unintentionally entered. I think you might want `if(_size > 0)`

trandi: But that's exactly my use case (which from existing comment I assume is valid and not just some obscure problem on my side)  : size is -1 because the server hasn't put the relevant header in the response, BUT there is valid data to be read from the server so I do want to enter this block and read it.

trandi: Maybe I should put `if(_size > 0 || _size == -1)` to be super explicit ??

atanisoft: I'm not sure what server you are connecting to but nearly all do send the Content-Length header. I think your explicit check should be better.

atanisoft: Another option is to remove the else clause and leave the if alone

trandi: Google scripts (I use this from my ESP32 to pull data from Google Drive, etc. ...):
https://script.googleusercontent.com/macros/echo?user_content_key=g_uWqKQm9SRnYzPUr8zlBKhb4xJe2aMNKC_WtktO-5JD9ibF04HScAPl-4KWqei6ad0sK5CgJ1y5GGi6kjQ-_VujQEJrTSj1m5_BxDlH2jW0nuo2oDemN9CCS2h10ox_1xSncGQajx_ryfhECjZEnEXX6cJoH7fZw5b8BrSbrnico0hlOeVt3nbQiyiUYOk816bc8RKBUVWSsOSSCSTGmrxy_u9Yn2NR&lib=MBFYKdoFPkY3yJEvlYtZAhIyQU_TzVcy6

Re the "else" clause that's exactly what I did, I removed it:

```
    // _size can be -1 when Server sends no Content-Length header
    if(_size != 0) {
        StreamString sstring;
        writeToStream(&sstring);
        return sstring;
    }

    return "";
```

atanisoft: I'd suggest revert the change and just remove the else block since the if block is used to reserve X bytes of memory before trying to read the page in. With your update it will do a number of re-allocations as the size of the stream increases which is not ideal when the size is known.

trandi: Ok, fair enough, updated as suggested, with both  the more explicit `if(_size > 0 || _size == -1)` condition and the advance reservation of memory optimization.

trandi:
@atanisoft  / @me-no-dev or others more knowledgeable, what is the process now in order to have this merged ?
Are there more checks / reviews to be done ?

PR #4387, opened 2020-10-04
Add type's definition

destroyedlolo:
Make easier file inclusion

### DIFF
--- a/cores/esp32/esp32-hal-psram.h
+++ b/cores/esp32/esp32-hal-psram.h
@@ -19,6 +19,10 @@
 extern "C" {
 #endif
 
+  /* Prototyping */
+#include <stddef.h>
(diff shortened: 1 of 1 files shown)


lbernstone:
Can you explain why?  These utilities are already included in Arduino.h.

destroyedlolo:
For example, if you're trying to compile the examples provided by TTGO smart watch, you need to include /esp32/core/esp32-hal-psram.h
But if you do so, you need to include before stddef and stdbool otherwise it will fail as types are undefined.
So, as a thumb of rule it would easier for user project to add all needs inside header which is harmless :)

me-no-dev:
what you need to do is `#include "Arduino.h"` and not the hal header itself.

PR #4392, opened 2020-10-05
Support for ethernet PHY IP101

HilverinkJ:
Hi,
The IP101 ethernet phy is supported in the ESP IDF. Testing the ESP32-Ethernet-Kit V1.1, I noticed that the IP101 is not officially supported in Arduino ESP32.
I'm hoping this is possible.

Joop.


me-no-dev:
This PR has a lot more changes than adding support for IP101. Also I do think that it already has support for it :)

Issue #4335, opened 2020-09-12
WiFiClientSecure does not free memory after a failed connection.


VoLinhTruc:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			1.0.3
IDE name:						Platform.io
Flash Frequency:					80Mhz
PSRAM enabled:                                      no
Upload Speed:					115200
Computer OS:                                         Windows 10

### Description:
WiFiClientSecure does not free memory after a failed connection.

### Update: In the ssl_client.cpp, in start_ssl_client function.
```
if (rootCABuff != NULL) {
        log_v("Loading CA cert");
        mbedtls_x509_crt_init(&ssl_client->ca_cert);
        mbedtls_ssl_conf_authmode(&ssl_client->ssl_conf, MBEDTLS_SSL_VERIFY_REQUIRED);
        ret = mbedtls_x509_crt_parse(&ssl_client->ca_cert, (const unsigned char *)rootCABuff, strlen(rootCABuff) + 1);
        mbedtls_ssl_conf_ca_chain(&ssl_client->ssl_conf, &ssl_client->ca_cert, NULL);
        //mbedtls_ssl_conf_verify(&ssl_client->ssl_ctx, my_verify, NULL );
        if (ret < 0) {
            return handle_error(ret);
        }
    }
```

You can see that after ` mbedtls_x509_crt_init(&ssl_client->ca_cert);`, 
 `if (ret < 0)` has no `mbedtls_x509_crt_free(&ssl_client->ca_cert);` to free memory.
So the free memory in RAM decreasing continuously.

The same thing (I guess) may occur with the 
```
ret = mbedtls_x509_crt_parse(&ssl_client->client_cert, (const unsigned char *)cli_cert, strlen(cli_cert) + 1);
if (ret < 0) {
       return handle_error(ret);
}
```

and the 
```
ret = mbedtls_pk_parse_key(&ssl_client->client_key, (const unsigned char *)cli_key, strlen(cli_key) + 1, NULL, 0);

if (ret != 0) {
     return handle_error(ret);
}
```

So please update the code. My project is stuck here. I need your help.
Thank you so much.

### Sketch: 

```
void loop() {
  // put your main code here, to run repeatedly:

  Serial.println(esp_get_free_heap_size());

  WiFiClientSecure client;
  client.setCACert(test_root_ca);
  client.setCertificate(client_crt);
  client.setPrivateKey(client_key);

[...]

PR #4412, opened 2020-10-13
Certificate isn't be free in case parse failure.

VoLinhTruc:
I met problem while I was working with the WiFiClientSecure.
I tried to found the source of the problem, and I found it in the sll_client.cpp.
Please check my contribution.

I've open this problem in #4335 but received no response.


me-no-dev:
thanks for figuring it out :)

Issue #3460, opened 2019-11-07
WifiClientSecure hangs on write - socket error


Onigiri-Kun:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			ESP32 Arduino 1.0.3 (IDF 3.2) Release
IDE name:						Visual Studio with Visual Micro add on
Flash Frequency:					40Mhz - tried a few and problem persists.
PSRAM enabled:                                     no
Upload Speed:					115200
Computer OS:                                        Windows 10

### Description:
I have an ESP32 publishing code to a Mosquitto MQTT server, using the PubSubClient library.
This works fine, and publishes many short messages (around 50 float values, to one per topic) at a 10 seconds interval (50 mqtt write every 10 seconds).
I had problems with the board hanging occasionally, and managed to isolate the problem to the WebClientSecure library.

The log files of the server would give "Socket Error on the client".
The ESP32 debug log would give "Writing HTTP request", before hanging.

The point where the code hangs is the following:

```
int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, uint16_t len)
{
    log_v("Writing HTTP request...");  //for low level debug
    int ret = -1;

    while ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0) {
        if (ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE) {
            return handle_error(ret);
        }
    }

    len = ret;
    //log_v("%d bytes written", len);  //for low level debug
    return ret;
}
```

The sketch hangs in the while loop, because mbedtls_ssl_write does not return an error code.
I attribute this to some occasional write failure/socket failure, which is not handled properly.

I've tried adding some timeout code here, which does catch the problem. However, I am not able to reset the code correctly and restart successfully transmission of data without a board reset.

My question is: Can you have code to handle a timeout of the mbedtls_ssl_write?
In such a way that, if some socket error occurs, the client can restart transmission quickly?

### Debug Messages:
```
"Writing HTTP request.."
It then hangs
```

lbernstone:
Are you checking to make sure the connection is active before you send?

Onigiri-Kun:
Sorry for the late reply.
Yes, I do. I have:

```	
if (WiFi.status() != WL_CONNECTED) {
	Serial.println(F("trying to publish, but wifi disconnected"));
	return;
}
if (mqttclient.connected() == false) {
[...]

PR #4424, opened 2020-10-18
Fix for espressif#3460 issue

igolubic:
Fixes: #3460

This code has been run in production for 1 month and it looks stable, no data dropped and it definitely fixes the issue described. I think that this can be merged to avoid using custom package referencing in PlatformIO that has been used in quite a few projects for now.


Bmooij:
It seems logic to remove the while statement to prevent infinite looping.
I will test this solution the coming days

bhupiister:
@igolubic  If found this as more suitable, as it may take multiple attempts before it gets successful.. i have locked trials to 100

```
int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, size_t len)
{
    log_v("Writing HTTP request with %d bytes...", len); // for low level debug
    int ret = -1;
    int retry = 0;
    while ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0)
    {
        if ((ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE && ret < 0) || retry >= 100)
        {
            log_v("Trial #  %d", retry);     // for low level debug // Show number of attempts taken 
            log_v("Handling error %d", ret); // for low level debug
            return handle_error(ret);
        }
        retry++;
        // wait for space to become available
        vTaskDelay(2);
    }
    log_v("Trial #  %d", retry); // for low level debug // Show number of attempts taken 
    return ret;
}
```

PR #4441, opened 2020-10-24
Added isKey method to Preferences

### LINKED ISSUE #4440: Feature Request. In Preferences library expose ability to check if a key exists already
Feature Request only.

A common requirement for ESP32 projects is to have default values, for instance for settings, that can be overridden by the user.
The preferences library provides a great way to conveniently store the settings.
However, on boot, via the Preferences library, there seems to be no way to test if the user has already configured a value of their own, and therefore whether to adopt the hardcoded default instead.
I feel it would be helpful to expose an additional method for preferences that allows the program to check if a key has been set, so the value can then be set to the default, whereas if it has been set that old saved value would be appropriate.
At the moment the get methods do allow a value to be provided as a default, which is returned. But this value is not itself saved, so the program doesn't know if the value is saved (but happens to be the default) or the default has been returned.

lbernstone:
Checks to see if a string is a key in the namespace. Fixes #4440


libraries/Preferences/src/Preferences.cpp
```diff
@@ -280,6 +280,26 @@ size_t Preferences::putBytes(const char* key, const void* value, size_t len){
...
 }
 
+bool Preferences::isKey(const char* key) {
+    if(!_started || !key){
+        return false;
+    }
+    int8_t mt1; uint8_t mt2; int16_t mt3; uint16_t mt4;
+    int32_t mt5; uint32_t mt6; int64_t mt7; uint64_t mt8;
+    size_t len = 0;
+    if(!nvs_get_i8(_handle, key, &mt1)) return true;
```
atanisoft, line 290: these checks would be better written as:
```
if(nvs_get_i8(_handle, key, &mt1) == ESP_OK) return true;
...
```
Using an explicit check for ESP_OK.

me-no-dev:
you know what I'm thinking? :D Since you have that whole if tree, maybe change that function to return item type (enum) and then implement isKey again, by checking if the returned type is valid.

What do you think?

lbernstone:
Can you give me an example of what you want the elements to look like?

me-no-dev:
your branch is writable, I'll just do it :)

me-no-dev:
done :)

PR #4443, opened 2020-10-24
Add partition label argument to SPIFFS

thewavelength:
SPIFFS currently assumes there is only ever one partition.

This change allows a user to pass the label argument (defaults to NULL)
to SPIFFS::begin() so a specific SPIFFS partition can be referenced.

This change does not break compatibility.


me-no-dev:
why not use `char * _partitionLabel;` for class member instead? You can default it to NULL and `strdup` it if it's given. That would save the whole check for lengths and so on.

thewavelength:
Would be fine with me. I'm not that experienced with Arduino Core internals and saw it uses `String` just about everywhere, so I did it as well. I'll change it.

thewavelength:
Done :)

libraries/SPIFFS/src/SPIFFS.cpp
```diff
@@ -43,23 +43,29 @@ bool SPIFFSImpl::exists(const char* path)
...
 
-SPIFFSFS::SPIFFSFS() : FS(FSImplPtr(new SPIFFSImpl()))
+SPIFFSFS::SPIFFSFS() : FS(FSImplPtr(new SPIFFSImpl())), partitionLabel_(NULL)
 {
 
 }
 
+SPIFFSFS::~SPIFFSFS()
+{
+    delete partitionLabel_;
```
atanisoft, line 53: add a not-null check for this before delete as it may throw an exception otherwise

thewavelength: The standard says it is safe to call `delete` on `NULL`. Is this different on micro controllers?

atanisoft: It's more of a general check and ensures that the underlying memory manager doesn't throw for a null being passed in (though you are right, it shouldn't!)

thewavelength: Thanks for your hint :) Done!

thewavelength: I replaced `delete` with `free` according to [https://stackoverflow.com/a/39935254/1433564](this StackOverflow answer). I hope this is correct.

libraries/SPIFFS/src/SPIFFS.cpp
```diff
@@ -43,21 +43,31 @@ bool SPIFFSImpl::exists(const char* path)
...
-    if(esp_spiffs_mounted(NULL)){
+    if (partitionLabel_){
+        free(partitionLabel_);
+        partitionLabel_ = NULL;
+    }
+}
+
+bool SPIFFSFS::begin(bool formatOnFail, const char * basePath, uint8_t maxOpenFiles, const char * partitionLabel)
+{
+    partitionLabel_ = partitionLabel ? strdup(partitionLabel) : NULL;
```
me-no-dev, line 61: it's maybe possible to call `begin` more than once. In that case make sure that if `partitionLabel_` is not NULL, to free it first :)

thewavelength: Done.

Issue #4454, opened 2020-10-28
Http Client does not accept "ICY 200 OK" as valid header


FrankBoesing:
We can not use the Http client for some icecast servers (audio-streaming), because the response ICY 200 OK is not considered as valid response.
These servers behave as HTTP servers, so everything else is identical.

Is there a workaround, or, better, can you fix this?

`15:36:08.264 -> [V][HTTPClient.cpp:235] beginInternal(): url: http://194.183.138.206:40250/
15:36:08.264 -> [D][HTTPClient.cpp:276] beginInternal(): host: 194.183.138.206 port: 40250 url: /
15:36:08.304 -> [D][HTTPClient.cpp:1025] connect():  connected to 194.183.138.206:40250
15:36:08.345 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'ICY 200 OK'
15:36:08.345 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-notice1:<BR>This stream requires <a href="http://www.winamp.com/">Winamp</a><BR>'
15:36:08.345 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-notice2:SHOUTcast Distributed Network Audio Server/win32 v1.9.8<BR>'
15:36:08.382 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-name:Radio Especially - Das Radio aus dem Westen ⸮sterreichs - Austria / ⸮sterreich Vorarlberg Feldkirch'
15:36:08.382 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-genre:Misc'
15:36:08.382 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-url:http://www.radio-especially.at'
15:36:08.382 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'content-type:audio/aacp'
15:36:08.425 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-pub:1'
15:36:08.425 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-metaint:8192'
15:36:08.425 -> [V][HTTPClient.cpp:1123] handleHeaderResponse(): RX: 'icy-br:96'
[...]

PR #4455, opened 2020-10-28
Fix header parsing

FrankBoesing:
Valid responses can have other content than HTTP1.x as first line. A valid response is e.g. "ICY 200 OK".
This fix allows other responses. 

This makes it easy to use the client e.g. for streaming audio.

fixes https://github.com/espressif/arduino-esp32/issues/4454

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -1199,7 +1199,6 @@ int HTTPClient::handleHeaderResponse()
                     _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
                 }
     }
-    headerLine.substring(headerLine.indexOf(' '));
     _returnCode = headerLine.substring(headerLine.indexOf(' ')).toInt();
 	
     while(connected()) {
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -1186,23 +1186,33 @@ int HTTPClient::handleHeaderResponse()
 
     _transferEncoding = HTTPC_TE_IDENTITY;
     unsigned long lastDataTime = millis();
+	
+    String headerLine = _client->readStringUntil('\n');
+    headerLine.trim(); // remove \r
 
+    lastDataTime = millis();
+
+    log_v("RX: '%s'", headerLine.c_str());
+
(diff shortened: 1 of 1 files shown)


FrankBoesing:
Any thoughts?

FrankBoesing:
It is no fun to contribute if there is not the slightest reaction.

I know why I don't like ESP.

me-no-dev:
Attitude and unrealistic expectations aside, I would suggest that you use a boolean `responseHeadParsed` or similar to properly parse the first line inside the while (thus having minimal code duplication).

FrankBoesing:
The first line is, yes, the first line. No need to parse it inside a loop (Why should one do that? Slower than needed)
Then, my Patch has minimal code duplication.

Anyway, as long it gets fixed I don't care.
Thanks.

FrankBoesing:
I'll open a new PR, which will be slower and larger, as you requested 👍

FrankBoesing:
Hm, nö, es widerstrebt mir zutiefst in der Schleife immer wieder auf den Header zu parsen.
Lieber einige (wenige!) Bytes mehr, als das.

Ich mach das hier wieder auf :)
Du hast schon gesehen, daß aus der Schleife jede Menge code entfernt wurde?

me-no-dev:
change is exactly 4 lines. I'm not sure what you mean about slower and larger...

```cpp
    bool firstLine = true;
    while(connected()) {
         size_t len = _client->available();
         if(len > 0) {
             String headerLine = _client->readStringUntil('\n');
             headerLine.trim(); // remove \r

             lastDataTime = millis();

             log_v("RX: '%s'", headerLine.c_str());

             if(firstLine) {
                 firstLine = false;
                 if(headerLine.startsWith("HTTP/1.") && _canReuse) {
                     _canReuse = (headerLine[sizeof "HTTP/1." - 1] != '0');
                 }
                 int codePos = headerLine.indexOf(' ') + 1;
                 _returnCode = headerLine.substring(codePos, headerLine.indexOf(' ', codePos)).toInt();
             } else if(headerLine.indexOf(':')) {
                 String headerName = headerLine.substring(0, headerLine.indexOf(':'));
                 String headerValue = headerLine.substring(headerLine.indexOf(':') + 1);
                 headerValue.trim();

```

FrankBoesing:
Ok, if you like that more :)

me-no-dev:
done :)

PR #4464, opened 2020-10-30
Fixing BLE GATT Characteristic notification and Characteristic Descriptor read

JimmyDurandWesolowski:
These commits fix some issues with GATT Characteristic and Characteristic Descriptor issues.

### DIFF
--- a/libraries/BLE/src/BLERemoteCharacteristic.cpp
+++ b/libraries/BLE/src/BLERemoteCharacteristic.cpp
@@ -468,7 +468,7 @@ void BLERemoteCharacteristic::registerForNotify(notify_callback notifyCallback, 
 		if(!notifications) val[0] = 0x02;
 		BLERemoteDescriptor* desc = getDescriptor(BLEUUID((uint16_t)0x2902));
 		if (desc != nullptr)
-			desc->writeValue(val, 2);
+			desc->writeValue(val, 2, true);
 	} // End Register
 	else {   // If we weren't passed a callback function, then this is an unregistration.
 		esp_err_t errRc = ::esp_ble_gattc_unregister_for_notify(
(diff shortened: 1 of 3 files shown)


libraries/BLE/src/BLERemoteCharacteristic.cpp
```diff
@@ -468,7 +475,7 @@ void BLERemoteCharacteristic::registerForNotify(notify_callback notifyCallback,
 		if(!notifications) val[0] = 0x02;
 		BLERemoteDescriptor* desc = getDescriptor(BLEUUID((uint16_t)0x2902));
 		if (desc != nullptr)
-			desc->writeValue(val, 2);
+			desc->writeValue(val, 2, true);
```
chegewara, line 478: What is the reason to response await here?

JimmyDurandWesolowski: I gave the details in the commit that introduces the change. For convenience, the commit message is:

> When registering a notification on a characteristic, the 2902 descriptor
> (CCCD) value is set to 1 (or 2 for indication).
> According to the BLUETOOTH CORE SPECIFICATION Version 5.2, Revision Date
> 2019-12-31, section 4.12.3 "Write Characteristic Descriptors" (page 1588),
> the characteristic descriptor write must expect a response.
> Currently, the descriptor write is performed without expecting a reponse,
> which prevents the notification to be functional with some BLE stacks.
> This commit modify the write to expect the response.

chegewara: Ok, thanks. 
Commit aproved.

chegewara (approved):

libraries/BLE/src/BLERemoteCharacteristic.cpp
JimmyDurandWesolowski: Thank you.

PR #4096, opened 2020-06-17
Adding LITTLEFS w/o rebuild the IDF

lorol:
based on https://github.com/joltwallet/esp_littlefs

This will be my first PR, hope will do it right.
Details in README

PR #4483, opened 2020-11-03
Adding LITTLEFS after esp_littlefs (IDF) is built-in 

lorol:
See: https://github.com/espressif/arduino-esp32/pull/4096
Also Tools idea:  https://github.com/lorol/arduino-esp32fs-plugin


me-no-dev:
please pull the latest change. it's required to make CI work :)

tools/sdk/esp32/include/esp_littlefs/include/esp_littlefs.h
```diff
@@ -13,7 +13,6 @@
 #include <sys/reent.h>
 #include <sys/stat.h>
 #include <sys/time.h>
-#include <sys/termios.h>
```
me-no-dev, line 16: this will unfortunately not work. header comes from upstream.

libraries/LITTLEFS/src/LITTLEFS.cpp
```diff
@@ -0,0 +1,105 @@
...
+static constexpr const char LFS_NAME[] = "spiffs";
+
+#include "vfs_api.h"
+
+extern "C" {
+#include <sys/unistd.h>
+#include <sys/stat.h>
+#include <dirent.h>
+#include "esp_littlefs.h"
+}
```
me-no-dev, line 24: you can add the following lines here to clear the issue with termios:
```cpp
#undefine B110
#undefine B1000000
```

lorol: Can you fix it from your side, please :) I am almost loosing my job trying to understand how this GIT works :)

lorol:
Great. Now testing, feedback and tools mess.

lorol:
@me-no-dev 
It looks like to suppress the B110 and B1000000 redefining warnings. #undeff's should be after #include "vfs_api.h" and before #include "esp_littlefs.h".
I guess the easiest way is just you to correct directly on next commit instead of PR only for that?

```
#include "vfs_api.h"

extern "C" {
#include <sys/unistd.h>
#include <sys/stat.h>
#include <dirent.h>
  <<< undefs to be moved here 
#include "esp_littlefs.h"
#undef B110
#undef B1000000
}
```

PR #4512, opened 2020-11-09
Add two new hardware boards

ladyada:
two ESP32S2 boards


me-no-dev:
@ladyada please add this commit to the `idf-release/v4.2` branch instead. This one is still here just in case while I am merging stuff around :)

ladyada:
ok

PR #4518, opened 2020-11-10
Added `#define LED_BUILTIN` to all pins_arduino.h that need it.

### LINKED ISSUE #4134: LED_BUILTIN must be macro not const int
The symbol LED_BUILTIN is supposed to be a macro not a const int
This was discussed at length on the Arduino developers mailing list years ago and final conclusion is that it must be a macro.
The reason is that some boards have a builtin led and some don't. 
The only way for code, particularly library code that runs a many different platforms/core/boards, to "know" if the board that the code is being compiled for has a built in led is to use a macro for LED_BUILTIN so that the code can do compile time conditionals and enable / disable any led code appropriately when being built for various boards.

If you go look at the variants that arduino.cc supplies and even the esp8266 variants, you will see that LED_BUILTIN is a macro rather than a const int

variants that use const int instead of a macro will "trick" code into thinking that there is no built in led.

lbernstone:
Fixes #4134

### DIFF
--- a/variants/Microduino-esp32/pins_arduino.h
+++ b/variants/Microduino-esp32/pins_arduino.h
@@ -13,6 +13,7 @@
 
 static const uint8_t LED_BUILTIN = -1;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN
 
 #define MTDO 15
 #define MTDI 12
--- a/variants/S_ODI_Ultra_v1/pins_arduino.h
+++ b/variants/S_ODI_Ultra_v1/pins_arduino.h
@@ -14,7 +14,9 @@
 static const uint8_t LED_BUILTIN = 2;
 static const uint8_t LED_BUILTINB = 4;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN
 #define BUILTIN_LED2  LED_BUILTINB
+#define LED_BUILTIN
 
 static const uint8_t TX = 1;
 static const uint8_t RX = 3;
--- a/variants/alksesp32/pins_arduino.h
+++ b/variants/alksesp32/pins_arduino.h
@@ -15,6 +15,7 @@
 
 static const uint8_t LED_BUILTIN = 23;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN
 
 static const uint8_t TX = 1;
 static const uint8_t RX = 3;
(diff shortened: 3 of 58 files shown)


variants/S_ODI_Ultra_v1/pins_arduino.h
```diff
@@ -14,7 +14,9 @@
 static const uint8_t LED_BUILTIN = 2;
 static const uint8_t LED_BUILTINB = 4;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN
 #define BUILTIN_LED2  LED_BUILTINB
+#define LED_BUILTIN
```
me-no-dev, line 19: `LED_BUILTINB`?

me-no-dev:
how does a define and a const char coexist?
```cpp
static const uint8_t LED_BUILTIN = 4;
#define LED_BUILTIN
```

variants/Microduino-esp32/pins_arduino.h
```diff
@@ -13,6 +13,7 @@
 
 static const uint8_t LED_BUILTIN = -1;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN
```
atanisoft, line 16: this and other cases should be `#define LED_BUILTIN LED_BUILTIN`

atanisoft:
> how does a define and a const char coexist?
> 
> ```c++
> static const uint8_t LED_BUILTIN = 4;
> #define LED_BUILTIN
> ```

The `#define` variant is used by the pre-processor and the other by the compiler during code generation. The pre-processor can not see `static const` variables unfortunately (even `static constexpr ... ` would be invisible)

me-no-dev:
this works:
```cpp

static const uint8_t LED_BUILTIN = 4;
#define LED_BUILTIN LED_BUILTIN

int main(){
#ifdef LED_BUILTIN
	printf("LED_BUILTIN: %u\n", LED_BUILTIN);
#endif
    return 0;
}

```

atanisoft:
Yes, but this won't:
```
static const uint8_t LED_BUILTIN = 4;

int main(){
#ifdef LED_BUILTIN
	printf("LED_BUILTIN: %u\n", LED_BUILTIN);
#endif
    return 0;
}
```

The ifdef block will be excluded from compilation.

me-no-dev:
yes yes :) i got the problem. PR needs fixing :) likle find and replace 😄

atanisoft:
> Can you back out this commit so I can submit a PR to fix it?

This PR hasn't been merged yet and is still pending here for fixes.

lbernstone:
No good deed goes unpunished.

PR #4534, opened 2020-11-14
Added usedBytes to match other filesystems

lbernstone:


me-no-dev:
we should maybe make them all uint64_t instead, so that we can have them the same across the file systems

PR #4545, opened 2020-11-17
Add ESP RainMaker support to Arduino (Only for esp32) [idf-release/v4.2]

sweetymhaiske:
End to End solution for Arduino users to realize their IoT ideas faster using phone apps, without making any configuration to the cloud.


me-no-dev:
Merged: https://github.com/espressif/arduino-esp32/commit/c743f22ba603fc459dce60b4c01961293cfa30d5

PR #4564, opened 2020-11-23
[1.0.5] Make looptask stack size configurable

coylen:
Added guarded define to set the stacksize on the main looptask.
Advantage of this is that build_flags can be used to provide a different value for the stack size should it be neccessary
default behaviour is unaffected

### DIFF
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -3,6 +3,10 @@
 #include "esp_task_wdt.h"
 #include "Arduino.h"
 
+#ifndef LOOP_STACK_SIZE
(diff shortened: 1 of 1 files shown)


cores/esp32/main.cpp
```diff
@@ -3,6 +3,10 @@
 #include "esp_task_wdt.h"
 #include "Arduino.h"
 
+#ifndef LOOP_STACK_SIZE
```
me-no-dev, line 6: please rename to `CONFIG_ARDUINO_ LOOP_STACK_SIZE` so I can add this to the IDF Kconfig as well

me-no-dev:
thanks :)

PR #4591, opened 2020-11-29
[2.0.0] Modified WString.h to allow for larger strings with PSRAM

lbernstone:
This is more for conversation, as I haven't tested very thoroughly.  Since psram is now mapped into malloc, it should be fairly straight forward to increase the maximum size of String.


me-no-dev:
merged: https://github.com/espressif/arduino-esp32/commit/9cbf4cb07629b3a6ac2bac663d32aa2902bbee6c

PR #4612, opened 2020-12-07
Create Blink_ET_Board

11dhking23:


me-no-dev:
I am not sure what is the purpose of this PR. Seems you are trying to add an example for some board, but the example is wrong and is just blink.ino that already exists :)

PR #4634, opened 2020-12-14
BluetoothSerial SSP Authentication with callbacks

### LINKED ISSUE #4622: SSP Functionality Incomplete
I have tried to make use of SSP and setup ESP32 device as slave to accept binding request from smart phones, PCs, etc.  However I figure out that the SSP implementation is incomplete.  Normally the binding process involves the Smart Phone initiate a binding request with a key code.  User can verify the code in the ESP32 device screen, and select the desire action (either accept or reject) in the ESP32 device.

However, just make use of the BluetoothSerial API cannot make the job done.  The API seems missing some callbacks mechanism for the binding process.

I need to copy and modify the BluetoothSerial source code to make this works, and I think I can contribute my work to the source as a pull request.  So, if I do this, is there anybody who will review the code and merge it?

richardclli:
Added authentication callbacks and example, resolves #4622.

### DIFF
--- a/libraries/BluetoothSerial/examples/SerialToSerialBT_SSP_pairing/SerialToSerialBT_SSP_pairing.ino
+++ b/libraries/BluetoothSerial/examples/SerialToSerialBT_SSP_pairing/SerialToSerialBT_SSP_pairing.ino
@@ -1,7 +1,7 @@
 //This example code is in the Public Domain (or CC0 licensed, at your option.)
-//By Evandro Copercini - 2018
+//By Richard Li 2020
 //
-//This example creates a bridge between Serial and Classical Bluetooth (SPP)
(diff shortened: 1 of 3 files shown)


me-no-dev:
Thanks :)

luc-github:
@richardclli  @me-no-dev  Hi I have tested the example and looks great, thank you - but is it normal that the setPin function as no effect and it is always a random pin that is requested ?

richardclli:
For SSP pairing, the pin is always randomly generated by the master device.  The setPin function is for old BT pairing mechanism and need to have some pre-compiled library replaced before it can work properly.  That is why I develop this function for new pairing mechanism that can work on the fly.  You can use it in 2 modes, required confirm (like pairing the phone with a watch) and auto-confirm (like BT speakers).

PR #4653, opened 2020-12-23
add ET-Board

misery-tt:
We have developed an ET board for educational esp32 and would like to add it.


misery-tt:
We have developed an ET board for educational esp32 and would like to add it.

misery-tt:
There's a bit of a different part of the ET-ga esp32 pin map that we're registering this time, so where should we contact? For example, A0 is 36, A1 is 39, A2 is 32, A3 is 34, A4 is 34, A5 is 35, A6 is 25, A7 is 26, D2 is 27, D3 is 14, D5 is 13, D5 is 15, D7 is 167, D8 is 49, D9 is 49.

me-no-dev:
where is the variant header that defines pins and such?

PR #4656, opened 2020-12-23
[2.0.0] Adds the setRXInterrupt

timkoers:
Should fix pull request https://github.com/espressif/arduino-esp32/pull/3376

I haven't tested the code yet myself, but feel free to review

### DIFF
--- a/.github/scripts/install-arduino-core-esp32.sh
+++ b/.github/scripts/install-arduino-core-esp32.sh
@@ -15,13 +15,14 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
 		pip install requests > /dev/null
 	fi
 
-	if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
-		echo "Linking Core..."
-		ln -s $GITHUB_WORKSPACE esp32
-	else
-		echo "Cloning Core Repository..."
-		git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
-	fi
+# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
+#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+	echo "Linking Core..."
+	ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+#else
+#	echo "Cloning Core Repository ..."
+#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
+#fi
 
 	echo "Updating Submodules ..."
 	cd esp32
--- a/.github/scripts/install-arduino-core-esp32.sh
+++ b/.github/scripts/install-arduino-core-esp32.sh
@@ -15,8 +15,13 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
 		pip install requests > /dev/null
 	fi
 
-	echo "Linking Core..."
-	ln -s $GITHUB_WORKSPACE "$ARDUINO_ESP32_PATH"
+	if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+		echo "Linking Core..."
+		ln -s $GITHUB_WORKSPACE esp32
+	else
+		echo "Cloning Core Repository..."
+		git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
+	fi
 
 	echo "Updating Submodules ..."
 	cd esp32
--- a/.github/scripts/install-arduino-core-esp32.sh
+++ b/.github/scripts/install-arduino-core-esp32.sh
@@ -18,7 +18,7 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
 # I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
 #if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
 	echo "Linking Core..."
-	ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+	ln -s $GITHUB_WORKSPACE "$ARDUINO_ESP32_PATH"
 #else
 #	echo "Cloning Core Repository ..."
 #	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
--- a/.github/scripts/install-arduino-core-esp32.sh
+++ b/.github/scripts/install-arduino-core-esp32.sh
@@ -15,14 +15,8 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
 		pip install requests > /dev/null
 	fi
 
-# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
-#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
 	echo "Linking Core..."
 	ln -s $GITHUB_WORKSPACE "$ARDUINO_ESP32_PATH"
-#else
-#	echo "Cloning Core Repository ..."
-#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
-#fi
 
 	echo "Updating Submodules ..."
 	cd esp32
--- a/.github/scripts/install-platformio-esp32.sh
+++ b/.github/scripts/install-platformio-esp32.sh
@@ -14,8 +14,13 @@ python -m platformio platform install https://github.com/platformio/platform-esp
 echo "Replacing the framework version ..."
 python -c "import json; import os; fp=open(os.path.expanduser('~/.platformio/platforms/espressif32/platform.json'), 'r+'); data=json.load(fp); data['packages']['framework-arduinoespressif32']['version'] = '*'; del data['packages']['framework-arduinoespressif32']['owner']; fp.seek(0); fp.truncate(); json.dump(data, fp); fp.close()"
 
-echo "Linking Core..."
-ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+	echo "Linking Core..."
+	ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+else
+	echo "Cloning Core Repository ..."
+	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
+fi
 
 echo "PlatformIO for ESP32 has been installed"
 echo ""
--- a/.github/scripts/install-platformio-esp32.sh
+++ b/.github/scripts/install-platformio-esp32.sh
@@ -14,13 +14,14 @@ python -m platformio platform install https://github.com/platformio/platform-esp
 echo "Replacing the framework version ..."
 python -c "import json; import os; fp=open(os.path.expanduser('~/.platformio/platforms/espressif32/platform.json'), 'r+'); data=json.load(fp); data['packages']['framework-arduinoespressif32']['version'] = '*'; del data['packages']['framework-arduinoespressif32']['owner']; fp.seek(0); fp.truncate(); json.dump(data, fp); fp.close()"
 
-if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
-	echo "Linking Core..."
-	ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
-else
-	echo "Cloning Core Repository ..."
-	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
-fi
+# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
+#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+echo "Linking Core..."
+ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+#else
+#	echo "Cloning Core Repository ..."
+#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
+#fi
 
 echo "PlatformIO for ESP32 has been installed"
 echo ""
--- a/.github/scripts/install-platformio-esp32.sh
+++ b/.github/scripts/install-platformio-esp32.sh
@@ -14,14 +14,8 @@ python -m platformio platform install https://github.com/platformio/platform-esp
 echo "Replacing the framework version ..."
 python -c "import json; import os; fp=open(os.path.expanduser('~/.platformio/platforms/espressif32/platform.json'), 'r+'); data=json.load(fp); data['packages']['framework-arduinoespressif32']['version'] = '*'; del data['packages']['framework-arduinoespressif32']['owner']; fp.seek(0); fp.truncate(); json.dump(data, fp); fp.close()"
 
-# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
-#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
 echo "Linking Core..."
 ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
-#else
-#	echo "Cloning Core Repository ..."
-#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
-#fi
 
 echo "PlatformIO for ESP32 has been installed"
 echo ""
--- a/.github/scripts/on-push.sh
+++ b/.github/scripts/on-push.sh
@@ -42,18 +42,15 @@ if [ "$BUILD_PIO" -eq 0 ]; then
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/WiFiClientSecure/examples/WiFiClientSecure/WiFiClientSecure.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/BLE/examples/BLE_server/BLE_server.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/AzureIoT/examples/GetStarted/GetStarted.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/NoInterrupt/NoInterrupt.ino"
+		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino"
+
 	elif [ "$OS_IS_MACOS" == "1" ]; then
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/WiFi/examples/WiFiClient/WiFiClient.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/WiFiClientSecure/examples/WiFiClientSecure/WiFiClientSecure.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/BluetoothSerial/examples/SerialToSerialBT/SerialToSerialBT.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/BLE/examples/BLE_server/BLE_server.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/AzureIoT/examples/GetStarted/GetStarted.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/NoInterrupt/NoInterrupt.ino"
+		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino"
 	else
 		# CMake Test
 		if [ "$CHUNK_INDEX" -eq 0 ]; then
--- a/.github/scripts/on-push.sh
+++ b/.github/scripts/on-push.sh
@@ -61,14 +61,12 @@ if [ "$BUILD_PIO" -eq 0 ]; then
 else
 	# PlatformIO Test
 	source ./.github/scripts/install-platformio-esp32.sh
-	BOARD="esp32dev"
-	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino" && \
+	BOARD="esp32dev"	
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/WiFi/examples/WiFiClient/WiFiClient.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/WiFiClientSecure/examples/WiFiClientSecure/WiFiClientSecure.ino" && \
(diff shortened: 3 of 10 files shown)


timkoers:
I'll take a look at the failing tests after X-mass

AlessandroAU:
Thanks for doing this! I've wanted this feature for awhile

timkoers:
I'm working on a rewrite, it seems that the last byte is not received properly or something. It is being received properly when using uartRead().

.github/scripts/install-arduino-core-esp32.sh
```diff
@@ -15,13 +15,14 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
...
-		git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
-	fi
+# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
+#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+	echo "Linking Core..."
+	ln -s $GITHUB_WORKSPACE "$ARDUINO_ESP32_PATH"
+#else
+#	echo "Cloning Core Repository ..."
+#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
+#fi
```
me-no-dev, line 25: please do not edit those scripts. They are used by CI

.github/scripts/install-platformio-esp32.sh
```diff
@@ -14,13 +14,14 @@ python -m platformio platform install https://github.com/platformio/platform-esp
...
-	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
-fi
+# I don't get the code below, you'd want to test the source code in the current repository (forked or not), not in the espressif repo right?
+#if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
+echo "Linking Core..."
+ln -s $GITHUB_WORKSPACE "$PLATFORMIO_ESP32_PATH"
+#else
+#	echo "Cloning Core Repository ..."
+#	git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
+#fi
```
me-no-dev, line 24: Same here. Those files are not to be edited by users.

.github/scripts/on-push.sh
```diff
@@ -42,14 +42,18 @@ if [ "$BUILD_PIO" -eq 0 ]; then
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/WiFiClientSecure/examples/WiFiClientSecure/WiFiClientSecure.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/BLE/examples/BLE_server/BLE_server.ino" && \
 		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/AzureIoT/examples/GetStarted/GetStarted.ino" && \
-		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino"
+		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino" && \
+		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino" && \
+		build_sketch "$FQBN" "$ARDUINO_ESP32_PATH/libraries/ESP32/examples/Serial/NoInterrupt/NoInterrupt.ino"
```
me-no-dev, line 47: these tests do not need to be added for Windows and Mac, they are already checked in Linux

me-no-dev:
Please note my comments and revers all code style changes. Stick to the style in the files and change only what is necessary for this to work. I can not review this request as-is.

.github/scripts/install-arduino-core-esp32.sh
timkoers: The scripts as they where did not pull my repo and therefore tested the espressif/arduino-esp32 git repository, excluding my code changes. I think this is undesired behaviour for a CI pipeline as a CI pipeline is used to test the current repository and should be setup to test the repository that the configuration file is stored in.

timkoers: I'll remove the commented lines, but I won't revert back to the original script as that doesn't make sense at all.

.github/scripts/on-push.sh
timkoers: Ok, but I don't get why all the others are tested then.

me-no-dev: because they are specific tests that require larger partitions and such. There for making sure that between the operating systems everything is correct.

.github/scripts/install-arduino-core-esp32.sh
```diff
@@ -15,13 +15,8 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
...
 
-	if [ "$GITHUB_REPOSITORY" == "espressif/arduino-esp32" ];  then
-		echo "Linking Core..."
-		ln -s $GITHUB_WORKSPACE esp32
-	else
-		echo "Cloning Core Repository..."
-		git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
-	fi
+	echo "Linking Core..."
+	ln -s $GITHUB_WORKSPACE "$ARDUINO_ESP32_PATH"
```
me-no-dev, line 19: please do not edit those scripts! you do not need them! they are for CI and specific tasks.

me-no-dev:
Please DO NOT EDIT ANY FILES INSIDE `.github` FOLDER!!!!

cores/esp32/esp32-hal-uart.c
```diff
@@ -69,34 +76,45 @@ static uart_t _uart_bus_array[3] = {
...
+                (*uart_interrupt->func)(c, uart_interrupt->user_arg);
+            }else if (uart->queue != NULL) {
                 xQueueSendFromISR(uart->queue, &c, &xHigherPriorityTaskWoken);
             }
         }
+
+        // Clear the interrupts after everything was read
+        uart->dev->int_clr.rxfifo_full = 1;
+        uart->dev->int_clr.frm_err = 1;
+        uart->dev->int_clr.rxfifo_tout = 1;
```
me-no-dev, line 109: there is a reason why interrupts are cleared BEFORE reading the bytes. Please revert.

cores/esp32/esp32-hal-uart.c
```diff
@@ -107,6 +125,14 @@ void uartEnableInterrupt(uart_t* uart)
     uart->dev->int_ena.rxfifo_tout = 1;
     uart->dev->int_clr.val = 0xffffffff;
 
+    if(arg != NULL){
+        (*arg) = malloc(sizeof(uart_interrupt_t));
+        (*arg)->func = func;
+        (*arg)->dev = uart;
+        (*arg)->user_arg = user_arg;
+        _uart_interrupt_array[uart->num] = (*arg);
```
me-no-dev, line 133: what if `_uart_interrupt_array[uart->num]` already exists? Leak?

.github/scripts/install-arduino-core-esp32.sh
timkoers: I understand what they are for, please take a look once at my edits and imagine the actual functioning of these scripts from a forked repo.

When the CI pipeline is ran from github.com/timkoers/arduino-esp32.git, the repository owner expects it's repository to be used during the CI pipeline. As of your original .sh files, the github.com/espressif/arduino-esp32.git repository is cloned and the various tests will be ran on the espressif/arduino-esp32 repository, **NOT** on the owners/forkers repository.

I had been searching for over 60+ minutes on why the various tests did result in anything on my repo, so please kindly accept this since it will not break anything on your side

cores/esp32/esp32-hal-uart.c
```diff
@@ -118,6 +144,9 @@ void uartDisableInterrupt(uart_t* uart)
     uart->dev->int_ena.val = 0;
     uart->dev->int_clr.val = 0xffffffff;
 
+    // Free uart rx interrupt
+    _uart_interrupt_array[uart->num] = NULL;
```
me-no-dev, line 148: Setting this to NULL means that if there was an allocated structure, you will now leak it. You need to FREE it first.

cores/esp32/esp32-hal-uart.c
timkoers: What is the reason? With all the respect that I have for you, trust me that is a lot since this Arduino framework is excellent, there are a lot of things like this that are not documented properly. 
Please comment the code properly to explain why the interrupts need to be cleared before reading the bytes

cores/esp32/esp32-hal-uart.h
```diff
@@ -80,6 +83,9 @@ unsigned long uartDetectBaudrate(uart_t *uart);
 
 bool uartRxActive(uart_t* uart);
 
+void uartDisableInterrupt(uart_t* uart);
```
me-no-dev, line 86: How about `uartDisableRxInterrupt ` instead?

cores/esp32/esp32-hal-uart.c
timkoers: uartDisableInterrupt get's called before the uartEnableInterrupt function get's called, so the `_uart_interrupt_array[uart->num]` should be deleted.
Just for safety I'll create a if statement that deletes the value if it is still present

cores/esp32/esp32-hal-uart.h
timkoers: Good one!

.github/scripts/on-push.sh
timkoers: Ah ok, will remove

cores/esp32/esp32-hal-uart.c
me-no-dev: there are cases where bytes might come in between the time you check and handle the bytes and the time you clear the interrupt. In that case you will not get an ISR for those bytes. We had that happen and scratched heads for quite some time. There was another case that I do not recall at this time as well. Generally your code had nothing to do with that, so I do not know why you would go and change the lines anyway :) I do accept the criticism that comment should be added for such cases.

.github/scripts/install-arduino-core-esp32.sh
me-no-dev: tests can run outside of github CI. in that case `GITHUB_REPOSITORY` will not be defined. Maybe detect if in Github CI instead of checking if `GITHUB_REPOSITORY` is the original repository. I can accept that.

cores/esp32/esp32-hal-uart.c
```diff
@@ -107,17 +138,45 @@ void uartEnableInterrupt(uart_t* uart)
     uart->dev->int_ena.rxfifo_tout = 1;
     uart->dev->int_clr.val = 0xffffffff;
 
+    if(arg != NULL){
+        (*arg) = malloc(sizeof(uart_interrupt_t));
+        (*arg)->func = func;
+        (*arg)->dev = uart;
+        (*arg)->user_arg = user_arg;
+		
+		if(_uart_interrupt_array[uart->num])
```
me-no-dev, line 147: please keep the style and add brackets. Also something went wrong with tabs/spaces. We use 4 spaces for tab

cores/esp32/esp32-hal-uart.c
```diff
@@ -107,17 +138,45 @@ void uartEnableInterrupt(uart_t* uart)
...
+  */
+void uartDisableRxInterrupt(uart_t* uart)
 {
     UART_MUTEX_LOCK();
     uart->dev->conf1.val = 0;
     uart->dev->int_ena.val = 0;
     uart->dev->int_clr.val = 0xffffffff;
 
+    // Free uart rx interrupt
+	if(_uart_interrupt_array[uart->num])
```
me-no-dev, line 176: same here for short `if`

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,21 @@
+HardwareSerial hwSerial(0);
+HardwareSerial hwSerial2(2);
+
+bool ok = true;
+
+static void IRAM_ATTR onSerialRX(uint8_t c, void* user_arg){  
+	hwSerial.print(c);
+	((HardwareSerial*)user_arg)->print(c);
```
me-no-dev, line 8: this line may be too much for some beginner users. 
```cpp
HardwareSerial* serial = (HardwareSerial*)user_arg;
serial->print(c);
```

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,21 @@
+HardwareSerial hwSerial(0);
```
me-no-dev, line 1: Why not use `Serial` instead?

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,21 @@
...
+HardwareSerial hwSerial2(2);
```
me-no-dev, line 2: This also exists as `Serial2`, though I will recommend you to use `Serial1` instead, because some newer chips like S2 and C3 do not have third UART port.

timkoers: But on those chips, will compiling Serial2 fail? Or will it just use Serial?

me-no-dev: on those chips `Serial` and `Serial1` exists. `Serial2` or `HardwareSerial hwSerial2(2)` do not and will cause error.

.github/scripts/install-arduino-core-esp32.sh
me-no-dev: how about you revert all changes to `.github/` files and open a new PR for that? Already having the PR here means that your changes do go through CI anyway

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,24 @@
+// Serial, Serial1 and Serial2 are all of the type HardwareSerial.
+
+static void IRAM_ATTR onSerialRX(uint8_t c, void* user_arg){  
+	Serial.print(c);
```
me-no-dev, line 4: There is another thing that comes to mind. If `Serial::print` is not in IRAM, calls to it from IRAM interrupt will cause Flash to be read and then that will cause exception. In future Arduino ISRs will not be in IRAM, but currently they are.

cores/esp32/esp32-hal-uart.c
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 
-//     http://www.apache.org/licenses/LICENSE-2.0
+//	   http://www.apache.org/licenses/LICENSE-2.0
```
me-no-dev, line 7: something went wrong here again :)

timkoers: Bloody autoformatter xD

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
timkoers: I'm changing the example to something else!

.github/scripts/install-arduino-core-esp32.sh
timkoers: So a simple,

	if [ ! -z "$GITHUB_REPOSITORY" ];  then
		echo "Linking Core..."
		ln -s $GITHUB_WORKSPACE esp32

Would suffice?

.github/scripts/on-push.sh
```diff
@@ -61,11 +62,13 @@ else
...
 	BOARD="esp32dev"
+	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/WiFi/examples/WiFiClient/WiFiClient.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/WiFiClientSecure/examples/WiFiClientSecure/WiFiClientSecure.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/BluetoothSerial/examples/SerialToSerialBT/SerialToSerialBT.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/BLE/examples/BLE_server/BLE_server.ino" && \
 	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/AzureIoT/examples/GetStarted/GetStarted.ino" && \
-	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino"
+	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino" && \
+	build_pio_sketch "$BOARD" "$PLATFORMIO_ESP32_PATH/libraries/ESP32/examples/Serial/NoInterrupt/NoInterrupt.ino"
```
me-no-dev, line 72: you can also remove changes in this file as well :) as we discussed, the examples are already built by GitHub CI in Linux (where all examples are built)

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,42 @@
+#define BUFFER_SIZE 8
+
+static volatile char inputBuffer[BUFFER_SIZE];
+static volatile size_t inputBufferLength = 0;
+
+// Please keep in mind, since the ESP32 is dual core, 
+// the interrupt will be running on the same core as the setRxInterrupt function was called on.
+static void IRAM_ATTR onSerialRX(uint8_t character, void* user_arg){
+
```
me-no-dev, line 9: this is a dangerous code, because the ISR might run in the middle of your for loop in `loop()`. Gotchas like this is the reason why this functionality was never added :) Using some sort of ring buffer can help, especially if it's protected by semaphore/mutex.

timkoers: Oh yeah sure, pardon me!

cores/esp32/esp32-hal-uart.c
```diff
@@ -30,273 +30,349 @@
 #include "soc/rtc.h"
 #include "esp_intr_alloc.h"
 
-#define UART_REG_BASE(u)    ((u==0)?DR_REG_UART_BASE:(      (u==1)?DR_REG_UART1_BASE:(    (u==2)?DR_REG_UART2_BASE:0)))
-#define UART_RXD_IDX(u)     ((u==0)?U0RXD_IN_IDX:(          (u==1)?U1RXD_IN_IDX:(         (u==2)?U2RXD_IN_IDX:0)))
-#define UART_TXD_IDX(u)     ((u==0)?U0TXD_OUT_IDX:(         (u==1)?U1TXD_OUT_IDX:(        (u==2)?U2TXD_OUT_IDX:0)))
+#define UART_REG_BASE(u)	((u==0)?DR_REG_UART_BASE:(		(u==1)?DR_REG_UART1_BASE:(	  (u==2)?DR_REG_UART2_BASE:0)))
```
timkoers, line 33: I don't understand why these are all back, I did not run any formatting tool.

me-no-dev: it is very much possible that your Editor/IDE does that on it's own

cores/esp32/esp32-hal-uart.c
```diff
@@ -69,23 +76,48 @@ static uart_t _uart_bus_array[3] = {
 
 static void uart_on_apb_change(void * arg, apb_change_ev_t ev_type, uint32_t old_apb, uint32_t new_apb);
 
-static void IRAM_ATTR _uart_isr(void *arg)
+static uart_interrupt_t *_uart_interrupt_array[3] = {NULL, NULL, NULL};
+
+static void IRAM_ATTR _uart_isr()
```
me-no-dev, line 81: here `_uart_isr` lost the `void *arg` argument. Is that intentional?

libraries/ESP32/examples/Serial/NoInterrupt/NoInterrupt.ino
```diff
@@ -0,0 +1,13 @@
+HardwareSerial hwSerial(0);
```
me-no-dev, line 1: Is this example necessary now?

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
```diff
@@ -0,0 +1,54 @@
...
+	  // Clear the buffer
+	  inputBuffer[i] = '\0';
+	}
+	// Clear the bufferLength
+	inputBufferLength = 0;
+	
+	xSemaphoreGive(bufferSemaphore);
+  }
+  
+  delay(1000); // Wait for one second
```
me-no-dev, line 53: I would suggest not to wait that long. Maybe 10ms would be enough. I'm afraid someone will run the example as is and will then complain that it's missing data :)

cores/esp32/esp32-hal-uart.c
timkoers: No, in my complete auto pilot I removed the parameter after setting NULL as an argument in esp_intr_alloc.
Fixed

libraries/ESP32/examples/Serial/Interrupt/Interrupt.ino
timkoers: I'll write another sketch that uses a queue to wait for data. In that sketch the delay function is not needed at all

timkoers:
Are you able to merge it?

ianr34:
it looks good - and when this issue was first raised (in another post, a year or two ago) I got so frustrated I just junked the whole serial class and did my own.. Polling for input in your top program is just so not the way to go..

me-no-dev:
Coming in 2.0.0 almost everything will be event based. I'm thinking that events are probably the best solution here also. They are thread-safe and can still access the RX Queue. Events can come based on different factors (byte received, transmission done, queue full, etc.). What do you think?

ianr34:
Yes, Yes, and Yes. Events are the way to go (not interrupts, that was OK for a short term solution) - the one I wrote to replace the arduino-esp32 one was event based, with a queue.

We have to get everyone away from thinking about a single task with polling, it simply isn't the way to use hardware this powerful, let alone the way to use more than 1 core, and all those nice hardware bits..

timkoers:
> 
> 
> Coming in 2.0.0 almost everything will be event based. I'm thinking that events are probably the best solution here also. They are thread-safe and can still access the RX Queue. Events can come based on different factors (byte received, transmission done, queue full, etc.). What do you think?

For the sake of unified code, I think that is good practice. Are you going to change it up to events?

me-no-dev:
@timkoers yes I will :) do not close this PR so it reminds me of it.

SuGlider:
SerialHardware and HAL have changed to work over IDF 4.4 layer.
This PR won't fit anymore in the new refactoring effort.

Another way to solve it would be creating ```Serial.onReceive(callback_function)```
This follows Arduino style and works as way to ```setRCInterrupt``` as well.

Example:

A potential solution is being implemented.
It's the ```Serial.onReceive(function)```

As soon as any data arrives UART RX, ```function()``` will called. 

Example:
``` dart
void UART_RX_IRQ(){
  uint16_t size = Serial.available();
  Serial.printf("Got %d bytes on Serial to read\n", size);
  while(Serial.available()) {
    Serial.write(Serial.read());
  }
  Serial.printf("\nSerial data processed!\n");
}

void setup() {
  Serial.begin(115200);
  Serial.onReceive(UART_RX_IRQ);
  Serial.println("Send data to UART0 in order to activate the RX callback");
}

void loop() {
  Serial.println("Sleeping for 10 seconds...");
  delay(10000);
}
```

ianr34:
I I said above - it shouldn't be an interrupt, it should be either a queue that you can wait on ie xQueueReceive, or an event that you can also wait on. 

I wrote my own quite a while ago, you do a uart_driver_install, then have a task waiting on that queue, then a task that waits on that to copy it over to another queue for the user program (doing a logical packet at a time), with an optional event. Works great and NO polling OR interrupts.. 

Waiting on something is much better than interrupts for these types of things, though maybe I think that because I've been writing multi threaded programs for 40 years... It's too limiting trying to process things in an interrupt, which often means you'd just have to raise an event in that anyway... (And I don't mean a volatile bool that you do polling on type of event!!)

SuGlider:
> I I said above - it shouldn't be an interrupt, it should be either a queue that you can wait on ie xQueueReceive, or an event that you can also wait on.
> 
> I wrote my own quite a while ago, you do a uart_driver_install, then have a task waiting on that queue, then a task that waits on that to copy it over to another queue for the user program (doing a logical packet at a time), with an optional event. Works great and NO polling OR interrupts..
> 
> Waiting on something is much better than interrupts for these types of things, though maybe I think that because I've been writing multi threaded programs for 40 years... It's too limiting trying to process things in an interrupt, which often means you'd just have to raise an event in that anyway... (And I don't mean a volatile bool that you do polling on type of event!!)

Hi @ianr34 

The Arduino API will be the ```Serial.onReceive(callback_func)``` but internally it will use a task waiting for UART Events to know when to execute the UART callback. IDF UART has an EventQueue that is populated from the UART ISR, which will be read by a separated task.

So everything will be done exactly as you described above. It will work correctly with FreeRTOS, IDF and Arduino.
You will be able to see it in the code next week, when I publish this PR.

Thanks for your contribution!

ianr34:
thanks - I'd missed the fact it was a callback, not an interrupt (or a callback from an interrupt) - which means they can spend as much time in it as they need to.. Though of course they will need to be processing fast enough that the serial queue doesn't fill up waiting for this callback to return (I assume in most cases they will be reading the queue in this callback) .. :-)

SuGlider:
> thanks - I'd missed the fact it was a callback, not an interrupt (or a callback from an interrupt) - which means they can spend as much time in it as they need to.. Though of course they will need to be processing fast enough that the serial queue doesn't fill up waiting for this callback to return (I assume in most cases they will be reading the queue in this callback) .. :-)

Yes, exactly. The only potential issue is when the callback never returns and blocks the UART Event Task. But it won't cause any issue to the UART ISR and neither to the regular Arduino Sketch that is running in a separated task as well (or to any other task running within the application).

I think it's a safe solution. Let me know if you think about any other potential issue in order to consider it when implementing it.
Thanks again!

PR #4657, opened 2020-12-25
ArduinoOTA upload intermittent failure fixed

leifclaesson:
OTA upload often fails when client.read() return -1 and we subsequently try to write 4 gigabytes to flash. I fixed it by adding a signed comparison of the client.read() return value, and a continue. Completely solves the intermittent upload problem for me.


libraries/ArduinoOTA/src/ArduinoOTA.cpp
```diff
@@ -315,6 +315,10 @@ void ArduinoOTAClass::_runUpdate() {
         size_t r = client.read(buf, available);
         if(r != available){
             log_w("didn't read enough! %u != %u", r, available);
+			if((int32_t) r<0){
+				delay(50);
```
me-no-dev, line 319: why delay 50 though? maybe delay less?

lucasssvaz:
@leifclaesson Could you please sign the CLA ?

lucasssvaz (approved):

libraries/ArduinoOTA/src/ArduinoOTA.cpp
```diff
@@ -315,6 +315,10 @@ void ArduinoOTAClass::_runUpdate() {
         size_t r = client.read(buf, available);
         if(r != available){
             log_w("didn't read enough! %u != %u", r, available);
+			if((int32_t) r<0){
+				delay(1);
+				continue;  	//let's not try to write 4 gigabytes when client.read returns -1
+			}
         }
```
P-R-O-C-H-Y, line 322:
```suggestion
        if(r != available){
            log_w("didn't read enough! %u != %u", r, available);
	    if((int32_t) r<0) {
	        delay(1);
	        continue;  	//let's not try to write 4 gigabytes when client.read returns -1
	    }
        }
```

PR #4659, opened 2020-12-26
Permit `notify` event subscription when BLE servers don't require registration

xtrasimplicity:
**Background**
I have an iTag that I'm wanting to use as a BLE button to control an application on my ESP32.

**Problem**
The iTag doesn't require explicit registration for a client to be able to listen to `notify` events. Currently, the `BLERemoteCharacteristic::registerForNotify` function tries to update the remote descriptor at `0x2902` and register/unregister the client as required.

As the iTag doesn't have this descriptor, it causes an unhandled exception error and crashes the application.

**Changes**
To support devices that don't require you to explicitly register to follow such events, I've added a simple `descriptorRequiresRegistration` flag to the `BLERemoteCharacteristic::registerForNotify` function. This defaults to true, for backward compatibility.

Thanks!

### DIFF
--- a/libraries/BLE/src/BLERemoteCharacteristic.cpp
+++ b/libraries/BLE/src/BLERemoteCharacteristic.cpp
@@ -453,7 +453,7 @@ std::string BLERemoteCharacteristic::readValue() {
  * unregistering a notification.
  * @return N/A.
  */
-void BLERemoteCharacteristic::registerForNotify(notify_callback notifyCallback, bool notifications) {
+void BLERemoteCharacteristic::registerForNotify(notify_callback notifyCallback, bool notifications, bool descriptorRequiresRegistration) {
 	log_v(">> registerForNotify(): %s", toString().c_str());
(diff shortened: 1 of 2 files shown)


chegewara:
Even if your iTag does not require to write that descriptor it does not hurt to do it. BLE specification says that characteristic which supports notify/indication shall have descriptor 0x2902, which means you can read/write it.
That change does nothing useful in library.

xtrasimplicity:
Writing to that descriptor with my iTag results in an unhandled exception as the descriptor does not exist. This change allows the rest of the event subscription process to take place, without writing to that descriptor.

chegewara (approved):

chegewara:
Im still not sure about that change, but i accept explanation and code is backward compatible, so approved.

xtrasimplicity:
Thanks, @chegewara! Are you able to merge this, or is this something that will need to be done by someone else?

Issue #4668, opened 2020-12-29
Sketch using both BLE and WiFi exceeds available space


tko:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			1.0.4
IDE name:						Arduino IDE
Flash Frequency:					80MHz
PSRAM enabled:                                    no
Upload Speed:					921600
Computer OS:                                        macOS Mojave 10.14.6

### Description:

I'm trying to build a thing to forward BLE announcements to an HTTP endpoint in WiFi network. However, just combining minimal BLE and WiFi code already exceeds available space. Compiling fails with error:
```
Sketch uses 1311266 bytes (100%) of program storage space. Maximum is 1310720 bytes.
text section exceeds available space in board
Global variables use 57716 bytes (17%) of dynamic memory, leaving 269964 bytes for local variables. Maximum is 327680 bytes.
Sketch too big; see http://www.arduino.cc/en/Guide/Troubleshooting#size for tips on reducing it.
Error compiling for board ESP32 Dev Module.
```

(My first attempt in firmware / arduino, not too familiar how restrictive the environment really is but this is feeling like hitting the limits much too quickly.)

### Sketch:
```cpp
#include <BLEAdvertisedDevice.h>
#include <BLEDevice.h>
#include <BLEScan.h>
#include <BLEUtils.h>
#include <WiFi.h>

void setup() {
  Serial.begin(115200);
  WiFi.begin("", "");
  BLEDevice::init("");
  auto scan = BLEDevice::getScan();
  auto result = scan->start(0);
}
void loop() {
}
```

chegewara:
https://github.com/espressif/arduino-esp32/issues?q=is%3Aissue+of+program+storage+space.+Maximum+is+is%3Aclosed+ble+wifi

h2zero:
As an alternative solution I welcome you to checkout my library here:

https://github.com/h2zero/NimBLE-Arduino

It uses half of the flash and memory of the built-in BLE library.

Very few changes to make to your code as the classes are all aliased to the built in BLE library. For the most part it's a just a change to the include header.

Rhodion:
https://www.youtube.com/watch?v=5VoXNloOwZE&ab_channel=XTronical
case you are on mac (caso esteja no mac) /Users/$(whoami)/Projects/bt4ml/data_manager_page

PR #4673, opened 2020-12-29
Mention size limitation in default configuration

tko:
Given that BLE and WiFi are kind of a selling point of ESP32 to begin with it's worth mentioning they can't be used together without some tweaks to default configuration.

---

The alternate partitioning scheme I picked is one I think is the _least_ drastically different from the default. I have no experience to evaluate the alternatives other than things like dropping OTA completely, or switching to FATFS just feel completely unwarranted when all one needs is some more bytes. (I can guess the implications of no-OTA but I've absolutely no idea about FATFS.)

IMO the default configuration should "just work" out of the box to minimize any friction with newcomers while making as much of the chip accessible as possible, but I've no idea of all the things default configuration needs to cater to.

Ref: https://github.com/espressif/arduino-esp32/issues/4668

### DIFF
--- a/docs/arduino-ide/boards_manager.md
+++ b/docs/arduino-ide/boards_manager.md
@@ -10,3 +10,5 @@ Starting with 1.6.4, Arduino allows installation of third-party platform package
 - Start Arduino and open Preferences window.
 - Enter one of the release links above into *Additional Board Manager URLs* field. You can add multiple URLs, separating them with commas.
 - Open Boards Manager from Tools > Board menu and install *esp32* platform (and don't forget to select your ESP32 board from Tools > Board menu after installation).
+
+**Note:** The default configuration does not work for using both BLE and WiFi in the same sketch. If you need both or even more modules at the same time other Partition Schemes _(Tools -> Partition Scheme)_ may work (for example "Minimal SPIFFS (1.9MB APP with OTA/190K SPIFFS)")


lbernstone:
What does this have to do with installing the board package?  Perhaps this belongs at https://github.com/espressif/arduino-esp32/tree/master/libraries/BLE

chegewara:
I have a question. When you will have another project, not related to ble and wifi, and binary size will exceed default partition, you will come back again and add new description? Changing partition scheme is most basic thing in arduino, not only esp32.

BTW not everyone is buying esp32 because of ble + wifi, there is many more features including 2 cores, psram, RMT etc

tko:
@lbernstone the change is to the README that is linked from the "Using Arduino IDE Boards Manager (preferred)" from the repo root and so following the instructions gives some visibility to the unobvious restrictions. libraries/BLE is buried deeper and not even mentioned at repo root so it's relatively invisible.

@chegewara Depends on the context. From what I vaguely understand only BLE is is so massive that combining it with anything else is likely to exceed default partition size. Or are there some other surprises that simply won't work by default?

My concern is about the out of the box surprises, IMO for a beginner friendly project like Arduino it is important that the defaults are good ones and not add any unnecessary roadblocks. As it currently stands BLE+WiFi just plain does not work (more below) -- if changing partitioning scheme fixes it, couldn't the default partitioning scheme support it already and save everyone some trouble? Would changing the defaults to accommodate an important(?) use case break other use cases? If not then why not change it so that getting started is easier, and beginners wouldn't be _required_ to know about partitions when starting?

Regarding the wording in the PR I noticed the instruction is incomplete as apparently each board has its own partitioning schemes. My chip just arrived and depending on which details I read it's either a NodeMCU (default partition doesn't work and there is no easy alternative either as there is no "Partition Scheme" menu option at all) or WROOM (not present as a board alternative)

me-no-dev:
While I do not mind such notes to be added, that should happen in another place. Like some general help/FAQ page

PR #4682, opened 2021-01-02
Use non-deprecated DPORT values in SPI HAL

kabir-plod:
According to comments in `tools\sdk\include\soc\soc\dport_reg.h`, the DPORTs were deprecated. They caused compile errors in a downstream project.


me-no-dev:
Thanks :)

PR #4685, opened 2021-01-04
Create ET-Board

ketri2484:
I'd like to put the header file of the ET board.


me-no-dev:
But this is not how board support is added. Also not how a library is added. What do you want to achieve?

PR #4737, opened 2021-01-21
Added callbacks to BLERemoteCharacteristic

morsisko:
Hello,

currently it is not possible to subscribe for the remote characteristic's notification with additional context. This way it's really hard to distinguish the source of callback in case if you are connected to multiple devices that shares the same callback function. Thanks to this pull request one can implement the BLERemoteCharacteristicCallbacks interface and then they can easily determine the device connected to particular notification. It is very useful for my case.

Let me know what do you think about it

### DIFF
--- a/libraries/BLE/src/BLERemoteCharacteristic.h
+++ b/libraries/BLE/src/BLERemoteCharacteristic.h
@@ -11,6 +11,7 @@
 #if defined(CONFIG_BT_ENABLED)
 
 #include <string>
+#include <functional>
 
 #include <esp_gattc_api.h>
 
@@ -21,8 +22,7 @@
 
 class BLERemoteService;
 class BLERemoteDescriptor;
-typedef void (*notify_callback)(BLERemoteCharacteristic* pBLERemoteCharacteristic, uint8_t* pData, size_t length, bool isNotify);
-
+typedef std::function<void(BLERemoteCharacteristic* pBLERemoteCharacteristic, uint8_t* pData, size_t length, bool isNotify)> notify_callback;
 /**
  * @brief A model of a remote %BLE characteristic.
  */


kind3r:
After https://github.com/espressif/arduino-esp32/pull/4735 was merged you can now access characteristic's service and from there the device so you know which device and service the notification was send from.

morsisko:
Yeah, this is definitely some sort of solution, but this way you need to keep global array of pairs like (device_id, associated_object_pointer) (unless I'm missing something) which is not the cleanest solution I would say

kind3r:
Oh, I think I see what you mean now. The callback is now a function and you want to be able to provide a method (part of an object). Wouldn't std::bind work in this case ?

morsisko:
Yes, that's what I want to do. Unfortunately std::bind won't work here, because the notification callback parameter is raw function pointer, std::bind need std::function to work, but this change would break backward compatibility. Additionally a lot of BLE classes use the interface solution, for example you got `BLEClientCallbacks` or `BLEServerCallbacks`

me-no-dev:
What I do not like in the situation proposed is that there are two methods with different names that seem to do the same/very similar thing. On top of that it's possible to have both defined and if you first `registerForNotify(cb)` and then `setCallbacks(nullptr)`, then notifications will be disabled for both.

Please find some common ground and propose a solution that will work in both cases and will have a common internal exit route. Lambdas come into mind :) You can have a look at `AsyncUDP::onPacket` as example.

morsisko:
Ok, thanks for the feedback. I've just pushed new commit. This time I've changed the typedef to std::function. This change is backward compatible, as the function pointers may be implicitly casted to std::function. To add additional context variable to the notification call, user may just bind the variable using std::bind.

me-no-dev:
Thanks :)

Issue #4596, opened 2020-12-01
[1.0.5-rc3] AdvertisedDevice::getServiceDataUUID() excepts when no UUID is present


mkfrey:
I'm preparing the migration of some of my code from 1.0.4 to 1.0.5 and found that the behaviour of `AdvertisedDevice::getServiceDataUUID()` has now changed.

I don't know if this is actually a bug, but this new behaviour could lead to some confusion for beginners,
since function calls like this should not throw an exception my opinion. So I'm opening this issue.

In 1.0.4 a call to `AdvertisedDevice::getServiceDataUUID()` would return the advertised devices member `m_serviceDataUUID`. If the device did not advertise any service, this would be a `BLEUUID` created with the default constructor.

As of 1.0.5-rc3, this call returns the first element of the vector `m_serviceUUIDs`. This vector contains no entries if no UUIDs were advertised, which means that `m_serviceUUIDs[0]` might throw an exception.

There also seems to be no function for getting the count of advertised UUIDs. This is an issue, because there is no way to determine beforehand whether `AdvertisedDevice::getServiceDataUUID()` or `BLEAdvertisedDevice::getServiceDataUUID(int i)` will cause an exception.

PR #4739, opened 2021-01-22
Refactor BLEAdvertisedDevice

mkfrey:
* Prevent possible undefined behaviour by get methods not taking an index as parameter
* Add methods to get the count of service data UUIDs and service UUIDs
* Various code improvements


mkfrey:
Will also fix #4596.

me-no-dev:
Thanks :)

Issue #4047, opened 2020-06-01
deleting BLEClient is not safe


jackjansen:
My BLE client code was structured more-or-less as

```
BLEClient *client = new BLEClient();
client->connect(address, type);
// interact with client
client->disconnect();
delete client;
```

but this is unsafe. Sometimes I get a corrupted heap, sometimes I get other crashes, possibly after running this code again (I am creating many connections to many BLE devices over a long time).

I _think_ the problem may be that `disconnect()` is asynchronous, because often my crashes are in `BLEClient::gattClientEventHandler()` in the `ESP_GATTC_DISCONNECT_EVT` case handler, but they can also happen in the BLE thread, or in completely unrelated code.

The workaround I have is to never delete my BLEClient objects but to cache and reuse them, but from the code (and the minimal documentation) it seems like I should have been able to create and delete them on-the-fly.

yongkimleng:
From what I tested, you have to wait until client->isConnected returns false before you can delete it. I added a spinloop which checks every 50ms before deleting.
Other methods such as keeping old instances in a list for later garbage-collection does not yeild anything stable. Seems risky to have more than one BLEClient at any time.

chegewara:
BLEClient is deleted inside library after disconnecting.

yongkimleng:
oh? is it a recent change?

chegewara:
No, its been added months ago, before BLE library has been merged.
BLEClient:
https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/src/BLEClient.cpp#L191
and BLEServer:
https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/src/BLEDevice.cpp#L589-L628

buxtronix:
I have just had this issue also. It turns out that GATTC_DISCONNECT (and other events) are propagated to ALL clients, not just the one that disconnected.

Fix is in #4055 - my multi-client code has been pretty stable since i applied it.

jackjansen:
Hmm. Looking at the BLEClient code a bit more. 

I think that _maybe_ adding a `BLEDevice::removePeerDevice(m_appId, true);` to the destructor is enough to fix the original issue (but unfortunately given the structure of my code I cannot test this hypothesis easily, sorry....).

[...]

PR #4742, opened 2021-01-25
Improve cleanup in BLEClient

kind3r:
- Remove client from the list of devices in case registration fails
- Filter other events not related to registration during registration phase
- Cleanup if connect fails
- Reset if after disconnect
- Disconnect callback *after* cleanup is done so object can be deleted

This fixes some of the issues I had like:
- `BLEClient::connect` hangs up and never recovered because registration failed
- `BLEClient` could not be deleted after disconnect or deletion creating ghost events https://github.com/espressif/arduino-esp32/issues/4047
- `BLEClient` could not be properly reused after a connection was attempted (successful or not)


me-no-dev:
Thanks for the fix :)

PR #4748, opened 2021-01-26
etboard pinmap upload

ketri2484:
I want to add a device.
Device name is ET_Board.


lbernstone:
You will need to add it to [boards.txt](https://github.com/espressif/arduino-esp32/blob/master/boards.txt) as well.

variants/ET_Board/pins_arduino.h
```diff
@@ -0,0 +1,20 @@
+// #include "pinmap.h"
+
```
me-no-dev, line 2: please check `variants/esp32/pins_arduino.h` for proper pin header. You need to have a few defines plus proper declaration of I2C, SPI and ADC pins. You can leave the Dx pins as defines.

variants/ET_Board/pins_arduino.h
```diff
@@ -0,0 +1,47 @@
...
+static const uint8_t SDA = 33;
+static const uint8_t SCL = 36;
+
+static const uint8_t SS    = 29;
+static const uint8_t MOSI  = 37;
+static const uint8_t MISO  = 31;
+static const uint8_t SCK   = 30;
+
+static const uint8_t A0 36   // BUILTIN_Potentiometer
```
me-no-dev, line 25: you have forgotten the equal sign and period at the end. This will not compile :)

`static const uint8_t A0 = 36;`

ketri2484: thank you for your advice.

Issue #4773, opened 2021-02-04
SimpleWiFiServer-Example hangs on first connect #esp32s2


newHeiko:
I'm trying to get the SimpleWiFiServer example to run so I can get my feet wet with the ESP32-S2.

The code compiles and uploads, connects to my WiFi and I get a few responses from the server when connecting with my phone. After a few of them, typically hitting both links on the page once or twice, it just hangs and doesn't even answer to pings any more.

I saw a couple of issue reports and commits in the master branch, so I cherry-picked all the commits that sounded like HTTPServer or WiFiClient into a separate PR which I will link to in my next post.

Changed the example sketch to include two calls to esp_wifi_set_ps(WIFI_PS_NONE).

Added another delay to the end of the main loop.

Commented out the digitalWrite calls in case there is some HW on the board.

Still the same error.

Debug shows (among others):

[ 28339][D][WiFiClient.cpp:522] connected(): Disconnected: RES: 0, ERR: 128

...but I can't seem to find what errno 128 (or -1?) is...

Careful reading of the debug output might indicate the issue pops up when the browser tries to open two connections at once to read the favicon?

### Hardware:
Board:							ESP32-S2-WROOM on ESP32-S2-DevKit-LiPo by Olimex https://www.olimex.com/Products/IoT/ESP32-S2/ESP32-S2-DevKit-Lipo/open-source-hardware
Core Installation version:			branch: esp32s2 and esp32s2 with 10 commits cherry-picked from master
IDE name:						Arduino IDE 1.8.13
CPU Frequency:                                      240MHz
Flash Frequency:					80MHz
PSRAM enabled:                                     no
Upload Speed:				        921600
Computer OS:                                         Slackware Linux

### Sketch:
```cpp
/*
 WiFi Web Server LED Blink

 A simple web server that lets you blink an LED via the web.
 This sketch will print the IP address of your WiFi Shield (once connected)
 to the Serial monitor. From there, you can open that address in a web browser
 to turn on and off the LED on pin 5.

 If the IP address of your shield is yourAddress:
 http://yourAddress/H turns the LED on
 http://yourAddress/L turns it off

 This example is written for a network using WPA encryption. For
 WEP or WPA, change the Wifi.begin() call accordingly.

 Circuit:
 * WiFi shield attached
 * LED attached to pin 5

 created for arduino 25 Nov 2012
 by Tom Igoe

ported for sparkfun esp32 
31.01.2017 by Jan Hendrik Berlin
 
[...]

PR #4774, opened 2021-02-04
Just for Reference: Cherry-picked WiFi-related commits from master...

newHeiko:
...see #4773  - this is the code I used to try and find out if it helps the issue...

Heiko


newHeiko:
Sorry for the noise - switching to idf-release/v4.2 fixes this.

PR #4775, opened 2021-02-04
Update README.md to include ESP32-S2 instructions

newHeiko:


me-no-dev:
we are moving very soon to 4.2 so this message would be very short lived :)

PR #4776, opened 2021-02-05
[2.0.0] Fix issue with init I2C for esp32 S2

chegewara:


chegewara:
Actually without this fix I2C crash even with arduino IDE and most recent idf-release/v4.2.
@me-no-dev

makermelissa:
You could probably fix it by adding:
```
conf.clk_flags = 0;
```
I found the answer at https://github.com/espressif/esp-idf/issues/6293#issuecomment-747901072 and tested it out successfully.

chegewara:
> You could probably fix it by adding:
> 
> ```
> conf.clk_flags = 0;
> ```
> 
> I found the answer at [espressif/esp-idf#6293 (comment)](https://github.com/espressif/esp-idf/issues/6293#issuecomment-747901072) and tested it out successfully. 

Thats true, but next time when esp-idf will be updated with new parameter you will have to update this library, and again and again.  This way of init variable is good practice:
`i2c_config_t conf = {0};`

makermelissa:
Ah, I get what you're saying. :)

makermelissa (approved):
Tested and this fix worked for me.

me-no-dev:
I would say, use memset to set the whole struct to zero, not just the first element :)

```cpp
i2c_config_t conf;
memset(conf, 0, sizeof(i2c_config_t)); //Might require to #include "string.h"
```

chegewara:
Well, i always thought it is the way to initialize all elements in struct  with 0.

me-no-dev:
https://stackoverflow.com/questions/6891720/initialize-reset-struct-to-zero-null/6891731#comment46703800_6891773

The answer below that comment also gives some info. And given that this has caused issues on ESP32 before, I suggest you change it to memset :)

ncohen17:
Hi, 

I am having a similar issue. I am trying to run the ESP-NOW protocol together with the I2C interface but I am getting an error. 

This is what I am getting on the serial terminal. 

`E (39) i2c: i2c_param_config(671): i2c clock choice is invalid, please check flag and frequency
I2C slave init failed`

I have tried adding this conf.clk_flags = 0; and this i2c_config_t conf;
memset(conf, 0, sizeof(i2c_config_t)); to the code but it is still not working. I am getting the same error message on ther serial terminal. 

Can someone please advise on the fix to this issue ?

PR #4777, opened 2021-02-05
[2.0.0] Add support to read/write raw sectors from/to sd card

chegewara:
This is added to easy use with usb msc.

### DIFF
--- a/libraries/SD/src/SD.cpp
+++ b/libraries/SD/src/SD.cpp
@@ -103,4 +103,15 @@ uint64_t SDFS::usedBytes()
 	return size;
 }
 
+bool SDFS::readRAW(uint8_t* buffer, uint32_t sector)
+{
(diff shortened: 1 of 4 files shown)


cores/esp32/esp32-hal-i2c.c
```diff
@@ -1816,7 +1816,7 @@ i2c_t * i2cInit(uint8_t i2c_num, int8_t sda, int8_t scl, uint32_t clk_speed){
 		i2c_driver_delete((i2c_port_t)i2c_num);
 	}
 
-    i2c_config_t conf;
+    i2c_config_t conf = {0};
```
me-no-dev, line 1819: this should not be here :P

chegewara: Right, my mistake.

chegewara: ok, fixed.

PR #4786, opened 2021-02-09
[AsyncUDP] Fix: free memory after processing data receive

TienHuyIoT:
The memory handle data receive must be free after processing in _recv() function.

As follows:
The NetBIOS application is using AsyncUDP library to handle PORT(137). The heap memory will be decreasing after get a NetBIOS package from network.


me-no-dev:
packet should be freed in NetBIOS instead :)

me-no-dev:
Hmm maybe not... some tests are needed to make sure that other code will work as expected too.

me-no-dev:
Issue is fixed here: https://github.com/espressif/arduino-esp32/commit/8134a42162f11cb01155038a2465848d6b2b84bc

Problem was that if a handler passes the packet to another function not by reference, then a copy will be created, it's pbuf will not be reffed, but it will be freed/un-reffed on destruct.

PR #4789, opened 2021-02-10
AsyncUDP: Added lastErr helper variable

alorbach:
The variable is useful when debugging AsyncUDP send problems.
The upper application can read and analyze the error reason.


me-no-dev:
Nice!

PR #4791, opened 2021-02-10
Update PlatformIO build script to handle spaces in env names

valeros:
This PR fixes a possible issue when a user specifies an environment with a whitespace character in its name, e.g.:
```ini
[env:esp wrover kit]
platform = espressif32
framework = arduino
board = esp-wrover-kit
```

https://github.com/platformio/platform-espressif32/pull/483 depends on this fix.

cc @me-no-dev


me-no-dev:
thanks @valeros :)

PR #4811, opened 2021-02-15
[2.0.0] BtClassic Discovery with info without connect

ArcticSnowSky:
Hey guys,
so I wanted to do a BtClassic Discovery without the need to call connect
and to list all found devices on a display and continue work with that list.

I wasn't capable to test the example code with my file structure, but I did use the discovery already in some different situations.

However when I noted that the Bluedroid stack won't let me enforce an RfComm SPP connection to a GPS Device (Skytraxx 2 plus, I guess its interface is built so simple that it doesn't advertise its SPP over SDP), I will probably have to switch to BtStack (BlueKitchen) and stop on this side meanwhile


ArcticSnowSky:
Hi, today I moved my changes into a separated branch and merged master into.
Do I have to create a new Pull request in this scenario?

me-no-dev:
Just please make sure that you fix the merge conflict with `libraries/BluetoothSerial/src/BluetoothSerial.cpp`

patfelst:
I'm interested in using this, it will be very useful for my project - thanks @ArcticSnowSky. How do I use this code before it is merged into master, or will that happen soon?

ArcticSnowSky:
> will be very useful for my project - thanks @ArcticSnowSky. How do I use this code before it is merged into master, or will that happe

Hi, I for my part - at the beginning - copied the BluetoothSerial folder into my project folder and used those as direct include (Simple approach).
Later on I used this forked and modified repo in my .platformio with a @3.6 (as the newest platform-framework is used by platformio in my case), this approach is probably more complicated, but might be cleaner from the perspective of your project.

patfelst:
Thanks I've worked out how to clone the repository (I think!) and will replace the "BluetoothSerial" folder with your work.

PR #4819, opened 2021-02-17
Add Lolin32 Lite and TTGO T7 boards

balrog-kun:
Add the Wemos Lolin32 Lite (no longer being produced like the original Lolin32) and the TTGO T7 v1.3 and v1.4 boards (status unknown). All three are rather generic ESP32 boards the main difference being pin layouts and ESP32 module type.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -700,6 +700,238 @@ ttgo-t1.menu.DebugLevel.verbose.build.code_debug=5
 
 ##############################################################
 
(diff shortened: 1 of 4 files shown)


me-no-dev:
Thanks :)

RA-NGR:
TTGO T7 v1.4 based on WROVER module, so it must has PSRAM, please correct

Issue #3193, opened 2019-09-09
WiFiClientSecure::write only sends the first 16KB


marcelstoer:
### Description:
Shouldn't [`WiFiClientSecure::write`](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L183) send the whole buffer even beyond SSL fragment length? Currently it only writes up to mbedTLS maximum fragment length which is 16KB by default.

According to [ARM mbed TLS](https://tls.mbed.org/api/ssl_8h.html#a5bbda87d484de82df730758b475f32e5) `mbedtls_ssl_write` has the following limitation

> If the requested length is greater than the maximum fragment length (either the built-in limit or the one set or negotiated with the peer), then:
>- with TLS, less bytes than requested are written.

It appears that `mbedtls_ssl_write` is only called once at https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/ssl_client.cpp#L280

### Sketch
While the (simplified) workaround below works ok it does feel like such a use case should be handled by `WiFiClientSecure` internally. Opinions?

```cpp
size_t sent = 0; 
size_t remaining = fileSize; 
size_t offset = 0;

while (true) {
  sent = client->write(buffer + offset, remaining);
  remaining = remaining - sent;
  offset += sent;
  if (remaining <= 0 || sent <= 0) {
    break;
  }
}
```

marcelstoer:
@me-no-dev according to the commit history you might be familiar with the subject?

marcelstoer:
@me-no-dev what can I do to get some feedback from a developer to find out if my expectations are justified or not?

me-no-dev:
This should definitelly not be the case and I will put it on my list of things to check. Doing such while loops in FreeRTOS might not be a good idea, but I need to have a better look at what goes underneath to know if this can do or not.

marcelstoer:
"Hey bot, shut up. The grown-ups are talking." Not frequently but they do 😜

marcelstoer:
nudge

atanisoft:
@marcelstoer perhaps submit a PR with your proposal?

marcelstoer:
Re: PR; Normally not an issue but I'm not familiar with the inner workings of the affected class(es) to make a meaningful proposal.

Besides @me-no-dev stated that

> Doing such while loops in FreeRTOS might not be a good idea

[...]

PR #4820, opened 2021-02-17
Fix dropped SSL connection when buffer gets full.

everslick:
mbedTLS requires repeated calls to mbedtls_ssl_write() whenever it returns MBEDTLS_ERR_SSL_WANT_READ or MBEDTLS_ERR_SSL_WANT_WRITE. this happens when the client sends data faster then the server or the connection can handle.

### DIFF
--- a/libraries/WiFiClientSecure/src/ssl_client.cpp
+++ b/libraries/WiFiClientSecure/src/ssl_client.cpp
@@ -300,7 +300,8 @@ int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, uint16_t l
             log_v("Handling error %d", ret); //for low level debug
             return handle_error(ret);
         }
-	vTaskDelay(2);
+        //wait for space to become available
+        vTaskDelay(2);
     }
 
     return ret;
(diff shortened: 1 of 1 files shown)


libraries/WiFiClientSecure/src/ssl_client.cpp
```diff
@@ -295,11 +295,11 @@ int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, uint16_t l
     log_v("Writing HTTP request with %d bytes...", len); //for low level debug
     int ret = -1;
 
-    if ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0){
-        log_v("Handling error %d", ret); //for low level debug
-        return handle_error(ret);
-    } else{
-        log_v("Returning with %d bytes written", ret); //for low level debug
+    while ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0) {
```
me-no-dev, line 298: I left a note on this part in the 2.0.0 PR :) could you please have a look?

everslick: done both! :)

marcelstoer:
This likely also fixed the (largely) ignored #3193.

Issue #2494, opened 2019-02-22
WiFiClientSecure gets stuck, when send buffers get full


everslick:
When we send too fast (faster then the TLS connection allows), eventually the send buffers will fill up and write() returns 0 (indicating, that nothing could be written). Normally this means we just retry until all data is sent, but in this case the TLS socket never recovers from this condition (buffers stay full). The only thing we can do is close the socket and go home.

everslick:
still valid, don't close.

everslick:
https://github.com/esp8266/Arduino/pull/6104

could be the same bug as on ESP8266. @me-no-dev ?

everslick:
still relevant, don't close.

everslick:
still relevant, don't close.

everslick:
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L183

My guess is, when `send_ssl_data()` returns an error we silently close the socket, and do not inform the caller. write should return int, but returns size_t.

everslick:
@me-no-dev what do you think?

everslick:
Doing more tests with idf-release/v4.2 shows the bug still present. So this could very easily be a issue with WiFiClientSecure. It hink this is serious and worth looking into before next stable release. What can I do to bring this foward besides the what I commented above?

sansillusion:
wow I feel bad for you, not even a single aknowledgment of you issue and you have to fight the stale bot to keep it open. I hope you get the attention you need on this issue, unfortunately I can't help.

PR #4821, opened 2021-02-17
[2.0.0] Fix dropped SSL connection when buffer gets full.

everslick:
mbedTLS requires repeated calls to mbedtls_ssl_write() whenever it returns MBEDTLS_ERR_SSL_WANT_READ or MBEDTLS_ERR_SSL_WANT_WRITE. this happens when the client sends data faster then the server or the connection can handle.

Fixes: https://github.com/espressif/arduino-esp32/issues/2494

### DIFF
--- a/libraries/WiFiClientSecure/src/ssl_client.cpp
+++ b/libraries/WiFiClientSecure/src/ssl_client.cpp
@@ -294,6 +294,7 @@ int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, uint16_t l
             log_v("Handling error %d", ret); //for low level debug
             return handle_error(ret);
         }
+	vTaskDelay(2);
     }
 
     return ret;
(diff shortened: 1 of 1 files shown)


libraries/WiFiClientSecure/src/ssl_client.cpp
```diff
@@ -289,11 +289,11 @@ int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, uint16_t l
     log_v("Writing HTTP request with %d bytes...", len); //for low level debug
     int ret = -1;
 
-    if ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0){
-        log_v("Handling error %d", ret); //for low level debug
-        return handle_error(ret);
-    } else{
-        log_v("Returning with %d bytes written", ret); //for low level debug
+    while ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0) {
```
me-no-dev, line 292: I am a bit worried about such busy `while` loop. Add at least `vTaskDelay(2);` to ensure that task switching will work

PR #4823, opened 2021-02-17
[2.0.0] Remove temp buffer in lastError

everslick:
The temp buffer serves no purpose here. As a side-note mbedtls_strerror can be called with size == 0 safely. No need to check.


me-no-dev:
already in :)

PR #4832, opened 2021-02-18
[2.0.0] Add BLE characteristic callbacks overloads

mixa3607:
Add BLE characteristic callbacks overloads with esp_ble_gatts_cb_param_t* param.
Example:
```
class BleCharactCallback : public BLECharacteristicCallbacks
{
    void onRead(BLECharacteristic *pCharacteristic, esp_ble_gatts_cb_param_t *param)
    {
        auto addr = param->read.bda;
        ESP_LOGV(TAG, "Device " ESP_BD_ADDR_STR " request data", ESP_BD_ADDR_HEX(addr));
    }
    void onWrite(BLECharacteristic *pCharacteristic, esp_ble_gatts_cb_param_t *param)
    {
        auto addr = param->write.bda;
        ESP_LOGV(TAG, "Device " ESP_BD_ADDR_STR " transmit data", ESP_BD_ADDR_HEX(addr));
    }
};
```


me-no-dev:
Hi @mixa3607 :) Thanks for the PR!
I have a few things that do bother me a bit though, like calling both functions one after another.
Here is what I propose:
1. Mark the callbacks without param as deprecated
2. Call the callbacks without param from the default handlers with param
3. Remove logging from the default default handlers with param

What do you think?

mixa3607:
Hi @me-no-dev!
Marked overloads with single arg as "DEPRECATED" only in doc comments, and didn't add `__attribute__ ((deprecated))` to the definition.
Moved call `onXXX(pCharact)` to `onXXX(pCharact, param)`.
And also fix logging messages like `log_d("BLECharacteristicCallbacks", "<< onStatus");` by change to `log_d("<< onStatus");`

PR #4839, opened 2021-02-19
Added BLEAddress operator overload methods

Bascy:
Allows BLEAddress to be used as key in std::map etc


me-no-dev:
Nice! Thanks!

PR #4845, opened 2021-02-21
Added possibility to use ESP32-IDF log insted of redefined one

martirius:
With this PR user can select to use the original ESP-IDF log instead of the redefined one.

User can also redefine the log function as per [Logging Library](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/log.html#_CPPv419esp_log_set_vprintf14vprintf_like_t) so he can for example redirect logs to a file.

To enable this change just add -DUSE_ESP32_LOG to build flags.
User can also change the default TAG (that now is ES32) to whatever it wants adding '-DTAG="tag_value"' to build flags

### DIFF
--- a/cores/esp32/esp32-hal-log.c
+++ b/cores/esp32/esp32-hal-log.c
@@ -2,41 +2,18 @@
 #define __MY_LOG__
 #include "stdio.h"
 #include "esp32-hal-log.h"
-#ifdef USE_ESP32_LOG
-    void log_to_esp(esp_log_level_t level, const char *format, ...)
-    {   
-        va_list va_args;
-        va_start(va_args, format);
+void log_to_esp(char* tag, esp_log_level_t level, const char *format, ...)
+{
+    va_list va_args;
+    va_start(va_args, format);
 
-        char log_buffer[512];
-        int len = vsnprintf(log_buffer, sizeof(log_buffer), format, va_args);
-        if (len > 0)
-        {
-            switch (level)
-            {
-            case ESP_LOG_ERROR:
-                ESP_LOGE(TAG, "%s", log_buffer);
-                break;
-            case ESP_LOG_DEBUG:
-                ESP_LOGD(TAG, "%s", log_buffer);
-                break;
-            case ESP_LOG_WARN:
-                ESP_LOGW(TAG, "%s", log_buffer);
-                break;
-            case ESP_LOG_INFO:
-                ESP_LOGI(TAG, "%s", log_buffer);
-                break;
-            case ESP_LOG_VERBOSE:
-                ESP_LOGV(TAG, "%s", log_buffer);
-                break;
-            case ESP_LOG_NONE:
-                //do nothing
-                break;
-            }
-        }
-
-        va_end(va_args);
+    char log_buffer[512];
+    int len = vsnprintf(log_buffer, sizeof(log_buffer), format, va_args);
+    if (len > 0)
+    {
+        ESP_LOG_LEVEL_LOCAL(level, tag, "%s", log_buffer);
     }
-#endif
-#endif
 
+    va_end(va_args);
+}
+#endif
--- a/cores/esp32/esp32-hal-log.c
+++ b/cores/esp32/esp32-hal-log.c
@@ -1,0 +1,42 @@
+#ifndef __MY_LOG__
+#define __MY_LOG__
+#include "stdio.h"
+#include "esp32-hal-log.h"
+#ifdef USE_ESP32_LOG
+    void log_to_esp(esp_log_level_t level, const char *format, ...)
+    {   
+        va_list va_args;
+        va_start(va_args, format);
+
+        char log_buffer[512];
+        int len = vsnprintf(log_buffer, sizeof(log_buffer), format, va_args);
+        if (len > 0)
+        {
+            switch (level)
+            {
+            case ESP_LOG_ERROR:
+                ESP_LOGE(TAG, "%s", log_buffer);
+                break;
+            case ESP_LOG_DEBUG:
+                ESP_LOGD(TAG, "%s", log_buffer);
+                break;
+            case ESP_LOG_WARN:
+                ESP_LOGW(TAG, "%s", log_buffer);
+                break;
+            case ESP_LOG_INFO:
+                ESP_LOGI(TAG, "%s", log_buffer);
+                break;
+            case ESP_LOG_VERBOSE:
+                ESP_LOGV(TAG, "%s", log_buffer);
+                break;
+            case ESP_LOG_NONE:
+                //do nothing
+                break;
+            }
+        }
+
+        va_end(va_args);
+    }
+#endif
+#endif
+
--- a/cores/esp32/esp32-hal-log.h
+++ b/cores/esp32/esp32-hal-log.h
@@ -1,4 +1,4 @@
-// Copyright 2015-2021 Espressif Systems (Shanghai) PTE LTD
+// Copyright 2015-2016 Espressif Systems (Shanghai) PTE LTD
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


cores/esp32/esp32-hal-log.h
```diff
@@ -125,9 +130,24 @@ int log_printf(const char *fmt, ...);
...
 #define isr_log_n(format, ...)
 #endif
+#endif
 
 #include "esp_log.h"
 
+#ifdef USE_ESP32_LOG
+
+#ifndef TAG
+#define TAG "ESP32"
```
me-no-dev, line 140: using TAG like "arduino" makes more sense here

cores/esp32/esp32-hal-log.h
```diff
@@ -37,6 +36,9 @@ extern "C"
 #define ARDUHAL_LOG_LEVEL CONFIG_ARDUHAL_LOG_DEFAULT_LEVEL
 #else
 #define ARDUHAL_LOG_LEVEL CORE_DEBUG_LEVEL
+#ifdef USE_ESP32_LOG
```
me-no-dev, line 39: I think that renaming the name to `USE_ESP_IDF_LOG` would describe a bit better what the switch does :)

cores/esp32/esp32-hal-log.h
martirius: No problem , i will change it for sure :)

cores/esp32/esp32-hal-log.h
martirius: You are right, let me change it

cores/esp32/esp32-hal-log.c
```diff
@@ -0,0 +1,42 @@
+#ifndef __MY_LOG__
+#define __MY_LOG__
+#include "stdio.h"
+#include "esp32-hal-log.h"
+#ifdef USE_ESP32_LOG
```
me-no-dev, line 5: you can remove the `#ifdef` switch. Function will not be included in the final binary if not called within the application and will at the same time allow `USE_ESP32_LOG` to be defined on per file basis.

cores/esp32/esp32-hal-log.c
```diff
@@ -0,0 +1,42 @@
...
+    void log_to_esp(esp_log_level_t level, const char *format, ...)
+    {   
+        va_list va_args;
+        va_start(va_args, format);
+
+        char log_buffer[512];
+        int len = vsnprintf(log_buffer, sizeof(log_buffer), format, va_args);
+        if (len > 0)
+        {
+            switch (level)
```
me-no-dev, line 15: this switch can be replaced by the following line:
```cpp
ESP_LOG_LEVEL_LOCAL(level, tag, "%s", log_buffer);
```

cores/esp32/esp32-hal-log.c
```diff
@@ -0,0 +1,42 @@
+#ifndef __MY_LOG__
+#define __MY_LOG__
+#include "stdio.h"
+#include "esp32-hal-log.h"
+#ifdef USE_ESP32_LOG
+    void log_to_esp(esp_log_level_t level, const char *format, ...)
```
me-no-dev, line 6: please pass also the TAG as argument:
```cpp
void log_to_esp(esp_log_level_t level, const char *tag, const char *format, ...)
```

cores/esp32/esp32-hal-log.c
martirius: It was my decision to doesn't call directly ESP_LOG_LEVEL_LOCAL, instead using the provided MACROs.
Bu i agree that the switch it's resolved in this way.

me-no-dev: by the way, the following should also work:
```cpp
ESP_LOG_LEVEL_LOCAL(level, tag, (const char *)log_buffer);
```

cores/esp32/esp32-hal-log.c
martirius: Here i wanted to replicate the actual behavior of the arduino library, so the tag is fixed.
But I understand it's better to put it as a parameter ( and passing it fixed when calling MACROs)

cores/esp32/esp32-hal-log.c
martirius: Done! :)

cores/esp32/esp32-hal-log.c
martirius: No, in this way the compiler says :
` expression preceding parentheses of apparent call must have (pointer-to-) function typeC/C++(109)`

me-no-dev: ok :)

martirius:
@me-no-dev I have noticed a strange behaviour; if i enable the esp-idf logs and lower the define of CORE_DEBUG_LEVEL to a value under DEBUG (4), the code doesn't compile anymore and this message is shown:
`.platformio/packages/framework-arduinoespressif32/libraries/WiFi/src/WiFiSTA.cpp:738:25: error: 'sc_status_strings' was not declared in this scope`.
However if I don't enable my defined log_* and use the arduino log, the code compile.

I checked that line that raises the error but i cannot understand why it compilessince it's calling an undefined variable

me-no-dev:
Because you also need to make the defines based on the log level selected:
```cpp
#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_ERROR
#define log_e(format, ...) do {log_to_esp(ESP_LOG_ERROR, TAG, format, ##__VA_ARGS__);}while(0)
#define isr_log_e(format, ...) do {ets_printf(LOG_FORMAT(E, format), esp_log_timestamp(), tag, ##__VA_ARGS__);}while(0)
#else
#define log_e(format, ...)
#define isr_log_e(format, ...)
#endif
```

jfbuggen:
Many thanks for developing the link with the ESP IDF log.

I have been looking at the code (as I had a buggy implementation of my own vprintf to install via esp_log_set_vprintf...) and was wondering why it is necessary to create the log_to_esp function in esp32-hal-log.c, instead if directly referring to ESP_LOG_LEVEL_LOCAL (which ultimately calls esp_log_write) in the definition of log_x macros in esp32-hal-log.h like this:

#define log_n(format, ...) ESP_LOG_LEVEL_LOCAL(ESP_LOG_ERROR, TAG, format, ##__VA_ARGS__)

I made some tests and it looks to work very fine like that.

It spares a function call, the 512b buffer of log_to_esp (esp_log_write in log.c is more prudent with a 64b buffer and malloc if bigger size is necessary) and a call to vsnprintf (esp_log_write in log.c already calls it twice).

martirius:
Hi @jfbuggen, thanks for appreciating.
I'm honest, i don't remember why i took this decision, however you are free to made a PR to with your proposed solution :)

lbernstone:
If your method makes it easier to redirect with esp_log_set_vprintf, then by all means submit it.

jfbuggen:
Thank you @martirius. I will have to learn how to do that on GitHub but I will have a try.

@lbernstone it will not change much in terms of functionality, esp_log_set_vprintf works fine already with the current code, it brings just a small performance improvement.

cyberman54:
Even after this merge esp32-idf logging is still not usable, because of lots of conflicts with TAG macro in libraries, e.g.

```
packages/framework-arduinoespressif32/libraries/WiFiClientSecure/src/esp_crt_bundle.c:25:20: note: in expansion of macro 'TAG'
 static const char *TAG = "esp-x509-crt-bundle";
                    ^~~
```
/cc @coratron @lbernstone @martirius

zekageri:
Any news on this?

VojtechBartoska:
Hello @zekageri and @cyberman54, if this implementation doesn't work properly, please consider opening a new issue with behavior description. Thanks.

zekageri:
Thank you for your response. I will test it soon.

PR #4853, opened 2021-02-23
Add div by zero check back into WMath::map

lbernstone:
The check for divisor==0 was removed in last commit.  This is still necessary.


me-no-dev:
CI is not happy :) https://github.com/espressif/arduino-esp32/pull/4853/checks?check_run_id=1964101781#step:4:38

Also... how would someone end up in situation where min and max are equal?

lbernstone:
> Also... how would someone end up in situation where min and max are equal?

Clearly an error situation, but it should not cause a crash.  Let's say you are calibrating something, and you stick the sensor in the low level solution twice instead of the high.  You should get an error, without the div by zero crash.

PR #4867, opened 2021-02-27
Update pins_arduino.h

ZCB1208:
Fix the bug that the IDE show: "Error compiling for the board FireBettle-ESP32" 
Detial"error : redefinition of const uint8_t D0~D9"

### DIFF
--- a/variants/firebeetle32/pins_arduino.h
+++ b/variants/firebeetle32/pins_arduino.h
@@ -59,16 +59,16 @@ static const uint8_t A14 = 26;
 
 
 
-static const uint8_t D0 = 3;
-static const uint8_t D1 = 1;
-static const uint8_t D2 = 25;
-static const uint8_t D3 = 26;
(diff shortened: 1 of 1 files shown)


variants/firebeetle32/pins_arduino.h
```diff
@@ -59,16 +59,16 @@ static const uint8_t A14 = 26;
...
+// static const uint8_t D0 = 3;
+// static const uint8_t D1 = 1;
+// static const uint8_t D2 = 25;
+// static const uint8_t D3 = 26;
+// static const uint8_t D4 = 27;
+// static const uint8_t D5 = 9;
+// static const uint8_t D6 = 10;
+// static const uint8_t D7 = 13;
+// static const uint8_t D8 = 5;
+// static const uint8_t D9 = 2;
```
me-no-dev, line 71: why not remove the lines and add `D10` up with the other definitions?

Issue #4640, opened 2020-12-17
Possible issue in cacheing files with spaces in them


nitko12:
### Hardware:
Board:							https://github.com/e-radionicacom/Inkplate-6-hardware
Core Installation version:			1.0.5 RC4
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                    yes
Upload Speed:					115200
Computer OS:                                        Mac OSX Big Sur

### Description:
Hi there,

We've been working on a library using your Arduino board definition.

We are using an ESP32 board as specified here. 
https://github.com/e-radionicacom/Inkplate-6-hardware

For some reason the bash script specified in platform.txt fails when theres a space in the file name.

We've diagnosed the issue to be a non escaped space here which causes cp command to fail.
https://github.com/espressif/arduino-esp32/blob/b05bdf690418efd28c8468b5cb8345d95ba77e9f/platform.txt#L69

You can try uploading any sketch as long as there's a space in the name to recreate.

Here's the error:
```
[...]

Issue #4579, opened 2020-11-25
Spaces in path causes compilation error


jpasqua:
### Hardware:
Board:							ESP32 D1 Mini
Core Installation version:			1.0.5 rc2
IDE name:						Arduino IDE 1.8.13
Flash Frequency:					80Mhz
PSRAM enabled:                                    N/A
Upload Speed:					921600
Computer OS:                                       macOS 10.15.7

### Description:
I'm trying to compile a project in a directory that contains spaces in the path:
`/Users/joe/Dropbox/Dev/Arduino/libraries/TFT_eSPI-master/examples/320 x 240/TFT_Meters`
This is a standard example that comes with the TFT_eSPI library and works fine with other cores like the ESP8266. Removing all spaces from the path causes the error to go away.

### Sketch:  [Any sketch will fail]
```cpp

void setup() {}
void loop() {}
```

### Debug Messages:
```
bash -c "[ ! -f /Users/joe/Dropbox/Dev/Arduino/libraries/TFT_eSPI-master/examples/320 x 240/TFT_Meters/partitions.csv ] || cp -f /Users/joe/Dropbox/Dev/Arduino/libraries/TFT_eSPI-master/examples/320 x 240/TFT_Meters/partitions.csv /var/folders/97/nw96g6nn0k33z212yw4by4yr0000gn/T/arduino_build_953387/partitions.csv"
bash: line 0: [: too many arguments
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory
exit status 64
```

Gius-8:
Same problem on Mac OS Big Sur (11.1)

coelner:
https://github.com/espressif/arduino-esp32/issues/4640#issuecomment-748051578

PR #4868, opened 2021-02-28
Quote {build.source.path} to allow spaces in path

tttapa:
Previously sketches or examples that had spaces anywhere in their absolute path caused a total build failure. By adding quotes around the path in `platform.txt`, they now build correctly.

Consider a “bare minimum” sketch in `/home/pieter/Arduino/folder with spaces/Sketch/Sketch.ino`. Previously, the bash command generated for copying the `partitions.csv` file was:
```sh
bash -c "[ ! -f /home/pieter/Arduino/folder with spaces/Sketch/partitions.csv ] || cp -f /home/pieter/Arduino/folder with spaces/Sketch/partitions.csv /tmp/arduino_build_324565/partitions.csv"
bash: line 0: [: too many arguments
cp: target '/tmp/arduino_build_324565/partitions.csv' is not a directory
exit status 1
```
With the proposed changes, the path in the bash command is now correctly quoted, and the build succeeds:
```sh
bash -c "[ ! -f \"/home/pieter/Arduino/folder with spaces/Sketch\"/partitions.csv ] || cp -f \"/home/pieter/Arduino/folder with spaces/Sketch\"/partitions.csv /tmp/arduino_build_324565/partitions.csv"
```

This pull request addresses issues #4579 and #4640.

People use spaces in folder names, and this works correctly in version 1.0.4 of this ESP32 core and all other Arduino cores I've tried. It no longer works in version 1.0.5 (after commit 9ef3e2d2a6b79c951f894dba1dd71c811317ca6e), so I believe this is well worth fixing, especially since it's such a simple change.


me-no-dev:
quotation does not look alright. this needs testing :)

tttapa:
The confusing thing about the quotation is that the Arduino IDE seems to handle quotes differently when they are next to a placeholder:

```
"outer string "inner string {placeholder} more inner string" more outer string"
```
gets converted into 
```
"outer string "inner string value more inner string" more outer string"
```
which is obviously incorrect. On the other hand,
```
"outer string "{placeholder}" more outer string"
```
becomes 
```
"outer string \"value\" more outer string"
```
which is correct in this case.

This is kind of uncommon, though, and I couldn't find any mention of this behavior in the [platform specification docs](https://arduino.github.io/arduino-cli/platform-specification/).

I've tested this using the Arduino IDE v1.8.9 and 1.8.13 on Ubuntu 20.04, as well as with the `arduino-cli` directly for my CI builds and it works correctly. The bash commands it generates look correct to me (enable verbose output in the preferences to see them, the output is also in my previous message) and the `partitions.csv` file gets copied to the build folder if it exists, no matter if the sketch path contains spaces or not.

me-no-dev:
what if we pre-escape the whole paths so that `partitions.csv` is also escaped. I also need to test this with Sloeber and make sure that paths expand as they should

tttapa:
Why does `partitions.csv` need to be escaped? It doesn't contain any spaces or special characters.

Bash should handle this correctly, multiple strings (quoted or not quoted) that are right next to each other are combined into a single string.

For example:
```
$ echo "quoted&:'string"suffix
> quoted&:'stringsuffix
$ ls "$HOME"/.profile
> /home/pieter/.profile
```

me-no-dev:
because the commands are used by other builders and IDEs. It makes most sense that the full path is escaped, not just part of it.

PR #4879, opened 2021-03-01
Fixed use of Bluedroid instead of BT in HAL.

AronRubin:
Fixed use of CONFIG_BLUEDROID_ENABLED instead of CONFIG_BT_ENABLED in HAL. This prevented compilation with Nimble-only configuration without apparent benefit.


me-no-dev:
thanks :)

PR #4883, opened 2021-03-02
Back out duplicate definitions from #4220

lbernstone:
D[0-9] pins had duplicate definitions.  #4220 was invalid.


me-no-dev:
@lbernstone same as in: https://github.com/espressif/arduino-esp32/pull/4867 ?

lbernstone:
Apparently so.

PR #4891, opened 2021-03-04
Remove tools/sdk/include/nimble from include path

maxgerhardt:
For the PlatformIO build script. 

Per https://github.com/espressif/arduino-esp32/tree/master/tools/sdk/include this folder doesn't exist exist anymore, so users will see a warning in VSCode regarding this non-existent folder (see [here](https://community.platformio.org/t/fatal-error-pins-arduino-h-no-such-file-or-directory/19634/7?u=maxgerhardt) for example).

I don't know whether nimble has been removed or renamed, but this change just removes the reference to the folder in the platformio build script.

It looks like it's also gone from other branches (v4.0, v4.2), so.. it should also be merged there I guess.

### DIFF
--- a/tools/platformio-build.py
+++ b/tools/platformio-build.py
@@ -138,7 +138,6 @@ env.Append(
         join(FRAMEWORK_DIR, "tools", "sdk", "include", "mqtt"),
         join(FRAMEWORK_DIR, "tools", "sdk", "include", "newlib"),
         join(FRAMEWORK_DIR, "tools", "sdk", "include", "nghttp"),
-        join(FRAMEWORK_DIR, "tools", "sdk", "include", "nimble"),
         join(FRAMEWORK_DIR, "tools", "sdk", "include", "nvs_flash"),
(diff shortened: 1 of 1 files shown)


me-no-dev:
this is done automatically. will accept the PR now and fix it in the builder

maxgerhardt:
Thanks - I see the commit at https://github.com/espressif/esp32-arduino-lib-builder/commit/45370ab77a9e87f6e0c5f53414eecc34a9a18e09

Just to reiterate: It is correct that arduino-esp32 doesn't contain the `nimble` include folder anymore? If the builder script iterates through all built components that end up in Arduino-ESP32, it's kinda weird how there's only exception.

me-no-dev:
that is an old script to work with the `make` build system for IDF. The 4.2 branch uses a different approach that actually extracts the build parameters from the `cmake` system itself. Nimble should not be there because it's not built with Arduino (bluedroid is)

PR #4907, opened 2021-03-07
Update debian_ubuntu.md

serhiidedik:
ERROR: This script does not work on Python 2.7 The minimum supported Python version is 3.6. Please use https://bootstrap.pypa.io/2.7/get-pip.py instead.


chegewara:
python 2.7 is end of life, so what is purpose to use pip for 2.7?

me-no-dev:
maybe something like:
```bash
wget https://bootstrap.pypa.io/get-pip.py && \
sudo python3 get-pip.py && \
sudo pip3 install pyserial
```

PR #4942, opened 2021-03-14
Update HTTPUpdate to add setFollowRedirects

DTTerastar:
This is a mirror of a change in esp8266 needed to update with a url that redirects.

### DIFF
--- a/libraries/HTTPUpdate/src/HTTPUpdate.h
+++ b/libraries/HTTPUpdate/src/HTTPUpdate.h
@@ -63,6 +63,15 @@ public:
     {
         _rebootOnUpdate = reboot;
     }
+    
+    /**
+      * set redirect follow mode. See `followRedirects_t` enum for avaliable modes.
+      * @param follow
+      */
+    void setFollowRedirects(followRedirects_t follow)
+    {
+        _followRedirects = follow;
+    }
 
     void setLedPin(int ledPin = -1, uint8_t ledOn = HIGH)
     {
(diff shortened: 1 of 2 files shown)


lbernstone:
Already merged
https://github.com/espressif/arduino-esp32/commit/ee88c42c3b5e87a2fc09303f9e82dd8172e99990

DTTerastar:
I don't believe you looked at this close enough.  It is a change to HTTPUpdate NOT to HTTPClient.  It is not already merged and is a case where the esp8266 has better functionality than the esp32.

libraries/HTTPUpdate/src/HTTPUpdate.h
```diff
@@ -89,6 +98,7 @@ class HTTPUpdate
     bool _rebootOnUpdate = true;
 private:
     int _httpClientTimeout;
+    followRedirects_t _followRedirects = HTTPC_DISABLE_FOLLOW_REDIRECTS;
```
me-no-dev, line 101: default value should be set in the class constructors instead.
```cpp
HTTPUpdate::HTTPUpdate(void)
        : _httpClientTimeout(8000), _followRedirects(HTTPC_DISABLE_FOLLOW_REDIRECTS), _ledPin(-1)
{
}
HTTPUpdate::HTTPUpdate(int httpClientTimeout)
        : _httpClientTimeout(httpClientTimeout), _followRedirects(HTTPC_DISABLE_FOLLOW_REDIRECTS), _ledPin(-1)
{
}
```

DTTerastar: Requested changes made.

me-no-dev:
Thanks :)

PR #4954, opened 2021-03-16
Allows you to specify which channels are used to advertise.

raulcigil:
In some use cases getting rssi signal from one channel is more stable (less variance) than rssi from the three advertising channels. 

This change allows you to specify which channels are used to advertise.

### DIFF
--- a/libraries/BLE/src/BLEAdvertising.cpp
+++ b/libraries/BLE/src/BLEAdvertising.cpp
@@ -91,6 +91,10 @@ void BLEAdvertising::setAdvertisementType(esp_ble_adv_type_t adv_type){
 	m_advParams.adv_type = adv_type;
 } // setAdvertisementType
 
+void BLEAdvertising::setAdvertisementChannelMap(esp_ble_adv_channel_t channel_map) {
+	m_advParams.channel_map = channel_map;
+} // setAdvertisementChannelMap
+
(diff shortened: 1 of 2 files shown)


me-no-dev:
thanks :)

selman-nus:
an example code would be very nice to understand @raulcigil

raulcigil:
Good idea.
Using BleServer.ino from examples:

```
BLEAdvertising *pAdvertising = BLEDevice::getAdvertising();
  pAdvertising->addServiceUUID(SERVICE_UUID);
  pAdvertising->setScanResponse(true);
  pAdvertising->setMinPreferred(0x06);  // functions that help with iPhone connections issue
  pAdvertising->setMaxPreferred(0x12);

  //Setting the advertising channel
  pAdvertising->setAdvertisementChannelMap(ADV_CHNL_37);
  
  BLEDevice::startAdvertising();
```
You can set the channel with this enumeration:

[https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/bluetooth/esp_gap_ble.html#_CPPv421esp_ble_adv_channel_t](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/bluetooth/esp_gap_ble.html#_CPPv421esp_ble_adv_channel_t)

PR #4958, opened 2021-03-17
Update getLocalTime to a more reasonable range

aovestdipaperino:
It seems like getLocalTime is expecting a valid date to be in the year range 1900..2016
The problem is the retry logic, which defaults to 5000 msec.
So in the meanwhile at least the right range should be fixed.

### DIFF
--- a/cores/esp32/esp32-hal-time.c
+++ b/cores/esp32/esp32-hal-time.c
@@ -84,7 +84,7 @@ bool getLocalTime(struct tm * info, uint32_t ms)
     while((millis()-start) <= ms) {
         time(&now);
         localtime_r(&now, info);
-        if(info->tm_year > (2016 - 1900)){
+        if(info->tm_year > (2030 - 1900)){
             return true;
         }
         delay(10);


me-no-dev:
it is actually expecting the valid year to be after 2016. Could you please provide more info on the problem?

me-no-dev:
To be more clear, `info->tm_year` has the value `years since 1900`, so `if(info->tm_year > (2016 - 1900))` checks if he returned year is after 2016 (which was the year before the one I wrote this code in, thus could not return any earlier year).

I will close this PR now, because your change checks for 2030, which is 9 years ahead :) Feel free to comment here and explain the issue that you are having, so that we can look into it's resolution :)

derLalla:
I'd like to reopen this case. So when the time I want to set is below 2016, the call takes 5 seconds and returns false. Why is this behavior needed? 
So in my case, I am either using the UTC time which I am getting from LTE or the internal time of the ESP which defaults to 1970 in case the LTE is not working. 

This means for the not available UTC time, I have to wait 5 seconds to get the current time.

PR #4959, opened 2021-03-18
Allow passing custom HTTPClient to HTTPUpdate

MohammedNoureldin:
This enables customizing HTTP headers which adds some extra flexibility.
This does not break anything of course because this change introduces a new constructor with a new additional HTTPClient argument for HTTPUpdate class.


me-no-dev:
Wouldn't it make sense to just add something like `t_httpUpdate_return update(HTTPClient& http, const String& currentVersion);` and `t_httpUpdate_return updateSpiffs(HTTPClient& http, const String& currentVersion);` which would call `handleUpdate` directly and not pass a WiFiClient (which might not be necessary when initializing the HTTPClient)?

MohammedNoureldin:
Well, why not. I pushed this edit as you suggested.

libraries/HTTPUpdate/src/HTTPUpdate.cpp
```diff
@@ -58,6 +58,11 @@ HTTPUpdateResult HTTPUpdate::update(WiFiClient& client, const String& url, const
     return handleUpdate(http, currentVersion, false);
 }
 
+HTTPUpdateResult HTTPUpdate::updateSpiffs(HTTPClient& httpClient, const String& url, const String& currentVersion)
```
me-no-dev, line 61: Does not need the `url` argument :)

libraries/HTTPUpdate/src/HTTPUpdate.h
```diff
@@ -86,6 +86,10 @@ class HTTPUpdate
 
     t_httpUpdate_return updateSpiffs(WiFiClient& client, const String& url, const String& currentVersion = "");
 
+    t_httpUpdate_return update(HTTPClient& httpClient, const String& host, uint16_t port, const String& uri = "/",
+                               const String& currentVersion = "");
+
+    t_httpUpdate_return updateSpiffs(HTTPClient &httpClient, const String &url, const String &currentVersion = "");
```
me-no-dev, line 92: Does not need the `url` argument :)

libraries/HTTPUpdate/src/HTTPUpdate.cpp
```diff
@@ -68,6 +73,12 @@ HTTPUpdateResult HTTPUpdate::updateSpiffs(WiFiClient& client, const String& url,
     return handleUpdate(http, currentVersion, true);
 }
 
+HTTPUpdateResult HTTPUpdate::update(HTTPClient& httpClient, const String& host, uint16_t port, const String& uri,
```
me-no-dev, line 76: Does not need the `host`, `port` and `uri` arguments :)

libraries/HTTPUpdate/src/HTTPUpdate.h
```diff
@@ -86,6 +86,10 @@ class HTTPUpdate
 
     t_httpUpdate_return updateSpiffs(WiFiClient& client, const String& url, const String& currentVersion = "");
 
+    t_httpUpdate_return update(HTTPClient& httpClient, const String& host, uint16_t port, const String& uri = "/",
```
me-no-dev, line 89: Does not need the `host`, `port` and `uri` arguments :)

me-no-dev:
you forgot to remove some arguments there :)

MohammedNoureldin:
@me-no-dev oh sorry! I did that it in parallel with other work. I removed these redundant arguments. This should be fine now.

me-no-dev:
Thanks :) all good now!

PR #4977, opened 2021-03-26
Update platform.txt

fdivitto:
This change avoids the error "bash: line 0: [: ------: binary operator expected" when the source path contains spaces (Linux, Mac)


me-no-dev:
More or less dup of https://github.com/espressif/arduino-esp32/pull/4868 :)

PR #4983, opened 2021-03-28
FIX max SSID length is 32

dzsoni:
Same bug was in the WiFiSTA.cpp.
Maybe a definition file could contian these standard related boundaries.

### DIFF
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -50,7 +50,7 @@ bool WiFiMulti::addAP(const char* ssid, const char *passphrase)
 {
     WifiAPlist_t newAP;
 
-    if(!ssid || *ssid == 0x00 || strlen(ssid) > 31) {
+    if(!ssid || *ssid == 0x00 || strlen(ssid) > 32) {
         // fail SSID too long or missing!
         log_e("[WIFI][APlistAdd] no ssid or ssid too long");
         return false;


lbernstone:
Dupe https://github.com/espressif/arduino-esp32/commit/e7a2759b65e1bf41227aad46df57425a92dadc50

lbernstone:
This is not as easy as WiFiSTA.  You will need to add the ssid_len parameter.  Otherwise it assumes the string is null terminated.

me-no-dev:
@lbernstone could you please have a second look? Seems like it's a proper fix/change. SSID will later be passed to WiFi.begin, which will accept 32 chars (NULL terminated string of max strlen of 32)

lbernstone:
It is certainly possible I am misinterpreting.  I didn't test to see, just reading the def.
https://github.com/espressif/esp-idf/blob/release/v3.3/components/esp32/include/esp_wifi_types.h#L212-L214
```
    uint8_t ssid[32];           /**< SSID of ESP32 soft-AP */
    uint8_t password[64];       /**< Password of ESP32 soft-AP */
    uint8_t ssid_len;           /**< Length of SSID. If softap_config.ssid_len==0, check the SSID until there is a termination character; otherwise, set the SSID length according to softap_config.ssid_len. */
```

dzsoni:
When I wrote that same bug was few days before in the WiFiSTA.cpp a thougt to this:
[https://github.com/espressif/arduino-esp32/commit/e7a2759b](url)

We allow store only 31 char long SSID in  the std::vector<WifiAPlist_t> APlist  against 32.

PR #4999, opened 2021-04-01
Add setMTU function to BLEClient.cpp/.h

ericalbers:
The current implementation has a getMTU function which returns the mtu sent in a message.

This function allows you to set the MTU value on the connected device, it first sets the MTU locally by calling esp_ble_gatt_set_local_mtu.   It then calls esp_ble_gattc_send_mtu_req  to have the connected device also change its MTU size.


h2zero:
Instead of duplicating all the code, why not just call `BLEDevice::setMTU()` ?

ericalbers:
> 
> 
> Instead of duplicating all the code, why not just call `BLEDevice::setMTU()` ?

the setMTU function in BLEDevice specifically says it only sets the local copy.  I didn't think modifying it to also set the remote copy would be a good idea as there may be code out there dependent upon it only setting the local copy.

h2zero:
The MTU is determined based on the lowest value between the connecting devices. In this library all you need to do is set the local MTU once before connecting to a peer device and the MTU gets negotiated after the connection is established, this happens automatically (you can see the code that does this in the client on connected event handler). 

After that there is nothing to be done. The local preferred MTU is a global setting for all connections so you could simple set it once to the max (517) and the actual used MTU will be determined at connection.
I'm not sure if this PR is going to accomplish what you're looking for beyond what is already available.

chegewara:
You can set local MTU from BLEDevice, then after connecting to peripheral just call `esp_ble_gattc_send_mtu_req`.
As it is event driven process you will have to add event that will get response from peer device, because the value can be smaller than you are trying to set.

ericalbers:
I tried setting the MTU value before connection and it didn't set it on the device, the device refused to respond with data...once I added the call too esp_ble_gattc_send_mtu_req the device would respond with packets.
Not sure if the negotiation was failing somehow or if perhaps the device I was talking too just didn't negotiate correctly.  I assumed it was only setting it locally and not on the remote, so added a function to set it explicitly.

h2zero:
I must apologize, I was not looking at this repos code and instead basing it on cpp_utils that this came from.

In this code base the call to exchange MTU is not automatic on connection so I withdraw my previous comment.

Considering this fact I would think it better to  change the code in this PR to just call the BLEDevice method then call the MTU exchange function save a few bytes instead.

Alternatively, pulling the code from cpp_utils would be more seamless. 
Good change either way.

PR #5003, opened 2021-04-01
Added method to change the ledc PWM frequency programmatically

iTacco:
Currently there is no real option to change the pwm frequency programmatically. The only "hack" solution is to use the `ledcWriteTone(uint8_t chan, double freq)` method.

But there is problem with this hack. Inside this method the bit resolution value is hard coded to 10. Which means if you had a different bit resolution inside your setup it would be overrided here.

So I added a simple method to change the pwm frequency programmatically. This method takes three parameters:
`uint8_t chan, double freq, uint8_t bit_num`

### DIFF
--- a/cores/esp32/esp32-hal-ledc.c
+++ b/cores/esp32/esp32-hal-ledc.c
@@ -279,7 +279,7 @@ void ledcDetachPin(uint8_t pin)
     pinMatrixOutDetach(pin, false, false);
 }
 
-void ledcChangeFrequence(uint8_t chan, double freq, uint8_t bit_num)
+void ledcChangeFrequency(uint8_t chan, double freq, uint8_t bit_num)
 {
     if (chan > 15) {
         return 0;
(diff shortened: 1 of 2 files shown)


lbernstone:
How about adding a boolean to ledcSetup that determines whether to reset the duty cycle instead?

iTacco:
But you could not change the frequency during runtime.
Or I am not getting what do you want to archive with resetting the duty cycle.
@lbernstone

lbernstone:
You would just call the setup a second time to change the frequency.  Your code is almost identical to what ledcSetup calls, it just doesn't set the duty to 0.  If that is not clear enough, you could then have your setFrequency function, but it would just be an alias to ledcSetup.

iTacco:
Ah okay. Now I understand what you meant.
But still. I guess it would be more intuitive and cleaner if there would be a seperate method to change the frequency with instead of add another paramater to the setup method.

But of couse. It could be solved by adding another parameter to the setup method.

me-no-dev:
any development here? :)

iTacco:
All development done in commits ;)
But I guess it needs to be discussed whether it should be done like my approach or it could be done with the approach by @Ibernstone

me-no-dev:
I will accept this if you return the frequency as done in `ledcSetup`

iTacco:
> I will accept this if you return the frequency as done in `ledcSetup`

Done.

PR #5037, opened 2021-04-10
Removed ProS2 and added TinyS2 to boards.txt

UnexpectedMaker:
I wont be shipping my ProS2 and instead am shipping my TinyS2.


lbernstone:
Looks like a nice board :champagne:

UnexpectedMaker:
> Looks like a nice board 🍾

Cheers :)

lbernstone:
Modifying hal code is outside of the scope of a variant change, and I don't think you are correct.  pin46 is input only (so NUM_OUTPUT_PINS really be 46).

UnexpectedMaker:
This is NOT a variant change. This is specific to all S2 and S3 chips. Either way that number is wrong. It should at least be 46 as having it as 45 excludes IO45 which is a JTAG pin, so this is a bug that needs to be fixed.

UnexpectedMaker:
> Modifying hal code is outside of the scope of a variant change, and I don't think you are correct. pin46 is input only (so NUM_OUTPUT_PINS really is 46).

This is NOT a variant change. This is specific to all S2 and S3 chips. Either way that number is wrong. It should at least be 46 as having it as 45 excludes IO45 which is a JTAG pin, so this is a bug that needs to be fixed.

chegewara:
It all depends what you mean by `NUM_OUPUT_PINS`. Is it highest number of the output pin or total number of output pins. But actually in both cases it is wrong value. 
- highest number of output pins is 45, pin IO46 is IN only,
- total number of OUT pins is 42, because pin IO46 is IN only and there is no IO22-25 on S2 (why?)

Am i missing something?

UnexpectedMaker:
> It all depends what you mean by `NUM_OUPUT_PINS`. Is it highest number of the output pin or total number of output pins. But actually in both cases it is wrong value.
> 
> * highest number of output pins is 45, pin IO46 is IN only,
> * total number of OUT pins is 42, because pin IO46 is IN only and there is no IO22-25 on S2 (why?)
> 
> Am i missing something?

Yeah, it's an ambiguous name for sure, but in code it limits the MAX GPIO value that can be used... and that's just incorrect, as it prevents anyone from using IO45 in their Arduino code.

chegewara:
would rather propose to fix this line:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-gpio.c#L212

UnexpectedMaker:
> I would rather propose to fix this line:
> https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-gpio.c#L212

Is that the only place it's used? 
I don't mind how it's fixed, just that it gets fixed (and quickly) as blocking this IO is nasty - especially for my FeatherS2 :)

chegewara:
It is also used in this code:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-gpio.c#L255-L270

So, in that case your suggestion of changing it in define is good, but value shall be 46 not 47.

PR #5042, opened 2021-04-11
Added example for connecting ESP32 to Mosquitto (mqtt) broker through TLS encrypted channel.

norbertg1:
I spent long hours to make working TLS encrypted communication with ESP32 <----> Mosquitto broker (MQTT) also opened for that an issue where others too suffered from a similar problem: https://github.com/espressif/arduino-esp32/issues/5021

I decided to make a useful example, I hope it can be useful in the future, please include it. 

In WiFiClientSecureClientAuthentication.ino soure file also contains short manual.
certificate_generator.sh is an easy tool for generating certificates


me-no-dev:
I think that this example should go into the PubSub library and not here. CI can not build it because it requires external lib.

PR #5060, opened 2021-04-14
Esp32c3 Support

me-no-dev:


Jason2866:
I think this should be esp32c3
https://github.com/espressif/arduino-esp32/blob/master/tools/platformio-build-esp32c3.py#L252

me-no-dev:
good catch. Address of bootloader is 0x00 also

PR #5065, opened 2021-04-16
WiFiClientSecure fix flipped cert/key in comment

Deffendor:
The key and cert were flipped. It's just in the comments, but might avoid confusion.


me-no-dev:
thanks :)

PR #5077, opened 2021-04-18
BLE5 features to use with C3/S3

chegewara:
Added new BLE5 features to use on C3/S3 family:
- extended scan,
- extended/multi advertising

New code is not fancy (no feedback from events), but i think it is functional. To get feedback from events i am suggesting to use custom GAP callback, which is already implemented in BLEDevice.


chegewara:
@me-no-dev Should i skip examples for esp32s2 and c3?

h2zero:
Looks nice, wish I had a C3 to test it out with 😢

chegewara:
@h2zero  Please contact me, and i can send you one.

h2zero:
That would be amazing! I will be in touch, thanks.

PR #5081, opened 2021-04-19
In esp32-hal-log, direct calls to ESP_LOG_x macros is more efficient …

jfbuggen:
…than using intermediate function log_to_esp

As indicated in https://github.com/espressif/arduino-esp32/issues/4845#issuecomment-820540642_ it is more efficient to call directly the ESP LOG macros. This spares a function call, a 512b buffer and a call to vsnprintf. No change in functionality.


me-no-dev:
thanks :)

coratron:
Hi, thanks for addressing this. I found a couple of issues
@jfbuggen 

1) there is a bug on line 161 which breaks compilation if log levels are set to ERROR or higher:
`#define log_e(format, ...) do {log_to_esp(TAG, ESP_LOG_ERROR, format, ##__VA_ARGS__);}while(0)`
should be changed to
`#define log_e(format, ...) do {ESP_LOG_LEVEL_LOCAL(ESP_LOG_ERROR, TAG, format, ##__VA_ARGS__);}while(0)`

2) Also, not sure why the ARDUHAL_LOG_LEVEL_NONE is formatted as ARDUHAL_LOG_LEVEL_ERROR . Perhaps not an issue but intentional?

3) Main reason that #4845 was trying to address is ability to use ESP log instead of Arduino so that esp_log_set_vprintf could be used. This is now broken. Not 100% sure it is due to this change or something else in IDF. See below:

Example code flow:
`
//function used to reroute the logs
int websocket_printf(char * message, va_list args)
{
  //Serial.printf("%s\r\n", message); //this line is commented but the system crashes anyway
}

esp_log_level_set("*", ESP_LOG_ERROR);
esp_log_set_vprintf((vprintf_like_t)&websocket_printf);
log_e("%s\r\n", "hi");
`
This results in the following error
`0x4014b063:0x3ffb27500 in s_log_level_get_and_unlock at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/log/log.c:162`

If the esp_log_set_vprintf() line is commented out then the logs work via serial normally (given point 2) above is fixed)

PR #5085, opened 2021-04-19
BLE5 features to use with C3/S3

chegewara:
Added new BLE5 features to use on C3/S3 family:

extended scan,
extended/multi advertising
New code is not fancy (no feedback from events), but i think it is functional. To get feedback from events i am suggesting to use custom GAP callback, which is already implemented in BLEDevice.


chegewara:
Ready to merge.

PR #5092, opened 2021-04-20
RainMaker library: Minor changes

shahpiyushv:
- Use Serial.print instead of log_i for QR code helper information,
  so that it is always printed by default.
- Expose the RainMaker factory reset and wifi reset APIs.
- Simplify example to have only a Switch device.
- Enable push button based Factory reset and Wi-Fi reset.
- Moved API doc to RainMaker library's top level README.
- Other minor doc changes.


me-no-dev:
CI is breaking for RMakerSwitch sketch. Please fix :)

me-no-dev:
done :)

PR #5098, opened 2021-04-21
Still finding missing bits for the tinys2 board definition :(

UnexpectedMaker:
Sorry @me-no-dev  :(


UnexpectedMaker:
@me-no-dev thanks! Hopefully the last one ;)

PR #5113, opened 2021-04-27
add board KB32-FT in boards.txt

woodif:
add board KB32-FT


woodif:
Can add my board KB32-FT for esp32 arduino are use ESP-WROOM-32 

ref.
https://kb32ft.makerasia.com/

from Thailand thank.

PR #5114, opened 2021-04-27
softAP should only start modern WPA2

Rotzbua:
* AP with old WPA should not be used due security flaws.
* restore behavior from 1.0.6, see https://github.com/espressif/arduino-esp32/blob/46d5afb17fb91965632dc5fef237117e1fe947fc/libraries/WiFi/src/WiFiAP.cpp#L127


me-no-dev:
Thanks :)

PR #5123, opened 2021-04-28
Added condition to only use tinyusb as needed

AronRubin:
Tinyusb was required on platforms where it was not available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,8 +161,11 @@ set(includedirs
 set(srcs ${CORE_SRCS} ${LIBRARY_SRCS} ${BLE_SRCS})
 set(priv_includes cores/esp32/libb64)
 set(requires spi_flash mbedtls mdns esp_adc_cal)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -161,8 +161,11 @@ set(includedirs
 set(srcs ${CORE_SRCS} ${LIBRARY_SRCS} ${BLE_SRCS})
 set(priv_includes cores/esp32/libb64)
 set(requires spi_flash mbedtls mdns esp_adc_cal)
-set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt arduino_tinyusb main)
+set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt main)
 
+if(IDF_TARGET MATCHES "esp32s[23]|esp32c3" AND CONFIG_TINYUSB_ENABLED)
```
atanisoft, line 166: we can limit this to only esp32s2/s3 as the c3 does not support tinyusb at this time.

AronRubin: Ok, done

PR #5133, opened 2021-04-30
Added new boards pinout images for ESP32, ESP32-S2 and ESP32-C3

pedrominatel:
Updated images (suggestion) for the ESP32 DevKitC, ESP32-S2 Saola and ESP32-C3-DevKitM.


pedrominatel:
I can also add the SVG files if needed :)

igrr (commented):
Looks like docs/_old_esp32_pinmap.png and docs/_old_esp32s2_pinmap.png aren't used anywhere, can they be removed?

atanisoft:
@pedrominatel is it possible to add the default pull-up/down and other indicators to the new images that are present on the current ones?

pedrominatel:
Yes, it is @atanisoft. I'll update the images to include this information.

lbernstone:
pin46 on esp32-s2 is GPI, and really needs to be indicated somehow that it is input only.

pedrominatel:
Can you review @atanisoft?
@lbernstone, see the new indication on the GPIO46. Same as the ESP32-DevKitC.

atanisoft:
1. esp32 34,35,36,39 are input only but listed as input/output (GPIO vs GPI). Same comment on esp32-s2 for 46.
2. Instead of WPU/WPD just use PU/PD (for all boards).
3. If you can also add a legend to the images for what PU/PD/IE (input enabled)/ID (input disabled) all mean it would be great.
4. a note should be added for esp32-s2 IO 26 as used for PSRAM on WROVER modules.

pedrominatel:
Hi @atanisoft thanks for your feedback!

1. Why don't we use the same abbreviation found in the datasheet? It's easy to correlate when using the datasheet/TRM as reference.
2. Same as above.
3. Legend sounds good to have.
4. I can add it.

atanisoft:
> 1. Why don't we use the same abbreviation found in the datasheet? It's easy to correlate when using the datasheet/TRM as reference.

Fine with me, just needs to be easy to read and understand in the image.

pedrominatel:
I'll change the color and add it to the legend :)

atanisoft:
@pedrominatel the latest images look really nice!

One more rather minor change to the esp32 picture, if possible add a note about IO 6-11 being not usable by the application code.

pedrominatel:
I've added the warning in the pins used by Flash/PSRAM on both boards (ESP32/S2). I also did some minor changes in some labels, according to the datasheet naming.

PR #5135, opened 2021-04-30
MDNSResponder: Add method to get TXT key values.

### LINKED ISSUE #5132: [ESPmDNS] MDNSResponder does not have methods to receive TXT record keys (only values).
[ESPmDNS] MDNSResponder does not have methods to receive TXT record keys (only values).

`MDNSResponder` should have a `String MDNSResponder::key(int idx, int txtIdx)` identical to `String MDNSResponder::txt(int idx, int txtIdx)` but for keys

```
String MDNSResponder::key(int idx, int txtIdx) {
    mdns_result_t * result = _getResult(idx);
    if(!result){
        log_e("Result %d not found", idx);
        return "";
    }
    if (txtIdx >= result->txt_count) return "";
    return result->txt[txtIdx].key;
}
```

filimonic:
Add method `MDNSResponder::txtKey(int idx, int txtIdx)` aka `MDNS.txtKey(int idx, int txtIdx)` to be able to get mDNS TXT record name (key) and explore TXT records. Closes #5132

### DIFF
--- a/libraries/ESPmDNS/src/ESPmDNS.cpp
+++ b/libraries/ESPmDNS/src/ESPmDNS.cpp
@@ -245,6 +245,16 @@ mdns_result_t * MDNSResponder::_getResult(int idx){
     return result;
 }
 
+mdns_txt_item_t * MDNSResponder::_getResultTxt(int idx, int txtIdx){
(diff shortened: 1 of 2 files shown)


me-no-dev:
thanks :)

PR #5136, opened 2021-04-30
Separated library sources in cmake for selective.

AronRubin:
Separated library sources and requirements in cmake for selective compilation with Kconfig system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,44 +40,105 @@ set(CORE_SRCS
   cores/esp32/WString.cpp
   )
 
-set(LIBRARY_SRCS
-  libraries/ArduinoOTA/src/ArduinoOTA.cpp
-  libraries/AsyncUDP/src/AsyncUDP.cpp
+set(ARDUINO_ALL_LIBRARIES
+  ArduinoOTA
+  AsyncUDP
+  BLE
+  BluetoothSerial
+  DNSServer
+  EEPROM
+  ESPmDNS
+  FFat
+  FS
+  HTTPClient
+  HTTPUpdate
+  LITTLEFS
+  NetBIOS
+  Preferences
+  RainMaker
+  SD_MMC
+  SD
+  SPIFFS
+  SPI
+  Ticker
+  Update
+  WebServer
+  WiFiClientSecure
+  WiFi
+  WiFiProv
+  Wire
+  )
+
+set(ARDUINO_LIBRARY_ArduinoOTA_SRCS libraries/ArduinoOTA/src/ArduinoOTA.cpp)
+set(ARDUINO_LIBRARY_ArduinoOTA_REQUIRES esp_https_ota)
+
+set(ARDUINO_LIBRARY_AsyncUDP_SRCS libraries/AsyncUDP/src/AsyncUDP.cpp)
+
+set(ARDUINO_LIBRARY_BluetoothSerial_SRCS
   libraries/BluetoothSerial/src/BluetoothSerial.cpp
   libraries/BluetoothSerial/src/BTAddress.cpp
   libraries/BluetoothSerial/src/BTAdvertisedDeviceSet.cpp
-  libraries/BluetoothSerial/src/BTScanResultsSet.cpp
-  libraries/DNSServer/src/DNSServer.cpp
-  libraries/EEPROM/src/EEPROM.cpp
-  libraries/ESPmDNS/src/ESPmDNS.cpp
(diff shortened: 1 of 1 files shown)


me-no-dev:
@projectgus could you have a look please

CMakeLists.txt
```diff
@@ -124,51 +187,28 @@ set(BLE_SRCS
...
 set(requires spi_flash mbedtls mdns esp_adc_cal)
-set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt arduino_tinyusb main)
-
-if(NOT CONFIG_ARDUINO_SELECTIVE_COMPILATION OR CONFIG_ARDUINO_SELECTIVE_ArduinoOTA)
-  list(APPEND priv_requires esp_https_ota)
-endif()
-if(NOT CONFIG_ARDUINO_SELECTIVE_COMPILATION OR CONFIG_ARDUINO_SELECTIVE_LITTLEFS)
-  list(APPEND priv_requires esp_littlefs)
-endif()
+set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt main ${ARDUINO_LIBRARIES_REQUIRES})
```
projectgus, line 211: This looks like a good way to structure the source file dependencies, nice and clean!

The only difficulty is `ARDUINO_LIBRARIES_REQUIRES`, because in the ESP-IDF build system component requirements can't depend on config. This is actually already a problem in the current CMakeLists.txt file, though - it's not a problem added in this PR.

What happens in the build system is the component CMakeLists.txt files are expanded twice - once to build the dependency tree of component requirements and determine which components are included in the build, and then again to do everything else (add source files, compiler command line arguments, etc.). The first expansion doesn't have the project config available (because the project config depends on which components are included in the build, so it would otherwise be a cyclic dependency.)

As a result, I expect that the test ` if(NOT CONFIG_ARDUINO_SELECTIVE_COMPILATION OR CONFIG_ARDUINO_SELECTIVE_${libname})` is always true on the first pass, and all of the component requirements (esp_littlefs, esp_https_ota, etc.) will be added to ARDUINO_LIBRARIES_REQUIRES even if these options are disabled.

On the second pass, the config *is* available so the SRCS files are filtered correctly as expected.

Long term, we know this is not ideal and we have a plan  to add "weak" requirements of the variety "component X requires component Y, but only if component Y is available". But this isn't available right now, sorry!

Suggest for now making the requirements list constant (ie remove ARDUINO_LIBRARIES_REQUIRES, place esp_littlefs and esp_https_ota on this line.)

The rest LGTM!

projectgus (changes requested):

CMakeLists.txt
AronRubin: I have actually been able to get around that limitation with another project by using 


idf_build_get_property(__hack_component_targets __COMPONENT_TARGETS)

And then testing both __hack_component_targets and BUILD_COMPONENTS in the rest of the CMake file. esp_littlefs requires subprojects and esp_https_ota should not be required for projects that do not use OTA (e.g. are BLE only).

me-no-dev: @projectgus it seems to work though. I tested this exact case when fixing arduino to compile as stand-alone component.

@AronRubin could you please rebase to the current CMakeLists.txt?

projectgus: @me-no-dev Oh you're right - sorry. I see now that in current master that if CONFIG_LITTLEFS_PAGE_SIZE and CONFIG_TINYUSB_ENABLED are not set then the relevant component is not added, so that will work fine. (In the first expansion these variables are always empty as there is no config yet, in the second expansion their value will depend on the config.)

Looking at `__COMPONENT_TARGETS` to check the list at build time should work as well. You can also do the check this way:

```cmake
if (NOT CMAKE_BUILD_EARLY_EXPANSION)
    idf_build_get_property(components_to_build BUILD_COMPONENTS)
    if(some_component IN_LIST components_to_build)
      message(STATUS "Including optional component some_component")
      list(APPEND priv_requires some_component)
   endif()
endif()
```

(The only significant differences: checking CMAKE_BUILD_EARLY_EXPANSION means this is skipped on the initial pass before any config is available, and the BUILD_COMPONENTS property is documented so it shouldn't change in an update.)

P-R-O-C-H-Y (approved):
Rebased, fixed conflicts and tested within IDF.

AronRubin:
I apologize for not seeing these emails. I trust that you have done the right thing. Thank you for your time.

PR #5144, opened 2021-05-02
Add uPesy ESP32 DevKit boards support

uPesy:
Please add uPesy ESP32 DevKit boards support


me-no-dev:
I think both boards can use the same variant, no?

uPesy (commented):
Yes it's the same pins_arduino.h file for both uPesy ESP32 Wroom DevKit board and uPesy ESP32 Wrover DevKit board.

PR #5173, opened 2021-05-12
Make LOOP_STACK_SIZE user configurable at compile-time.

everslick:
This PR lets the user override the LOOP STACK SIZE set in sdkconfig (for example with -DARDUINO_LOOP_STACK_SIZE=12288).

### DIFF
--- a/cores/esp32/main.cpp
+++ b/cores/esp32/main.cpp
@@ -6,8 +6,12 @@
 #include "USB.h"
 #endif
 
+#ifndef ARDUINO_LOOP_STACK_SIZE
 #ifndef CONFIG_ARDUINO_LOOP_STACK_SIZE
-#define CONFIG_ARDUINO_LOOP_STACK_SIZE 8192
+#define ARDUINO_LOOP_STACK_SIZE 8192
+#else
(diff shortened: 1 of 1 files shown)


lonerzzz:
Another option from within the code is the following:

```
SET_LOOP_TASK_STACK_SIZE( 16*1024 ); // 16KB

setup()
{
	// setup code here
}

loop()
{
	// loop code here:
}
```

ullix:
https://github.com/espressif/arduino-esp32/pull/5173#issuecomment-1546701008

> SET_LOOP_TASK_STACK_SIZE( 16*1024 ); // 16KB

That would be marvelous, but it does not work for me. What kind of versions of what are needed to make it work, and what else needs to be set/configured?

ullix:
I stand corrected: it DOES work! 

What it does NOT do, and which tripped me, is that it does NOT change the value `CONFIG_ARDUINO_LOOP_STACK_SIZE`. This remains at the default value, in my case 8192. But you see it from `uxTaskGetStackHighWaterMark(<looptask>)`.

So thanks, this little command `SET_LOOP_TASK_STACK_SIZE( 16*1024 ); // 16KB` is very helpful!

Issue #3659, opened 2020-01-23
HttpClient crashes after HTTP request


KenthJ:
Board:							ESP-WROOM-32
IDE name:						Arduino IDE version 1.0.4
Flash Frequency:					80Mhz
PSRAM enabled:                                     No
Upload Speed:					921600
Computer OS:                                        Windows 10

Hi

Please bear with me, I'm completely new in this ESP32 world.

I'm trying to make a very simple HTTP request from my ESP32 towards a webserver which is running on a Raspberry Pi running Win 10 IOT, I have been banging my head for several hours and days trying to solve my problem below.

In most cases the code succeeds getting a response as expected, but when I call http.end() it crashes, sometimes with Guru Meditation Error: Core 1 panic'ed (InstrFetchProhibited), other times with (LoadProhibited).

If I instead use any other URLs instead of my own webserver, everything works fine.
When I test the URL of my Raspberry PI webserver with a normal internet browser on my PC, or using Fiddler everything works fine too.

The strange thing is that it only crashes when it sends a http req to my homemade webserver, it works fine with all other webservers.

I double checked my http header from the server, and also added "Content-Type: text/html" "Date:....." and "Server:..." to the header.

Further more I checked the header and html response using a program called Fiddler and using FireFoxs Inspector tool, and I really can't see any problem with it, but there must be something in the response from my webserver causing the crash

```
#include <WiFi.h>
#include <HTTPClient.h>
 
const char* ssid = "mySSID";
const char* password =  "myPassword";

void setup(void)
{
  Serial.begin(9600);
  delay(4000);

  WiFi.begin(ssid, password);
 
  while (WiFi.status() != WL_CONNECTED) {
    delay(1000);
    Serial.println("Waiting for Wifi");
  }
 
  Serial.println("Wifi connected");
  delay(1000);
}



void loop()
{ 
  if ((WiFi.status() == WL_CONNECTED)) //Check the current connection status
  {  
    HTTPClient http;;

    bool httpInitResult = http.begin("http://192.168.0.105:485/esp32.html");

    http.addHeader("Content-Type", "text/html");

    if( httpInitResult == false )
    {
      Serial.println("http.begin() failed!");
    }
    else
    {
      int httpCode = http.GET();
    
      Serial.print("httpCode: ");
      Serial.println(httpCode);
     
      if (httpCode > 0) //Check for the returning code
      { 
        String payload = http.getString();
    
        Serial.print("Payload: ");
        Serial.println(payload);
[...]

PR #5197, opened 2021-05-19
Fix crash in WiFiClient when read() called after stop()

ppescher:
Thi may happen if `read()` gets called repeatedly (such as in `HttpClient` to parse response headers) and the connection is closed unexpectedly or the remote peer may have unexpected behavior that causes the underlying socket to report an error. In that case `read()` itself calls `stop()`, which invalidates the receive buffer object. Then when `read()` is called again without checking, such as inside `readStringUntil()`, the `_rxBuffer` is null and ESP32 crashes.

See also #3659


me-no-dev:
thanks :)

PR #5219, opened 2021-05-28
Added support for Trueverit Universal IoT Driver

Simonefardella:
Add support for the Trueverit Universal IoT Driver (https://www.trueverit.com)

The board will be released on market using electronic distributors soon


Simonefardella:
@me-no-dev, it's possible to get a verification and add this board?

PR #5274, opened 2021-06-10
Enable clock output for Ethernet RMII on ESP32

### LINKED ISSUE #5239: Ethernet PHY and arduino-esp32 2.0.0/IDF 4.4
Hi guys,

In these days i'm trying to use the arduino-esp32 2.0.0 beta version on my test lab boards ([OLIMEX ESP32-POE](https://www.olimex.com/Products/IoT/ESP32/ESP32-POE/open-source-hardware)). 

Most of the functionalities are okay, but when i try to use the onboard ETH lan chip (LAN8720) nothing works.

I know that "nothing works" is not correct in the IT/Electronics field, but i'll try to explain more in depth in the following paragraphs.

Preliminary informations: this board uses the `GPIO12` as `ETH_PHY_POWER` and the `GPIO17` as `ETH_CLK_MODE`.

I'm using the standard sketch that works for other versions of arduino-esp32 in order to power up the ETH, acquire an IP from DHCP server and do a sample request to google: [here it's the sketch](https://github.com/OLIMEX/ESP32-POE/blob/master/SOFTWARE/ARDUINO/ESP32_PoE_Ethernet_Arduino/ESP32_PoE_Ethernet_Arduino.ino)

but when i open the arduino console i see those messages:

Screenshot 2021-06-03 at 09 09 44

At this point, it seems that the ETH phy chip isn't initialized correctly, so i've picked up my oscilloscope and this is what i've seen:

Older esp32-arduino (1.0.4 and 1.0.6), probe on GPIO17, as you can see, there is a clean and nice looking 50MHZ signal coming out from GPIO17, as expected:

bmp_36_001gpio17-1 0 6

When i use the esp32-arduino 2.0.0, probing GPIO17 there isn't any signal coming out from that pin, as you can see:

bmp_36_003gpio17-2 0 0

...so, i've asked to gitter, and @lbernstone clearly says that starting from esp-arduino 2.0.0 the eth clock [gpio output is hardcoded in sdkconfig](https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/sdkconfig#L616) and if i want to modify it i need to compile my own libraries. 

Anyhow...i've repeated the same probing test on GPIO0 (...thinking that the phy clock out comes out from the "default" GPIO0 pin...), but also in this case, nothing comes out from that:

bmp_36_002gpio0-2 0 0

This seems odd...i mean, in the past using a define (`#define ETH_CLK_MODE ETH_CLOCK_GPIO17_OUT`) everything was okay and no custom libraries were needed...how it's possible that now anyone who is not using the same pinout definition for ETH phy chips of the ESP32-Ethernet-Kit needs a custom compiled libraries?

Please clarify those aspects and help us to make the phy chips work again, because i think that i've misinterpreted those changes and maybe is more simple than reported!

me-no-dev:
fixes #5239


lbernstone:
Aren't there a lot more devices out there using gpio0 in (eg esp32 ethernet kit) than gpio17 out?

me-no-dev:
@lbernstone working on it :)

me-no-dev:
BTW more are using IO17. Don't know how many ESP Ethernet boards are out there

PR #5279, opened 2021-06-11
Fix incorrect Flash Size in Esp32c3

Jason2866:
original PR #5278 from @s-hadinger, which is not accepted because of failure of CLAassistant

`ESP.getFlashChipSize()` always returns `0` on Esp32c3 due to a change in the Flash layout. It looks like to Espressif documentation was not updated accordingly.

https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/storage/spi_flash.html#spi-flash-size

- Esp32 and Esp32s3 have the flash information update by esptool.py
and decribed here
- https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/spi_flash.html
Esp32c3 has no image information at 0x0000 and contrary to the docu
this information is at offset 0x0000

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -35,15 +35,19 @@ extern "C" {
 #if CONFIG_IDF_TARGET_ESP32 // ESP32/PICO-D4
 #include "esp32/rom/spi_flash.h"
 #include "soc/efuse_reg.h"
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks @s-hadinger and @Jason2866 :)

Issue #5300, opened 2021-06-17
NetBIOS library only works in STA mode, and not in AP mode.


pabloandresm:
Title says it all. The problem is the same as the ESP8266NetBIOS library (https://github.com/esp8266/Arduino/issues/8139).

If you look at the code, file NetBIOS.cpp, line 95, in the netbios answer, you can see:
		nbnsa.addr = WiFi.localIP();

Which works only if WiFi is in STA mode, but not AP mode.

The fix would be something like this:
	if (WiFi.getMode()==WIFI_STA)
		nbnsa.addr = WiFi.localIP();
	else
		nbnsa.addr = WiFi.softAPIP();

PR #5301, opened 2021-06-17
fix 5300 netbios in AP mode not working

pabloandresm:
fixes bug where netbios not answering properly in case of wifi mode = AP

https://github.com/espressif/arduino-esp32/issues/5300


me-no-dev:
if mode is AP_STA then it will not work on STA

libraries/NetBIOS/src/NetBIOS.cpp
```diff
@@ -92,7 +92,10 @@ void NetBIOS::_onPacket(AsyncUDPPacket& packet){
                 append_32((void *)&nbnsa.ttl, 300000);
                 append_16((void *)&nbnsa.data_len, 6);
                 append_16((void *)&nbnsa.flags, 0);
-                nbnsa.addr = WiFi.localIP();
+				if (WiFi.getMode()==WIFI_STA)
```
me-no-dev, line 95: you need to change this to `if (WiFi.getMode() & WIFI_STA)`

me-no-dev:
closing due to no response

PR #5304, opened 2021-06-18
[FEATURE] Basic implementation of Arduino's I2S library

PilnyTomas:
Basic version of Arduino's I2S library. We currently support only 16bps + 16000 Hz sampling frequency. Other bitrates and sample rates will print warning and continue to operate, however the resulting audio quality may be poor.
There will be further effort to fix these issues.

**Tasks lists:**
- [x] Upload to GitHub 1st version of I2S Library [PR](https://github.com/espressif/arduino-esp32/pull/5304)
- [x] Fix and specify basic functions of I2S Library
- [x] Verify thread safety
- [x] Fix CI errors
- [x] Auto config of ADC for selecting different pins. [Done](https://github.com/PilnyTomas/arduino-esp32/commit/b3fdeecce22a7ab41b7eb0786277d748d551c213)
  - [x] Implement GPIO to ADC unit & channel conversion methods in IDF i2s driver + edit Arduino I2S acordingly (Will be implemented by responsible team)
- [x] create a workaround about sample rate limitations
- [x] waiting review (consider configuration with Sample rate) @SuGlider
- [x] Implementing proposed changes of @SuGlider by @TomasPilny
- [ ] evaluate EVENTS calls (ESP-IDF)


PilnyTomas:
Commit f366160 improved performance by implementing internall buffer. This buffer is usefull especially when using single-byte writes and reads.
Input should be generally ok, however when reading silent channel I'm getting value 1, but I would expect value 0. I'm not sure if it is expected in ESP, but Arduino has 0 value for silent channel.
Output is little distorted, but much better than without the buffer. I will try using queue instead of buffer and see if it improves performance further.

SuGlider:
@PilnyTomas , I have read the code and those are some thoughts:

- Instead of using Serial.print() in the I2S.cpp, or as in the places removed in  the PR, I'd suggest to use ESP_LOGx(...) instead. Please look at https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-log.h

- I think i2s_write() and i2s_read are already thread safe (based on the code of i2c.c (IDF dirver)). Thus, maybe, the use of MUTEX in I2SClass::read() and I2SClass::write() may not need another semaphore. Please check it.

- In Examples (ADCPlotter and SimpleTone) it will be a lot better to read/write a whole buffer to I2S instead of sample by sample in loop(). Specially for SimpleTone.ino using a buffer with samples already in a buffer will improve performance of I2S and maybe reproduce a clear sound output.

SuGlider:
> Commit [f366160](https://github.com/espressif/arduino-esp32/commit/f3661604ceb9e8ecebb9c783b8f9fe9eba7b63a3) improved performance by implementing internall buffer. This buffer is usefull especially when using single-byte writes and reads.

--Buffering helps in storing data using single-byte operations. But what happens, during single write, when the sketch is faster in producing than the I2S in consuming? Will it overwrite data, for instance? Would this explain output distortion?

------------------------------
> Output is little distorted, but much better than without the buffer. I will try using queue instead of buffer and see if it improves performance further.

-- It may help, as commented above about a potential racing condition. Another way to improve it would be the usual "whole sample array" being sent at once to I2S Write().

SuGlider (changes requested):
Instead of using Serial.print() in the I2S.cpp, or as in the places removed in the PR, I'd suggest to use ESP_LOGx(...) instead. Please look at https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-log.h

I think i2s_write() and i2s_read are already thread safe (based on the code of i2c.c in IDF driver). Thus, maybe, the use of another MUTEX in I2SClass::read() and I2SClass::write() may be redundant. Please check it.

Examples that use single sample Write/Read may produce a racing condition with possible internal buffer overwrite, for instance. For instance, continuous writing in loop() may be faster than I2S data consumption. It may need to use Queues in sigle sample mode or Interrupts with full sample array usage.

PilnyTomas:
> --Buffering helps in storing data using single-byte operations. But what happens, during single write, when the sketch is faster in producing than the I2S in consuming? Will it overwrite data, for instance? Would this explain output distortion?
Yes, it might explain the distortion.

> -- It may help, as commented above about a potential racing condition. Another way to improve it would be the usual "whole sample array" being sent at once to I2S Write().
Yes, filling the buffer and then writing it would be faster, but I'm trying to implement it in compliance with the original Arduino examples which generate and write (or read and print) samples one-by-one, see the examples in [original Arduino library](https://github.com/arduino/ArduinoCore-samd/tree/master/libraries/I2S/examples)


> Instead of using Serial.print() in the I2S.cpp, or as in the places removed in the PR, I'd suggest to use ESP_LOGx(...) instead. Please look at https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-log.h
Thanks for the tip. I was going to do that in near future :)

> I think i2s_write() and i2s_read are already thread-safe (based on the code of i2c.c in IDF driver). Thus, maybe, the use of another MUTEX in I2SClass::read() and I2SClass::write() may be redundant. Please check it.
Mutex is not for the read/write, but for the buffers.
 
> Examples that use single sample Write/Read may produce a racing condition with possible internal buffer overwrite, for instance. For instance, continuous writing in the loop() may be faster than I2S data consumption. It may need to use Queues in single sample mode or Interrupts with full sample array usage.
Yes, you are right, thank you, I did not realize this when I was writing it. In the current implementation the write function will simply refuse to write into the buffer (and return 0 Bytes written) and the samples will be lost.

pedrominatel:
Hi @PilnyTomas,

Due the I2S sample rate issue, we need to use the sample rate divided by 2 on the `i2s_set_clk` function for mono/single channel.

`i2s_set_clk(CHANNEL, SAMPLE_RATE / 2, I2S_BITS_PER_SAMPLE_16BIT, I2S_CHANNEL_MONO);`

I suggest adding the `i2s_set_clk` function available on our API since this issue will be fixed on ESP-IDF soon (MR pending).

This may affect only for mono.

SuGlider (approved):
Approved as is, for  the community's valuation. 
Future revisions will be necessary.

VojtechBartoska:
@pedrominatel Can you please help with review of initial I2S Documentation here?

Issue #4972, opened 2021-03-23
Update Preferences.cpp to include nvs_commit() on three methods


beamholder:
In [Preferences.cpp](https://github.com/espressif/arduino-esp32/blob/master/libraries/Preferences/src/Preferences.cpp),  the functions:
```
Preferences::clear()
Preferences::remove()
Preferences::end()
```
should be revised to include a call to 
`nvs_commit()`
as required per 
[Non-volatile storage library](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html)
when using 
```
nvs_erase_all()
nvs_erase_key()
nvs_close()
```
Suggested changes:

```
void Preferences::end(){                                         // modified to add an nvs_commit()
    if(!_started){
        return;
    }
    esp_err_t err = nvs_commit(_handle);                         // to undo changes: delete the lines from here...
    if(err){
        log_e("nvs_commit fail: %s %s", key, nvs_error(err));
    }                                                            // to here.
    nvs_close(_handle);
    _started = false;
}
```

```
bool Preferences::clear(){                                  // modified to add an nvs_commit()
    if(!_started || _readOnly){
        return false;
    }
    esp_err_t err = nvs_erase_all(_handle);
    if(err){
        log_e("nvs_erase_all fail: %s", nvs_error(err));
        return false;
    }
    // return true;                                         // to undo changes: uncomment this line and...
    err = nvs_commit(_handle);                              // delete the lines from from here...
    if(err){
        log_e("nvs_commit fail: %s %s", key, nvs_error(err));
        return false;
    }
    return true;                                            // to here.
}
```

```
bool Preferences::remove(const char * key){                 // modified to add an nvs_commit()
    if(!_started || !key || _readOnly){
        return false;
    }
    esp_err_t err = nvs_erase_key(_handle, key);
    if(err){
        log_e("nvs_erase_key fail: %s %s", key, nvs_error(err));
        return false;
    }
    // return true;                                         // to undo changes: uncomment this line and...
    err = nvs_commit(_handle);                              // delete the lines from from here...
    if(err){
        log_e("nvs_commit fail: %s %s", key, nvs_error(err));
        return false;
    }
    return true;                                            // to here.
}
```

Xylopyrographer:
Correction to my suggested changes for the end() and clear() methods to fix a coding error in writing to log_e().
```
void Preferences::end(){                                         // modified to add an nvs_commit()
    if(!_started){
[...]

PR #5309, opened 2021-06-19
Include nvs_commit() on three methods

beamholder:
In response to Issue #4972 

In [Preferences.cpp](https://github.com/espressif/arduino-esp32/blob/master/libraries/Preferences/src/Preferences.cpp),  the functions:
```
Preferences::clear()
Preferences::remove()
Preferences::end()
```
should be revised to include a call to 
`nvs_commit()`
as required per 
[Non-volatile storage library](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html)
when using 
```
nvs_erase_all()
nvs_erase_key()
nvs_close()
```


libraries/Preferences/src/Preferences.cpp
```diff
@@ -53,10 +53,14 @@ bool Preferences::begin(const char * name, bool readOnly, const char* partition_
     return true;
 }
 
-void Preferences::end(){
+void Preferences::end(){                                         // modified to add an nvs_commit()
```
igrr, line 56: The comment isn't necessary, as the file change history is visible in `git log` or `git blame`, and any commit that has modified a file can be found and reverted if needed. The comments describing the changes (`// added by XXX 06/2021`) only distract from the purpose of the code.

libraries/Preferences/src/Preferences.cpp
```diff
@@ -74,14 +78,20 @@ bool Preferences::clear(){
         log_e("nvs_erase_all fail: %s", nvs_error(err));
         return false;
     }
-    return true;
+    // return true;                                         // to undo changes: uncomment this line and...
```
igrr, line 81: ditto, and please remove this commented out line

libraries/Preferences/src/Preferences.cpp
```diff
@@ -53,10 +53,14 @@ bool Preferences::begin(const char * name, bool readOnly, const char* partition_
     return true;
 }
 
-void Preferences::end(){
+void Preferences::end(){                                         // modified to add an nvs_commit()
     if(!_started){
         return;
     }
+    esp_err_t err = nvs_commit(_handle);                         // to undo changes: delete the lines from here...
```
igrr, line 60: Since every other method which does some modifications already calls nvs_commit, is this call actually necessary?

libraries/Preferences/src/Preferences.cpp
```diff
@@ -74,14 +78,20 @@ bool Preferences::clear(){
...
+    }
+    return true;                                            // ... to here.
 }
 
 /*
  * Remove a key
  * */
 
-bool Preferences::remove(const char * key){
+bool Preferences::remove(const char * key){                 // modified to add an nvs_commit()
```
igrr, line 94: same here, please remove trailing comments and the commented out return statement.

igrr (changes requested):
Thanks for the PR @beamholder, it is useful! Some minor requests below...

beamholder:
@igrr As I'm new two this whole Github/ Arduino-ESP32 thing, just wondering if this PR is now all good or is there is something else I need to do before it is merged? Thanks.

Xylopyrographer:
Hi. Has this been merged into the master yet? Anything missing if not? Still figuring out the whole GitHub thing.

On Jul 2, 2021, at 12:26 AM, Ivan Grokhotkov ***@***.***> wrote:


@igrr requested changes on this pull request.

Thanks for the PR @beamholder <https://github.com/beamholder>, it is useful! Some minor requests below...

In libraries/Preferences/src/Preferences.cpp <https://github.com/espressif/arduino-esp32/pull/5309#discussion_r662765845>:

> @@ -53,10 +53,14 @@ bool Preferences::begin(const char * name, bool readOnly, const char* partition_
     return true;
 }
 
-void Preferences::end(){
+void Preferences::end(){                                         // modified to add an nvs_commit()
The comment isn't necessary, as the file change history is visible in git log or git blame, and any commit that has modified a file can be found and reverted if needed. The comments describing the changes (// added by XXX 06/2021) only distract from the purpose of the code.

In libraries/Preferences/src/Preferences.cpp <https://github.com/espressif/arduino-esp32/pull/5309#discussion_r662766131>:

> @@ -74,14 +78,20 @@ bool Preferences::clear(){
         log_e("nvs_erase_all fail: %s", nvs_error(err));
         return false;
     }
-    return true;
+    // return true;                                         // to undo changes: uncomment this line and...
ditto, and please remove this commented out line

In libraries/Preferences/src/Preferences.cpp <https://github.com/espressif/arduino-esp32/pull/5309#discussion_r662766547>:

>      if(!_started){
         return;
     }
+    esp_err_t err = nvs_commit(_handle);                         // to undo changes: delete the lines from here...
Since every other method which does some modifications already calls nvs_commit, is this call actually necessary?

In libraries/Preferences/src/Preferences.cpp <https://github.com/espressif/arduino-esp32/pull/5309#discussion_r662766716>:

>  }
 
 /*
  * Remove a key
  * */
 
-bool Preferences::remove(const char * key){
+bool Preferences::remove(const char * key){                 // modified to add an nvs_commit()
same here, please remove trailing comments and the commented out return statement.

igrr (approved):

igrr:
@me-no-dev this PR looks good to me, please consider it for the next release.

me-no-dev:
@igrr @beamholder done :)

Xylopyrographer:
Whoot! Thank-you.

On Aug 23, 2021, at 12:24 PM, Me No Dev ***@***.***> wrote:


@igrr <https://github.com/igrr> @beamholder <https://github.com/beamholder> done :)

PR #5310, opened 2021-06-20
Prefer POSIX over GNU extensions.

dok-net:
Also, commit https://github.com/espressif/esp-idf/commit/624725d6e0f5d4557f3dbbb8f3c7597b17698163 suggests that "WITH_POSIX" shall be removed from the compiler flags.
This PR aligns ESP32 to ESP8266 PR https://github.com/esp8266/Arduino/issues/8150


me-no-dev:
all of the changes that you have made are to files that are auto generated by the lib builder. They will not stick. The lib builder get's the flags from IDF's compile process.

dok-net:
@me-no-dev OK. So you don't feel the idea itself is worth carrying forward to wherever the generation process starts? Just asking, and suggesting. Thanks for the quick response!

me-no-dev:
The proper way to do something like that is to have it configured at build time of the IDF libs in the lib-builder repo, so that when it builds IDF, it will do it with those same flags, which will then carry over to here :)

dok-net:
I am not so sure what is best anymore, anyway, now both ESP8266 and ESP32 allow the full set by defining _GNU_SOURCE . All things even, it might be more worth of consideration to migrate ESP32 to a current GCC release, than spent time pondering the "best" subset of extensions. Again, just suggesting off the top of my head.

PR #5332, opened 2021-06-26
[backport-v1] Fix softap_config_equal compare auth_mode/cipher

### LINKED ISSUE #5038: SoftAP suddenly change back to no password protected (Arduino Release 1.0.6 based on ESP-IDF v3.3.5)
I have tried below code from ESP32 Arduino boards manager v1.0.6 library. I found that similar problem that password protected AP will disappear to become an OPEN softAP after reset. It will happen if softAP restarted, even reburn program (no change flash setting area).

It is working fine if using ESP32 Arduino boards manager v1.0.4 library.
Any guys know what is it going on?

```
#include <WiFi.h>
#include <WiFiClient.h>
#include <WiFiAP.h>

#define LED_BUILTIN 2   // Set the GPIO pin where you connected your test LED or comment this line out if your dev board has a built-in LED

// Set these to your desired credentials.
const char *ssid = "yourAP";
const char *password = "yourPassword";
bool wifi_onoff_toggle = true;
WiFiServer server(80);


void setup() {
  pinMode(LED_BUILTIN, OUTPUT);

  Serial.begin(115200);
  Serial.println();
  Serial.println("Configuring access point...");

  // You can remove the password parameter if you want the AP to be open.
  WiFi.softAP(ssid, password);
  IPAddress myIP = WiFi.softAPIP();
  Serial.print("AP IP address: ");
  Serial.println(myIP);
  server.begin();

  Serial.println("Server started");
}

void loop() {
  WiFiClient client = server.available();   // listen for incoming clients

  if (client) {                             // if you get a client,
    Serial.println("New Client.");           // print a message out the serial port
    String currentLine = "";                // make a String to hold incoming data from the client
    while (client.connected()) {            // loop while the client's connected
      if (client.available()) {             // if there's bytes to read from the client,
        char c = client.read();             // read a byte, then
        Serial.write(c);                    // print it out the serial monitor
        if (c == '\n') {                    // if the byte is a newline character

          // if the current line is blank, you got two newline characters in a row.
          // that's the end of the client HTTP request, so send a response:
          if (currentLine.length() == 0) {
            // HTTP headers always start with a response code (e.g. HTTP/1.1 200 OK)
            // and a content-type so the client knows what's coming, then a blank line:
            client.println("HTTP/1.1 200 OK");
            client.println("Content-type:text/html");
            client.println();

            // the content of the HTTP response follows the header:
            client.print("Click <a href=\"/H\">here</a> to turn ON the LED.<br>");
            client.print("Click <a href=\"/L\">here</a> to turn OFF the LED.<br>");
            client.print("Click <a href=\"/T\">here</a> to turn OFF the WiFi in 10 second & resume .<br>");
            // The HTTP response ends with another blank line:
            client.println();
            // break out of the while loop:
            break;
          } else {    // if you got a newline, then clear currentLine:
            currentLine = "";
          }
        } else if (c != '\r') {  // if you got anything else but a carriage return character,
          currentLine += c;      // add it to the end of the currentLine
        }

        // Check to see if the client request was "GET /H" or "GET /L":
        if (currentLine.endsWith("GET /H")) {
          digitalWrite(LED_BUILTIN, HIGH);               // GET /H turns the LED on
        }
        if (currentLine.endsWith("GET /L")) {
          digitalWrite(LED_BUILTIN, LOW);                // GET /L turns the LED off
        }
        if (currentLine.endsWith("GET /T")) {
            WiFi.softAPdisconnect(false);
            delay(10000);
            if(WiFi.enableAP(true)) Serial.println("WiFi.enableAP(true) == TRUE");
            else Serial.println("WiFi.enableAP(true) == false");
        }
      
      }
    }
    // close the connection:
    client.stop();
    Serial.println("Client Disconnected.");
  }
}
```

Rotzbua:
backport from v2 to v1 of fix #5121
fixes #5038

### DIFF
--- a/libraries/WiFi/src/WiFiAP.cpp
+++ b/libraries/WiFi/src/WiFiAP.cpp
@@ -67,6 +67,9 @@ static bool softap_config_equal(const wifi_config_t& lhs, const wifi_config_t& r
     if(lhs.ap.channel != rhs.ap.channel) {
         return false;
     }
+    if(lhs.ap.authmode != rhs.ap.authmode) {
+        return false;
+    }
     if(lhs.ap.ssid_hidden != rhs.ap.ssid_hidden) {
         return false;
     }


me-no-dev:
Thanks @Rotzbua !

PR #5339, opened 2021-06-28
Added function to retrieve the amount of data received from an BLECharacteristic

xBiohazardx:
Is there a specific reason for it not to be included for the end user? The code should be self explanatory.


chegewara (approved):

xBiohazardx:
Could you please approve these changes? I have made an embarrasing mistake.

chegewara (approved):

Issue #4964, opened 2021-03-21
Second and third NTP servers are not working


szerwi:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			1.0.5
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Windows 10

### Description:
I am using SimpleTime example. I have added second and third NTP server, but when I change first NTP server's address to wrong one, it cannot obtain time. ```configTzTime()``` and ```configTime()``` are not using second and third NTP servers.

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp
#include <WiFi.h>
#include "time.h"

const char* ssid       = "xxx";
const char* password   = "xxx";

const char* ntpServer = "wrong.ntp.org";
const char* ntpServer2 = "pl.pool.ntp.org";
const char* ntpServer3 = "3.pl.pool.ntp.org";
const long  gmtOffset_sec = 3600;
const int   daylightOffset_sec = 3600;

void printLocalTime()
{
  struct tm timeinfo;
  if(!getLocalTime(&timeinfo)){
    Serial.println("Failed to obtain time");
    return;
  }
  Serial.println(&timeinfo, "%A, %B %d %Y %H:%M:%S");
}

void setup()
{
  Serial.begin(115200);
  
  Serial.printf("Connecting to %s ", ssid);
  WiFi.begin(ssid, password);
  while (WiFi.status() != WL_CONNECTED) {
      delay(500);
      Serial.print(".");
  }
  Serial.println(" CONNECTED");
  
  configTzTime("CET-1CEST,M3.5.0,M10.5.0/3", ntpServer, ntpServer2, ntpServer3);
  //configTime(gmtOffset_sec, daylightOffset_sec, ntpServer, ntpServer2, ntpServer3); //same situation
  printLocalTime();

  WiFi.disconnect(true);
  WiFi.mode(WIFI_OFF);
}

void loop()
{
  delay(1000);
  printLocalTime();
}
```

### Debug Messages:
```
Connecting to xxx [D][WiFiGeneric.cpp:374] _eventCallback(): Event: 0 - WIFI_READY
[D][WiFiGeneric.cpp:374] _eventCallback(): Event: 2 - STA_START
[D][WiFiGeneric.cpp:374] _eventCallback(): Event: 4 - STA_CONNECTED
[...]

PR #5343, opened 2021-07-01
Examples update, add a note for configTime()

vortigont:
lwip lib bundled with esp32 Arduino supports only one ntp server. Any additional servers set are just silently ignored.
This default is different from esp8266 Arduino core and very confusing. Most of the examples provided uses 3 different ntp servers for redundancy while only the first one is used actually.
Addressing issue #4964


me-no-dev:
Hi @vortigont :) Thanks for taking the time! LITTLEFS was recently renamed to LittleFS. Could you please adjust your pull request?

vortigont:
@me-no-dev merged with recent master changes. Tnx for the hint.
I was thinking that maybe lwip build defines should be changed to match esp8266? So that by default all 3 ntp servers options could be used? esp8266 Arduino core bundle contains several versions of lwip, like full-featured, small-mss, IPv6. But esp2 offers only one version, and rebuilding your own version doesn't look that easy for everyone's. So, maybe it would be better to match the defauls? I can prepare a PR for the build-scripts changes instead of this PR. How do you think?

me-no-dev:
@vortigont if this is an IDF menuconfig option, then change should be added in the [lib-builder](https://github.com/espressif/esp32-arduino-lib-builder) sdkconfigs, which will then propagate to arduino the next time that IDF has some updates for us :)

vortigont:
OK, then I will check for lib-builder's defines and make a PR, then you can choose between this PR or lwip fix :)
Tnx!

vortigont:
@me-no-dev looks like lib-builder won't help here since esp-idf's lwip [Kconfig](https://github.com/espressif/esp-idf/blob/master/components/lwip/Kconfig) and [lwipopts.h](https://github.com/espressif/esp-idf/blob/master/components/lwip/port/esp32/include/lwipopts.h) doesn't have proper defines to adjust SNTP_MAX_SERVERS at build time. 
Would appreciate if this PR for examples could be accepted, meanwhile I'll try to prepare PR for esp-idf to implement required menuconfig option and maybe someday we could get back to lib-builder's default override for arduino-esp32
Tnx!

me-no-dev:
Merged @vortigont thanks!

PR #5387, opened 2021-07-15
PlatformIO updates for CI

valeros:
This PR is an initial attempt to switch the CI to the latest toolchains 8.4.0 for ESP32, ESP32S2, ESP32C3 from Espressif. Additionally, updated the CI to install PlatformIO from master branch for better robustness and freshened `package.json`.

cc @me-no-dev


me-no-dev:
Thanks @valeros ! Some questions: 
- how would that work for people that want to use the github master? 
- Did you make sure you will not forget to include the proper toolchains when adding a release package? I assume PlatformIO will be taking care of that

PR #5391, opened 2021-07-16
arduino as component: Remove circular dependency around 'main'

projectgus:
arduino-esp32 has to depend on main in autostart mode, for setup() and loop(),
but this can be done with undefined symbol entries to avoid a large dependency
cycle and other linker errors.

Closes https://github.com/espressif/esp-idf/issues/6968

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -161,7 +161,7 @@ set(includedirs
 set(srcs ${CORE_SRCS} ${LIBRARY_SRCS} ${BLE_SRCS})
 set(priv_includes cores/esp32/libb64)
 set(requires spi_flash mbedtls mdns esp_adc_cal wifi_provisioning nghttp)
-set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt main)
+set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt esp_ipc)
 
 if(IDF_TARGET MATCHES "esp32s2|esp32s3" AND CONFIG_TINYUSB_ENABLED)
   list(APPEND priv_requires arduino_tinyusb)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -183,3 +183,15 @@ endif()
...
+if(CONFIG_AUTOSTART_ARDUINO)
+    # in autostart mode, arduino-esp32 contains app_main() function and needs to
+    # reference setup() and loop() in the main component. If we add main
+    # component to priv_requires then we create a large circular dependency
+    # (arduino-esp32 -> main -> arduino-esp32) and can get linker errors, so
+    # instead we add setup() and loop() to the undefined symbols list so the
+    # linker will always include them.
+    #
+    # (As they are C++ symbol, we need to add the C++ mangled names.)
+    target_link_libraries(${COMPONENT_LIB} INTERFACE "-u _Z5setupv -u _Z4loopv")
```
me-no-dev, line 196: How do you know what is the proper mangle prefix/suffix? Asking because we might need to add some more functions to the list

me-no-dev (approved):

me-no-dev:
@projectgus Unfortunately this change caused arduino to not find the other components. I guess this means that sdkconfig was not expanded when it ran through CMakeLists and the components were not "required"

CMakeLists.txt
igrr: You can run something like:
```bash
echo "void setup() {}" | g++ -c -x c++ -o - - | nm -
```
Output:
```
0000000000000000 T __Z5setupv
```

igrr:
@me-no-dev as I mentioned on Wednesday, this pattern:

https://github.com/espressif/arduino-esp32/blob/a9bd39de666557c9ff0daf3540a6d02afd45a05b/CMakeLists.txt#L169-L178

doesn't work in IDF build system. This is because IDF build system processes each component CMakeLists.txt twice:
1. first time, in a process called "early expansion", to determine component requirements. All arguments except REQUIRES and PRIV_REQUIRES that are passed to idf_component_register are ignored.
2. second time, when the components list is known, and CONFIG options are expanded. At this stage, REQUIRES and PRIV_REQUIRES arguments are ignored.

Before this PR, `arduino` component had `main` as a requirement, while `main` implicitly required all components included into the build. Even though the lines in CMakeLists.txt which I linked above didn't do anything (CONFIG_* items were not defined at the early expansion pass), this transitive dependency `arduino` -> `main` -> {every other component} caused the compilation to succeed.

Now that the dependency from `arduino` to every other component through `main` has been removed, `arduino` component dependencies on arduino_tinyusb, esp_littlefs, esp_https_ota no longer works.

I'll open a PR to correct this by adding these dependencies at a later stage in the component CMakeLists file.

Issue #5319, opened 2021-06-23
Unable to compile arduino as idf component


tsctrl:
Hardware:
Board: ESP32 Dev Module
Core Installation version: ESP-IDF v4.4-dev-1594-g1d7068e4b-dirty
IDE name: IDF component
Flash Frequency: 40Mhz
PSRAM enabled: no
Upload Speed: 115200
Computer OS: Windows 10

Description:
Unable to compile arduino as idf component.

I am doing library update from 4.2 to the current master 4.4 for idf and arduino but having issue with the arduino esp32 lib to compile with error:
I am able to replicate this issue using simple blink project with one additional idf components.
Arduino WiFi / Server need to be call from idf component cpp file.

Step to reproduce:

1. create new blink project
2. create components folder
3. create new .cpp app_main file and replace .c file
4. download the arduino-esp32
5. copy the sdkconfig from tools folder
6. create new component
7. create new clinkage .cpp static file in components with header
8. call arduino WiFi.begin / Server.begin in component static function
9. call component static function from main function

Observation:
can compile and Serial.println works in main files, also work if added from component.
compile failed if i add WiFi.begin() in the component files.

This issue not happening on the 4.2 release

build the project with error as below:
```
Building in: ~\eclipse-workspace\blink\build
cmake --build . -- -v
[1/9] cmd.exe /C "cd /D ~\eclipse-workspace\blink\build\bootloader && ~\.espressif\tools\cmake\3.16.4\bin\cmake.exe --build ."
[1/1] cmd.exe /C "cd /D ~\eclipse-workspace\blink\build\bootloader\esp-idf\esptool_py && python C:/esp-idf/components/partition_table/check_sizes.py --offset 0x8000 bootloader 0x1000 ~/eclipse-workspace/blink/build/bootloader/bootloader.bin"
Bootloader binary size 0x4300 bytes. 0x2d00 bytes (67%) free.
[...]

PR #5401, opened 2021-07-16
add optional component dependencies after Kconfig options are known

igrr:
Follow-up to https://github.com/espressif/arduino-esp32/pull/5391#issuecomment-881629624.

Until this commit, Kconfig options (e.g. `CONFIG_TINYUSB_ENABLED`) were
used in conditions preceding `idf_component_register` to determine which
components need to be added to `arduino` component requirements.
However the Kconfig options aren't known at the early expansion stage,
when the component CMakeLists.txt files are expanded the first time
and requirements are evaluated. So all the conditions evaluated as if
the options were not set.

This commit changes the logic to only add these components as
dependencies when the Kconfig options are known. Dependencies become
"weak", which means that if one of the components isn't included into
the build for some reason, it is not added as a dependency.
This may happen, for example, if the component is not present in the
`components` directory or is excluded by setting `COMPONENTS` variable
in the project CMakeLists.txt file.
This also ensures that if the component is not present, it will not be
added as a dependency, and this will allow the build to proceed.

Closes https://github.com/espressif/arduino-esp32/issues/5319.


me-no-dev:
Already implemented through idf-master branch

PR #5418, opened 2021-07-19
fix: Respecting connection timeout in WiFiClientSecure

### LINKED ISSUE #5398: WiFiClientSecure: connection timeout not working
When connecting to not responding server/port using `WiFiClientSecure` , connection timeout is always around 18s.
However, connection timeout works ok for the `WiFiClient`.

When looking at sources, [WiFiClient::connect](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiClient.cpp#L212) uses a common non-blocking socket connect pattern for receiving the desired timeout. This is not implemented in the [start_ssl_client](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/ssl_client.cpp#L54) function used by `WiFiClientSecure`.

Is there a reason, why connection timeout is not implemented in the `start_ssl_client` function?

I've tried copy the initial connection pattern from `WiFiClient::connect` pattern to `start_ssl_client` and it works ok (unsurprisingly).

Testing code:
```cpp
#include <WiFiClientSecure.h>

 const char* ssid     = "SSID ";     // your network SSID (name of wifi network)
 const char* password = "password"; // your network password

const char*  server = "www.howsmyssl.com";  // Server URL



void setup() {
  //Initialize serial and wait for port to open:
  Serial.begin(115200);
  delay(100);

  Serial.print("Attempting to connect to SSID: ");
  Serial.println(ssid);
  WiFi.begin(ssid, password);

  // attempt to connect to Wifi network:
  int i = 30;
  while (WiFi.status() != WL_CONNECTED && --i) {
    Serial.print(".");
    // wait 1 second for re-trying
    delay(500);
  }
  if(WiFi.isConnected()) {
    Serial.print("Connected to ");
    Serial.println(ssid);
  } else {
    Serial.println("Connection failed ");
    Serial.println("Restarting.. ");
    ESP.restart();
  }

  WiFiClientSecure secClient;


  secClient.setInsecure();

  Serial.println("\nStarting secure connection to server...");
  uint32_t start = millis();
  int r = secClient.connect(server, 440, 5000);
  Serial.printf("Connection took: %lums\n", millis()-start);
  if(!r) {
    Serial.println("Connection failed!");
  } else {
    Serial.println("Connected!");
    secClient.stop();
  }

  WiFiClient client;
  Serial.println("\nStarting unsecure connection to server...");
  start = millis();
  r = client.connect("192.168.1.3", 500, 3000);
  Serial.printf("Connection took: %lums\n", millis()-start);
  if(!r) {
    Serial.println("Connection failed!");
  } else {
    Serial.println("Connected!");
    client.stop();
  }
}

void loop() {
  // do nothing
}

```

vlastahajek:
Closes #5398

Using the same non-blocking socket connect pattern for respecting connection timeout, copied from [WiFiClient::connect](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiClient.cpp#L212).

`WiFiClient::connect` uses `lwip_connect_r`, whereas [start_ssl_client](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiClientSecure/src/ssl_client.cpp#L54) uses `lwip_connect`. I haven't found what is the difference between them. I tested both, both work ok, so I kept `lwip_connect`.

### DIFF
--- a/libraries/WiFiClientSecure/src/ssl_client.cpp
+++ b/libraries/WiFiClientSecure/src/ssl_client.cpp
@@ -76,30 +76,67 @@ int start_ssl_client(sslclient_context *ssl_client, const char *host, uint32_t p
         return -1;
     }
 
+    fcntl( ssl_client->socket, F_SETFL, fcntl( ssl_client->socket, F_GETFL, 0 ) | O_NONBLOCK );
     struct sockaddr_in serv_addr;
     memset(&serv_addr, 0, sizeof(serv_addr));
     serv_addr.sin_family = AF_INET;
     serv_addr.sin_addr.s_addr = srv;
     serv_addr.sin_port = htons(port);
 
-    if (lwip_connect(ssl_client->socket, (struct sockaddr *)&serv_addr, sizeof(serv_addr)) == 0) {
-        if(timeout <= 0){
-            timeout = 30000; // Milli seconds.
-        }
-        timeval so_timeout = { .tv_sec = timeout / 1000, .tv_usec = (timeout % 1000) * 1000 };
+    if(timeout <= 0){
+        timeout = 30000; // Milli seconds.
+    }
 
-#define ROE(x,msg) { if (((x)<0)) { log_e("LWIP Socket config of " msg " failed."); return -1; }}
-        ROE(lwip_setsockopt(ssl_client->socket, SOL_SOCKET, SO_RCVTIMEO, &so_timeout, sizeof(so_timeout)),"SO_RCVTIMEO");
-        ROE(lwip_setsockopt(ssl_client->socket, SOL_SOCKET, SO_SNDTIMEO, &so_timeout, sizeof(so_timeout)),"SO_SNDTIMEO");
+    fd_set fdset;
+    struct timeval tv;
+    FD_ZERO(&fdset);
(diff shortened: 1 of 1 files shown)


me-no-dev:
@david-cermak could you please have a look?

david-cermak (approved):
LGTM! Yes, the correct way of handling connection timeout is to set the socket to the non-blocking mode and `select()` on its writeability.

Some consistency comments:
* the code uses both `lwip_...` prefixed API (lwip-connect), as well as the standard posix like `select()` or `close()`. The former is faster, the latter is useful when writing generic code (e.g. with VFS), but in general better to stick to one API. On the other hand, this seems to be in line with the `WiFiClient::connect`, so it is consistent.
* `lwip_connect_r()` was an old re-entrant function variant in older IDF's, but in all currently supported versions, the socket API is fully reentrant and thread safe.
* As a side note: we typically switch the socket back to the blocking mode, after connect, so standard `SO_RCVTIMEO` and `SO_SNDTIMEO` options make sense, but this is in line with the previous code, as the socket was left in the non-blocking mode previously as well.

PR #5425, opened 2021-07-20
Fix ESP32-Solo WDT on HTTP OTA update

arendst:
ESP32-Solo without these changes will report below on executing HTTP OTA update:

``` 
E (72016) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (72016) task_wdt:  - IDLE0 (CPU 0)
E (72016) task_wdt: Tasks currently running:
E (72016) task_wdt: CPU 0: loopTask
E (72016) task_wdt: Aborting.
abort() was called at PC 0x401641e8 on core 0
```

With these changes HTTP OTA update finish siccesfully.

(Tested with Tasmota on ESP32-Solo-1)

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -42,7 +42,7 @@ extern "C" {
 #elif CONFIG_IDF_TARGET_ESP32C3
 #include "esp32c3/rom/spi_flash.h"
 #define ESP_FLASH_IMAGE_BASE 0x0000     // Esp32c3 is located at 0x0000
-#else 
+#else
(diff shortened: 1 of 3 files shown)


arendst:
The provided PR is actually more or less a bandaid solving possible WDT's on ESP32-Solo

The solo is bound to WDT as it has only one core and a while loop taking over 4 seconds will WDT the device.

In this case the delay(1) in ``HTTPUpdate.cpp`` would be better located in ``Esp.cpp`` in the ``getSketchMD5`` function ``while`` loop.

arendst:
Close PR in favor of a better one.

PR #5426, opened 2021-07-20
Fix ESP32-Solo WDT on HTTP OTA update

arendst:
ESP32-Solo without these changes will report below on executing HTTP OTA update:
```
E (72016) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (72016) task_wdt:  - IDLE0 (CPU 0)
E (72016) task_wdt: Tasks currently running:
E (72016) task_wdt: CPU 0: loopTask
E (72016) task_wdt: Aborting.
abort() was called at PC 0x401641e8 on core 0
```

With these changes HTTP OTA update finish succesfully.

(Tested with Tasmota on ESP32-Solo-1)

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -42,7 +42,7 @@ extern "C" {
 #elif CONFIG_IDF_TARGET_ESP32C3
 #include "esp32c3/rom/spi_flash.h"
 #define ESP_FLASH_IMAGE_BASE 0x0000     // Esp32c3 is located at 0x0000
-#else 
+#else
 #error Target CONFIG_IDF_TARGET is not supported
 #endif
 #else // ESP32 Before IDF 4.0
@@ -187,7 +187,7 @@ static uint32_t sketchSize(sketchSize_t response) {
         return data.image_len;
     }
 }
-    
+
 uint32_t EspClass::getSketchSize () {
     return sketchSize(SKETCH_SIZE_TOTAL);
 }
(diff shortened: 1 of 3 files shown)


cores/esp32/Esp.cpp
```diff
@@ -42,7 +42,7 @@ extern "C" {
 #elif CONFIG_IDF_TARGET_ESP32C3
 #include "esp32c3/rom/spi_flash.h"
 #define ESP_FLASH_IMAGE_BASE 0x0000     // Esp32c3 is located at 0x0000
-#else
```
me-no-dev, line 45: please revert all styling changes. They are not related to the PR :)

cores/esp32/Esp.cpp
```diff
@@ -226,6 +226,8 @@ String EspClass::getSketchMD5()
         md5.add(buf.get(), readBytes);
         lengthLeft -= readBytes;
         offset += readBytes;
+
```
me-no-dev, line 229: You can use the following to delay only when single core (also applies to ESP32-S2 and ESP32-C3)
```cpp
#if CONFIG_FREERTOS_UNICORE
delay(1);  // Fix solo WDT
#endif
```

mrengineer7777:
I wonder if this would work just as well to prevent WDT timeouts? I use this in a couple of background threads.
`vPortYield();                                                           //Allow other tasks to run`

lucasssvaz (approved):

P-R-O-C-H-Y (approved):

VojtechBartoska:
Hello @arendst, can you please sign CLA so we can merge this PR. Thanks a lot!

PR #5433, opened 2021-07-21
Replace the libesp32-camera.a file fix the issue with ov2640 camera driver

nqh2412:
Newly compiled esp32-camera driver from https://github.com/espressif/esp32-camera


me-no-dev:
did you compile the very latest camera lib? it has 270K difference

nqh2412:
Yes. It is the latest version of the esp32-camera. 
It was compiled with esp-idf

me-no-dev:
latest ESP-IDF with the latest toolchain esp-2021r1?

me-no-dev:
I wonder why there is such big size difference...

Issue #5375, opened 2021-07-11
LEDC on ESP32-S2 frequency calculation


simon-jouet:
### Hardware:
Board:							es32-s2 wrover
Core Installation version:			2.0.0 alpha 21947ebe76cf6380fe06aa885390b86cc477709a
IDE name:						platform.io
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					460800
Computer OS:                                        Xubuntu

### Description:

The frequency of the ledc driver when running on an esp32-s2 is off by a factor of 80. If you configure to have a fairly slow frequency (50Hz in my case) the ledc output frequency will be 0.62

![image](https://user-images.githubusercontent.com/832495/125203135-24b4ad80-e26f-11eb-8bf6-d0e7b2b54c1a.png)

My guess is that the culprit is https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-ledc.c#L154, the ESP32-S2 doesn't have a high speed mode for ledc.

If i multiply the clock by 80 then I get the desired frequency

![image](https://user-images.githubusercontent.com/832495/125203273-a9073080-e26f-11eb-9940-3c0a4804d105.png)

### Sketch
```cpp
#include <Arduino.h>
#include "driver/ledc.h"

void setup() {
  ledcSetup(LEDC_CHANNEL_7, 50, LEDC_TIMER_14_BIT);
  ledcAttachPin(38, LEDC_CHANNEL_7);
  ledcWrite(LEDC_CHANNEL_7, 8191);
}

void loop() {
}
```

Issue #5050, opened 2021-04-13
ESP32-S2 PWM for a Servo pulse


StuartsProjects:
ESP32-S2 PWM settings do not work as expected. 

The code bellow when run on an ESP32 WROOM Development board, generates 1ms and 2ms pulses at 50hz, works as intended. It correctly produces a pulse to control a servo that shifts between 1m2 and 2mS.  

```
/*
On ESP32 generates a sweeping pulse of 1mS to 2mS at 50hz
*/

const int servoPin = 17;

int dutyCycle = 0;

const int PWMFreq = 50;                                //50hz
const int PWMChannel = 0;
const int PWMResolution = 12;                       //12 bits 0-4095
 
void setup()
{  
  ledcSetup(PWMChannel, PWMFreq, PWMResolution);
  ledcAttachPin(servoPin, PWMChannel);
  ledcWrite(PWMChannel, dutyCycle);
}

void loop()
{
    ledcWrite(PWMChannel, 200);                     //approx 1ms
    delay(2000);    
  
    ledcWrite(PWMChannel, 400);                     //approx 2mS
    delay(2000);    
}
```

When the above code is run on a ESP32-S2 Saola board the pulse produced is not correct, it has a period of circa 1500mS and periods of 80mS and 160mS. 

If in the code I change PWMFreq  from 50 to 4096, I end up with a pulse that is 1ms to 2ms @ 50hz on the ESP32-S2.

lbernstone:
It seems like something is wrong at the higher resolutions.  Can you test at 10 bit and scale your timings to fit?

StuartsProjects:
Using this;

const int PWMFreq = 50;                             //50hz
const int PWMChannel = 0;
const int PWMResolution = 10;                      
 
Gives me a pulse period of 250uS. 

I need to set PWMFreq = 4096; to get back to a 20mS period.

Dlloydev:
In testing an analogWrite library I wrote (using a ESP32S2-Saola-1M) that uses the latest commit here (it wraps the ledc functions) here from 2 days ago, I find that the set frequency is off by a factor of 80. Need to use 80000 to get 1kHz PWM signal.
```
  analogWriteFrequency(35, 80000);
  analogWriteResolution(35, 8);
  analogWrite(35, 31);
```
The bit resolution and duty value work as expected, but not the frequency.

![image](https://user-images.githubusercontent.com/63488701/115160671-79c59680-a067-11eb-82ff-f7396294dc3e.png)

PR #5452, opened 2021-07-25
Fixes ESP32-S2 LEDC PWM #5375 #5050

SuGlider:
Fixes: #5375
Fixes: #5050

### DIFF
--- a/cores/esp32/esp32-hal-ledc.c
+++ b/cores/esp32/esp32-hal-ledc.c
@@ -115,6 +115,10 @@ static void _ledcSetupTimer(uint8_t chan, uint32_t div_num, uint8_t bit_num, boo
     uint8_t group=(chan/8), timer=((chan/2)%4);
     static bool tHasStarted = false;
     static uint16_t _activeChannels = 0;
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-ledc.c
```diff
@@ -115,6 +115,10 @@ static void _ledcSetupTimer(uint8_t chan, uint32_t div_num, uint8_t bit_num, boo
     uint8_t group=(chan/8), timer=((chan/2)%4);
     static bool tHasStarted = false;
     static uint16_t _activeChannels = 0;
+#if CONFIG_IDF_TARGET_ESP32S2
```
me-no-dev, line 118: is this true just for S2 or also for C3 (S3 maybe?)

SuGlider: I tested it on C3 as well. It works fine on it given that C3 has not such bit, thus thre is no issue/fix to be done.

SuGlider: S3 not tested.

PR #5466, opened 2021-07-27
Add initial support for USB MSC

me-no-dev:
Adds initial support for USB MSC
- General class USBMSC on which drives can be realized
- FirmwareMSC class to add a drive from which the current firmware can be read and a new one written
- Option in the board menu to enable CDC, MSC and DFU on boot
- Multiple MSC drives can be added at the same time


me-no-dev:
@SuGlider @pedrominatel @igrr @lbernstone @atanisoft @chegewara PTAL :)

me-no-dev:
@ladyada @UnexpectedMaker please check this upcoming change and how the general ESP32S2 board is defined. Will be great if you can give those a try as well :)

PR #5486, opened 2021-08-01
HTTPClient: Support for Transfer-Encoding headers that specify "identify"

t-oot:
In [HTTPClient](https://github.com/espressif/arduino-esp32/tree/master/libraries/HTTPClient), if the `Transfer-Encoding` header is set to `identity`, an error (Transfer-Encoding not supported) will occur.

HTTPClient will consider the request as `identity` if the `Transfer-Encoding` header is not set. But it is also defined a response with `identity` explicitly set in the `Transfer-Encoding` header (ref:[MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding)).

This pull request will allow the request to be processed normally even when `identity` is explicitly set. Please consider this fix.


me-no-dev:
Thanks @t-oot :)

PR #5496, opened 2021-08-04
Increase default timeout for WiFiClient from 3ms to 3s

drzony:
## Summary
https://github.com/espressif/arduino-esp32/pull/5487 introduced a default timeout for WiFiClient, however the default was specified in milliseconds instead of seconds, see https://github.com/espressif/arduino-esp32/commit/be84c8219c251210afd87cac71b8ef3527b601b3#commitcomment-54358731
This 3ms timeout breaks OTA when the processor is busy.

## Impact
Sets the default to a saner value, fixes OTA.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -23,10 +23,10 @@
 #include <lwip/netdb.h>
 #include <errno.h>
 
-#define WIFI_CLIENT_DEF_CONN_TIMEOUT  (3)
-#define WIFI_CLIENT_MAX_WRITE_RETRY   (10)
-#define WIFI_CLIENT_SELECT_TIMEOUT_US (1000000)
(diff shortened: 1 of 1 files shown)


libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -23,7 +23,7 @@
 #include <lwip/netdb.h>
 #include <errno.h>
 
-#define WIFI_CLIENT_DEF_CONN_TIMEOUT  (3)
+#define WIFI_CLIENT_DEF_CONN_TIMEOUT  (3000)
```
Rotzbua, line 26: The name should include the time scale to avoid such issues eg `WIFI_CLIENT_DEF_CONN_TIMEOUT_MS`

drzony: Done

me-no-dev:
@drzony thanks for fixing this :)

Issue #5506, opened 2021-08-07
WebServer serveStatic () can cause LoadProhibited exception in _svfprintf_r


timr49:
### Hardware:
Board:							Sparkfun ESP32Thing
Core Installation version:			2.0.0-rc1
IDE name:						Arduino IDE 1.8.5
Flash Frequency:					80Mhz
PSRAM enabled:                                    ?PSRAM?
Upload Speed:					921600
Computer OS:                                        Mac OSX 11.5.1

### Description:
Using a Core Debug Level of Verbose and the WebServer serveStatic() function with the default value of nullptr for its cache_header argument, results in a LoadProhibited exception in _svfprintf_r().
This is because serveStatic() calls log_v() with cache_header corresponding to a "%s" in its format but without checking that cache_header is not nullptr, and then logv() (indirectly) calls _svfprintf_r().
On the other hand, with a Core Debug Level other than Verbose, this does not occur.

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp

//Change the code below by your sketch
[code]
#include <WiFi.h>
#include <WebServer.h>
#include <SPIFFS.h>
#include "secret.h" // for definitions of MySSID and MyPASSWORD

WebServer server;

void setup() {
  // put your setup code here, to run once:
  // serial
  Serial.begin(115200);
  while (!Serial)
    delay(100);
  Serial.println("setting up");

  // wifi
  WiFi.begin(MySSID, MyPASSWORD); // as defined in "secret.h"
  while (!WiFi.isConnected())
  {
    Serial.print('.');
    delay(100);
  }
  Serial.println();
  Serial.print("connected to "); Serial.print(WiFi.SSID());
  Serial.print(" with address "); Serial.println(WiFi.localIP());

  // www server
  server.begin();
  SPIFFS.begin();
  server.serveStatic("/index.htm", SPIFFS, "/index.htm");

  Serial.println("listening");
}

void loop() {
  // put your main code here, to run repeatedly:
  server.handleClient();
}
[/code]```

### Debug Messages:
```
ets Jun  8 2016 00:22:57

rst:0x10 (RTCWDT_RTC_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[...]

PR #5508, opened 2021-08-08
Fix issue #5506 "WebServer serveStatic () can cause LoadProhibited ex…

timr49:
…ception in _svfprintf_r"

"Using a Core Debug Level of Verbose and the WebServer serveStatic() function with the default value of nullptr for its cache_header argument, results in a LoadProhibited exception in _svfprintf_r().
This is because serveStatic() calls log_v() with cache_header corresponding to a "%s" in its format but without checking that cache_header is not nullptr, and then logv() (indirectly) calls _svfprintf_r().
On the other hand, with a Core Debug Level other than Verbose, this does not occur."

Changed serveStatic() to check the value of cache_header and if it is nullptr, instead pass an empty string to log_v().

### Checklist

## Summary
Fix issue #5506 "WebServer serveStatic () can cause LoadProhibited exception in _svfprintf_r"

## Impact
This change adds null-ness checking to the cache_header argument in WebServer::serveStatic() before passing it to log_v(). If it is null, it passes an empty string ("") instead of nullptr. The only intended/expected impact is to prevent a LoadProhibited exception in _svfprintf_r when serveStatic() is called with a nullptr value for its argument cache_header, which for example, occurs when the default value for cache_header is used.

### DIFF
--- a/libraries/WebServer/src/detail/RequestHandlersImpl.h
+++ b/libraries/WebServer/src/detail/RequestHandlersImpl.h
@@ -69,7 +69,7 @@ public:
     , _cache_header(cache_header)
     {
         _isFile = fs.exists(path);
-        log_v("StaticRequestHandler: path=%s uri=%s isFile=%d, cache_header=%s\r\n", path, uri, _isFile, cache_header);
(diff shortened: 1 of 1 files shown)


igrr (approved):

me-no-dev:
Thanks @timr49 :)

VojtechBartoska:
Close #5506

Issue #5507, opened 2021-08-08
Constructor WebServer::WebServer(IPAddress addr, int port) produces an unexpected result


timr49:
### Hardware:
Board: Sparkfun ESP32Thing
Core Installation version: 2.0.0-rc1
IDE name: Arduino IDE 1.8.5
Flash Frequency: 80Mhz
PSRAM enabled: ?PSRAM?
Upload Speed: 921600
Computer OS: Mac OSX 11.5.1

### Description:
The class Webserver is declared with two explicit constructors, one  with signature:
   WebServer::WebServer(IPAddress addr, int port)
Using this results in a server listening on the port number obtained by converting the value of the IPAddress addr argument (in host byte order) to a uint32_t and then to a uint16_t, which is manifestly not the result that would be expected.

The reason for this is that the definition of this constructor calls the WiFiServer constructor as _server(addr, port).
The class WiFiServer has exactly one explicit constructor with signature:
  WiFiServer(uint16_t port=80, uint8_t max_clients=4).
This does not result in a compile-time error because class IPAddress declares an "operator uint32_t" and hence addr is implicitly cast to uint32_t from which it is implicitly case to uint16_t as required by WiFiserver constructor.

This can be demonstrated with the sketch below. A browser was used to make the HTTP request to the URL:
  http://192.168.10.157:43200/
where the port number 43200 was chosen from the number displayed by the line:
  Serial.print("(uint16_t)WiFi.localIP()="); Serial.println((uint16_t)WiFi.localIP(), DEC);
in the sketch, namely:
  (uint16_t)WiFi.localIP()=43200

The Core Debug Level was set to Verbose so that the port number actually used in the HTTP request is exposed by the line:
  log_v("headerValue: %s", headerValue.c_str());
in Parsing.cpp, namely:
  [395236][V][Parsing.cpp:228] _parseRequest(): headerValue: 192.168.10.157:43200

In this case the port number is 43200 because the IP address is 192.168.10.157 but in general it will vary depending on the server's IP address.

As for a fix, we can assume from these results that this constructor is not being used and therefore could simply be deleted.

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp

//Change the code below by your sketch
[code]
#include <WiFi.h>
#include <WebServer.h>
#include "secret.h" // for definitions of MySSID and MyPASSWORD

WebServer *server;

void setup() {
  // put your setup code here, to run once:
[...]

PR #5509, opened 2021-08-08
Fix issue #5507 "Constructor WebServer::WebServer(IPAddress addr, int…

timr49:
… port) produces an unexpected result"

"The class Webserver is declared with two explicit constructors, one with signature:
WebServer::WebServer(IPAddress addr, int port)
Using this results in a server listening on the port number obtained by converting the value of the IPAddress addr argument (in host byte order) to a uint32_t and then to a uint16_t, which is manifestly not the result that would be expected.
...
As for a fix, we can assume from these results that this constructor is not being used and therefore could simply be deleted."

### Checklist

## Summary
Fix issue #5507 "Constructor WebServer::WebServer(IPAddress addr, int port) produces an unexpected result"

## Impact
This change deletes the declaration and invalid definition of an apparently unused constructor of class WebServer. The intended/expected impact is that (attempted) use of this constructor will result in a compile-time error rather than erroneous run-time behaviour.


Rotzbua:
Wouldn't it be better to implement the missing function like in old esp:
https://github.com/esp8266/Arduino/blob/3.0.2/libraries/ESP8266WiFi/src/WiFiServer.h#L80

Furthermore removing maybe break derived classes.. so better fix `, _server(addr, port)` and print a warning at runtime?

me-no-dev:
I agree with @Rotzbua . Issue should be fixed in `WiFiServer`

me-no-dev:
@timr49 even though there are no examples, in reality we can/should add this functionality, because ESP-IDF supports it and could be useful in some cases. ESP32 can have more than 3 interfaces at the same time :) For now the code can just go to the  other constructor.

timr49:
> @timr49 even though there are no examples, in reality we can/should add this functionality, because ESP-IDF supports it and could be useful in some cases. ESP32 can have more than 3 interfaces at the same time :) For now the code can just go to the other constructor.

It occurred to me after commenting above that the ESP WiFi can have two IP interfaces, one the station and one the soft AP.  So, we do have a way of testing multi-homed servers, at least for n=2.

timr49:
I have pushed a different change onto the same branch, which adds support for multi-homed web servers. It  includes a new examples script (WebServer/MultiHomedServers), which test this change. The testing method is in the comments. The expected results were achieved.

At about the same time as my push, I got some error notifications from CI, but am not sure whether they are related to my previous fix or this one. Unfortunately, my GitHub skills are new, so am not sure what to do next.

timr49:
> At about the same time as my push, I got some error notifications from CI, but am not sure whether they are related to my previous fix or this one. Unfortunately, my GitHub skills are new, so am not sure what to do next.

I successfully guessed what I had done wrong and fixed it in a further commit to the pull request branch ("All checks have passed").

I believe that this change is now ready for review.

libraries/WebServer/examples/MultiHomedServers/MultiHomedServers.ino
```diff
@@ -0,0 +1,173 @@
...
+  Serial.print(" IP address: ");
+  Serial.println(WiFi.softAPIP());
+
+  if (MDNS.begin("esp32")) {
+    Serial.println("MDNS responder started");
+  }
+
+  server0 = new WebServer(8080);
+  server1 = new WebServer(WiFi.localIP(), 8081);
+  server2 = new WebServer(WiFi.softAPIP(), 8082);
```
me-no-dev, line 150: for this test to be valid, both servers should be on the same port. That is how you can make sure that the functionality works.

timr49: Changed port numbers as suggested. Expected results returned by test script. Pushed new commit. "All checks have passed".

me-no-dev:
Thanks @timr49 !!! all merged :)

PR #5538, opened 2021-08-13
Implement USB HID Device Support for ESP32-S2

me-no-dev:


me-no-dev:
@pedrominatel @SuGlider @igrr @lbernstone @atanisoft PTAL

SuGlider:
Nice! I'll test it.

SuGlider:
Tested it with Windows 10.
LGTM

libraries/USB/src/USBHIDVendor.cpp
```diff
@@ -0,0 +1,235 @@
...
+static uint8_t feature[64];
+static xQueueHandle rx_queue = NULL;
+static bool prepend_size = false;
+
+USBHIDVendor::USBHIDVendor(uint8_t report_size, bool prepend): hid(){
+	static bool initialized = false;
+	if(!initialized){
+		initialized = true;
+		hid.addDevice(this, TUD_HID_REPORT_DESC_GENERIC_INOUT_FEATURE_LEN);
+        memset(feature, 0, 64);
```
atanisoft, line 69: it looks like a mix of tab/spaces in this method.

libraries/USB/src/USBHID.h
```diff
@@ -0,0 +1,79 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+#include <stdint.h>
+#include <stdbool.h>
+#include "esp32-hal.h"
+#include "USB.h"
+#if CONFIG_TINYUSB_HID_ENABLED
```
atanisoft, line 20: Any reason why both `CFG_TUD_HID` is checked in the cpp files but `CONFIG_TINYUSB_HID_ENABLED` is checked in headers? Shouldn't they be the same value?

me-no-dev: they generally should. done in https://github.com/espressif/arduino-esp32/pull/5538/commits/196ce947789c1d8f3a86047110049ab0edbef5ec

libraries/USB/src/USBHIDVendor.cpp
me-no-dev: also fixed :)

PR #5541, opened 2021-08-13
Fixes uartAvailable() in case UART_READ_RX_FIFO is set

### LINKED ISSUE #5512: uartAvailable() has race condition
### Hardware:
Board: DOIT ESP32 DEVKIT v1
Core Installation version: 1_0_6
IDF Version: 3.3.5-1-g85c43024c
IDE name: VS Code with platform.io
Flash Frequency: 40MHz
PSRAM enabled: no
Upload Speed: 115200
Computer OS: macOS

### Description:
I am experiencing race conditions with `HardwareSerial.available()` and `HardwareSerial.read()`. The following single-threaded code is returning -1 from `HardwareSerial.read()`, which shouldn't be possible after checking that `HardwareSerial.available() > 0`.

This seems to be caused by the underlying HAL `uartAvailable()` function.

### Sketch:
```cpp

//Change the code below by your sketch
#include <Arduino.h>

void setup() {
    Serial.begin(115200);
}

void loop() {
    if (Serial.available() > 0) {
        int b = Serial.read();
        if (b == -1) {
            Serial.println("error reading");
        }
    }
    doOtherStuff()
}
```

**Expected result:** Nothing output on serial console
**Actual result:** Eventually "error reading" output on serial console

### Possible cause

This issue seems to be caused by a race condition with the following line:

https://github.com/espressif/arduino-esp32/blob/5bb8177aa1feec665938be97607b526a4f0ec0bb/cores/esp32/esp32-hal-uart.c#L344

I think that `uartAvailable()` should turn off interrupts and call `UART_MUTEX_LOCK()` in order to avoid the race condition.

SuGlider:
## Summary
Fixes #5512 
In version 2.0.0, in case UART_READ_RX_FIFO is set, a race condition may happen causing the result of uartAvailable() to fail.
If two tasks read the same UART, uartAvailable() may be preempted right when the sum is being executed, generating a sporadic failed result.

## Impact
Fixes the specific condition detailed in issue #5512

### DIFF
--- a/cores/esp32/esp32-hal-touch.c
+++ b/cores/esp32/esp32-hal-touch.c
@@ -95,6 +95,7 @@ void __touchInit()
     SET_PERI_REG_MASK(SENS_SAR_TOUCH_CTRL2_REG, SENS_TOUCH_MEAS_EN_CLR);
     //clear touch enable
     WRITE_PERI_REG(SENS_SAR_TOUCH_ENABLE_REG, 0x0);
+    SET_PERI_REG_MASK(RTC_CNTL_STATE0_REG, RTC_CNTL_TOUCH_SLP_TIMER_EN);
     __touchSetCycles(__touchMeasureCycles, __touchSleepCycles);
     esp_intr_alloc(ETS_RTC_CORE_INTR_SOURCE, (int)ARDUINO_ISR_FLAG, __touchISR, NULL, &touch_intr_handle);
 #else


rob-deutsch:
Thank you for the PR, @SuGlider.

Do interrupts also have to be turned off? For example, what happens if `_uart_isr()` is called right after `uxQueueMessagesWaiting()` but before `uart->dev->status.rxfifo_cnt` is fetched?

cores/esp32/esp32-hal-uart.c
```diff
@@ -341,7 +341,9 @@ uint32_t uartAvailable(uart_t* uart)
         return 0;
     }
 #ifdef UART_READ_RX_FIFO
+    UART_MUTEX_LOCK();
     return (uxQueueMessagesWaiting(uart->queue) + uart->dev->status.rxfifo_cnt) ;
+    UART_MUTEX_UNLOCK();
```
atanisoft, line 346: this line will never be called and the lock will potentially be left locked. You will need to store the value in a temporary variable and unlock prior to return.

SuGlider: Absolutelly right!

PR #5583, opened 2021-08-23
Adds HardwareSerial::setRxBufferSize()

### LINKED ISSUE #5580: 2.0.0-rc2 - Serial.setRxBufferSize() is missing
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			2.0.0-rc2
IDE name:						Arduino IDE
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Windows 10

### Description:
`Serial.setRxBufferSize()` function is missing in HardwareSerial library in 2.0.0-rc2.

SuGlider:
## Summary
It adds HardwareSerial::setRxBufferSize() with default value of 256 bytes.
Fixes #5580 

## Impact
It allows to define a Rx Buffer Size for UART, before begin().

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -105,7 +105,7 @@ void serialEventRun(void)
 
 HardwareSerial::HardwareSerial(int uart_nr) : _uart_nr(uart_nr), _uart(NULL) {}
 
-void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd)
+void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd, size_t rxBufferSize)
 {
     if(0 > _uart_nr || _uart_nr >= SOC_UART_NUM) {
         log_e("Serial number is invalid, please use numers from 0 to %u", SOC_UART_NUM - 1);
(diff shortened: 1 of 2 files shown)


bertmelis:
`uint8_t rxfifo_full_thrhd = 112, size_t rxBufferSize = 256`

Isn't this inconsistent?

szerwi:
I think it will be better to keep similar parameter order that is in the `uartBegin()` function. So instead of:
`unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd, size_t rxBufferSize`
I propose:
`unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, size_t rxBufferSize, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd`

SuGlider:
> `uint8_t rxfifo_full_thrhd = 112, size_t rxBufferSize = 256`
> 
> Isn't this inconsistent?

Why would it be? 
rxfifo_full_thrhld is realted to UART RX FIFO (max size is 128 bytes)
rxBufferSize is related to RX Queue size feed by RX ISR.

SuGlider:
> I think it will be better to keep similar parameter order that is in the `uartBegin()` function. So instead of:
> `unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd, size_t rxBufferSize`
> I propose:
> `unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, size_t rxBufferSize, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd`

Adding to the end of parameter list will preserve retro compatibility.

me-no-dev:
I would suggest adding back `setRxBufferSize(size_t)` which would set a class variable that can then be used in `begin`, so code like the one below would work. Similar to `setPins()`
```cpp
Serial.setRxBufferSize(1024);
Serial.begin(115200);
```

SuGlider:
> I would suggest adding back `setRxBufferSize(size_t)` which would set a class variable that can then be used in `begin`, so code like the one below would work. Similar to `setPins()`
> 
> ```c++
> Serial.setRxBufferSize(1024);
> Serial.begin(115200);
> ```

Added and commited.

PR #5586, opened 2021-08-24
Add string constructor and concat routines taking explicit length args

rcombs:
## Summary
Applies the upstream changes here: https://github.com/arduino/ArduinoCore-API/compare/3b88acac8%5E...0d83f1afc3367037dbde5323c2abd0ae1bd2c583

## Impact
Adds new String convenience methods that are now available in the mainline Arduino implementation, simplifying interoperability with C code that uses pointer+length strings rather than 0-termination. Also includes a change to avoid mutating the source string when taking a substring.

### DIFF
--- a/cores/esp32/WString.cpp
+++ b/cores/esp32/WString.cpp
@@ -35,6 +35,12 @@ String::String(const char *cstr) {
         copy(cstr, strlen(cstr));
 }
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
thanks @rcombs :)

PR #5598, opened 2021-08-26
Update ESP.getChipModel() to support ESP32-S2

kintel:
## Summary

Until now, ESP.getChipModel() only supported ESP32. This PR adds support for ESP32-S2 as well.


cores/esp32/Esp.cpp
```diff
@@ -270,7 +271,17 @@ const char * EspClass::getChipModel(void)
...
+    uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SPI_SYS_3_REG, EFUSE_PKG_VERSION);
+    switch (pkg_ver) {
+    case 0:
+      return "ESP32-S2";
+    case 1:
+      return "ESP32-S2FH16";
+    case 2:
+      return "ESP32-S2FH32";
+    default:
+      return "ESP32-S2 (Unknown)";
```
TD-er, line 283: Why are strings like these not using `F()` ?  (it wasn't in the initial code and also is found in other functions too)

PR #5614, opened 2021-08-30
Add support for the hardware CDC in ESP32-C3

me-no-dev:


chegewara:
@me-no-dev  Is it intended to use `Serial.begin()` to get serial logs working?
```
void setup() {
Serial.begin(115200);
}

void loop() {
Serial.println("test");
delay(1000);
}
```
Yes, i do have devkit with native USB, and yes i did set tools->... options to use it. And here is log output without `Serial.begin()`:
```
20:14:53.968 -> ESP-ROM:esp32c3-api1-20210207
20:14:53.968 -> 
```
nothing more.

s-light:
@chegewara maybe distantly related to my problem?
https://www.esp32.com/viewtopic.php?f=19&t=24280

PR #5633, opened 2021-09-05
Add ALPN support to WiFiClientSecure 

torntrousers:
This adds a function to WiFiClientSecure to set the ALPN protocol.

This is required for an MQTT client to connect to AWS IoT when using an AWS Custom Authorizer, as described [here](https://docs.aws.amazon.com/iot/latest/developerguide/custom-auth.html).

### DIFF
--- a/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
+++ b/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
@@ -127,7 +127,7 @@ int WiFiClientSecure::connect(const char *host, uint16_t port, const char *CA_ce
     if(_timeout > 0){
         sslclient->handshake_timeout = _timeout;
     }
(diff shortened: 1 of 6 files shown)


libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -341,3 +341,8 @@ void WiFiClientSecure::setHandshakeTimeout(unsigned long handshake_timeout)
 {
     sslclient->handshake_timeout = handshake_timeout * 1000;
 }
+
+void WiFiClientSecure::setAlpnProtocols(const char **alpn_protos)
+{
+    _alpn_protos = alpn_protos;
```
me-no-dev, line 347: `_alpn_protos` must be set to NULL in the class constructor.

PR #5634, opened 2021-09-05
Append '/' to an URL with empty path in HTTPClient::begin

long-long-float:
## Summary

Append '/' to an URL with empty path in HTTPClient::begin. For example, `http.begin("http://example.com")` is the same as `http.begin("http://example.com/")`.

~Add an error check for empty path of HTTPClient::begin. For example, `http.begin("http://example.com")` returns false because the path of it is empty (`http.begin("http://example.com/")` returns true).~

RFC2616 and RFC2396 say this url is invalid. Furthermore, most servers return `400 Bad Request`.

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -262,8 +262,8 @@ bool HTTPClient::beginInternal(String url, const char* expectedProtocol)
 
     index = url.indexOf('/');
     if (index == -1) {
-        log_e("the path of url must not be empty");
-        return false;
+        index = url.length();
+        url += '/';
     }
     String host = url.substring(0, index);
     url.remove(0, index); // remove host part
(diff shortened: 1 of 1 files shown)


libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -261,6 +261,10 @@ bool HTTPClient::beginInternal(String url, const char* expectedProtocol)
     url.remove(0, (index + 3)); // remove http:// or https://
 
     index = url.indexOf('/');
+    if (index == -1) {
+        log_e("the path of url must not be empty");
+        return false;
```
atanisoft, line 266: I agree a check is needed here to prevent using `-1` for `index`, but I don't think it should be an error. If you type in "example.com" into a browser it defaults the path to `/` for you. Likely this code should do the same in this case. Perhaps something like this inside this new condition block:
```
  url.append("/");
  index = url.indexOf('/');
```
It will incur the cost of calling indexOf a second time but I don't think there are many options around this.

long-long-float: I think so too. I'll fix this.
> I don't think there are many options around this.  

The method `length` can be also used.

PR #5650, opened 2021-09-09
Arduino component build process fixed

AlexeyYukhin:
## Summary
* esp_eth_phy_new_ksz8081 call issue fixed, which was leading to build process error with error "error: 'esp_eth_phy_new_ksz8081' was not declared in this scope"
* INADDR_NONE const variable name conflict resolved in IPAdrdress.h module 

## Impact
This PR will allow other users to easily compile their ESP-IDF projects with arduino as a component

### DIFF
--- a/cores/esp32/IPAddress.h
+++ b/cores/esp32/IPAddress.h
@@ -91,6 +91,6 @@ public:
     friend class DNSClient;
 };
 
-const IPAddress INADDR_NONE(0, 0, 0, 0);
+// const IPAddress INADDR_NONE(0, 0, 0, 0);
 
 #endif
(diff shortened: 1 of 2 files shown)


Jason2866:
What is different to #5599 ?

AlexeyYukhin:
> What is different to #5599 ?

didn't see that PR. actually it does the same thing, but why it's still not applied? And also i've added IPAddress fix, there was variable name conflict with define macros

atanisoft:
> And also i've added IPAddress fix, there was variable name conflict with define macros

By removing that variable you are potentially breaking any projects which depend on it's existence. It is only a name conflict in some circumstances otherwise it would have been changed long ago, usually this would only be seen if you are doing raw socket manipulation in the same file where this is included.

me-no-dev:
#5599 is merged. You can not "fix" IPAddress like this :)

PR #5623, opened 2021-09-02
Fix boards.txt options for atmegazero_esp32s2 deneyapmini

manchoz:
## Summary

Fixes https://github.com/espressif/arduino-esp32/issues/5618

## Impact

None

PR #5673, opened 2021-09-14
Fixes boards.txt entries for the atmegazero_esp32s2, and also the pla…

eddieespinal:
## Summary
This PR fixes the compiling issue with the ATMegaZero ESP32-S2 board.  I had entered some values wrong in the boards.txt definition for this board but I fixed them in this PR.

I'm also having issues compiling because of the `-flash-size detect` flag which is not respecting the flash-size entered in the boards.txt file and instead is auto-detecting the size of memory on the board. This will fail on the ATMegaZero since this board comes with 32MB of Flash memory but at the moment we are only using 16MB due to esptool.py limitations.

## Impact
It will fix the atmegazero_esp32s2 board, which is affecting a lot of our users who want to use Arduino vs CircuitPython.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -9386,19 +9386,32 @@ atmegazero_esp32s2.build.core=esp32
 atmegazero_esp32s2.build.variant=atmegazero_esp32s2
 atmegazero_esp32s2.build.board=atmegazero_esp32s2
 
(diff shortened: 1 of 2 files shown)


manchoz:
LGTM. @pedrominatel, please, merge this. I will refactor my #5623 to fix the deneyapmini board only.

Issue #5682, opened 2021-09-20
Olimex ESP32 EVB. SDMMC with LAN can cause LAN freezes.


dukess:
I tried read files from SD card and send they through self-written web server (https://github.com/fredlcore/BSB-LAN). When files transferred through WiFi all works fine (on Arduino Mega and Due boards code works without problem too), but when i start to use built-in LAN i got these problems:
1. ICMP packets are lost time-after-time.
2. Low LAN performance when transferring application-generated data.

The investigation allowed us to solve problem by reducing SDMMC frequency (https://github.com/fredlcore/BSB-LAN/pull/390#pullrequestreview-753091482) from SDMMC_FREQ_HIGHSPEED to SDMMC_FREQ_DEFAULT.

I want propose to add one more parameter in begin() function which allow to change SDMMC frequency from application rather SDK. Backward compatibility will be preserved. Also i want to propose to add ETH_CAN_INTERFERE_WITH_SDMMC definition which can switch default speed between SDMMC_FREQ_HIGHSPEED and SDMMC_FREQ_DEFAULT for boards with LAN adapter.

Patch: https://github.com/dukess/arduino-esp32/commit/f068626c12f7cdbe16077a758fe59d043f65eea8

### Hardware:
Board:							Olimex ESP32 EVB
Core Installation version:			1.0.6
IDE name:						Arduino IDE
Flash Frequency:					40Mhz
Upload Speed:					115200
Computer OS:                                        Windows 7/10

PR #5631, opened 2021-09-03
SD_MMC max_freq_khz is set to HIGHSPEED by default

dizcza:
## Summary
I noticed that `max_freq_khz` was set twice to different values; I picked the last one. I made a fix because previously `SDMMC_FREQ_DEFAULT` was used in the host initialization, and some users elsewhere suggest lowering `SDMMC_FREQ_DEFAULT` from 20_000 to 5_000. Now it's clearly seen that changing `SDMMC_FREQ_DEFAULT` won't affect the initialization of an SD card with the SD_MMC driver.

## Impact
It has no impact other than improving code clarity.

------

BTW, I don't understand why `host` initialization has been changed from
```c
    sdmmc_host_t host = {
        .flags = SDMMC_HOST_FLAG_4BIT,
        .slot = SDMMC_HOST_SLOT_1,
        .max_freq_khz = SDMMC_FREQ_DEFAULT,
        .io_voltage = 3.3f,
        .init = &sdmmc_host_init,
        .set_bus_width = &sdmmc_host_set_bus_width,
        .get_bus_width = &sdmmc_host_get_slot_width,
        .set_bus_ddr_mode = &sdmmc_host_set_bus_ddr_mode,
        .set_card_clk = &sdmmc_host_set_card_clk,
        .do_transaction = &sdmmc_host_do_transaction,
        .deinit = &sdmmc_host_deinit,
        .io_int_enable = &sdmmc_host_io_int_enable,
        .io_int_wait = &sdmmc_host_io_int_wait,
        .command_timeout_ms = 0
    };
```
to
```c
    sdmmc_host_t host;
    host.flags = SDMMC_HOST_FLAG_4BIT;
    host.slot = SDMMC_HOST_SLOT_1;
    host.max_freq_khz = SDMMC_FREQ_DEFAULT;
    host.io_voltage = 3.3f;
    host.init = &sdmmc_host_init;
    host.set_bus_width = &sdmmc_host_set_bus_width;
    host.get_bus_width = &sdmmc_host_get_slot_width;
    host.set_bus_ddr_mode = &sdmmc_host_set_bus_ddr_mode;
    host.set_card_clk = &sdmmc_host_set_card_clk;
    host.do_transaction = &sdmmc_host_do_transaction;
    host.deinit_p = &sdspi_host_remove_device;
    host.io_int_enable = &sdmmc_host_io_int_enable;
    host.io_int_wait = &sdmmc_host_io_int_wait;
    host.command_timeout_ms = 0;
```
[...]

PR #5688, opened 2021-09-21
SDMMC frequency selection based on board type

dukess:
On Olimex ESP32 EVB I/O operations with SD card can cause error when LAN is used in same time.
Problem is disappearing if SD MMC frequency lower down from SDMMC_FREQ_HIGHSPEED to SDMMC_FREQ_DEFAULT.

No problem if WiFi used instead LAN.

For issue https://github.com/espressif/arduino-esp32/issues/5682

### DIFF
--- a/variants/esp32-evb/pins_arduino.h
+++ b/variants/esp32-evb/pins_arduino.h
@@ -29,5 +29,6 @@ static const uint8_t MISO  = 15;
 static const uint8_t SCK   = 14;
 
 #define BOARD_HAS_1BIT_SDMMC
+#define ETH_CAN_INTERFERE_WITH_SDMMC
 
 #endif /* Pins_Arduino_h */
--- a/variants/esp32-evb/pins_arduino.h
+++ b/variants/esp32-evb/pins_arduino.h
@@ -29,6 +29,6 @@ static const uint8_t MISO  = 15;
 static const uint8_t SCK   = 14;
 
 #define BOARD_HAS_1BIT_SDMMC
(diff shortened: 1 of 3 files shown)


variants/esp32-evb/pins_arduino.h
```diff
@@ -29,5 +29,6 @@ static const uint8_t MISO  = 15;
 static const uint8_t SCK   = 14;
 
 #define BOARD_HAS_1BIT_SDMMC
+#define ETH_CAN_INTERFERE_WITH_SDMMC
```
igrr, line 32: Perhaps defining something like `BOARD_MAX_SDMMC_FREQ` would be a bit more generic? Then in sdmmc library you can include pins_arduino.h, check if BOARD_MAX_SDMMC_FREQ isn't defined, and define it to the default value. That would eliminate the ifdef around the begin function.

igrr (changes requested):

variants/esp32-evb/pins_arduino.h
dukess: Thanks for idea. Will try it right now.

dukess:
this pull request include this request too: https://github.com/espressif/arduino-esp32/pull/5631

variants/esp32-evb/pins_arduino.h
dukess: @igrr Sorry for disturb, i do not know how to correct declare about changes, so just calling you in this PR.

Issue #5691, opened 2021-09-21
ESP32-S2 analogRead function not returning correct value


brentru:
### Hardware:
Board:							Adafruit FunHouse ESP32-S2
Core Installation version:			2.0.0
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                        Yes
Upload Speed:					115200
Computer OS:                                            macOS and Ubuntu

### Description:
Calls to Arduino's `analogRead` (https://www.arduino.cc/reference/en/language/functions/analog-io/analogread/) function are not working on BSP v2.0.0. It seems to always return a value of `0`, instead of the correct value from the ADC. This functionality was working prior to 2.0.0.

![image](https://user-images.githubusercontent.com/322428/134194992-b2f405fa-e231-4e07-917c-12b91b906503.png)

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp

void setup() {
  Serial.begin(9600);
}

void loop() {
  Serial.println(analogRead(A3));
  delay(100);
}
```

felmue:
Hello @brentru 

do you know which GPIO corresponds to A3? I am only seeing A0 to A2 looking at the FunHouse [pinout](https://learn.adafruit.com/adafruit-funhouse/pinouts).

Some analog inputs share the same ADC module with WiFi and fail to work when WiFi is turned on at the same time. Have a look  [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/peripherals/adc.html).

That said, have you tried to turn off WiFi? Have you tried with a different analog input?

Thanks
Felix

brentru:
@felmue  - looking at the board variant (https://github.com/espressif/arduino-esp32/blob/master/variants/adafruit_funhouse_esp32s2/pins_arduino.h#L53), it'd be GPIO18.

> Some analog inputs share the same ADC module with WiFi and fail to work when WiFi is turned on at the same time. Have a look here.

We've had success reading the analog input using the same code, prior to the v2.0.0 BSP update. 

> Have you tried with a different analog input?

We've also tried reading the value of a pot. with analog pin A0.

> That said, have you tried to turn off WiFi? 

I'm not sure how to turn off the ESP32's WiFi via the Arduino IDE.. The sketch attached with the issue doesn't perform a network setup.

felmue:
Hello @brentru 

[...]

Issue #5658, opened 2021-09-11
[ESP32-S2] analogRead(A10) WIDTH ERR


touchgadget:
### Hardware:
Board:							ESP32S2 Saola, ESP32S2 Dev Module
Core Installation version:			2.0.0
IDE name:						Arduino IDE 1.8.16
Flash Frequency:					80Mhz
PSRAM enabled:                                    no
Upload Speed:					921600
Computer OS:                                        Ubuntu 18.04

### Description:
Analog input on ADC2 returns WIDTH ERR. A0-A9 work fine.

### Sketch:  (leave the backquotes for [code formatting](https://help.github.com/articles/creating-and-highlighting-code-blocks/))
```cpp
void setup() {
  Serial.begin(115200);
  Serial.println(analogRead(A10));
}

void loop() {
}
```

### Debug Messages:
```
E (90) ADC: adc2_get_raw(578): WIDTH ERR: see `adc_bits_width_t` for supported bit width
[   559][E][esp32-hal-adc.c:174] __analogRead(): GPIO11: ESP_ERR_INVALID_ARG
```

touchgadget:
Setting the default analogWidth to 4 for ESP32S2 seems to solve this problem. adc2_get_raw rejects all other values.
```
diff --git a/portable/packages/esp32/hardware/esp32/2.0.0/cores/esp32/esp32-hal-adc.c b/portable/packages/esp32/hardware/esp32/2.0.0/cores/esp32/esp32-hal-adc.c
index 0530f1d..72d051f 100644
--- a/portable/packages/esp32/hardware/esp32/2.0.0/cores/esp32/esp32-hal-adc.c
+++ b/portable/packages/esp32/hardware/esp32/2.0.0/cores/esp32/esp32-hal-adc.c
@@ -46,7 +46,11 @@ static uint8_t __analogVRefPin = 0;
 #endif
 
 static uint8_t __analogAttenuation = 3;//11db
+#if CONFIG_IDF_TARGET_ESP32
 static uint8_t __analogWidth = 3;//12 bits
+#elif CONFIG_IDF_TARGET_ESP32S2
+static uint8_t __analogWidth = 4;   // 13 bits
+#endif
 static uint8_t __analogClockDiv = 1;
 static adc_attenuation_t __pin_attenuation[SOC_GPIO_PIN_COUNT];
 

```

touchgadget:
@chegewara, @caternuson, @atanisoft, @ladyada 

[...]

PR #5711, opened 2021-09-27
Fix analogWidth for ESP32S2 in esp32-hal-adc.c

caternuson:
## Summary
Potential fix for #5691 and #5658.

## Impact
Sets a different default value for `analogWidth` in `esp32-hal-adc.c` for ESP32S2 based boards. This fixes issue with performing `analogRead` on that platform.

Test sketch run on an [Adafruit FunHouse](https://www.adafruit.com/product/4985) which has an analog light sensor attached to `A3`:
```cpp
void setup() {
  Serial.begin(9600);
}

void loop() {
  Serial.println(analogRead(A3));
  delay(100);
}
``` 

Screenshot from 2021-09-27 08-35-21

### DIFF
--- a/cores/esp32/esp32-hal-adc.c
+++ b/cores/esp32/esp32-hal-adc.c
@@ -46,7 +46,11 @@ static uint8_t __analogVRefPin = 0;
 #endif
 
 static uint8_t __analogAttenuation = 3;//11db
+#if CONFIG_IDF_TARGET_ESP32
 static uint8_t __analogWidth = 3;//12 bits
+#elif CONFIG_IDF_TARGET_ESP32S2
+static uint8_t __analogWidth = 4;   // 13 bits
+#endif
 static uint8_t __analogClockDiv = 1;
 static adc_attenuation_t __pin_attenuation[SOC_GPIO_PIN_COUNT];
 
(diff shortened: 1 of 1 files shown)


ladyada:
cc @felmue @me-no-dev @atanisoft @touchgadget

atanisoft:
As noted in https://github.com/espressif/arduino-esp32/issues/5691#issuecomment-927926749 this will break ESP32-C3 ADC support.

cores/esp32/esp32-hal-adc.c
```diff
@@ -46,7 +46,11 @@ static uint8_t __analogVRefPin = 0;
 #endif
 
 static uint8_t __analogAttenuation = 3;//11db
+#if CONFIG_IDF_TARGET_ESP32
 static uint8_t __analogWidth = 3;//12 bits
+#elif CONFIG_IDF_TARGET_ESP32S2
```
atanisoft, line 51: This should be listed first and have an `#else` with the existing 12bit analog value.  With the proposed change this will fail to compile on the ESP32-C3.

atanisoft (changes requested):

caternuson:
@atanisoft What changes are needed? Can update the PR.

me-no-dev:
thanks @caternuson

VojtechBartoska:
close #5691

Issue #5671, opened 2021-09-14
How to disable Auto-detection of flash size when compiling from Arduino?


eddieespinal:
Make your question, not a Statement, inclusive. Include all pertinent information:

What you are trying to do?

I'm trying to compile the ATMegaZero ESP32-S2 board but I'm getting an error about Flash size not being supported.  The thing is that the ATMegaZero comes with 32MB flash memory but we are using 16MB at the moment due to this limitation.  This used to work before but now it's doing some auto-detection and ignoring the flash size entry in my configuration file.

esptool.py v3.1
Serial port /dev/cu.usbmodem01
Connecting...
Chip is ESP32-S2
Features: WiFi, ADC and temperature sensor calibration in BLK2 of efuse
Crystal is 40MHz
MAC: 7c:df:a1:4a:12:8c
Uploading stub...
Running stub...
Stub running...
Changing baud rate to 921600
Changed.
Configuring flash size...
**Auto-detected Flash size: 32MB**
Flash will be erased from 0x0000e000 to 0x0000ffff...
Flash will be erased from 0x00001000 to 0x00006fff...
Flash will be erased from 0x00010000 to 0x00046fff...
Flash will be erased from 0x00008000 to 0x00008fff...
Flash will be erased from 0x002d0000 to 0x002effff...
Compressed 8192 bytes to 47...
Writing at 0x0000e000... (100 %)
Wrote 8192 bytes (47 compressed) at 0x0000e000 in 0.1 seconds (effective 717.4 kbit/s)...
Hash of data verified.

**A fatal error occurred: Flash size '32MB' is not supported by this chip type. Supported sizes: 1MB, 2MB, 4MB, 8MB, 16MB
A fatal error occurred: Flash size '32MB' is not supported by this chip type. Supported sizes: 1MB, 2MB, 4MB, 8MB, 16MB**

### Hardware:
Board:							ATMegaZero ESP32-S2
Core Installation version:			?1.0.0? ?1.0.1-rc4? ?1.0.1? ?1.0.1-git? ?1.0.2? ?1.0.3?
IDE name:						Arduino IDE
Flash Frequency:					40Mhz
PSRAM enabled:                                      yes
Upload Speed:					115200
Computer OS:                                           Mac OSX

### Debug Messages:
```
Enable Core debug level: Debug on tools menu of Arduino IDE, then put the serial output here 

[...]

PR #5717, opened 2021-09-28
ESP32S2 Fix analogRead of adc2 channels

touchgadget:
See also #5658 " [ESP32-S2] analogRead(A10) WIDTH ERR #5658" and #5671 "ESP32-S2 analogRead function not returning correct value"


me-no-dev:
This is the same as #5711 :)

PR #5719, opened 2021-09-29
include pin_arduino.h for variant USB defines

hathach:
## Summary

add `pin_arduino.h` include in core's USB.cpp. This allows macro defined by variant e.g USB_VID, USB_PID to be used correct per board.

## Impact

Correct the USB defines such as VID/PID, manufacturer in usb descriptor. Following is `lsusb` output with Adafruit Metro ESP32-S2 board before and after the PR

Before

```
Bus 003 Device 094: ID 303a:0002 Espressif Systems METRO_ESP32S2
```

After

```
Bus 003 Device 096: ID 239a:80df Adafruit Metro ESP32-S2
```

@ladyada

### DIFF
--- a/cores/esp32/USB.cpp
+++ b/cores/esp32/USB.cpp
@@ -15,6 +15,7 @@
 
 #if CONFIG_TINYUSB_ENABLED
 
+#include "pins_arduino.h"
 #include "esp32-hal.h"
 #include "esp32-hal-tinyusb.h"
 #include "common/tusb_common.h"


me-no-dev:
@hathach let me know if you have any issues with Arduino's USB stack :)
BTW if Adafruit's lib uses tinyusb from here directly, you can set CDC control endpoint to 0x85 and it will be assigned an empty fifo, thus allowing one more device to be added (CDC+MSC+HID+Vendor)

hathach:
thanks @me-no-dev , the adafruit library actually use   ESP core's configuration descriptor building and USBCDC Serial.  Therefore it is all good. For empty fifo for CDC's notification endpoint, yeah, I did notice you have some modification to have 0x85 as empty fifo. I actually plan to do it in a more generic way for tinyusb, basically adding an extra api for cdc driver to call if it is esp32s2 but haven't got time to get into the detail. will tag you if there is an PR for it.

PR #5720, opened 2021-09-29
Optimized getErrorText() in Wire.cpp library

ketukil:
ERRORTEXT string was converted to an array of strings. Each substring is now an independent array element. That provides faster and more reliable execution w/o weird pointer arithmetic.

Original code in x86_64 assembly
original

Optimized code in x68_64 assembly
optimized


me-no-dev:
a new I2C implementation is about to be merged :) it no longer has such interface

PR #5743, opened 2021-10-08
added support for readError on Stream *Unitl functions

imwhocodes:
This pull request add suppor for `*readError()` on all  `*Unitl(..., char terminator)` function of class `Stream`

Given the fact that these functions don't include the `char terminator` on the result this information/functions are really helpfull to know if the cause of function exit was a timeout on `timedRead()` or if the `char terminator`  was actually found

This feature is inspired by [TeensyCore](https://github.com/PaulStoffregen/cores/blob/master/teensy4/Stream.h) (Arduino for Teensy boards) from Paul Stoffregen

**This feature is NOT breaking change**

Luca


me-no-dev:
closing due to lack of response

PR #5746, opened 2021-10-09
I2C Slave Implementation

me-no-dev:
This pull request adds support for I2C Slave to the ESP32 Arduino Framework

Fixes: https://github.com/espressif/arduino-esp32/issues/118

### DIFF
--- a/libraries/Wire/src/Wire.h
+++ b/libraries/Wire/src/Wire.h
@@ -67,6 +67,7 @@ private:
     void (*user_onReceive)(int);
     static void onRequestService(uint8_t, void *);
     static void onReceiveService(uint8_t, uint8_t*, size_t, bool, void *);
+    bool initPins(int sdaPin, int sclPin);
 
 public:
     TwoWire(uint8_t bus_num);
--- a/cores/esp32/esp32-hal-i2c-slave.c
+++ b/cores/esp32/esp32-hal-i2c-slave.c
@@ -181,6 +181,7 @@ static inline bool i2c_ll_slave_rw(i2c_dev_t *hw)//not exposed by hal_ll
 }
 
 //-------------------------------------- PRIVATE (Function Prototypes) ------------------------------------------------
+static void i2c_slave_free_resources(i2c_slave_struct_t * i2c);
 static void i2c_slave_delay_us(uint64_t us);
 static void i2c_slave_gpio_mode(int8_t pin, gpio_mode_t mode);
 static bool i2c_slave_check_line_state(int8_t sda, int8_t scl);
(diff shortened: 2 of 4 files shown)


libraries/Wire/src/Wire.h
```diff
@@ -70,6 +77,7 @@ class TwoWire: public Stream
     bool setPins(int sda, int scl);
     
     bool begin(int sda=-1, int scl=-1, uint32_t frequency=0); // returns true, if successful init of i2c bus
+    bool begin(uint8_t slaveAddr, int sda=-1, int scl=-1, uint32_t frequency=0);
```
platypii, line 80: This PR added an overload for `Wire.begin` which is now causing errors downstream in [Heltec's BMP180.cpp](https://github.com/HelTecAutomation/Heltec_ESP32/blob/master/src/BMP180.cpp#L12).

```
.pio/libdeps/heltec_wifi_lora_32_V2/Heltec ESP32 Dev-Boards/src/BMP180.cpp:13:29: error: call of overloaded 'begin(int&, int, int)' is ambiguous
   Wire.begin(sda, 12, 100000);
                             ^
In file included from .pio/libdeps/heltec_wifi_lora_32_V2/Heltec ESP32 Dev-Boards/src/BMP180.h:5,
                 from .pio/libdeps/heltec_wifi_lora_32_V2/Heltec ESP32 Dev-Boards/src/BMP180.cpp:1:
/home/platypii/.platformio/packages/framework-arduinoespressif32/libraries/Wire/src/Wire.h:79:10: note: candidate: 'bool TwoWire::begin(int, int, uint32_t)'
     bool begin(int sda=-1, int scl=-1, uint32_t frequency=0); // returns true, if successful init of i2c bus
          ^~~~~
/home/platypii/.platformio/packages/framework-arduinoespressif32/libraries/Wire/src/Wire.h:80:10: note: candidate: 'bool TwoWire::begin(uint8_t, int, int, uint32_t)'
     bool begin(uint8_t slaveAddr, int sda=-1, int scl=-1, uint32_t frequency=0);
          ^~~~~
```

I plan to file an issue and a PR in that repo, but the problem is I can't actually figure out how to disambiguate between these two overloaded constructors. How do I even tell C++ which to prefer?

It seems like it would have been better to add the `uint8_t slaveAddr` argument to the END of the `Wire.begin` function, to avoid breaking lots of code in the wild?

me-no-dev: the easiest way is to cast the arguments, like `Wire.begin((int)SDA, (int)SCL);`, but the best would be to use:
```cpp
Wire.setPins(SDA, SCL);
Wire.begin();
Wire.setClock(frequency); // if necessary
```

I intend to look into this and see if I can find a better way to resolve it.

platypii: Thanks for looking into it. Actually though, casting the first argument to (int) doesn't work. GCC still thinks the `int` and `uint8_t` are ambiguous. The only way I was able to get it to compile was to cast the _third_ argument to unsigned:
```diff
-  Wire.begin(13, 12, 100000);
+  Wire.begin(13, 12, 100000U);
```

PR #5764, opened 2021-10-14
Add another Wire requestFrom function signature

torntrousers:
Arduino HardwareI2C defines a requestFrom function
```
    virtual size_t requestFrom(uint8_t address, size_t len, bool stopBit) = 0;
```
whereas the ESP32 Wire doesn't include any requestFrom using size_t types, so this can cause a compile error, eg when using the Arduino ATECC608 library with an ESP32, see https://github.com/arduino-libraries/ArduinoECCX08/issues/25


me-no-dev:
seems you did not start from clean Master? There are conflicts :)

PR #5804, opened 2021-10-25
[FEATURE]  add -D ARDUINO_PARTITION_xxx build flag to platform.txt

tobozo:
Proposition: adding `-DARDUINO_PARTITION_{build.partitions}` flag to make the build partition information available at compile time.

Use case examples:

```C
#if defined ARDUINO_PARTITION_default
  // prevent compilation
  #error "This sketch needs 'Minimal SPIFFS' partition scheme to compile" 
  // or disable sketch features that need flash space
  #define USE_HUGE_BITMAP_IMAGES false
#endif
```

```C
#if !defined ARDUINO_PARTITION_default_ffat
  #error "This sketch needs 'Default 4MB with ffat (1.2MB APP/1.5MB FATFS)' partition scheme to compile" 
#endif
```


tobozo:
Does this seem correct? 

```python
    '\\"ARDUINO_PARTITION_%s\\"' % env.BoardConfig().get("build.partitions").replace('"', "")
```

me-no-dev:
maybe something like this? can you test it?
```python
"ARDUINO_PARTITION_%s" % env.BoardConfig().get("build.partitions", "default.csv").replace('.csv', "")
```

me-no-dev:
@ivankravets @valeros could you propose what is correct here?

valeros:
Hi guys, the variant @me-no-dev suggested looks good, so in the script it should look something like this:
```python
env.Append(
    CPPDEFINES=[
       ...
       "ARDUINO_PARTITION_%s" % env.BoardConfig().get("build.partitions", "default.csv").replace(".csv", "")
   ]
)
```

PR #5807, opened 2021-10-26
Adds feature to decrypt uploaded image bin files. Used esp-idf to encrypt a bin file. 

theeprawn:
This adds decryption buffer before writing images to Flash in Update.cpp.  Uses esp-idf 'espsecure.py encrypt_flash_data' with AES 256bit key(32bytes) address offset for tweaking/salting the key when flash_crypt_conf is non-zero.

Also added example sketch, with ota_key, unencrypted image file and same file encrypted using same setting as in example file.

I been using this method for over a year for production code for upgrading files either from webserver using HTTPClient or as in example using webserver on ESP32 to upload image files via a webbrowser. This way the production code is always encrypted. The ESP32 flash encryption is enabled as well, using different AES key set in efuses, the Update writes to encrypted partition.
So flow would be as OTA encrypted image is uploaded, Update decrypts it back to uncrypted data using ota_key, ota_address and flash_crypt_conf setting, then Update encrypts the unencrpted data to flash memory it using the efuse key and settings.

Check beginning of  _writeBuffer() this is where image is decrypted then pass it on as before.

add function for setting up decryption
setupCrypt(const uint8_t *cryptKey, size_t cryptAddress, uint8_t cryptConfig, int cryptMode)

and some helper functions
setCryptKey(const uint8_t *cryptKey)
setCryptAddress(const size_t cryptAddress)
setCryptConfig(const uint8_t cryptConfig)
setCryptMode(const int cryptMode)

Hope that all makes sense, I not so great at explaining things or how to contribute


libraries/Update/src/Update.h
```diff
@@ -19,14 +20,25 @@
...
 #define U_FLASH   0
 #define U_SPIFFS  100
 #define U_AUTH    200
 
-#define ENCRYPTED_BLOCK_SIZE 16
+#define ENCRYPTED_BLOCK_SIZE       16
+#define ENCRYPTED_TWEAK_BLOCK_SIZE 32
+#define ENCRYPTED_KEY_SIZE         32
+
+#define U_AES_DECRYPT_NONE       0
```
me-no-dev, line 35: I would suggest to separate the option for app and spiffs and use something like:
```cpp
#define U_AES_DECRYPT_NONE       0
#define U_AES_DECRYPT_AUTO       1
#define U_AES_DECRYPT_ON         2
```

This will make the example and option selection much easier for the users

theeprawn: Sound good with me, I will simplify user options. Probably just go with those three as you listed and then user can decide image tyoe(app, spiffs, ..) with command U_FLASH, U_SPIFFS when calling Upate.begin() function. I make changes & retest. Note auto only works with app image atm, using magic byte check, I will check but dont think spiffs using a special byte at begining of image.
One question, I see only latest Arduino test are successfull, version 14  or earlier fail because #include <hwcrypto/aes.h> not found. I not got earlier version to test with to see if there a simple solution.

libraries/Update/src/Update.h
```diff
@@ -5,6 +5,7 @@
 #include <MD5Builder.h>
 #include <functional>
 #include "esp_partition.h"
+#include <hwcrypto/aes.h>
```
me-no-dev, line 8: change to "aes/esp_aes.h"

theeprawn: Updated files with changes and hopefully made more understandable comments and easier to follow tweak code. I tested with encrypted image files. The image file was encrypted using espsecure, tested with crypt config value 0x0 to 0xf and a range of address values from 0x00000000 throught to 0x00800000. All were succesfully loaded and ran on esp-wroom32 module.
I noticed encrypted addresses above 0x00fffff0 aren't decrypt correctly, it seems espsecure.py tweak code doesnt limit lower 24bit address bits.

torntrousers:
Hi @theeprawn , I'm really interested in this, especially the updating from webserver using HTTPClient that you mentioned but don't have an example of here. Do you have the code for an HTTClient update example that you could add to this PR so I could play around with it?

theeprawn:
Hi @torntrousers, sorry things abit busy here atm. I try write a example for firmware updating, but this PR should work on current HTTPclients that use 'update.h', just setup Crypt key and other settings as necessary, before image starts transfering, by calling 'Update.setupCrypt(OTA_KEY, ... );'

theeprawn:
Hi @torntrousers. Eventually found some time to write an example for using HTTPClient, included simplified server PHP code that can be used to by sketches to update firmware with newer versions. Also included direct download.
Note I removed '_verifyHeader' function from Update which only checks magic byte, but before buffer is decrypted so fails. '_verifyHeader' function wasnt needed because _writeBuffer also checks magic byte a little later after buffer been decrypted.

libraries/Update/src/Update.h
theeprawn: Hi @me-no-dev I've updated Ardunio-esp32 v2.0.1 it was on v1.0.6 before, hence why I was using <hwcrypto/aes.h>. I see the additional boards manage url to json has changed. So now updated to test with 2.0.1, the "aes/esp_aes.h" removed the compiler error. Would it be possible to run workflows to check all is good.

theeprawn:
Hi @VojtechBartoska. Hopefully signed CLA now, sorry I still a bit of newie to how use github for posting code.

VojtechBartoska:
thanks @theeprawn, CLA is fine. Feel free to review @lucasssvaz & @P-R-O-C-H-Y. Thanks

lucasssvaz (approved):

Jason2866:
@theeprawn @lucasssvaz @me-no-dev 
This PR brakes compile with the esp32-c2.
```
/home/runner/.platformio/packages/toolchain-riscv32-esp/bin/../lib/gcc/riscv32-esp-elf/12.2.0/../../../../riscv32-esp-elf/bin/ld: .pio/build/tasmota32c2/lib4b6/libUpdate.a(Updater.cpp.o): in function `UpdateClass::_decryptBuffer()':
/home/runner/.platformio/packages/framework-arduinoespressif32/libraries/Update/src/Updater.cpp:316: undefined reference to `esp_aes_init'
/home/runner/.platformio/packages/toolchain-riscv32-esp/bin/../lib/gcc/riscv32-esp-elf/12.2.0/../../../../riscv32-esp-elf/bin/ld: /home/runner/.platformio/packages/framework-arduinoespressif32/libraries/Update/src/Updater.cpp:325: undefined reference to `esp_aes_crypt_ecb'
/home/runner/.platformio/packages/toolchain-riscv32-esp/bin/../lib/gcc/riscv32-esp-elf/12.2.0/../../../../riscv32-esp-elf/bin/ld: /home/runner/.platformio/packages/framework-arduinoespressif32/libraries/Update/src/Updater.cpp:321: undefined reference to `esp_aes_setkey'
```
CI run where it fails https://github.com/Jason2866/platform-espressif32/actions/runs/7842916607/job/21402162987

lucasssvaz:
@Jason2866 Thanks for the heads up, I'll open a PR to fix it

butaikis:
@theeprawn Hi, can you explain why here 
https://github.com/espressif/arduino-esp32/blob/4123e20a2ca28db70ffeb4db3d5157c8e5b1a058/libraries/Update/src/Updater.cpp#L336
why is _MBEDTLS_AES_ENCRYPT_ used? There is a decryption process going on... (_MBEDTLS_AES_DECRYPT_)

theeprawn:
@butaikis Hi, FYI, copied from Flash Encryption Algorithm section. Hope that helps
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/security/flash-encryption.html#manual-encryption

> AES algorithm is used inverted in flash encryption, so the flash encryption "encrypt" operation is AES decrypt and the "decrypt" operation is AES encrypt. This is for performance reasons and does not alter the effeciency of the algorithm.

butaikis:
> @butaikis Hi, FYI, copied from Flash Encryption Algorithm section. Hope that helps https://docs.espressif.com/projects/esp-idf/en/stable/esp32/security/flash-encryption.html#manual-encryption
> 
> > AES algorithm is used inverted in flash encryption, so the flash encryption "encrypt" operation is AES decrypt and the "decrypt" operation is AES encrypt. This is for performance reasons and does not alter the effeciency of the algorithm.

@theeprawn get it, thanks!

Issue #1114, opened 2018-02-14
How to set SNTP refresh rate?


CelliesProjects:
I have 2 ESP32s running as led controllers.
Both run the same code, and use SNTP to sync the RTC.

Today after running for 10 days there is a time difference between the 2 devices of about 1 minute.

I suspect the SNTP code does not update the time after I set it with the following code:

````cpp
...
...
  configTzTime( "CET-1CEST,M3.5.0/2,M10.5.0/3", "nl.pool.ntp.org" );
...
...
````
How do I check/set the NTP refresh rate?

lbernstone:
From lwip/sntp_opts.h:
/** SNTP update delay - in milliseconds
 * Default is 1 hour. Must not be beolw 15 seconds by specification (i.e. 15000)
 */
#ifndef SNTP_UPDATE_DELAY
#define SNTP_UPDATE_DELAY           3600000
#endif

MarkyAD:
Its set in tools/sdk/include/lwip/apps/sntp/sntp_opts.h and default is 1 hour. An esp32 clock here has been up for over a month and it keeps to within about a second compared to a gps-synced clock sitting next to it. One thing to note is that those pool servers are on a round-robin dns, so you get a different server every time a hostname is looked up. Server offsets and jitter/drift can vary widely, as does network latency, its best to use a local ntp server.

MarkyAD:
Something to test would be using known reachable ntp servers instead of the pool names. I'm not sure how lwip handles sntp dns lookups as far as whether it resolves names with each sync or whether it does so only once when sntp is initialized. If it does it only at init, then its possible one of the esp resolved an unreachable server and ran unsynced the whole time. One minute offset over 10 days sounds like a whole lot but running unsynced would explain that. Running a local ntpsec server here, ntpmon shows my esp32's checking in every hour, and they do keep reasonable time over long periods.

CelliesProjects:
@MarkyAD 
I just reset the slow running device and it is in sync with the other device after the reset.
I compare them to a radio clock.

The device running unsynced would explain the difference.

To test I let some devices run a long time ( a while ago) without syncing them and the result was that cold ESP32s run their RTC slower than warm ESPs.
The difference I noticed while testing was about the same magnitude as the difference I now see. (couple of seconds a day)

I have been testing my network code lately, so the devices were connected/disconnected to WiFi a lot, and I suspect something goes wrong with time server resolution after (repeated) WiFi disconnects.
[...]

PR #5824, opened 2021-11-01
Change sntp default update interval to 6 hours.

CelliesProjects:
## Summary
Change sntp default update interval from 1 to 6 hours.

## Impact
This PR aims at reducing the amount of requests sent to the (free as-in-beer and community run) ntp time servers.
The amount of requests for the ntp project should be a a lot less (per esp32) than with the current settings. 

## Rationale
Just from my home network with 5 esp32 controllers there are more than 100 ntp requests per day with the 1 hour update interval. 
The [ntp project](http://www.ntp.org/) is a community run service and open source projects should strive to minimize their impact on this service.

### DIFF
--- a/tools/sdk/esp32/include/lwip/lwip/src/include/lwip/apps/sntp_opts.h
+++ b/tools/sdk/esp32/include/lwip/lwip/src/include/lwip/apps/sntp_opts.h
@@ -165,10 +165,10 @@
 #endif
 
 /** SNTP update delay - in milliseconds
- * Default is 1 hour. Must not be beolw 60 seconds by specification (i.e. 60000)
+ * Default is 6 hours. Must not be below 60 seconds by specification (i.e. 60000)
  */
 #if !defined SNTP_UPDATE_DELAY || defined __DOXYGEN__
-#define SNTP_UPDATE_DELAY           3600000
+#define SNTP_UPDATE_DELAY           3600000 * 6
 #endif
 
 /** SNTP macro to get system time, used with SNTP_CHECK_RESPONSE >= 2


chegewara:
Did you check what is time drift with internal RTC on esp32?

atanisoft:
This change can not be applied here, it would need to be applied upstream (https://github.com/espressif/esp-lwip and/or https://github.com/espressif/esp-idf) *OR* change `CONFIG_LWIP_SNTP_UPDATE_DELAY` from the current one hour value to your required value.

Changing to six hours though is a bit excessive as typically one hour is what most systems use today.

CelliesProjects:
@atanisoft Thanks for the links. 
Maybe 6 hours is too long between updates, but every hour does seem excessive.  

@chegewara I already had some challenges with sntp and I measured a drift of about 10s per day.
See https://github.com/espressif/arduino-esp32/issues/1114
The main contribution is (as with any clock) the temperature.
Hot clocks run fast, cold clocks run slow.

chegewara:
On esp32 RTC drift is about 5%, which is 18s per hour.

atanisoft:
> Maybe 6 hours is too long between updates, but every hour does seem excessive.

Perhaps change to three hours? You can submit a PR to https://github.com/espressif/esp32-arduino-lib-builder and update all sdkconfig.<target> files and once @me-no-dev approves it will be available in arduino-esp32 via PR after about six hours.

CelliesProjects:
@atanisoft I am a bit confused by the different repos.
The most upstream is the esp-lwip repo? From there it would propagate to idf and arduino?

atanisoft:
`esp-lwip` is where the file you edited comes from, however, in `esp-idf` the config parameter is exposed and configured with a default which overrides the one you edited. `esp32-arduino-lib-builder` is where the sdkconfig files come from that are used to build the binary bits for this repository. LwIP and it's defaults are compiled into the binary and very likely in the default SNTP implementation your change would have little or no impact.

CelliesProjects:
Closing this PR and continuing on https://github.com/espressif/esp32-arduino-lib-builder

PR #5830, opened 2021-11-02
[Examples] WiFi and Ethernet examples organization

pedrominatel:
## Summary

This PR is to change the WiFi and Ethernet examples organization in a separated folder. Two folders were created for WiFi and Ethernet examples.

## Impact

WiFi examples now are in a different folder location. This may cause some impacts on external references to the examples.


pedrominatel:
PTAL @me-no-dev and evaluate the risks and pros and cons of this change.

me-no-dev:
There are no risks other than maybe breaking some links somewhere on the web.

libraries/WiFi/library.properties
```diff
@@ -1,4 +1,4 @@
-name=WiFi
+name=WiFi and Ethernet
```
Rotzbua, line 1: This breaks dependencies, do not change the name.

Rotzbua (changes requested):

pedrominatel:
PTAL @me-no-dev

PR #5844, opened 2021-11-03
Allow use of log_x() in if

FrankBoesing:
if (1) log_e("error") produces a warning ("missing braces") and probably wrong code if no log level is defined.
This fixes that.


me-no-dev:
Nice solution :)

marcboon:
I've seen this solution before, but why 'do {} while(0)' instead of just '{}' ?

PR #5845, opened 2021-11-03
Support additional authorization schemes

Scott8440:
## Summary
The client always appends "Basic" to the authorization header, however there are other auth schemes that can be used: https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication
For example "Bearer" when using OAuth.
This PR adds a `setAuthorizationType` method to the HTTPClient which allows this scheme to be configured by the caller. Authorization type is set to "Basic" by default so this will have no impact on existing usecases.

## Impact
This PR will allow callers to specify the authorization type when making requests, and will not impact any existing usecases as the default is "Basic".


me-no-dev:
Great idea :)

PR #5851, opened 2021-11-05
[Docs] Peripheral list updated to include all supported SoCs

pedrominatel:
## Summary

Peripheral list updated to include all supported SoCs.

## Impact

None.


docs/source/libraries.rst
```diff
@@ -7,47 +7,51 @@ Here is where the Libraries API's descriptions are located:
...
+| TWAI          | No            | No            | No            |                   |
++---------------+---------------+---------------+---------------+-------------------+
+| UART          | Yes           | Yes           | Yes           |                   |
++---------------+---------------+---------------+---------------+-------------------+
+| USB           | Not Supported | Yes           | Yes           | ESP32-C3 only CDC |
++---------------+---------------+---------------+---------------+-------------------+
+| Wi-Fi         | Yes           | Yes           | Yes           |                   |
++---------------+---------------+---------------+---------------+-------------------+
 
 Some peripherals are not available for all ESP32 families. To see more details about it, see the corresponding datasheet.
```
VojtechBartoska, line 56: Table looks better. Here is probably missing the link to the product selector?

pedrominatel: Done. Tks!!

docs/source/libraries.rst
```diff
@@ -7,49 +7,53 @@ Here is where the Libraries API's descriptions are located:
...
-| Ethernet    | Yes         | Yes     | Only for ESP32     |
-+-------------+-------------+---------+--------------------+
-| USB         | Yes         | Yes     | Only for ESP32-S2  |
-+-------------+-------------+---------+--------------------+
++---------------+---------------+---------------+---------------+--------------------+
+| Peripheral    | ESP32         | ESP32-S2      | ESP32-C3      | Comments           |
++===============+===============+===============+===============+====================+
+| ADC           | Yes           | Yes           | Yes           |                    |
++---------------+---------------+---------------+---------------+--------------------+
+| Bluetooth     | Yes           | Not Supported | Yes           |                    |
```
me-no-dev, line 17: I wonder if we should specify what type of BT is supported. C3 is BLE-Only, while the ESP32 also supports classic BT (BTSerial, A2DP, etc.)

docs/source/libraries.rst
```diff
@@ -7,49 +7,53 @@ Here is where the Libraries API's descriptions are located:
...
++---------------+---------------+---------------+---------------+--------------------+
+| DAC           | No            | No            | No            | Planned            |
++---------------+---------------+---------------+---------------+--------------------+
```
me-no-dev, line 20: DAC is supported

docs/source/libraries.rst
```diff
@@ -7,49 +7,53 @@ Here is where the Libraries API's descriptions are located:
...
+| Ethernet      | Yes           | Not Supported | Not Supported |                    |
```
me-no-dev, line 21: I think SPI ETH is supported on S2 and C3. let's check

docs/source/libraries.rst
```diff
@@ -7,49 +7,53 @@ Here is where the Libraries API's descriptions are located:
...
++---------------+---------------+---------------+---------------+--------------------+
+| I2C           | Yes           | Yes           | Yes           |                    |
++---------------+---------------+---------------+---------------+--------------------+
+| I2S           | No            | No            | No            | WIP                |
++---------------+---------------+---------------+---------------+--------------------+
+| LEDC          | Yes           | Yes           | Yes           |                    |
++---------------+---------------+---------------+---------------+--------------------+
+| Motor PWM     | No            | Not Supported | Not Supported |                    |
++---------------+---------------+---------------+---------------+--------------------+
+| Pulse Counter | Yes           | Yes           | Yes           |                    |
```
me-no-dev, line 35: Actually not supported (well, no Arduino API available)

docs/source/libraries.rst
```diff
@@ -7,49 +7,53 @@ Here is where the Libraries API's descriptions are located:
...
++---------------+---------------+---------------+---------------+--------------------+
+| RMT           | Yes           | Yes           | Yes           |                    |
++---------------+---------------+---------------+---------------+--------------------+
+| SDIO/SPI      | Yes           | Yes           | Yes           |                    |
```
me-no-dev, line 39: SDIO and SPI are not the same thing. SDIO AFAIK is supported only on ESP32 and does not have Arduino API

docs/source/libraries.rst
pedrominatel: But we are referring to the internal peripheral only. We can add a note or just keep the internal ones.
What do you think about it?

pedrominatel: PTAL

PR #5856, opened 2021-11-06
Update / HTTP-Update: Add callback for custom LED initialization and LED function

Kampi:

### Checklist

2. [ ] Please provide related links (eg. Issue, other Project, submodule PR..)

## Summary
This pull request contains a small modification for the Update and HTTP Update class to add a function as a parameter to toggle a LED during the update. This is i. e. useful when your device doesn´t use an LED that is directly connected with the ESP, for example, an LED that is connected by a port expander.

As an example:
The LED is connected with a port expander in my design

```
void LED_Init(void)
{
    uint8_t Data;

    Wire.begin(SDA, SCL);

    // Configure the pin for the green LED as output
    Wire.beginTransmission(ADDRESS);

    // Configuration register for port 1
    Wire.write(0x07);

    // '1' represents an input
    // '0' represents an output
    Data = 0xFF;
    Data &= ~(0x01 << LED_GREEN);
    Wire.write(Data);
    Wire.endTransmission();
}

// `true` means the HTTP Update will turn on the LED. `false` means the LED should be turned off
// The LED is connected active-low with the port expander. So the signal must be inverted.
void LED_Write(bool Value)
{
    Wire.beginTransmission(ADDRESS);
    Wire.write(PORT_EXPANDER_ADDR_OUTPUT_1);
    Wire.write(!Value << LED_GREEN);
    Wire.endTransmission();
}

HTTPUpdateResult Return = httpUpdate.update(client, ota, "2.0.0", LED_Init, LED_Write);
```

```
void LED_Init(void)
{
   pinMode(LED_GREEN, OUTPUT);
}

void LED_Write(bool Value)
{
    digitalWrite(LED_GREEN, Value);
}

HTTPUpdateResult Return = httpUpdate.update(client, ota, "2.0.0", LED_Init, LED_Write);
```

## Impact
Fix LED assignment is replaced by two function pointers. The default value for these pointers is `NULL`. An issue only occurs when the "old" style is used. But the necessary modifications are very small

### DIFF
--- a/libraries/Update/src/Update.h
+++ b/libraries/Update/src/Update.h
@@ -22,168 +22,169 @@
 
 #define UPDATE_SIZE_UNKNOWN 0xFFFFFFFF
 
+#define ENCRYPTED_BLOCK_SIZE 16
+
 #define U_FLASH   0
 #define U_SPIFFS  100
 #define U_AUTH    200
 
-#define ENCRYPTED_BLOCK_SIZE 16
+typedef void (*LED_Init_t)(void);
+typedef void (*LED_Write_t)(bool Value);
 
-class UpdateClass {
-  public:
-    typedef std::function<void(size_t, size_t)> THandlerFunction_Progress;
+class UpdateClass
+{
+	public:
+		typedef std::function<void(size_t, size_t)> THandlerFunction_Progress;
 
-    UpdateClass();
+		UpdateClass();
 
-    /*
-      This callback will be called when Update is receiving data
-    */
-    UpdateClass& onProgress(THandlerFunction_Progress fn);
+		/*
+		  This callback will be called when Update is receiving data
+		*/
+		UpdateClass& onProgress(THandlerFunction_Progress fn);
 
-    /*
-      Call this to check the space needed for the update
-      Will return false if there is not enough space
-    */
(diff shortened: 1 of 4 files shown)


libraries/Update/src/Update.h
```diff
@@ -22,168 +22,169 @@
...
-    uint32_t _paroffset;
-    uint32_t _command;
-    const esp_partition_t* _partition;
-
-    String _target_md5;
-    MD5Builder _md5;
-
-    int _ledPin;
-    uint8_t _ledOn;
+typedef void (*LED_Init_t)(void);
```
atanisoft, line 31: reformatting the code makes the diff bloated and a bit more challenging to know exactly what you are changing

me-no-dev:
If you want to toggle LED, why not use the callbacks for start/end and progress? This seems too specific of a case to be added to the Update class

Kampi:
> If you want to toggle LED, why not use the callbacks for start/end and progress? This seems too specific of a case to be added to the Update class

Why do you think this is too specific? You already have a function to toggle an LED during an update to signal a running update. But this solution is limited to an existing GPIO. My solution is a more common solution because you can use a wider range of components to fulfill this function i. e. using LEDs that aren´t connected to the target device or using LEDs that are connected to the target device.

PR #5888, opened 2021-11-14
EEPROM library. EEPROM_FLASH_PARTITION_NAME was never used but now can be used for partition name setup

dukess:
This PR can be used as workaround when user want to set name of EEPROM partition but PlatformIO return error
```
/tmp/tmpw9cod4iu/src/BSB_LAN.ino:510:49: error: invalid conversion from 'const char*' to 'uint32_t {aka unsigned int}' [-fpermissive]
 EEPROMClass EEPROM_ESP("nvs");
                                                 ^
In file included from /tmp/tmpw9cod4iu/src/BSB_LAN.ino:502:0:
/home/runner/.platformio/packages/framework-arduinoespressif32/libraries/EEPROM/src/EEPROM.h:37:5: note:   initializing argument 1 of 'EEPROMClass::EEPROMClass(uint32_t)'
     EEPROMClass(uint32_t sector);
     ^
*** [.pio/build/esp32-evb/src/BSB_LAN.ino.cpp.o] Error 1

``` 

Sketch compiles normally with Arduino Studio.


me-no-dev:
I do not see how your change fixes the error you are getting? The error is about the class constructor and you did not change it's signature at all.

dukess:
> I do not see how your change fixes the error you are getting? The error is about the class constructor and you did not change it's signature at all.

Sorry. It was mistake. I was sure that Platformio already use framework 2.0.x.
But EEPROM_FLASH_PARTITION_NAME definition still nowhere in use so we have three ways:
1. leave it as is
2. delete definition
3. apply this patch

PR #5898, opened 2021-11-17
Add build path sketch

tanakamasayuki:

### Checklist

## Summary
Added the sketch PATH to the build path.

## Impact
can read the .h in the sketch's PATH when compiling the library.
Right now, we can't pass parameters to the library.

### Add path
```
 "-I{build.path}/sketch"
```

### Library code
```
#include "user_config.h"
```

If you prepare user_config.h on the sketch side, you can configure the library.

**However, the problem occurs when the library and sketch have the same file name.**


me-no-dev:
I am not convinced that this is a good approach or that it makes sense to do. A library should never include anything in the sketch, it is an independent piece of code. Do you know of any other Arduino core that does something like that?

h2zero:
This can be quite useful to define macros to configure library build options. There are a few cores around that I have seen which have this functionality but they usually specify a specific file to include, not the whole folder. It is usually an option that can be enabled/disabled in the board menu.

Such as:
```
XXXX.menu.custom_config.disable=Disabled
XXXX.menu.custom_config.enable=Enabled
XXXX.menu.custom_config.enable.compiler.c.extra_flags=-include "{build.source.path}/custom_config.h"
XXXX.menu.custom_config.enable.compiler.cpp.extra_flags=-include "{build.source.path}/custom_config.h"
```

tanakamasayuki:
sorry. I'm not very familiar with other boards.

This is a use case.

## Library using Wi-Fi or BLE
It will be loaded even if it is not currently used in the project.
Therefore, it may be supported by dividing it into three libraries, core, Wi-Fi, and BLE, and loading them individually.

https://github.com/sharpk/LINX/tree/dev/esp8266-cfg-2.1/LabVIEW/vi.lib/MakerHub/LINX/Firmware/Source/libraries

In the above, it is divided into LinxESP8266 and LinxESP8266 WifiListener. With ESP32, Bluetooth Serial can be used, so there are three types.

Now that it is one library integrated into one, the size is large.

https://github.com/tanakamasayuki/LinxESP32

## Read configuration file from library
https://github.com/lvgl/lvgl

https://github.com/lvgl/lvgl/blob/216298d6e991ea27f61423c246901da0afd777cc/src/lv_conf_internal.h#L41

This library loads the user's configuration file.
Only PlatformIO works.

Alternatively, place it in C:\Users\%username%\Documents\Arduino\libraries.

PR #5918, opened 2021-11-22
using ksz8081 only from ESP-IDF 4.4 onwards

vitorrussi:
## Summary
The previous assertion only considerate the existence of ESP-IDF 4.3, but with the ESP-IDF 4.3.1 release this assertion would generate errors. Now only includes from ESP-IDF 4.4 onwards.

## Impact
Now don't generate esp_eth_phy_new_ksz8081 not included error

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -282,7 +282,7 @@ bool ETHClass::begin(uint8_t phy_addr, int power, int mdc, int mdio, eth_phy_typ
             break;
 #endif
         case ETH_PHY_KSZ8081:
-#if ESP_IDF_VERSION => ESP_IDF_VERSION_VAL(4,4,0)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4,4,0)
             eth_phy = esp_eth_phy_new_ksz8081(&phy_config);
 #else
             log_e("unsupported ethernet type 'ETH_PHY_KSZ8081'");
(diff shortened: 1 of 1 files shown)


libraries/Ethernet/src/ETH.cpp
```diff
@@ -282,7 +282,7 @@ bool ETHClass::begin(uint8_t phy_addr, int power, int mdc, int mdio, eth_phy_typ
             break;
 #endif
         case ETH_PHY_KSZ8081:
-#if ESP_IDF_VERSION > ESP_IDF_VERSION_VAL(4,3,0)
+#if ESP_IDF_VERSION_MAJOR >= 4 && ESP_IDF_VERSION_MINOR >= 4
```
me-no-dev, line 285: this will fail for `5.2.0` for example. Use `ESP_IDF_VERSION > ESP_IDF_VERSION_VAL(4,4,0)` instead. In any case, current Arduino does not support anything below 4.4 (and now has a cmake check added to ensure that the version of IDF is correct) which makes this change unnecessary. Has the API for 4.3 changed?

vitorrussi: No, 4.3 hasn't changed but they released 4.3.1, which fails the previous assertion. If it's ensured by cmake check, this assertion can be deleted, but as it's don't make any significant improvement it can be like that. Arduino 2.0.0 don't make this assertion, should I make this change and make a new PR?

atanisoft: If updating to v4.4+ it should be `ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(4,4,0)`

vitorrussi: Updated the commit as instructed

PR #5919, opened 2021-11-22
Update Ticker.cpp

yaroshd81:
fix error when call method active(): `undefined reference to Ticker::active()`

### DIFF
--- a/libraries/Ticker/src/Ticker.cpp
+++ b/libraries/Ticker/src/Ticker.cpp
@@ -56,3 +56,7 @@ void Ticker::detach() {
     _timer = nullptr;
   }
 }
+
+void Ticker::active() {
+  return (bool)_timer; 
+}


bertmelis:
I'm not sure about the behavioural specifics but I don't think this is right.

What should be the result when a one shot timer has already fires?
If your PR it would still be active but I'd say it is not active anymore.

everslick:
Still it is the same implementation as on ESP8266:

```
bool Ticker::active() const                                                     
{                                                                               
    return _timer;                                                              
}    
```

libraries/Ticker/src/Ticker.cpp
```diff
@@ -56,3 +56,7 @@ void Ticker::detach() {
     _timer = nullptr;
   }
 }
+
+void Ticker::active() {
```
everslick, line 60: The return type is wrong!!! Should be `bool` obviously.

everslick:
we could just return `bool esp_timer_is_active(esp_timer_handle_t timer)` and call it a day, though.

me-no-dev:
@everslick want to make a new PR with that? @yaroshd81 seems unresponsive?

PR #5935, opened 2021-11-28
Refactoring TwoWire::requestFrom() headers in Wire.cpp

Maelstrom96:

### Checklist

## Summary
Changing the header so the main TwoWire::requestFrom() definition uses a `size_t` instead of a `uint8_t`, removing the 255 bytes limit on I2C requests.

The esp8266 arduino library for Wire is already using size_t for the function implementation : https://github.com/esp8266/Arduino/blob/076a4edf1e8146ef7420018a0b5b3eadc9acf6af/libraries/Wire/Wire.cpp#L123

This fix is in order to allow us to read 284 bytes, which is necessary for [this project](https://github.com/flaviut/emporia-vue2-reversing/issues/1#issuecomment-980757602).

## Impact
This shouldn't have any impact on existing code using this function.

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -464,7 +464,7 @@ uint8_t TwoWire::requestFrom(uint16_t address, uint8_t quantity, uint8_t sendSto
 /* Added to match the Arduino function definition: https://github.com/arduino/ArduinoCore-API/blob/173e8eadced2ad32eeb93bcbd5c49f8d6a055ea6/api/HardwareI2C.h#L39
  * See: https://github.com/arduino-libraries/ArduinoECCX08/issues/25
 */
-uint8_t TwoWire::requestFrom(uint8_t address, uint8_t len, bool stopBit)
+uint8_t TwoWire::requestFrom(uint16_t address, uint8_t len, bool stopBit)
 {
     return requestFrom((uint16_t)address, (size_t)len, stopBit);
 }
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -445,19 +445,19 @@ void TwoWire::flush(void)
     //i2cFlush(num); // cleanup
 }
 
-size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop)
+size_t TwoWire::requestFrom(uint8_t address, size_t len, bool sendStop)
 {
-    return requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(quantity), static_cast<bool>(sendStop));
+    return requestFrom(static_cast<uint16_t>(address), static_cast<size_t>(len), static_cast<bool>(sendStop));
 }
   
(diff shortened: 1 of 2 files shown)


libraries/Wire/src/Wire.cpp
```diff
@@ -373,7 +373,8 @@ uint8_t TwoWire::requestFrom(uint16_t address, uint8_t size, bool sendStop)
         nonStop = false;
         rxIndex = 0;
         rxLength = 0;
-        err = i2cWriteReadNonStop(num, address, txBuffer, txLength, rxBuffer, size, _timeOutMillis, &rxLength);
+        err = i2cWriteReadNonStop(num, address, txBuffer, txLength, rxBuffer, size, _timeOutMillis, &
```
flaviut, line 376:
```suggestion
        err = i2cWriteReadNonStop(num, address, txBuffer, txLength, rxBuffer, size, _timeOutMillis, &rxLength
```

libraries/Wire/src/Wire.cpp
```diff
@@ -354,7 +354,7 @@ uint8_t TwoWire::endTransmission(bool sendStop)
     return 4;
 }
 
-uint8_t TwoWire::requestFrom(uint16_t address, uint8_t size, bool sendStop)
+size_t TwoWire::requestFrom(uint16_t address, size_t size, bool sendStop)
```
flaviut, line 357: Are we OK doing this from a backwards compatibility perspective? Will folks get compiler warnings/errors if they pass in a `uint8_t` rather than a `size_t`? Or if they try and assign the `size_t` to an existing `uint8_t` variable?

flaviut: Also, the header file ought to be updated.

flaviut (changes requested):

libraries/Wire/src/Wire.cpp
Maelstrom96: Oups, most likely deleted that by accident when doing searches

libraries/Wire/src/Wire.cpp
Maelstrom96: So I don't actually change the full header definition sets, If you check closely, I swapped two of them ([1](https://github.com/espressif/arduino-esp32/pull/5935/files#diff-90b03e94086217edc0bf5abd688554e4165b4d7401b8413dc18c27e3263378d8R357) & [2](https://github.com/espressif/arduino-esp32/pull/5935/files#diff-90b03e94086217edc0bf5abd688554e4165b4d7401b8413dc18c27e3263378d8R462) )so it will come out to the exact same thing in the end.

Maelstrom96: Wait, just noticed the difference in address (uint16_t vs uint8_t) so I guess I will need to modify the headers

flaviut (approved):

me-no-dev:
seems you did not compile your changes and just posted them here :) please test your code before submission next time.

Issue #5781, opened 2021-10-19
WiFiClientSecure: Memory leak when an SSL Handshake fails


Phsacar:
### Hardware:
Board:							ESP32 Dev Module
Core Installation version:			x
IDE name:						Platform.io
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Ubuntu

### Description:

Hi,
I found a leak in ssl_client.cpp:
If the Handshake fails in start_ssl_client(), and an error is returned, the following section is never reached:
```
  if (rootCABuff != NULL) {
        mbedtls_x509_crt_free(&ssl_client->ca_cert);
    }

    if (cli_cert != NULL) {
        mbedtls_x509_crt_free(&ssl_client->client_cert);
    }

    if (cli_key != NULL) {
        mbedtls_pk_free(&ssl_client->client_key);
    }   
```

There does not seem to be a deconstructor in WiFiClientSecure that handles freeing the memory either.

Background: I am using a CA to connect to an encrypted mqtt server; only after freeing ssl_client->ca_cert myself I stopped having memory leaks.

TD-er:
Are you absolutely sure the leak stopped, after manually freeing ca_cert?

I'm looking into this too and I even copied WiFiClientSecure + ssl_client into new classes (with a prefix in the name) to be able to do simple tests by freeing everything that's being accessed via _init functions.
But on average I keep loosing ~1880 bytes per iteration.

Simple test here is to use a CA root certificate (I used the let's encrypt one, but it should not matter) to verify against a self-signed certificate.
I do get `TLS error code: -9984 X509 - Certificate verification failed, e.g. CRL, CA or signature check failed` , which is to be expected. But on each new attempt I loose about 1880 bytes of free heap memory.

I use it for accessing MQTT.
I don't call `loadCACert` as that's obviously leaking memory (see: https://github.com/espressif/arduino-esp32/issues/5826 ) but just call `setCACert`.
Also tested by deleting the WiFiClientSecure object (and implementing the calls to free in the destructor, just to be sure) but does not fix it.

TD-er:
OK, scrap that... Apparently I already fixed it myself, but have changed multiple things at once and did not notice it.

I had switched to no longer deleting the WiFiClientSecure, but still did create it and at the same time I had added to free the certificate data.
So now I also have the memory usage stable too.

[...]

PR #5944, opened 2021-11-29
Fix WiFiClientSecure memory leaks when the connection fails (certificate verification, handshake, etc.)

ppescher:
This should fix issue #5781 and possibly others related to the same problem: the function `start_ssl_client()` does not free memory allocated by mbed-tls library for data structures holding CA certificate, client certificate and private key, in case of early failures, but only when the connection is successful.

The proposed fix adds some cleanup code to `stop_ssl_socket()` so that WiFiSecureClient does not leak memory when it finally calls `stop()`.


ppescher:
Closed because of wrong patch, repo was not in-sync with origin.

PR #5945, opened 2021-11-29
Fix WiFiSecureClient memory leaks when SSL/TLS connection fails

ppescher:
This should fix issue #5781 and possibly others related to the same problem: the function `start_ssl_client()` does not free memory allocated by mbed-tls library for data structures holding CA certificate, client certificate and private key, in case of early failures (handshake, verification, etc.), but only when the connection is successful.

The proposed fix adds some cleanup code to `stop_ssl_socket()` so that `WiFiSecureClient` does not leak memory when it finally calls `stop()`.

### DIFF
--- a/libraries/WiFiClientSecure/src/ssl_client.cpp
+++ b/libraries/WiFiClientSecure/src/ssl_client.cpp
@@ -45,6 +45,8 @@ static int _handle_error(int err, const char * function, int line)
 
 void ssl_init(sslclient_context *ssl_client)
 {
+    // reset embedded pointers to zero
+    memset(ssl_client, 0, sizeof(sslclient_context));
     mbedtls_ssl_init(&ssl_client->ssl_ctx);
     mbedtls_ssl_config_init(&ssl_client->ssl_conf);
     mbedtls_ctr_drbg_init(&ssl_client->drbg_ctx);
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks for finding this!

libraries/WiFiClientSecure/src/ssl_client.cpp
```diff
@@ -280,7 +283,6 @@ int start_ssl_client(sslclient_context *ssl_client, const char *host, uint32_t p
         memset(buf, 0, sizeof(buf));
         mbedtls_x509_crt_verify_info(buf, sizeof(buf), "  ! ", flags);
         log_e("Failed to verify peer certificate! verification info: %s", buf);
-        stop_ssl_socket(ssl_client, rootCABuff, cli_cert, cli_key);  //It's not safe continue.
```
everslick, line 283: Why is this call to stop_ssl_socket() not needed anymore? Was it unnecessary to begin with?

libraries/WiFiClientSecure/src/ssl_client.cpp
```diff
@@ -313,10 +315,20 @@ void stop_ssl_socket(sslclient_context *ssl_client, const char *rootCABuff, cons
         ssl_client->socket = -1;
     }
 
+    // avoid memory leak if ssl connection attempt failed
+    if (ssl_client->ssl_conf.ca_chain != NULL) {
```
everslick, line 319: Testing for NULL is superficial and considered bad taste by some when the pointer is only handed over to free() anyway. The pointers below are not tested and thus this introduces a confusing inconstancy. IMHO.

PR #5959, opened 2021-12-01
Implement DAC based on ESP-IDF API

P-R-O-C-H-Y:
## Summary
This PR is refactoring of DAC HAL in order to use IDF instead of current Register manipulation approach.

Edited dacWrite() to use ESP-IDF api.
Added dacDisable() so there is an option to disable dac channel.

## Impact
None.

### DIFF
--- a/cores/esp32/esp32-hal-dac.c
+++ b/cores/esp32/esp32-hal-dac.c
@@ -13,51 +13,41 @@
 // limitations under the License.
 
 #include "esp32-hal.h"
+#include "soc/soc_caps.h"
 
-#if CONFIG_IDF_TARGET_ESP32
-#include "soc/rtc_io_reg.h"
-#define DAC1 25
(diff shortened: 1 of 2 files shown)


cores/esp32/esp32-hal-dac.c
```diff
@@ -13,51 +13,41 @@
...
 
-#ifndef NODAC
-#include "esp_attr.h"
-#include "soc/rtc_cntl_reg.h"
-#include "soc/rtc_io_periph.h"
-#include "soc/sens_reg.h"
-#include "soc/sens_struct.h"
-#include "driver/dac.h"
+#define DAC1    DAC_CHANNEL_1_GPIO_NUM
+#define DAC2    DAC_CHANNEL_2_GPIO_NUM
```
me-no-dev, line 25: there is no reason to keep those defines anymore :)

PR #5962, opened 2021-12-02
Update esp32-hal-ledc.c

oycq:
IN LINE 209:
if(group) {
LEDC_CHAN(group, channel).conf0.low_speed_update = 1;}
->group 1 is a LOW speed group.

BUT WHY IN LINE 319:
(chan/8)?LEDC_LS_SIG_OUT0_IDX:LEDC_HS_SIG_OUT0_IDX) ?
->group 1 is a HIGH speed group.

I think there is a conflick


me-no-dev:
because `group = chan/8` will produce 1 if it's the low speed group. Code is correct, maybe re-think how  `(chan/8)?LEDC_LS_SIG_OUT0_IDX:LEDC_HS_SIG_OUT0_IDX` will evaluate :)

PR #5970, opened 2021-12-03
Makes BluetoothSerial optional for NimBLE compatibility

4rzael:
## Summary
Currently, BluetoothSerial makes it impossible to run both arduino-esp32 and NimBLE.

### How to reproduce the original problem
esp-idf version: release/v4.4

arduino-esp32 version: 2.0.1

Steps:
* run `idf.py menuconfig`
* go to `Component Configuration / Bluetooth / Bluetooth Host` and set it to `NimBLE`.
* try to compile the project. BluetoothSerial will cause issues, as, if I understood correctly, it requires a proper bluetooth host, not only BLE.

## Impact
This Pull-Request makes it possible to use arduino-esp32 with NimBLE, by making the compilation BluetoothSerial optional. To fix the issue, you thus need to:
* run `idf.py menuconfig`
* go to `Arduino Configuration`
  * enable `Include only specific Arduino libraries`
  * in the submenu, disable `Enable BluetoothSerial`

We could probably, additionally, make that option depend on the BLE host, but I'm not sure how to do so, nor whether it's a good idea to.


CMakeLists.txt
```diff
@@ -201,6 +204,10 @@ set(priv_includes cores/esp32/libb64)
 set(requires spi_flash mbedtls mdns esp_adc_cal wifi_provisioning nghttp)
 set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support openssl bt esp_ipc esp_hid)
 
+if(NOT CONFIG_ARDUINO_SELECTIVE_COMPILATION OR CONFIG_ARDUINO_SELECTIVE_BluetoothSerial)
```
me-no-dev, line 207: this should be done in the source files and not here

4rzael: Okay ! Wasn't sure which approach was the best, I'll change it :)

P-R-O-C-H-Y:
We may close this PR as the issue no longer exists. Tested with Arduino core master as an idf component without any related compilation errors/ warnings.

PR #5973, opened 2021-12-04
Don't clear custom HTTP headers on redirect in HTTPClient

wojtekka:
## Summary
Current implementation clears _headers when request was sent. If the user added custom request headers, they will be lost and the redirected request will not contain them. This commit changes the scope of cleanup (moved from beginInternal() to begin(), changed the amount of data cleared in handleHeaderResponse()), so that the headers survive redirects but don't survive connection reuse.

## Impact
Possible impact on connection reuse. Tested manually to behave properly.


me-no-dev:
Thanks!

PR #5521, opened 2021-08-11
Add dummy test for self hosted GitHub runners

tomassebestik:
This is simple test for newly installed self hosted RPI Github Runners in Espressif Brno site.
Before accepting this PR some selfhosted GH runner must be assigned into project.

Details in https://jira.espressif.com:8443/browse/RDT-43

PR #5983, opened 2021-12-06
Prevent self hosted jobs to run on forks.

Ouss4:
## Summary
Run jobs that require self hosted runners only on the main repository.
https://github.com/espressif/arduino-esp32/pull/5521#issuecomment-986145419
https://github.com/adafruit/arduino-esp32/pull/11

## Impact
Forks only.

### DIFF
--- a/.github/workflows/test_selfhosted_runner.yml
+++ b/.github/workflows/test_selfhosted_runner.yml
@@ -10,6 +10,7 @@ jobs:
   build:
     name: Dummy test - self hosted GHR
     runs-on: self-hosted
+    if: github.repository == 'espressif/arduino-esp32'
     steps:
       - name: Check out repo
         uses: actions/checkout@v2


igrr:
Does this mean that we can't test external contributions on self-hosted runners at all?

Ouss4:
We can as long as they open a pull request in this repo.  In their forks, it won't work.

 - Self-hosted runner of this PR: https://github.com/espressif/arduino-esp32/runs/4441950944?check_suite_focus=true
 - A PR in my fork with self-hosted runner still enabled: https://github.com/Ouss4/arduino-esp32/runs/4437402472?check_suite_focus=true been trying to find a runner since yesterday.
 - And this is another PR in my fork with self-hosted runner skipped: https://github.com/Ouss4/arduino-esp32/runs/4437364409?check_suite_focus=true

igrr:
Got it, thanks for explaining. LGTM then!

Issue #5864, opened 2021-11-09
ESP_ARDUINO_VERSION_PATCH not updated by build script in 2.0.1-RC1 and 2.0.1


tobozo:
Hello,

The new `ESP_ARDUINO_VERSION ` macro can't be used yet to distinguish 2.0.0 from 2.0.1-RC1 and 2.0.1.

Something is probably missing in the build scripts to increment the values.

![image](https://user-images.githubusercontent.com/1893754/140921537-8791f32a-71d7-4f3d-ab4f-f0a17be87da7.png)

![image](https://user-images.githubusercontent.com/1893754/140921502-18aa45bd-99d2-4717-90fb-bf975ece19d1.png)

[edit]

I'm using this macro to distinguish versions:

```C
#if defined ESP_ARDUINO_VERSION_VAL
  #if __has_include("core_version.h") // for platformio
    #include "core_version.h"
  #endif
  #if ESP_ARDUINO_VERSION_VAL(2,0,1) >= ESP_ARDUINO_VERSION || ARDUINO_ESP32_GIT_VER == 0x15bbd0a || ARDUINO_ESP32_GIT_VER == 0xd218e58f || ARDUINO_ESP32_GIT_VER == 0xcaef400f
    #pragma message "Filesystem can create subfolders on file creation"
    #define FS_CAN_CREATE_PATH
  #endif
#endif
```

[edit] updated snippet with 2.0.2 git version

Tomat7:
The same issue on 2.0.2 :-(

VojtechBartoska:
Hi, @Tomat7 and @tobozo we will fix this in 2.0.3. Adding the milestone.

tobozo:
@me-no-dev @VojtechBartoska 

it's remotely related but none of the bundled libraries that were recently modified had their version number increased in their respective library.properties file either 

e.g. SD.h and Wire.h are still at 2.0.0

me-no-dev:
@tobozo are those versions used somewhere?

tobozo:
@me-no-dev those version numbers are visible in the Arduino IDE compilation recap so there's at least an added value for debugging.

It's also a good signal when one of the bundled libraries has changed e.g. SimpleBLE is unchanged since arduino-esp32-2.0.0 whereas BlueToothSerial received changes for arduino-esp32-2.0.2, would it make sense to synchronize the version numbers of the modified libraries with their respective arduino-esp320xxx version instead of using semver?

[...]

PR #5988, opened 2021-12-07
SD Card status check fix

P-R-O-C-H-Y:
## Summary
Edited sd_diskio to work the same as SD_MMC. The ff_sd_status function send command to card to check its status.
If card is disconnected, command will fail.

Fix for #3603

## Impact
Might have very low impact on speed.


FrankBoesing:
This does not work for me. Now I get repetedly, const CRC Errors. It is not usable anymore  :(

VojtechBartoska:
seems related: #5864

PR #5989, opened 2021-12-07
Test with the latest Arduino IDE version.

Ouss4:
## Summary
Download the latest Arduino IDE version.  1.0.4 is quite old.

## Impact
Testing.

### DIFF
--- a/.github/scripts/install-arduino-ide.sh
+++ b/.github/scripts/install-arduino-ide.sh
@@ -82,7 +82,7 @@ if [ ! -d "$ARDUINO_IDE_PATH" ]; then
 fi
 
 function build_sketch(){ # build_sketch <fqbn> <path-to-ino> [extra-options]
-    if [ "$#" -lt 2 ]; then
+  if [ "$#" -lt 2 ]; then
 		echo "ERROR: Illegal number of parameters"
 		echo "USAGE: build_sketch <fqbn> <path-to-ino> [extra-options]"
 		return 1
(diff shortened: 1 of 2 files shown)


Ouss4:
Can someone please approve the workflow here.

me-no-dev:
the ide version is not 1.0.4, it is just "hosted" in esp32-arduino's 1.0.4 release page. this is because Arduino's servers sometimes fail, so the tests fail as well. It happened a few times that there was a bug in the nightly as well.

Please do not change the style of the file. Now it is impossible to see what you have changed.

Ouss4:
> the ide version is not 1.0.4, it is just "hosted" in esp32-arduino's 1.0.4 release page. this is because Arduino's servers sometimes fail, so the tests fail as well. It happened a few times that there was a bug in the nightly as well.

I see, what version is hosted?

> Please do not change the style of the file. Now it is impossible to see what you have changed.

I didn't change the style in the same commit.  The first commit https://github.com/espressif/arduino-esp32/pull/5989/commits/2b2d6f86851ff8d854d7577006037508fd7e637b changes the version and the second commit https://github.com/espressif/arduino-esp32/pull/5989/commits/4759d6de26e386a9b8ef5a626ca8f8cc8d237c94 fixes indentation.  Each are separate and easy to inspect.

me-no-dev:
version is something from 2 years ago. It is worth updating but we should stick to something hosted on GH infrastructure.
What is the reason for the style change?

Ouss4:
> version is something from 2 years ago. It is worth updating but we should stick to something hosted on GH infrastructure.
Alright.

> What is the reason for the style change?

I didn't change the style, I just fixed some indentations. Some were inconsistent (like some `if` and `fi` groups) and others were just not indented correctly.  I put some examples at the end.
That said I probably (unwillingly, as it's my editor's default) also changed some tabs to spaces, the file is also inconsistent in tabs and spaces and uses both for indentation.
In any case, feel free to close this PR, it doesn't have a lot of value to it.

https://github.com/espressif/arduino-esp32/blob/6d400df952b7f07ff05d41bcc43351f67a222aa1/.github/scripts/install-arduino-ide.sh#L85-L89
https://github.com/espressif/arduino-esp32/blob/6d400df952b7f07ff05d41bcc43351f67a222aa1/.github/scripts/install-arduino-ide.sh#L121-L123
https://github.com/espressif/arduino-esp32/blob/6d400df952b7f07ff05d41bcc43351f67a222aa1/.github/scripts/install-arduino-ide.sh#L148-L150

me-no-dev:
Fixing style is something we could look into in the future as a batch job. Also, we can agree on 4 spaces per tab. Could you check your editor's current settings?

Issue #5502, opened 2021-08-05
ESP32-C3 cant read all analog values with analogRead()


kramzarales:
Hello,

I am currently programing an ESP32 C3 devboard which has multiple ADC channels and i would like to read data from 3 different sensors.
The problem is that whenever i use analogRead() with any port it always only shows the measurment on GPIO1.
I have tried it on an ESP32C3-WRoom-02 and also on the ESP32-C3 devmodule 02 with the same result.

I simply connected an external power supply to the pins and when connected to GPIO0 it reads the correct value, but on any other pin it only shows an analog value of 700-900 (i.e. floating pin)
Someone on ESP forum has also reported the same issue

### Hardware:
Board:							ESP32C3 Dev Module
Core Installation version:			2.0.0. alpha1
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                    no
Upload Speed:					115200
Computer OS:                                        Windows 10

### Description:
When trying to read the analog value of any pin with analogRead(pin); it always reads only the value on GPIO0 (ADC1_CHANNAL0). I can can get any other analog-enabled pin to read the value

### Sketch:  
//Just a super basic analog read sketch
int sensorPin = 3;  
int sensorValue = 0;  

void setup() {
  // declare the sensorPin as an input doesnt change anything:

   Serial.begin(115200);
}

void loop() {
  // read the value from the sensor:
sensorValue = analogRead(sensorPin);
  Serial.println(sensorValue);
  delay(100);
}

### Debug Messages:
it compiles and uploads perffectly

wllis:
I got the same problem too !! so what should I do to solve this issue?

RudyFiero:
Same here. I tried to follow the code to see if it was a result of the IDF or if it was something in the Arduino code. I wasn't successful. I wanted to try just the IDF alone to check but have not yet found the time to do that.

kramzarales:
> I got the same problem too !! so what should I do to solve this issue?

I later figured out a solution. You can use the #include <driver/adc.h> library (dont need to download it) and then use the folowing example code:

setup:
adc1_config_width(ADC_WIDTH_BIT_12);
adc1_config_channel_atten(ADC1_CHANNEL_3,ADC_ATTEN_DB_11); //set channel atenuation to 11db (optional)
 ledcSetup(0, 1000, 8); //channel1, 1000Hz frequency, 8 bit resolution
[...]

PR #6005, opened 2021-12-10
add GPIO defines for C3

NiklasVoigt:
Problem: https://github.com/espressif/arduino-esp32/issues/5502#issuecomment-984592239
Solution: https://github.com/tasmota/arduino-esp32/pull/31 from @Jason2866


GYSS1204:
Hello, I changed the code to the code you updated, but the problem still appears.

Jason2866:
@GYSS1204  Try this complete Arduino framework. It is builded on latest IDF4.4 and latest Arduino master including this change. It does work with my C3.
Platformio setup
```
platform                    = https://github.com/tasmota/platform-espressif32.git#feature/arduino-idf-v4.4
platform_packages           = framework-arduinoespressif32 @ https://github.com/Jason2866/esp32-arduino-lib-builder/releases/download/532/framework-arduinoespressif32-release_v4.4-c3ff20355.tar.gz
```

me-no-dev:
I am a bit confused how this change helps (apart from maybe the adc init). For C3 `USE_ESP_IDF_GPIO` is defined as `1`, so it never uses the mux table.

atanisoft:
@me-no-dev Perhaps it would simplify it if all architectures defined `USE_ESP_IDF_GPIO`

me-no-dev:
@atanisoft that will have to be tested once we move all other peripherals to IDF API. It is the end goal though :)

cores/esp32/esp32-hal-gpio.c
```diff
@@ -192,7 +221,7 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
 #endif
 	} else if(mode == ANALOG){
 #if !CONFIG_IDF_TARGET_ESP32C3
-		//adc_gpio_init(ADC_UNIT_1, ADC_CHANNEL_0);
+		adc_gpio_init(ADC_UNIT_1, ADC_CHANNEL_0);
```
P-R-O-C-H-Y, line 224: Is there any reason for uncommenting this line?

PR #6014, opened 2021-12-13
Add some more adafruit boards and fix some board typos

ladyada:
we're making many more esp32-s2 boards - here's some fixes and new definitions


ladyada:
@kattni @jepler ^^

me-no-dev:
@ladyada since @hathach also added some changes, now there are conflicts. Could you please take a look?

hathach:
thanks @me-no-dev , I have resolved the conflict

me-no-dev:
@ladyada do you distribute in any way to EU? I would love a couple of those small S2 boards that I saw you prepare a few days back. Amazon does not have much...

ladyada:
@me-no-dev we ship to the EU all the time :) we should have the QT Py ESP32-S2 in stock in the next week or two - if you can remember, ping me when you see them released & we can get you one

me-no-dev:
Thanks @ladyada :) Customs in my country are PITA, so I was hoping for shipping from EU

ladyada:
we have a large number of stockists around the globe too!
https://www.adafruit.com/distributors

PR #6024, opened 2021-12-15
RMT refactoring based on IDF

### LINKED ISSUE #5905: Change RMT to use ESP-IDF API
This issue is tracking refactoring of RMT to use ESP-IDF API.

- [x] evaluate the current state
- [x] include all related issues
- [x] propose the solution

Related issues:
- #2885

SuGlider:
## Summary
RMT HAL refactoring based on IDF.

## Impact
Improves RMT by adding IDF v4.4 support.
Receiving RMT can handle any size of data.
rmtInit() has a new parameter - RxBufferSize - to hold any number of data when receiving RMT.
rmtWrite() has a new parameter - wait_tx_done - to block writing until sending all data.

## Related links
fix #5905


cores/esp32/esp32-hal-rmt.c
```diff
@@ -23,9 +23,15 @@
...
 
+#include "driver/rmt.h"
+
 /**
  * Internal macros
  */
+
+// USE SOC_RMT_CHANNELS_PER_GROUP
+// LOOK to https://github.com/espressif/esp-idf/blob/master/components/hal/include/hal/rmt_types.h#L32
+
```
me-no-dev, line 34: correct :) https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/include/soc/esp32/include/soc/soc_caps.h#L195-L202
```cpp
#include "soc/soc_caps.h"

#define MAX_CHANNELS (SOC_RMT_GROUPS * SOC_RMT_CHANNELS_PER_GROUP)
#define MAX_DATA_PER_CHANNEL SOC_RMT_MEM_WORDS_PER_CHANNEL
#define MAX_DATA_PER_ITTERATION (MAX_DATA_PER_CHANNEL - 2)
```

It is then chip independent. We should try to have as few chips specific things as possible

cores/esp32/esp32-hal-rmt.c
```diff
@@ -147,26 +192,131 @@ static xSemaphoreHandle g_rmt_block_lock = NULL;
...
+   rmt_get_tx_loop_mode(channel, &loop_en);
+   rmt_get_clk_div(channel, &div_cnt);
+   rmt_get_mem_block_num(channel, &memNum);
+   rmt_get_mem_pd(channel, &lowPowerMode);
+   rmt_get_memory_owner(channel, &owner);
+   rmt_get_rx_idle_thresh(channel, &idleThreshold);
+   rmt_get_status(channel, &status);
+   rmt_get_source_clk(channel, &srcClk);
+
+   ESP_LOGD(tag, "Status for RMT channel %d", channel);
```
me-no-dev, line 232: do not use ESP_LOGx in Arduino. Use log_x

cores/esp32/esp32-hal-rmt.c
```diff
@@ -616,446 +627,20 @@ rmt_obj_t* rmtInit(int pin, bool tx_not_rx, rmt_reserve_memsize_t memsize)
...
-    RMT.conf_ch[channel].conf1.rx_filter_en = 0;
-    RMT.conf_ch[channel].conf1.rx_filter_thres = 0;
-    RMT.conf_ch[channel].conf1.idle_out_lv = 0;     // signal level for idle
-    RMT.conf_ch[channel].conf1.idle_out_en = 1;     // enable idle
-    RMT.conf_ch[channel].conf1.ref_always_on = 0;     // base clock
-
-    RMT.apb_conf.fifo_mask = 1;
+        rmt_config_t config = RMT_DEFAULT_ARD_CONFIG_RX(pin, channel, buffers);
+        ESP_ERROR_CHECK(rmt_config(&config));
+        ESP_ERROR_CHECK(rmt_driver_install(config.channel, 1024, 0)); //, rxBufferSize, 0));
```
me-no-dev, line 638: `rxBufferSize` is ignored here?

me-no-dev:
Overall very good, but here are a few things:
- do not change the API, let's think how we can accommodate `rxBufferSize` while keeping init the same.
- same goes for write. Add a new function `rmtWriteBlocking`
- clean up all unused definitions and variables
- convert the ones left to `soc_caps` (I gave example in a comment)

cores/esp32/esp32-hal-rmt.c
SuGlider: OK. Done.

cores/esp32/esp32-hal-rmt.c
SuGlider: OK. Done.

cores/esp32/esp32-hal-rmt.c
SuGlider: Solved by keeping 1024 to all SoCs. Tested it with ESP32, S2 and C3. 
One possible solution is to use default values in the function signature, but that would require changing to change esp32-hal-rmt.c to cpp instead.

SuGlider:
> Overall very good, but here are a few things:
> 
>     * do not change the API, let's think how we can accommodate `rxBufferSize` while keeping init the same.
> 
>     * same goes for write. Add a new function `rmtWriteBlocking`
> 
>     * clean up all unused definitions and variables
> 
>     * convert the ones left to `soc_caps` (I gave example in a comment)

OK. All done.
rmtInit() is back to the original. I just removed rxBufferSize and it is being initialized with 1024 as buffer size always.
1024 works fine in all the SoCs.

cores/esp32/esp32-hal-rmt.c
```diff
@@ -290,84 +362,37 @@ bool rmtWrite(rmt_obj_t* rmt, rmt_data_t* data, size_t size)
...
-
-        // start the transation
-        //return _rmtSendOnce(rmt, data, MAX_DATA_PER_ITTERATION, true);
-        return _rmtSendOnce(rmt, data, MAX_DATA_PER_ITTERATION, false);
-    } else {
-        // use one-go mode if data fits one buffer 
-        return _rmtSendOnce(rmt, data, size, false);
-    }
+    RMT_MUTEX_LOCK(channel);
+    ESP_ERROR_CHECK(rmt_tx_stop(channel));
```
me-no-dev, line 366: `ESP_ERROR_CHECK` is also an IDF API that boils down to `ESP_LOGx`

SuGlider: Actually ```ESP_ERROR_CHECK``` boils down directly to ```esp_rom_printf()```, which seems to be a ROM function, thus, it should be fine for Arduino. What do you say?

https://github.com/espressif/esp-idf/blob/master/components/esp_common/include/esp_err.h#L121-L127
https://github.com/espressif/esp-idf/blob/master/components/esp_system/esp_err.c#L23-L33

me-no-dev: what it looks like is that it will abort and produce some IDF formatted message. Let's stick to Arduino constructs and log_e if not ESP_OK

SuGlider: OK. Done. 
I removed ```ESP_ERROR_CHECK()``` from IDF function calls that only checked IDF parameters and left it on the function call that really could return error on IDF operations.

Jason2866:
Compile with Arduino Lib Builder does fail with this PR

SuGlider:
> Compile with Arduino Lib Builder does fail with this PR

With IDF 4.4?

Could you please share the log?

Jason2866:
@SuGlider  Here is the log.
[logs_701.zip](https://github.com/espressif/arduino-esp32/files/7758697/logs_701.zip)

SuGlider:
> @SuGlider Here is the log. Yes with IDF4.4. Lib Builder https://github.com/Jason2866/esp32-arduino-lib-builder [logs_701.zip](https://github.com/espressif/arduino-esp32/files/7758697/logs_701.zip)

Thanks @Jason2866!
I just submitted a fixing PR.

@me-no-dev @VojtechBartoska - 
I think that the CI may not have the same compilation options as in the Lib Builder, otherwise this problem should have been caught in the CI testing. Could you please check it?

PR #6025, opened 2021-12-15
Adds support to change LoopTask Stack size

### LINKED ISSUE #6010: Error: Core  1 panic'ed (Unhandled debug exception)
Arduino IDE 1.8.9
ESP32 board 1.0.6 (edit; meanwhile updated to 2.0.1)
default settings

generally my program runs fine, but sometimes, unexpectedly, I get this error - 
but why and what does that mean....?

Guru Meditation Error: Core  1 panic'ed (Unhandled debug exception)
Debug exception reason: Stack canary watchpoint triggered (loopTask) 
Core 1 register dump:
PC      : 0x400d3254  PS      : 0x00060636  A0      : 0x800d3824  A1      : 0x3ffb0030  
A2      : 0x0000000e  A3      : 0x3ffcc4cc  A4      : 0xfffffe77  A5      : 0x00000080  
A6      : 0x00000000  A7      : 0x00000001  A8      : 0x3ffc1c24  A9      : 0x00000008  
A10     : 0xfffffce1  A11     : 0x00000002  A12     : 0x00000002  A13     : 0x00000353  
A14     : 0x0000000a  A15     : 0x3ffb08d0  SAR     : 0x00000011  EXCCAUSE: 0x00000001  
EXCVADDR: 0x00000000  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xfffffff8  

ELF file SHA256: 0000000000000000

Backtrace: 0x400d3254:0x3ffb0030 0x400d3821:0x3ffb0170 0x400d3821:0x3ffb02b0 0x400d3821:0x3ffb03f0 0x400d3821:0x3ffb0530 0x400d3821:0x3ffb0670 0x400d3821:0x3ffb07b0 0x400d3821:0x3ffb08f0 0x400d3821:0x3ffb0a30 0x400d3821:0x3ffb0b70 0x400d3821:0x3ffb0cb0 0x400d3821:0x3ffb0df0 0x400d3821:0x3ffb0f30 0x400d3821:0x3ffb1070 0x400d3821:0x3ffb11b0 0x400d3821:0x3ffb12f0 0x400d3821:0x3ffb1430 0x400d3821:0x3ffb1570 0x400d3821:0x3ffb16b0 0x400d3821:0x3ffb17f0 0x400d3821:0x3ffb1930 0x400d3821:0x3ffb1a70 0x400d3821:0x3ffb1bb0 0x400d493c:0x3ffb1cf0 0x400d683b:0x3ffb1e30 0x400d71a4:0x3ffb1ef0 0x400e352d:0x3ffb1fb0 0x4008a1fe:0x3ffb1fd0

Rebooting...

the program is this one:
https://github.com/dsyleixa/Arduino/tree/master/ESP32_GBox/ESP32_Box023
the error happens sometimes when running the "chess" subroutine.

just to mention:
the chess program (i.e., the move generator) is the same as for my Arduino Due and for my Raspberry Pi, and there it works absolutely fine without any problem ever. So IMO  the issue here on my ESP32 is probably not related to the chess algorithm itself as far as I can see.

PS, to clraify:
sometimes the move generator crashes already at the 2nd or 5th recursive ply after ~50000 move computations or even less,
sometimes it runs fine through the 7th recursive ply by more than 1 or 2 millions move computations and returns a valid and smart move, e.g.:

```
2 ply, searched:         9 
 3 ply, searched:       164 
 4 ply, searched:      1018 .
 5 ply, searched:     10045 ..........
 6 ply, searched:    138116 .........................................................................................................
 7 ply, searched:   1099725 
n b8-c6

0 ply, searched:         1 
 1 ply, searched:         2 
 2 ply, searched:        29 
 3 ply, searched:       466 
 4 ply, searched:      3262 ..
 5 ply, searched:     32422 ................
 6 ply, searched:    251225 
B c1-g5  

0 ply, searched:         1 
 1 ply, searched:         2 
 2 ply, searched:        63 
 3 ply, searched:       741 
 4 ply, searched:      2559 ....
 5 ply, searched:     31058 ..Guru Meditation Error: Core  1 panic'ed (Unhandled debug exception)
Debug exception reason: Stack canary watchpoint triggered (loopTask) 

```

SuGlider:
## Summary
Arduino ```setup()``` and ```loop()``` run under a Task with a fixed Stack size of 8KB.
Users may want to change this size.

This PR adds this possibility by just adding a line of code, as for example:
``` dart
ESP_LOOP_TASK_STACK_SIZE(16384);

void setup() { 
}

void loop() { 
}
```
## Impact
None. It adds a new functionality to ESP32 Arduino.
If ```ESP_LOOP_TASK_STACK_SIZE(newSize);``` is not declared/used, it will compile the sketch with the default stack size of 8KB.

## Related links
fix #6010 

https://github.com/espressif/arduino-esp32/issues/6010#issuecomment-992701658
Thanks @igrr for the suggestion!

### DIFF
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -179,6 +179,11 @@ uint16_t makeWord(uint8_t h, uint8_t l);
 
 #define word(...) makeWord(__VA_ARGS__)
 
+#define ESP_LOOP_TASK_STACK_SIZE(sz) \
+    size_t getArduinoLoopTaskStackSize(void) { \
+        return sz; \
+    }
+    
 unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
 unsigned long pulseInLong(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
 
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -180,7 +180,7 @@ uint16_t makeWord(uint8_t h, uint8_t l);
 #define word(...) makeWord(__VA_ARGS__)
 
 size_t getArduinoLoopTaskStackSize(void);
-#define SET_LOOPTASK_STACK_SIZE(sz) size_t getArduinoLoopTaskStackSize() { return sz;}
+#define SET_LOOP_TASK_STACK_SIZE(sz) size_t getArduinoLoopTaskStackSize() { return sz;}
     
 unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
 unsigned long pulseInLong(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
--- a/cores/esp32/Arduino.h
+++ b/cores/esp32/Arduino.h
@@ -179,10 +179,8 @@ uint16_t makeWord(uint8_t h, uint8_t l);
 
 #define word(...) makeWord(__VA_ARGS__)
 
-#define ESP_LOOP_TASK_STACK_SIZE(sz) \
-    size_t getArduinoLoopTaskStackSize(void) { \
-        return sz; \
-    }
+size_t getArduinoLoopTaskStackSize(void);
+#define SET_LOOPTASK_STACK_SIZE(sz) size_t getArduinoLoopTaskStackSize() { return sz;}
     
 unsigned long pulseIn(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
 unsigned long pulseInLong(uint8_t pin, uint8_t state, unsigned long timeout = 1000000L);
--- a/libraries/ESP32/examples/ArduinoStackSize/ArduinoStackSize.ino
+++ b/libraries/ESP32/examples/ArduinoStackSize/ArduinoStackSize.ino
@@ -6,19 +6,19 @@
     This should be plenty for most general sketches.
 
     There is a way to change the stack size of this task by using
-    SET_LOOPTASK_STACK_SIZE(size);
+    SET_LOOP_TASK_STACK_SIZE(size);
     It will bypass the default stack size of 8KB and allow the user to define a new size.
 
     It is recommend this value to be higher than 8KB, for instance 16KB.
     This increasing may be necessary for the sketches that use deep recursion for instance.
 
-    In this example, you can verify it by changing or just commenting out SET_LOOPTASK_STACK_SIZE();
+    In this example, you can verify it by changing or just commenting out SET_LOOP_TASK_STACK_SIZE();
 */
 
 #define ARDUINO_TASK_STACK_SIZE     (16*1024)       // 16KB
 
 // This sets Arduino Stack Size - comment this line to use default 8K stack size
-SET_LOOPTASK_STACK_SIZE(ARDUINO_TASK_STACK_SIZE);
+SET_LOOP_TASK_STACK_SIZE(ARDUINO_TASK_STACK_SIZE);
 
 void setup() {
   Serial.begin(115200);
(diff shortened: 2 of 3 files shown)


cores/esp32/Arduino.h
```diff
@@ -179,6 +179,11 @@ uint16_t makeWord(uint8_t h, uint8_t l);
 
 #define word(...) makeWord(__VA_ARGS__)
 
+#define ESP_LOOP_TASK_STACK_SIZE(sz) \
```
me-no-dev, line 182: How about we lose the `ESP_` prefix? Since it's not ESP's task, but rather Arduino's, I think we should stay clear of using `ESP_` as prefix for anything that is not ESP specific. @igrr any comment?

me-no-dev:
@SuGlider let's add one example that shows the usage :)

cores/esp32/Arduino.h
igrr: The reason I have suggested this name was that it's Espressif-specific, as in, this macro won't work on any other Arduino core (Atmel, STM, etc).

I'm okay with other naming as long as it's clear that it's an extension we define, not something coming from a library or available on other cores.

igrr:
+1 regarding adding an example. I think you can call some FreeRTOS function like `uxTaskGetStackHighWaterMark(NULL)` from `setup()` to demonstrate that the available stack space is increased when ESP_LOOP_TASK_STACK_SIZE is added.

pedrominatel:
Add to the docs as well.

pedrominatel:
Besides the macro, are we going to add this configuration through the Arduino IDE Tools menu with some predefined options?

cores/esp32/Arduino.h
me-no-dev: If we use more generic name, other cores that run or support RTOS could add the same functionality (it's not rare that we get asked to add something supported elsewhere).

me-no-dev:
> Besides the macro, are we going to add this configuration through the Arduino IDE Tools menu with some predefined options?

Not really. This is niche usage. For all the default should be plenty.

cores/esp32/Arduino.h
SuGlider: My suggestion is to call it 
```dart 
SET_LOOPTASK_STACK_SIZE(sz)
```

me-no-dev: I can get behind that if you split loop and task :D `SET_LOOP_TASK_STACK_SIZE(sz)`

SuGlider: Done.

SuGlider:
> +1 regarding adding an example. I think you can call some FreeRTOS function like `uxTaskGetStackHighWaterMark(NULL)` from `setup()` to demonstrate that the available stack space is increased when ESP_LOOP_TASK_STACK_SIZE is added. Once we have the test framework up and running, this would also make it easier for us to test that this functionality doesn't get broken — we can run this example in CI and check the a value larger than a certain threshold is printed.

Done. Example commited.
@pedrominatel - in the example, I added some commentaries explaining it. Maybe it can be used for documentations.

libraries/ESP32/examples/ArduinoStackSize/ArduinoStackSize.ino
```diff
@@ -0,0 +1,37 @@
...
+    SET_LOOP_TASK_STACK_SIZE(size);
+    It will bypass the default stack size of 8KB and allow the user to define a new size.
+
+    It is recommend this value to be higher than 8KB, for instance 16KB.
+    This increasing may be necessary for the sketches that use deep recursion for instance.
+
+    In this example, you can verify it by changing or just commenting out SET_LOOP_TASK_STACK_SIZE();
+*/
+
+#define ARDUINO_TASK_STACK_SIZE     (16*1024)       // 16KB
```
me-no-dev, line 18: I will suggest to not add a separate define here and just do `SET_LOOP_TASK_STACK_SIZE(16*1024); // 16KB`

SuGlider: done!

PR #6026, opened 2021-12-15
Forces UART Flush() to wait until all bits are sent

### LINKED ISSUE #5877: MODBUS communication randomly broken due to misbehaving Serial.flush() function
Using arduino-esp32 1.0.6, the library ModbusMaster (https://github.com/4-20ma/ModbusMaster) works as intended.
Using arduino-esp32 2.0.0 or 2.0.1 it randomly stops communicating with the slave Modbus device.

I investigated with a scope and these are my findings:
RED LINE = TX signal, BLUE LINE = TX enable signal (tied to DE/NOT RE pins)

1) Succesful transmission - the device replies as expected (Not seen here as I only have two channels) 
picoscope-success

2) First failed transmission - enable signal stays high for a long time, preventing reception of the device answer
picoscope-firstfail

3) Subsequent failed transmission - enable signal is prematurely turned to LOW. The time it stays HIGH varies.
picoscope-fail1
picoscope-fail2
picoscope-fail3

The mentioned library, when a packet is transmitted, turns HIGH the enable signal with preTransmission callback, write data to the serial port, flushes the serial port and then turns the enable signal LOW. Pretty straightforward. I assume Serial TX is buffered, then Serial.flush takes its time until the buffer is empty. 

modbus-fail

Given the recent big overhaul of HardwareSerial and some other issues, I think there is a bug in the new implementation with 2.0.0 and 2.0.1 releases.

This is similar to https://github.com/espressif/arduino-esp32/issues/4603

SuGlider:
## Summary
HardwareSerial flush() was returning before all data was sent out through serial port.
This is a problem to some RS485 libraries that depend on it to signaling.

This PR solves the issue by forcing it to block flush() until all data is sent.

## Impact
None, just fixes an issue.

## Related links
fix #5877

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -140,6 +140,9 @@ uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rx
         ESP_ERROR_CHECK(uart_set_line_inverse(uart_nr, UART_SIGNAL_TXD_INV | UART_SIGNAL_RXD_INV));    
     }
 
+    // Set RS485 half duplex mode on UART.  This shall force flush to wait up to sending all bits out
+    ESP_ERROR_CHECK(uart_set_mode(uart_nr, UART_MODE_RS485_HALF_DUPLEX));
+
     UART_MUTEX_UNLOCK();
 
     uartFlush(uart);


TD-er:
Just curious, why must the uart mode be set to half duplex for all use cases?
This doesn't seem to make sense.
It looks like this will break use cases where a device may be sending data while the ESP is still sending (or starts sending while the other device was still sending data)
Another use case that will very likely be broken by this, is my own project, where I do accept commands to be entered via the terminal and will send logs by the ESP.

It would make more sense to add such a set mode function to the class, or add the mode as a parameter to te `uartBegin` function, so it can be used when needed.

Now we're stuck with a change in mode settings we cannot adjust when you need full duplex.

So can you please explain why this can only be fixed by this, instead of fixing it where it (IMHO) should be fixed: the flush function.

VojtechBartoska:
@SuGlider Just sending a reminder, PTAL on last comment.

SuGlider:
@TD-er / @VojtechBartoska  (FYI)

IDF 4.4 supports some features RS-485 natively. 
Arduino SerialHardware is based on IDF but it doesn't implement RS485 and it doesn't use IDF RS485 features.
Checking the source code of IDF for uart, I found out that this Arduino Issue was related to a feature of IDF that returns from ```uart_wait_tx_done()``` whenever the data was pushed to the UART FIFO. The problem with it is that Arduino ```flush()```  should wait until the data is sent out completely and hold up to the TX FIFO is empty.

The way how IDF forces it is by setting the flag with ```uart_set_mode(uart_nr, UART_MODE_RS485_HALF_DUPLEX)```.
But it doesn't mean that Arduino UART will only work with half or even full duplex. It just makes the IDF driver work in the way Arduino needs it is necessary.

It is **not stuck** to any mode, because Arduino Libraries for MODBUS or RS-485 will actually control it in whatever it wants by software. This change doesn't force UART to act as full or half duplex. 

As said, it just forces IDF UART driver to wait until all data is flushed out when using arduino ```flush()``` function.
The sketches will be able to write and read UART and the same time in a non-blocking way, allowing it to work in full-duplex mode.

mrengineer7777:
This patch feels hacky and I suspect is what's causing issue #6126.

Digging into the IDF libraries:
uart.c::uart_set_mode() calls uart_hal_set_mode()
uart_hal.c::uart_hal_set_mode() calls uart_ll_set_mode()
uart_ll.h::uart_ll_set_mode() calls uart_ll_set_mode_rs485_half_duplex() for 485 half duplex.
uart_ll.h::uart_ll_set_mode_rs485_half_duplex() vs uart_ll_set_mode_normal():
For 485_half_duplex 
  hw->rs485_conf.rx_busy_tx_en = 1;
  hw->rs485_conf.en = 1;
  hw->conf0.sw_rts = 1;
For normal
  hw->rs485_conf.rx_busy_tx_en = 0;
  hw->rs485_conf.en = 1;
I was not able to find source files (compiled binary only?) where rx_busy_tx_en is used.

SuGlider:
@mrengineer7777 
Yes, it's probable. That's exactly what is described in the issue #6122.

SuGlider:
@mrengineer7777 
The issue has been reviewed and fixed by #6133

Issue #5784, opened 2021-10-20
[Docs] Add USB driver library on the documentation


pedrominatel:
Add USB driver library on the documentation.

PR #6036, opened 2021-12-17
[Docs] Added USB documentation

pedrominatel:
## Summary

Added USB documentation.

## Impact

None.

## Related links

Closes: https://github.com/espressif/arduino-esp32/issues/5784


docs/source/api/usb.rst
```diff
@@ -0,0 +1,606 @@
+#######
+USB API
+#######
+
+.. note:: This feature is only supported by ESP32-S2.
```
me-no-dev, line 5: I would suggest to reword this a bit, since very soon we will be adding support to S3 and all mentions for S2 will have to be changed. Maybe something like `This feature is only supported on ESP chips that have USB peripheral, like the ESP32-S2 and ESP32-S3. Some chips, like the ESP32-C3 include native CDC+JTAG peripheral that is not covered here`

docs/source/api/usb.rst
```diff
@@ -0,0 +1,606 @@
...
+
+About
+-----
+
+The **Universal Serial Bus** is a widely used peripheral to exchange data between devices. USB was introduced on the ESP32-S2, supporting both device and host mode.
```
me-no-dev, line 10: keeping ESP32-S2 here makes sense

docs/source/api/usb.rst
```diff
@@ -0,0 +1,606 @@
...
+
+To learn about the USB, see the `USB.org`_ for developers.
+
+USB as Device
+*************
+
+In the device mode, the ESP32-S2 acts as an USB device, like a mouse or keyboard to be connected to a host device, like your computer or smartphone.
```
me-no-dev, line 17: This and any further mention of the S2 should be changed to just ESP32

me-no-dev:
Seems you have added docs only for CDC and MSC so far. Will those show on separate pages? They should :) Also once you cover all exposed classes, maybe we can add doc on how to add a custom device as well.

pedrominatel:
> 

My suggestion is to create a separate page for CDC (`usb_cdc.rst`) and MSC (`usb_msc.rst`) and let the common API as the `usb.rst`. For the custom device, we can add it as a tutorial, not as part of the API.

docs/source/api/usb_cdc.rst
```diff
@@ -0,0 +1,204 @@
+#######
+USB CDC
+#######
+
+About
+-----
+
+USB Communications Device Class API. This class is used to enable communication between the host and the device.
+
+This class is often used to enable serial communication and can be used to flash the firmware on the ESP32-S2 without the external USB to Serial chip.
```
me-no-dev, line 10: ESP32-S2 mentioned here

docs/source/api/usb_cdc.rst
```diff
@@ -0,0 +1,204 @@
...
+
+Example Code
+------------
+
+There are a collection of USB device examples on the project GitHub, including Firmware MSC update, USB CDC, HID and composite device.
+
+HIDVendor
+*********
+
+.. literalinclude:: ../../../libraries/USB/examples/HIDVendor/HIDVendor.ino
```
me-no-dev, line 189: examples here seem unrelated to CDC. We do have CDC examples that you can add here

PR #6040, opened 2021-12-18
fix variant wt32-eth01 (initializer not constant)

mondbaron:
## Summary

This is the subsequent patch to the previous commit, wich introduced some errors (`error: initializer element is not constant`) due to the initialization of const vars using other const vars.

## Impact

The PR fixes a build error in variant wt32-eth01.


me-no-dev:
Thanks :)

Issue #5978, opened 2021-12-06
help with HID braille for open source devices for people with disabilities


discapacidad5:
help with HID braille for open source devices for people with disabilities

Hello, I am developing a device for people with disabilities. I need to use the HID braille protocol but I am not getting anything with HID for esp32, could you help me with a sample code of how I can implement the HID braille protocol?

here are references to HID braille

https://usb.org/sites/default/files/hutrr78_-_creation_of_a_braille_display_usage_page_0.pdf

https://github.com/nvaccess/nvda/blob/00cd67b737bc5b23a6f5e31cf28110b64ebf2fee/devDocs/hidBrailleTechnicalNotes.md

my repository is the following
https://github.com/brailletouch/Brailletouch

discapacidad5:
I found this library for esp32 on arduino https://github.com/T-vK/ESP32-BLE-Keyboard/blob/master/BleKeyboard.cpp  
And this other https://github.com/T-vK/ESP32-BLE-Mouse/blob/master/BleMouse.cpp I think that modifying it could adapt to the HID braille it is in charge of automatically making the description table for the controller you just have to load the list of the hid braille then one in the code adds the keys or functions that you need to use

the libraries that are called to do all the work are  "BLEHIDDevice"  y  "HIDTypes.h" en https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/src/


```
// BleKeyboard.cpp

#include "BleKeyboard.h"     // modify by Blebraille.h

static const uint8_t _hidReportDescriptor[] = {
  USAGE_PAGE(1),      0x01,   // USAGE_PAGE (Generic Desktop Ctrls)  // modify by HID braille   USAGE_PAGE(1),      0x41, 
  USAGE(1),           0x06,          // USAGE (Keyboard)  // modify by   USAGE(1),           0x01,  Braille Display
  COLLECTION(1),      0x01,          // COLLECTION (Application)
  
};
 ..............
```
and

```

// BleMouse.cpp

static const uint8_t _hidReportDescriptor[] = {
  USAGE_PAGE(1),       0x01, // USAGE_PAGE (Generic Desktop)   //  modify by HID braille     USAGE_PAGE(1),      0x41,
  USAGE(1),            0x02, // USAGE (Mouse)  // modify by  el HID  USAGE(1),           0x01,  Braille Display
  COLLECTION(1),       0x01, // COLLECTION (Application)
  USAGE(1),            0x01, //   USAGE (Pointer)
[...]

PR #6043, opened 2021-12-20
HID_BRAILLE_DISPLAY

discapacidad5:
## Summary

23 Braille Display Page (0x41)

Braille display allow visually impaired computer users to read out text using raised pins. The pins are electro-mechanically
activated. These devices also have support for controls that help navigate the computer screen. Typically, braille displays
interface with software known as a screen reader in order to perform this navigation.

## Impact
None

## Related links
#5978


VojtechBartoska:
@discapacidad5 just updated the description to fill the PR template.

PR #6045, opened 2021-12-20
Implement LEDC based on ESP-IDF API

### LINKED ISSUE #5089: PWM ledCWrite(255) or ledcWrite(1023) is not 100% on
### Hardware:
Board: ESP32

### Description:

Unable to light off PWM common anode RGB Led when setting value to latest value of range.

Old known issue of ESP8266 [#5957](https://github.com/esp8266/Arduino/issues/5957) and now seen is [Tasmota](https://github.com/arendst/Tasmota/pull/11786) and my sample test program

The fix is described in comments of official sample [code](https://github.com/espressif/arduino-esp32/blob/404a31f4455d2459dad0b6533b7bfa70347e0044/libraries/ESP32/examples/AnalogOut/ledcWrite_RGB/ledcWrite_RGB.ino#L22)

But from logical working, having value going from 0 to 255 should means that 255 is full scale so always 1 and not having a little pulse. This does now works on ESP8266 with analogWrite(255) and comments have been updated to code according in this [PR](https://github.com/esp8266/Arduino/pull/7456/files)
```
// Per the Arduino docs at https://www.arduino.cc/reference/en/language/functions/analog-io/analogwrite/
// val: the duty cycle: between 0 (always off) and 255 (always on).
// So if val = 0 we have digitalWrite(LOW), if we have val==range we have digitalWrite(HIGH)
```
I know it's not arduino but I think if PWM could work as is `0 (always off) and max range value (always on)` would be less confusing

my `ledOff()`sample function with 256 does works, but should works with 255, much easier if we use an `uint8_t` as scale range.

```
void ledOff() {
  ledcWrite(LED_RED_CHAN, 256);
  ledcWrite(LED_GRN_CHAN, 256);
  ledcWrite(LED_BLU_CHAN, 256);
}
```

P-R-O-C-H-Y:
## Summary
This PR is refactoring of LEDC HAL in order to use IDF instead of current Register manipulation approach.
Fixing duty -> if all bits in resolution are set -> FULL ON

## Impact
None

## Related links
Fixed #5089

### DIFF
--- a/cores/esp32/esp32-hal-ledc.c
+++ b/cores/esp32/esp32-hal-ledc.c
@@ -13,45 +13,25 @@
 // limitations under the License.
 
 #include "esp32-hal.h"
-#include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
-#include "freertos/semphr.h"
-#include "esp32-hal-matrix.h"
 #include "soc/soc_caps.h"
-#include "soc/ledc_reg.h"
-#include "soc/ledc_struct.h"
-#include "driver/periph_ctrl.h"
+#include "driver/ledc.h"
+#include "esp_err.h"
 
-#include "esp_system.h"
-#ifdef ESP_IDF_VERSION_MAJOR // IDF 4+
-#if CONFIG_IDF_TARGET_ESP32 // ESP32/PICO-D4
-#include "soc/dport_reg.h"
-#include "esp32/rom/ets_sys.h"
-#define LAST_CHAN (15)
-#elif CONFIG_IDF_TARGET_ESP32S2
-#include "soc/dport_reg.h"
-#include "esp32s2/rom/ets_sys.h"
-#define LAST_CHAN (7)
-#define LEDC_DIV_NUM_HSTIMER0_V LEDC_CLK_DIV_LSTIMER0_V
-#elif CONFIG_IDF_TARGET_ESP32C3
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-ledc.c
```diff
@@ -72,223 +52,89 @@ xSemaphoreHandle _ledc_sys_lock = NULL;
...
-    LEDC_MUTEX_UNLOCK();
-}
+    ledc_timer_config_t ledc_timer = {
+        .speed_mode       = group,
+        .timer_num        = timer,
+        .duty_resolution  = bit_num,
+        .freq_hz          = freq,
+        .clk_cfg          = LEDC_DEFAULT_CLK
+    };
+    ESP_ERROR_CHECK(ledc_timer_config(&ledc_timer));
```
me-no-dev, line 73: do not use ESP_ERR_CHECK in Arduino. It uses ESP_LOGx inside which we can not control in Arduino

P-R-O-C-H-Y: done

HubertHuber:
ERROR when setting FREQUENCY to less than 1 Hz ---> IntegerDivideByZero

me-no-dev:
ref: https://github.com/espressif/arduino-esp32/issues/6550

PR #6055, opened 2021-12-22
Introduce SSLClient library based on WiFiClientSecure

sgalabov:
## Summary
This PR reworks WiFiClientSecure into SSLClient so that it can be used with any Client-derived class.
The PR was tested with ESP32 WiFi as well as with a GSM module supported by [TinyGSM](https://github.com/vshymanskyy/TinyGSM).

## Impact
Any Client-derived class can get SSL applied to it, e.g. WiFiClient, but also TinyGsmClient and any other clients.
Protocol Clients can then be put on top of SSL, e.g. [PubSubClient](https://github.com/knolleary/pubsubclient) or [ArduinoHttpClient](https://github.com/arduino-libraries/ArduinoHttpClient).
This may also allow for DTLS to be supported as well in the same style in the future, e.g. [CoAP](https://github.com/hirotakaster/CoAP-simple-library) over both WiFi and GSM, etc.


sgalabov:
seems to build fine with the last commit (8b29310)

gonzabrusco:
Is this related to https://github.com/govorox/SSLClient ?

gonzabrusco:
This looks promising.

torntrousers:
Looks interesting to me too. Have you any ideas if it can also work with the mbedtls integration with the Microchip ATECC608 secure element which is presently only available with esp-idf?

torntrousers:
I've been trying this and it works well for me with TinyGSM. Would be great if it could be merged. @me-no-dev ?

me-no-dev:
yup :) in a bit. Wanted to merge S3 support first.

me-no-dev:
There seem to be some issues with S3. I'll have a look. So this seems like a new implementation to replace WiFiClientSecure? Why not modify the original class instead? Having two classes that do the same thing (for the majority of users) will be confusing.

igrr:
I think if WiFiClientSecure could be implemented as a thin wrapper over SSLClient, this would be a nice approach.

First, we would still have just one implementation (SSLClient).
Second, both the old code which uses WiFiClientSecure and the new code with SSLClient would continue to work.

However to achieve this we have to ensure that SSLClient behaves the same way as WiFiClientSecure, or that a backwards-compatible WiFiClientSecure wrapper would be possible to write.

Another point, I think this decision should be synchronized with esp8266/Arduino project. It wouldn't be great to introduce incompatibilities between the two cores intentionally.

igrr:
By the way, there is also https://github.com/OPEnSLab-OSU/SSLClient library which works with many official Arduino boards. @sgalabov do you know of this library? Is choosing the same name intentional?

sgalabov:
@igrr, @me-no-dev, my needs were to be able to use SSL on top of both WiFi and GSM (via TinyGSM), so when I had a closer look at WiFiClientSecure it was logical for me to try and rework it into something that could work with any Client based class and yes, the idea was to go on and convert WiFiClientSecure to work with SSLClient on top of WiFiClient, but it turned out I had to change HTTPClient as well and I didn't just haven't been able to find the time to do that yet (and don't know when/whether I'll be able to, to be honest)... other than that I basically used WiFiClientSecure as a base, including its README and its examples :-)

@igrr, I didn't know about https://github.com/OPEnSLab-OSU/SSLClient so the name is more coincidental than anything else - just thought it fitted in well with its purpose :-)

In the meantime someone posted (although I can't find the message here) about https://github.com/govorox/SSLClient which seems to be more or less the same approach and idea as what I have submitted. I wish I had known about this before - maybe it would have saved me a bit of work...

I thought about the correct place to have such a lib and my opinion at the time was that it belongs with the Arduino support for ESP32 as this is the main target for the lib and would be a logical place to look for something like this for others :-)
Also, I am not sure I will have time to support something like this as open source and would much rather see it integrated in something bigger and with a well established community, which is why I have submitted it here.

gonzabrusco:
> There seem to be some issues with S3. I'll have a look. So this seems like a new implementation to replace WiFiClientSecure? Why not modify the original class instead? Having two classes that do the same thing (for the majority of users) will be confusing.

```
WiFiClient wificlient;
SSLClient  client(wificlient); <--- this would be our new WiFiClientSecure
```
Maybe WiFiClientSecure could be just a wrapper of SSLClient(WiFiClient).

gonzabrusco:
For this to be complete as @sgalabov said, HTTPClient should use Client instead of WiFiClient (like the standard ArduinoHttpClient library). That way you could use an SSLClient on top of a WiFiClient in the included HTTPClient from this repository.

gonzabrusco:
[SSLClient.zip](https://github.com/espressif/arduino-esp32/files/8373096/SSLClient.zip)

This would be the same @sgalabov uploaded but with the latest WiFiSecureClient changes incorporated (specifically, certificate bundle).

The only example that does not work is the one relying on HTTPClient because it was developed to work with WiFiClient and not with Client

sgalabov:
It would probably be worth it to have a discussion on where and when this needs to go and then take the time to rework it based on the current WiFiClientSecure and rework everything else that needs to be touched, so that WiFiClientSecure becomes equivalent to SSLClient on top of WiFiClient and HTTPClient (at least) is taught about that. Probably SSLClient can be part of the WiFiClientSecure lib, although I personally tend to prefer them to be separate as it would be more visible and understandable what each one does...
What do you guys think?

torntrousers:
Sounds good to me. I'm doing some more testing with SSLClient this week as it looks like there might be a subtle difference it the way it behaves compared to WiFiClientSecure which I'm trying to get to the bottom of.

me-no-dev:
@sgalabov I have fixed and updated your branch. Please pull the changes on your end.

sgalabov:
> @sgalabov I have fixed and updated your branch. Please pull the changes on your end.

not sure what you mean - I can already see your changes in my branch :-)

me-no-dev:
> > @sgalabov I have fixed and updated your branch. Please pull the changes on your end.
> 
> not sure what you mean - I can already see your changes in my branch :-)

I guess your git client already pulled them (mine does too)

torntrousers:
> I'm doing some more testing with SSLClient this week as it looks like there might be a subtle difference it the way it behaves compared to WiFiClientSecure which I'm trying to get to the bottom of.

What this looks like is that when an SSLClient.connect fails then SSLClient calls stop() which frees up all the resources and you can't use that SSLClient instance again and if you try it fails "Client not initialized" from line 91 in ssl_client.cpp, whereas with WifiClientSecure you can reuse it after a failed connection.

gonzabrusco:
> "Client not initialized" from

Yes the behaviour is not the same. Because when you create an SSLClient, you pass the client for the connection. But when you stop() the SSLClient, it clears the passed client internally (in `stop_ssl_socket()` you will see `ssl_client->client = nullptr;` at the end). In WiFiClientSecure this worked because it didn't use clients, it used sockets and created the connections itself. But SSLClient needs a new client to work with if you stop it (you can use the operator =). 

Not sure if this is the best approach, but it is what it is haha. This SSLClient idea is great but it need more work.

torntrousers:
Another thing, SSLClient doesn't initialse ```_use_insecure```, it should have ```bool _use_insecure = false;``` in SSLClient.h, otherwise the first requests i do get "WARNING: Skipping SSL Verification. INSECURE!".

lucasssvaz:
@sgalabov Would it be possible to update this PR to 3.0.0 ?

VojtechBartoska:
@sgalabov will you be able to update this PR? Also please sign CLA. Thanks

Issue #5826, opened 2021-11-01
Memory leak on multiple calls to loadCACert and others on WiFiClientSecure


sguarin:
If you have a system that can accept certs configuration, then multiple calls to loadCACert or related functions will not reutilize the memory assigned.

https://github.com/sguarin/arduino-esp32/blob/caa8d07aafa04441bb85e6046a795249c01d9e39/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L304

_streamLoad()
allocates memory for new certificates. A better approach should be using realloc().

https://github.com/sguarin/arduino-esp32/blob/caa8d07aafa04441bb85e6046a795249c01d9e39/libraries/WiFiClientSecure/src/WiFiClientSecure.cpp#L290

sguarin:
This temporary fix work for me.
But I think the memory pointer should not be static and should live on the class.

https://github.com/espressif/arduino-esp32/compare/master...sguarin:fix_streamLoad?expand=1

TD-er:
Just a repeat of the comment I placed with that commit:

You're not checking the return value of `realloc`.
If the return value is not a nullptr, you should assign it to `*destPtr`.
If it is a nullptr, it depends on the C++ version whether it was a failed alloc or not changed (e.g. size was the same)

And no need to check if `*destPtr` is a nullptr, as `realloc` called with a nullptr acts like malloc.

sguarin:
You are right. Thanks!

PR #6062, opened 2021-12-25
Fix memory leak on stream load when updating CA Certificate

sguarin:
## Summary
This PR fix a memory leak when reconfiguring CA Certificate with multiple calls to WiFiClientSecure::setCACert()

## Impact
When reconfiguring multiple times a CA certificate, or if your systems refresh the connection calling multiple times this function, the system will run out of memory.

## Related links
https://github.com/espressif/arduino-esp32/issues/5826


mrengineer7777:
@VojtechBartoska The memory leak mentioned here is resolved in [PR #6387](https://github.com/espressif/arduino-esp32/pull/6387).

_I agree with adding `_use_insecure = false;` to setCACert(), setCertificate(), setPrivateKey().  setInsecure is used in setInsecure()._

@me-no-dev  setInsecure() is a dangerous function as it doesn't free memory allocated in loadCACert().  But....

Calling loadCACert() after setCACert() may free a pointer that doesn't belong to this module.  In my opinion loadCACert(), loadCertificate(), loadPrivateKey() should all have their own private storage variables.  Then call free() for each variable in setInsecure() and ~WiFiClientSecure().

lucasssvaz:
Changes already implemented.

PR #6076, opened 2021-12-28
Add variantInit setups for adafruit boards

ladyada:
this will let us activate hardware on boot - avoid customer confusion :)


me-no-dev:
Done :)

PR #6084, opened 2021-12-30
Print.flush() - Arduino API conformance

jandrassy:
In ARDUINO 1.0 (2011.11.30) Arduino changed Serial.flush() to "wait for outgoing data to be transmitted". In following years other classes derived from Stream (or Print) implemented flush() as "send the buffer out". In ARDUINO 1.8.4 (2017.08.23) they moved flush() from Stream to Print in the prominent Arduino AVR core. For backward compatibility of flush() in Print they changed the pure virtual function to empty implementation. The [ArduinoCore-API](https://github.com/arduino/ArduinoCore-API) recommended for new cores or major updates has flush() in Print too.

Most major cores already implemented this change. This PR moves flush() from Stream.h to Print.h in this core.

### DIFF
--- a/cores/esp32/Print.h
+++ b/cores/esp32/Print.h
@@ -108,6 +108,9 @@ public:
     size_t println(const Printable&);
     size_t println(struct tm * timeinfo, const char * format = NULL);
     size_t println(void);
+    
+    virtual void flush() { /* Empty implementation for backward compatibility */ }
+    
 };
 
 #endif
--- a/cores/esp32/Stream.h
+++ b/cores/esp32/Stream.h
@@ -48,7 +48,6 @@ public:
     virtual int available() = 0;
     virtual int read() = 0;
     virtual int peek() = 0;
-    virtual void flush() = 0;
 
     Stream():_startMillis(0)
     {


bxparks:
Thank you, I was about to create a PR to do this exact thing. The missing `Print::flush()` on the ESP32 core has been a minor annoyance to me for several years. I've been doing things like this to get around this inconsistency:

```C++
class Subclass : public Print {
  public:
  #if defined(ESP32)
    void flush() {...}
  #else
    void flush() override {...}
  #end

[...]
};
```

But I still cannot use `Subclass::flush()` polymorphically through the `Print` pointer or reference on the ESP32.

For reference, I am aware of the following Cores which *do* implement a `virtual void Print::flush()`:

* https://github.com/arduino/ArduinoCore-avr/blob/master/cores/arduino/Print.h#L90 
* https://github.com/arduino/ArduinoCore-API/blob/master/api/Print.h#L92
* https://github.com/esp8266/Arduino/blob/master/cores/esp8266/Print.h#L116
* https://github.com/stm32duino/Arduino_Core_STM32/blob/main/cores/arduino/Print.h#L110
* https://github.com/sparkfun/Arduino_Boards/blob/main/sparkfun/samd/cores/arduino/Print.h#L89
* https://github.com/PaulStoffregen/cores/blob/master/teensy/Print.h#L51
* https://github.com/PaulStoffregen/cores/blob/master/teensy3/Print.h#L62
* https://github.com/PaulStoffregen/cores/blob/master/teensy4/Print.h#L62
* https://github.com/MCUdude/MightyCore/blob/master/avr/cores/MCUdude_corefiles/Print.h#L65 
* https://github.com/MCUdude/MegaCore/blob/master/avr/cores/MCUdude_corefiles/Print.h#L65 
* https://github.com/MCUdude/MiniCore/blob/master/avr/cores/MCUdude_corefiles/Print.h#L65 

I am aware of only these Cores which do *not* implement a `Print::flush()`:

* https://github.com/SpenceKonde/ATTinyCore/blob/master/avr/cores/tiny/Print.h
* https://github.com/MCUdude/MicroCore/blob/master/avr/cores/microcore/Print.h
* https://github.com/adafruit/Adafruit_nRF52_Arduino/blob/master/cores/nRF5/Print.h

As far as I can tell, there are only 3 classes in the ESP32 core which inherit directly from `Print`. The rest inherit from `Stream` which will not be effected by this PR because `Stream` already defines a `virtual Stream::flush()`.

* `cores/esp32/Server.h`
* `libraries/AsyncUDP/src/AsyncUDP.h` (the `AsyncUDPMessage` class)
* `libraries/USB/src/USBHIDKeyboard.h`

The result is that these classes will consume a handful more bytes of flash memory, and maybe 4 extra bytes of static memory for the extra slot in the vtable? I hope this will not be the reason for rejecting this PR.

jandrassy:
and Arduino-API with Print.flush() is in:
https://github.com/arduino/ArduinoCore-samd
https://github.com/arduino/ArduinoCore-megaavr
https://github.com/arduino/ArduinoCore-mbed
https://github.com/earlephilhower/arduino-pico

me-no-dev:
Thanks!

PR #6088, opened 2022-01-01
consolidate duplicate libnames

MarkusAD:
## Summary
compiler.c.elf.libs.esp32 in platform.txt has grown to 425 libraries, nearly all are
duplicates. consolidate them to the unique existing 107 libnames. link order appears
to make no difference.

## Impact
None

## Related links
None


atanisoft:
This will need to be done over on https://github.com/espressif/esp32-arduino-lib-builder instead as the entries you have updated will be overwritten as part of the next esp-idf sync.

MarkusAD:
Ok, thanks.

PR #6091, opened 2022-01-02
Remove useless parameter check for unit < 0 in BTScanResultsSet.cpp

iltis42:
## Summary
Remove useless parameter check:

BTScanResultsSet.cpp:67:8: warning: comparison of unsigned expression < 0 is always false [-Wtype-limits]
  if (i < 0)
      ~~^~~
[ 83%] Linking C static library libarduino-esp32.a

## Impact
Please describe impact of your PR and it's function.

## Related links

### DIFF
--- a/libraries/BluetoothSerial/src/BTScanResultsSet.cpp
+++ b/libraries/BluetoothSerial/src/BTScanResultsSet.cpp
@@ -64,9 +64,6 @@ int BTScanResultsSet::getCount() {
  * @return The device at the specified index.
  */
 BTAdvertisedDevice* BTScanResultsSet::getDevice(uint32_t i) {
-	if (i < 0)
-		return nullptr;
-
 	uint32_t x = 0;
 	BTAdvertisedDeviceSet* pDev = &m_vectorAdvertisedDevices.begin()->second;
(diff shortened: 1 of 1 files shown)


VojtechBartoska:
@iltis42 Thanks for your PR. Description edited, please consider filling out PR description more specifically next time. It help us to understand more to your changes.

mrengineer7777:
This commit changes functionality: BTScanResultsSet::getDevice() will not be able to check for negative values.  Consider:
BTScanResultsSet::getCount() returns int.
BTScanResultsSet::dump() passes int to getDevice().

I believe the correct fix here is to change the function signature to getDevice(int32_t i).  It should also be updated in BTScan.h.

iltis42:
> This commit changes functionality: BTScanResultsSet::getDevice() will not be able to check for negative values. Consider: BTScanResultsSet::getCount() returns int. BTScanResultsSet::dump() passes int to getDevice().
> 
> I believe the correct fix here is to change the function signature to getDevice(int32_t i). It should also be updated in BTScan.h.

In an isolated view of this change, there is no change of functionality. But fully agree here. Changing the signature will make this change work. Bingo!

iltis42:
Closed

iltis42:
New RP is here:
https://github.com/espressif/arduino-esp32/pull/6109

Issue #3646, opened 2020-01-19
Feature Request: ESP32 SSL/TLS Certificate Store


yknivag:
Is it possible to handle https connections on the ESP32 **without** knowing in advance the root CA for the service one is trying to connect to?

For example, on the ESP8266 this is possible using `CertStoreBearSSL` (see https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266WiFi).

This is particularly important for https services which redirect from one domain to another where one would need to know all the certificates in advance on the ESP32 whereas on the ESP8266 one needs only to add `certs.ar` to SPIFFS and it is possible to connect to almost any https service.

I may have missed a way that this is already possible but I haven't been able to find one. How complex would it be to port the `CertStoreBearSSL` functionality to the ESP32?

lbernstone:
Porting BearSSL would be a very significant task, particularly to use the hardware acceleration.  However, if you attempt to create an https connection in WiFiClientSecure with no CA/cert information, it will make permissive requests, and should accept any certificate offered.  Perhaps posting your actual issue (with code and logs) instead of your proposed fix can resolve the root problem.

yknivag:
> make permissive requests, and should accept any certificate offered.
That really isn't a solution though is it?  That's just leaving a key under the doormat because you keep losing your keys - you may as well just leave the door open.

What I want to be able to do is connect to any arbitrary https web service from within a sketch without having to load an individual root CA for each domain and keep track of which is which.

I can do that on the ESP8266 using `CertStoreBearSSL` as the `certs.ar` file allows any certificate to be authenticated against it's root CA without knowing which domains are required at the time of writing the code.

I have no code to post as I have, as yet, found not a single example nor any documentation that proposes any method of securely hitting an arbitrary https web service.  I have code that works on the ESP8266 (and there are many official examples of it for the ESP8266) but, whilst I can happily follow the examples and documentation to use the `CertStoreBearSSL` library I will readily admit that I lack the knowledge to recreate it's functionality.

Essentially I would like to recreate (this ESP8266 sketch)[https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/examples/BearSSL_CertStore/BearSSL_CertStore.ino] on the ESP32 and have it function with the same level of security.

When I replace `#include <ESP8266WiFi.h>` with `#include <WiFi.h>` I get the following error in the Arduino IDE:

[...]

PR #6106, opened 2022-01-07
Add certificate bundle capability to WiFiClientSecure

esp32wrangler:
## Summary
Enable usage of the ESP32 IDF's certificate bundle for WiFiClientSecure connections.

Adds the ability to load a bundle or root certificates and use them for authenticating SSL servers.

Intends to fulfill the feature request https://github.com/espressif/arduino-esp32/issues/3646

## Impact
The changes impact only the WiFiClientSecure component. Unless the user invokes WiFiClientSecure::setCACertBundle, the behavior of WiFiClientSecure remains identical. If that function is called, and no CA certificate is provided to WiFiClientSecure, then the bundle will be used instead.

The actual certificate bundle parsing and authentication callback functionality is stable, proven ESP IDF code. Only the glue code that connects WiFiClientSecure to the bunde code is bespoke.

New code based on work from Onno-Dirkzwager, Duckle29, kubo6472, meltdown03, kinafu and others in issue 3646.

## Related links
- https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_crt_bundle.html

- libraries/WiFiClientSecure/README.md


libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -263,6 +265,17 @@ void WiFiClientSecure::setCACert (const char *rootCA)
...
 }
 
+ void WiFiClientSecure::setCACertBundle(const uint8_t * bundle)
+ {
+    if (bundle != NULL)
+    {
+        esp_crt_bundle_set(bundle);
+        _use_ca_bundle = true;
+    } else {
+        _use_ca_bundle = false;
```
me-no-dev, line 275: shouldn't any old bundle be freed here?

esp32wrangler: In the normal usage case, the "bundle" variable points to a static string in flash (see README), so trying to free it would not make sense, and also result in a crash (note that the variable is passed as a const, so it should not be in WiFiClientSecure's contract to free it in any case). Internally esp_crt_bundle_set frees s_crt_bundle.crts, which is the only dynamically allocated element in the mechanism (it is the index created for searching the bundle during authentication).

me-no-dev: The only reason I am asking is because I see in `esp_crt_bundle_set` you indeed free, but if bundle is `NULL`, `esp_crt_bundle_set` would not be called.

esp32wrangler: You are correct, it makes sense to clean up the old bundle index when a NULL bundle is attached. I added an esp_crt_bundle_detach() call on that branch.

me-no-dev:
Thanks @esp32wrangler :) Merged!

PR #6109, opened 2022-01-07
Consistently change device index to singed integer in BluetoothSerial lib

iltis42:
## Summary
Change device index consistently for getCount() parameter list to signed integer type.

## Impact
Get rid of compiler warning as below and get parameter check work that won't work with unsigned type:

BTScanResultsSet.cpp:67:8: warning: comparison of unsigned expression < 0 is always false [-Wtype-limits]
  if (i < 0)
      ~~^~~
[ 83%] Linking C static library libarduino-esp32.a

## Related links
This is related to discussion under:
https://github.com/espressif/arduino-esp32/pull/6091

### DIFF
--- a/libraries/BluetoothSerial/src/BTScan.h
+++ b/libraries/BluetoothSerial/src/BTScan.h
@@ -24,14 +24,14 @@ public:
 
     virtual void        		dump(Print *print = nullptr);
     virtual int         		getCount();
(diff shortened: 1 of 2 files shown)


mrengineer7777:
Reviewed code changes.  Looks good to me.  Just need to wait for a maintainer to commit.

me-no-dev:
Thanks @iltis42 :)

PR #6110, opened 2022-01-07
Add Core Debug Level option to Tools menu for all boards in boards.txt

gcweeks:
## Summary
Using an ESP32 board (namely an Olimex Gateway) I ran into an issue in an Ethernet library that could be debugged simply by setting the core debug level. Looking into it, most boards had this option configurable from the Arduino Tools menu, but a few (including mine) were missing that menu for some reason. I've added the menu for my board, but in addition I've added it for all boards in "boards.txt" because I couldn't think of a reason someone with any board wouldn't want this feature, but let me know if there is any reason to the contrary.

I also cleaned up some messy whitespace, sorry if that makes the commit harder to read.

## Impact
All boards offered by the Arduino ESP32 library will now provide a "Core Debug Level" option under the Arduino Tools menu that can be set to "None", "Error", "Verbose", etc.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -878,6 +878,7 @@ esp32wroverkit.menu.DebugLevel.verbose=Verbose
 esp32wroverkit.menu.DebugLevel.verbose.build.code_debug=5
 
 ##############################################################
+
 tinypico.name=UM TinyPICO
 
 tinypico.upload.tool=esptool_py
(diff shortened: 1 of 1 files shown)


me-no-dev:
Thanks @gcweeks :) looks great. Generally board options are left to the people adding them (users or manufacturers alike) It is also possible that the menu did not exist at the point of adding the board, but that is rather rare case.

PR #6113, opened 2022-01-08
WIFI STA AutoReconnect fix for assocfail reason,  Proposed fix for #2501

tablatronix:
Reliability fix for autoReconnect when assoc_fail, autoReconnect did not work before for these failures.

## Summary
Changes behavior of WIFI_REASON_ASSOC_FAIL event when autoReconnect is set, removes WIFI_REASON_ASSOC_FAIL/WL_CONNECT_FAILED so retry waitforconnectresult loop stays active for the retry, was not working before.

## Impact
Same behavior if autoReconnect is not enabled, suppress WL_disconnect when it is. 
Possible event loop, similar to other events, I see no special checks for avoiding, whole event method need a rewrite, currently prone to bugs

## Related links
#2501


me-no-dev:
Thanks @tablatronix :)

PR #6132, opened 2022-01-13
ci: Miscellaneous improvements

Ouss4:
1. Cache downloaded tools.
2. Cancel duplicate jobs.
3. Use current repo when linking the core, this allows users to test their code when creating PRs against their forks (for instance an error in an example sketch will build successfully in a fork workflow.)
4. Cache Arduino IDE. 
5. Add `workflow_dispatch` to be able to trigger the workflow manually.


.github/workflows/push.yml
```diff
@@ -22,6 +26,14 @@ jobs:
     - uses: actions/setup-python@v2
       with:
         python-version: '3.x'
+    - name: Cache tools
+      id: cache-linux
+      uses: actions/cache@v2
+      with:
+        path: |
+          ./tools/dist
+          /home/runner/arduino_ide
```
igrr, line 35: Is it possible to use some environment variable instead of `/home/runner`?

Ouss4: This is what I was referring to in the description.  I couldn't find an env variable for this.

.github/workflows/push.yml
```diff
@@ -22,6 +26,14 @@ jobs:
...
+        key: ${{ runner.os }}-${{ hashFiles('package/package_esp32_index.template.json', 'tools/get.py') }}
```
igrr, line 36: Probably also need to hash `.github/scripts/install-arduino-ide.sh` since that determines the contents of `/home/runner/arduino_ide`

.github/workflows/push.yml
Ouss4: I tried $HOME,it returns the home directory (i.e. /home).
I think what can be done here is to export the variable from our script as a Github Action output.

igrr: Oh, right, i understand that point now!
It seems that this action supports `~` (tilde) expansion, https://github.com/actions/cache/issues/664#issuecomment-1004662677.

.github/scripts/install-arduino-ide.sh
```diff
@@ -143,18 +143,18 @@ function count_sketches() # count_sketches <examples-path> <target-mcu>
...
-function build_sketches() # build_sketches <fqbn> <target-mcu> <examples-path> <chunk> <total-chunks> [extra-options]
+function build_sketches() # build_sketches <fqbn> <examples-path> <chunk> <total-chunks> [extra-options]
 {
     local fqbn=$1
-	local target="$2"
     local examples=$3
     local chunk_idex=$4
     local chunks_num=$5
     local xtra_opts=$6
+	  local target=`echo $fqbn | cut -d":" -f3`
```
igrr, line 153: nitpick, indentation

me-no-dev:
it is not necessary that the target mcu equals the third part of FQBN. This in our case currently is true, but if we want to test agains different boards (we will in near future) then scripts will not be able to tell what is the actual MCU and decide to skip the test.

Ouss4:
> it is not necessary that the target mcu equals the third part of FQBN. This in our case currently is true, but if we want to test agains different boards (we will in near future) then scripts will not be able to tell what is the actual MCU and decide to skip the test.

How would this look like? `espressif:esp32:esp32-devkitc` for instance?  What would the second part represent in this case?

me-no-dev:
first part is "manufacturer"m, second is "core" and third is the board. manufacturer and core come from the path of installation: `~/Arduino/hardware/[manufacturer]/[core]`

Ouss4:
> first part is "manufacturer"m, second is "core" and third is the board. manufacturer and core come from the path of installation: `~/Arduino/hardware/[manufacturer]/[core]`

Okay, thanks.  If the boards we want to test against are Espressif's, we can still extract the chip name from the board's, but other manufacturers won't necessary follow the sane naming conventions (and I think we also have few special cases).
I will drop this commit.

README.md
```diff
@@ -45,3 +45,4 @@ Finally, if you are sure no one else had the issue, follow the **Issue template*
 We welcome contributions to the Arduino ESP32 project!
 
 See [contributing](https://docs.espressif.com/projects/arduino-esp32/en/latest/contributing.html) in the documentation for more information on how to contribute to the project.
+TRIGGER
```
me-no-dev, line 48: would you please remove this so that we can merge your PR :)

Ouss4: This is just a dummy commit to trigger the workflow and see if that cache was restored correctly.  I'll drop it in a second.
In my last push, when the IDE was cached I saw that the sketches were skipped and not built.  I dropped the IDE commit, I'll reapply it and see what's going on.

Ouss4: I added `workflow_dispatch` to be able to trigger the workflow manually, this hack won't be needed anymore (i.e. got tired of force pushing...)

.github/workflows/push.yml
Ouss4: Tilde expansion seems to work.  I tried `$HOME` before but that didn't work.

PR #5549, opened 2021-08-16
Implements UART SerialHardware Refactoring on top of IDF

SuGlider:
## Summary
This PR is a complete reffactoring of UART Serial Hardware and respective HAL in order to use IDF instead of current Register manipulation approach. 

It  implements Arduino SerialEvent functionality. 

Fix #5287  
Fix #5273 
Fix #5519 
Fix #5247 
Fix #5403
Fix #5429
Fix #5047
Fix #5463
Fix #5362 
Fix #5112  
Fix #5443 

## Impact
It solves many reported issues related to UART.
It was tested and works fine for ESP32, ESP-S2 and ESP32-C3.

Issue #4603, opened 2020-12-03
ESP32S2 - Serial.Flash() is not waiting for transmission to end


DawidDiaco:
### Hardware:
Board:							ESP32-S2 Wrover
Core Installation version:			unofficial - branch - esp32s2, commit baaff8480ebbdcf5c8c9214283f538a3fcdba29d and release/4.2-beta1 of ESP-IDF
IDE name:						Visual Studio with VisualGDB
Flash Frequency:					80Mhz
PSRAM enabled:                                    I don't have a clue how to check it
Upload Speed:					115200
Computer OS:                                        Windows 10 Pro

### Description:
I'm using Serial1 for communicating with modbus device. For that I found a great library that relies on Serial.Flash() for ensuring that transmission has ended. What we found out in this issue: https://github.com/emelianov/modbus-esp8266/issues/85, Arduino liblaries are bugged, and Serial.Flush() is not waiting for the end of transmision.

### Sketch:
```cpp
#include <Arduino.h>

void setup() 
{
	Serial1.begin(9600, SERIAL_8E1, GPIO_RXD, GPIO_TXD);
	pinMode(GPIO_RTS, OUTPUT);
}

void loop() 
{
	digitalWrite(GPIO_RTS, HIGH);
	uint8_t data[] = { 0x0, 0x1, 0x2, 0x3, 0x4 };
	Serial1.write(data, 4);
	Serial1.flush();
	digitalWrite(GPIO_RTS, LOW);
	
	delay(1000);
}
```

### Debug Messages:
![IMG_20201201_193140](https://user-images.githubusercontent.com/30572679/101088324-cfffd480-35b3-11eb-9c02-dc8634b0aac4.jpg)

lbernstone:
Please update to idf-release/v4.2 and retest.  This is beta code and is being actively worked on.

DawidDiaco:
I've checkout branch idf-release/v4.2, but I'm unable to compile it with release/4.2 of esp-idf.  I'm getting multiple errors that look like this:

```
FAILED: esp-idf/arduino/CMakeFiles/__idf_arduino.dir/cores/esp32/esp32-hal-time.c.obj 
[...]

PR #6133, opened 2022-01-13
Fixes UART MODBUS and Loopback issue

### LINKED ISSUE #6126: UART Loopback doesn't work on 2.0.2
### Related Information
https://github.com/espressif/arduino-esp32/pull/5549#issuecomment-1009642889

### Board

ESP32, ESP32-S2 and ESP32-C3

### Hardware Configuration

ESP32 - loopback on UART2 : TX2 <--> RX2
ESP32-S2 and ESP32-C3 - loopback on UART1 : TX1 <--> RX1

### Description

When trying to send data in a UART loopback configuration, it fails and nothing is read.
Using Arduino Core 2.0.0 and 2.0.1 - it works fine. The issue is with v2.0.2
Most problably something related to last change on support to MODBUS https://github.com/espressif/arduino-esp32/pull/6026

### Sketch

```cpp
HardwareSerial LOOPBACK_UART(2);  // UART2 on ESP32 and UART1 on S2/C3

void setup() {
  Serial.begin(115200);
  LOOPBACK_UART.begin(115200);

  Serial.println("Serial Port 0 - Testing");
}


void loop() {
  static uint8_t count = 0;

  //Shorted Pins TX<->RX using ESP32/UART2 or S2/C3/UART1
  LOOPBACK_UART.write('0' + count);
  if (LOOPBACK_UART.available()) {
    Serial.write(LOOPBACK_UART.read());
  }
  count++;
  if ('0' + count > 'Z') {
    count = 0;
    LOOPBACK_UART.println();
  }
  delay(100);
}
```

### Debug Message

```plain
Correct Output should be:

ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
flash read err, 1000
ets_main.c 371 
ets Jun  8 2016 00:22:57

rst:0x10 (RTCWDT_RTC_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1420
ho 0 tail 12 room 4
load:0x40078000,len:13540
load:0x40080400,len:3604
entry 0x400805f0
Serial Port 0 - Testing
0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ
0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ
0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ
0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ
```

but with v2.0.2 there is no output after the message  ```Serial Port 0 - Testing```

SuGlider:
## Summary
This PR fixes an issue created by #6026 not allowing UART loopback to work correctly.
It fixes both at the same time:
- makes flush() to wait up to the end of UART transmission before returning, fixing issue related to MODBUS #5877 
- solves the UART loopback issue in https://github.com/espressif/arduino-esp32/pull/5549#issuecomment-1009642889 and #6126

## Impact
None.

## Related links
Fixes #6126 
https://github.com/espressif/arduino-esp32/pull/5549#issuecomment-1009642889
#5877 
#4603

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -153,9 +153,6 @@ uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rx
         ESP_ERROR_CHECK(uart_set_line_inverse(uart_nr, UART_SIGNAL_TXD_INV | UART_SIGNAL_RXD_INV));    
     }
 
-    // Set RS485 half duplex mode on UART.  This shall force flush to wait up to sending all bits out
-    ESP_ERROR_CHECK(uart_set_mode(uart_nr, UART_MODE_RS485_HALF_DUPLEX));
-
     UART_MUTEX_UNLOCK();
 
     uartFlush(uart);
@@ -306,7 +303,7 @@ void uartFlushTxOnly(uart_t* uart, bool txOnly)
     }
     
     UART_MUTEX_LOCK();
-    ESP_ERROR_CHECK(uart_wait_tx_done(uart->num, portMAX_DELAY));
+    while(!uart_ll_is_tx_idle(UART_LL_GET_HW(uart->num)));
 
     if ( !txOnly ) {
         ESP_ERROR_CHECK(uart_flush_input(uart->num));


TD-er:
Looks like a better approach.
Just waiting in the function that claims to actually just do that seems like a good idea :)

mrengineer7777:
Much better solution. I approve.

cores/esp32/esp32-hal-uart.c
```diff
@@ -306,7 +303,7 @@ void uartFlushTxOnly(uart_t* uart, bool txOnly)
     }
     
     UART_MUTEX_LOCK();
-    ESP_ERROR_CHECK(uart_wait_tx_done(uart->num, portMAX_DELAY));
+    while(!uart_ll_is_tx_idle(UART_LL_GET_HW(uart->num)));
```
avillacis, line 306: Could somebody explain why exactly was the `uart_wait_tx_done()` call replaced here with the `uart_ll_is_tx_idle()` loop?

The ModbusMaster library is now broken in ESP32S2 v2.0.3, and it was working correctly in v2.0.2 (checked, I switch between the two versions at will). Our project, like many others, uses RS485 and flips the data-enable and read-enable pins, and uses flush() to wait until TX done before switching to RX mode. I still have to do more tests, but this change in particular seems the most likely culprit.

SuGlider: @avillacis 

Have you read the issues related to this PR? They are related to MODBUS and `flush()` issues.

In summary:
`uart_wait_tx_done()` returns as soon as data is stored in TXRingBuffer, before the TX FIFO is really empty.

`uart_ll_is_tx_idle()` returns true just when TX FIFO is completely empty. 
The code for it is here - but it is the same for all other SoC, thus, the issue you see should happen for any ESP32-xx
https://github.com/espressif/esp-idf/blob/master/components/hal/esp32s2/include/hal/uart_ll.h#L714-L717

I tested `flush()` and I'm sure it works fine.
Maybe the issue you see isn't related to this PR.

avillacis: Objection withdrawn. With access to the actual hardware, I found out that our RS485 card was echoing back request commands, and this echo was previously hidden by the `uart_set_mode()` call. Now I have manually added the call in our own code, and the issue was fixed.

However, is there any upstream bug report documenting this issue with `uart_wait_tx_done()`? Is it a true software bug, or a documentation error?

SuGlider: Apparently `uart_wait_tx_done()` works correctly when used with `uart_set_mode(uart_num, UART_MODE_RS485_HALF_DUPLEX)`, considering an IDF environment.
 
But for Arduino, it isn't useful because of the its generic UART usage and APIs.
Thus we had to change it for the LL API.

SuGlider: By default, IDF starts UART with  `uart_set_mode(uart_num, enumerator UART_MODE_UART)` and in this case, it returns as soon as data is stored in TXRingBuffer, before the TX FIFO is really empty.

PR #6160, opened 2022-01-20
Fixing interrupts in LEDC

### LINKED ISSUE #6140: AttachInterrupt does not work on output of PWM (2.0.2)
### Board

ESP32Dev module

### Device Description

I'm using a PWM output, and setup a pin interrupt so i can detect the start of the PWM signal.
This works fine in 2.0.1, but not on 2.0.2

### Hardware Configuration

no

### Version

latest master

### IDE Name

PlatfromIO

### Operating System

Win10/Linux

### Flash frequency

80Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

`attachInterrupt` should be able to attach to the output signal of the PWM signal generated by ledC.
this does work in 2.0.1, but not in 2.0.2.

I noticed that there were changes made to the LEDC code (https://github.com/espressif/arduino-esp32/pull/6045)
when using the old (2.0.1) code for `ledcAttachPin` in my code, it works again.
So something in the new code, does not allow a PIN interrupt to be attached to the PWM output.

Below is test code to reproduce the problem.

### Sketch

```cpp
#include <Arduino.h>

#define PIN_PWM_OUT 19
#define PWM_CHANNEL 0

volatile uint32_t isrCounter = 0;

// PWM pin low to high transition ISR
void IRAM_ATTR onPulse() {
    // increment counter
    isrCounter++;
}

void setup() {

    pinMode(PIN_PWM_OUT, OUTPUT);            

    // Setup PWM on channel 0, 1000Hz, 10 bits resolution
    ledcSetup(PWM_CHANNEL, 1000, 10);             // channel 0  => Group: 0, Channel: 0, Timer: 0
    ledcAttachPin(PIN_PWM_OUT, PWM_CHANNEL);      // attach the channels to the GPIO to be controlled
    ledcWrite(PWM_CHANNEL, 512);                  // channel 0, duty cycle 50%

    // Setup PIN interrupt on rising edge
    attachInterrupt(PIN_PWM_OUT, onPulse, RISING);
    
    Serial.begin(115200);
    while (!Serial);
}

void loop() {
  
    Serial.printf("ISR Counter: %u\n",isrCounter);
    delay(1000);
}
```

### Debug Message

```plain
The Counter in above test code will not increase.
```

### Other Steps to Reproduce

_No response_

P-R-O-C-H-Y:
## Summary
Added `pinMode(pin,OUTPUT);` to ledcAttach function to ensure that GPIO pin interrupts will work again.

## Impact
Fix for ledc output pin interrupt

## Related links
Closes #6140

### DIFF
--- a/cores/esp32/esp32-hal-ledc.c
+++ b/cores/esp32/esp32-hal-ledc.c
@@ -168,6 +168,8 @@ void ledcAttachPin(uint8_t pin, uint8_t chan)
         .hpoint         = 0
     };
     ledc_channel_config(&ledc_channel);
+
+    pinMode(pin,OUTPUT);
 }
 
 void ledcDetachPin(uint8_t pin)


s-hadinger:
Hi, what is the reason behind this change? It actually breaks PMW in Tasmota. The GPIO seem not to be connected to ledc anymore.

P-R-O-C-H-Y:
Hi @s-hadinger, thanks for finding that. Linked PR which will fix the ledc for ESP32C3. The reason for that was that user can use attachInterrupt for ledC without calling pinMode himself.

s-hadinger:
@P-R-O-C-H-Y I did some more tests. I still can't get PWM working if pinMode is called after ledc_channel_config.

On the contrary, it works fine if we invert the orders of the calls:

```
    pinMode(pin,OUTPUT);
    ledc_channel_config(&ledc_channel);
```

Issue #6123, opened 2022-01-10
USBHID::SendReport takes two semaphores but only gives one back


tobozo:
### Board

ESP32-S2

### Device Description

AprilBrothers' WUD : https://wiki.aprbrother.com/en/wud.html 
Basically an ESP32-S2 with SD Card slot in a pendrive form factor.

### Hardware Configuration

Not really relevant for this issue, but there's a SD Card with the following pinout

```
#define SD_MISO  37
#define SD_MOSI  35
#define SD_SCK   36
#define SD_CS    34
```

### Version

other

### IDE Name

Arduino IDE 1.8.19 with arduino-esp32 v2.0.2

### Operating System

linux

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

912600

### Description

HID Keyboard is not working (detected but no keys received by the host OS) when multiple HID devices are created.

The cause appears to be a [bug in the library](https://github.com/espressif/arduino-esp32/blob/master/libraries/USB/src/USBHID.cpp#L336-L340).

@chegewara suggested the following fix, tested for a couple of days and solved it for me:

insert the following conditional statement at the end of `USBHID::SendReport`:

```C++
    if (tinyusb_hid_device_input_sem) {
        xSemaphoreGive(tinyusb_hid_device_input_sem);
    }
```

goes like:

```C++

bool USBHID::SendReport(uint8_t id, const void* data, size_t len, uint32_t timeout_ms){
    if(!tinyusb_hid_device_input_sem || !tinyusb_hid_device_input_mutex){
        log_e("TX Semaphore is NULL. You must call USBHID::begin() before you can send reports");
        return false;
    }

    if(xSemaphoreTake(tinyusb_hid_device_input_mutex, timeout_ms / portTICK_PERIOD_MS) != pdTRUE){
        log_e("report %u mutex failed", id);
        return false;
    }

    bool res = ready();
    if(!res){
        log_e("not ready");
    } else {
        res = tud_hid_n_report(0, id, data, len);
        if(!res){
            log_e("report %u failed", id);
        } else {
            xSemaphoreTake(tinyusb_hid_device_input_sem, 0);
            if(xSemaphoreTake(tinyusb_hid_device_input_sem, timeout_ms / portTICK_PERIOD_MS) != pdTRUE){
                log_e("report %u wait failed", id);
                res = false;
            }
        }
    }
    xSemaphoreGive(tinyusb_hid_device_input_mutex);
[...]

PR #6161, opened 2022-01-20
USBHID::SendReport takes more than it gives (see #6123)

tobozo:


libraries/USB/src/USBHID.cpp
```diff
@@ -340,8 +340,10 @@ bool USBHID::SendReport(uint8_t id, const void* data, size_t len, uint32_t timeo
             }
         }
     }
-
     xSemaphoreGive(tinyusb_hid_device_input_mutex);
+    if (tinyusb_hid_device_input_sem) {
```
chegewara, line 344: I think it is not needed to check it, since it will crash anyway in this code if semaphore is NULL
```
            xSemaphoreTake(tinyusb_hid_device_input_sem, 0);
            if(xSemaphoreTake(tinyusb_hid_device_input_sem, timeout_ms / portTICK_PERIOD_MS) != pdTRUE){
            ```

Comment (tobozo) [reply libraries/USB/src/USBHID.cpp]: makes sense, check removed

File: libraries/USB/src/USBHID.cpp [hunk 2]
```diff
@@ -340,8 +340,8 @@ bool USBHID::SendReport(uint8_t id, const void* data, size_t len, uint32_t timeo
             }
         }
     }
-
     xSemaphoreGive(tinyusb_hid_device_input_mutex);
+    xSemaphoreGive(tinyusb_hid_device_input_sem);
```
me-no-dev, line 344: this semaphore is given in `tud_hid_report_complete_cb`.

tobozo: does it mean I met a situation where `tud_hid_report_complete_cb` is never called?

me-no-dev: it maybe possible to get into such situation, but in the case you do, a timeout will be reached and you can continue with retry or else.

tobozo: So after a few days of testing I can't reproduce the problem alone, however I found some stability issues that affect all USB devices when USBMSC has an incomplete or failed init. 

I'm not sure if it's a bug or a feature but I can't late-init (on demand) the USBMSC driver without getting side effects.

example 1: comment out the `MSC.begin` in a composite scenario (MSC+Keyboard+Mouse) and try to use the other USB devices.

example 2: use USBMSC with SD.writeRAW and SD.readRAW and boot while no SD card is inserted, then try to use the other USB devices.

In both examples USBSerial eventually crashes and eats the exception message while dying, also HID (mouse or keyboard) freezes the ESP after sending a few reports.

Indeed using this PR does not fix the crash situation, it only makes it happen later.

Should I close this PR as it's obviously unrelated to HID?

me-no-dev: Yes, you should close this PR. As for MSC, maybe the USB protocol needs to be researched and figure out how to tigger later connect. I suspect that the other devices have issues because the HOST is trying to attach the MSC over and over and is eating the bandwidth. In any  case this is outside of the scope of Arduino as-is. Maybe try in TinyUSB?

PR #6179, opened 2022-01-22
Update RequestHandlersImpl.h to enable serveStatic on folders with LittleFS. 

mathertel:
## Summary
Using `serveStatic(...)` is broken when using folder names (not files) and LittleFS fil system.
This checks correctly for folders.

With LittleFS the `fs.exists(path)` returns true also on folders. A `isDirectory()` call is required to set _isFile to false on directories.
This enables serving all files from a folder like : `server.serveStatic("/", LittleFS, "/");`

## Impact
serveStatic is partially broken


me-no-dev:
Thanks @mathertel

Issue #4069, opened 2020-06-07
Multible http response headers with the same name get overwritten


yonie1:
If the http client gets multible headers with the same name in the response (for example set-cookie headers), they get overwritten. My way to fix it was to change 

https://github.com/espressif/arduino-esp32/blob/b92c58d74b151c7a3b56db4e78f2d3c90c16446f/libraries/HTTPClient/src/HTTPClient.cpp#L1149

to 

if(_currentHeaders[i].key.equalsIgnoreCase(headerName) && _currentHeaders[i].value.length() == 0) {

to check if the header value is not set to avoid overwriting when saving multible headers with the same name

lbernstone:
Make a pull request right at that page and mention @Jeroen88

Jeroen88:
I just took a quick review of the code. I am not in favor of this change:
- It does not fix the underlying problem: only one header with a specific name can exists, while a server may send more than one
- It only changes the behavior from keeping the last header sent with a certain name to keeping the first with a certain name

A real solution would involve quite some adaptations and very likely an API change, since [requestArgument ](https://github.com/espressif/arduino-esp32/blob/b92c58d74b151c7a3b56db4e78f2d3c90c16446f/libraries/HTTPClient/src/HTTPClient.h#L194) (not a very good name because in this context it is used as a responseArgument) needs to accommodate an array or vector of values. This implies changes in the functions for accessing the values as well.

kurenai-ryu:
why not implement like on the esp8266? 
https://github.com/esp8266/Arduino/blob/6feda9ebda8d5588c288145bff2f9f0d816b3ec2/libraries/ESP8266HTTPClient/src/ESP8266HTTPClient.cpp#L1277

PR #6183, opened 2022-01-24
HTTPClient: Support concatenation of headers

mattsches1:
## Summary
Adds support for multiple response headers with identical keys; the main goal is to support setting several cookies requested by a server, which result in several "set-cookie" response headers.

## Impact
If several headers with the same key are received in a response, they are concatenated using commas.

## Related links
* Alignment with ESP8266 implementation (see https://github.com/esp8266/Arduino/commit/1de0c341b55ba8c0993fd3d2e0c5696935578751#diff-977435a9cc4619fa0b8b995085f6ae683485cf563722756bab57108b362da316)
* Fixes https://github.com/espressif/arduino-esp32/issues/4069

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -1263,12 +1263,19 @@ int HTTPClient::handleHeaderResponse()
                     _location = headerValue;
                 }
 
-                for(size_t i = 0; i < _headerKeysCount; i++) {
-                    if(_currentHeaders[i].key.equalsIgnoreCase(headerName)) {
-                        _currentHeaders[i].value = headerValue;
-                        break;
+                for (size_t i = 0; i < _headerKeysCount; i++) {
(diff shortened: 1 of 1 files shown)


me-no-dev:
Could this be made to work with particular headers only? `set-cookie` in your case?

mattsches1:
It could, but if I'm honest, I don't really see a point in it. And I thought aligning this with the ESP8266 implementation (from where I actually copied the code snippet) would be sensible,

But I was actually looking into a solution to handle cookies in the client and found the header concatenation would not really solve that (or require too much efforts around for my taste). In the meantime, I therefore implemented a cookie jar  to enable the client to properly handle cookies. Just created a pull request for this: [https://github.com/espressif/arduino-esp32/pull/6216](url)

Feel free to discard any of these if they don't fit for whatever reason.

me-no-dev:
Closing in favor of https://github.com/espressif/arduino-esp32/pull/6216

PR #6191, opened 2022-01-25
Refactor the CI scripts

Ouss4:
## Summary
The reason behind this refactoring is that all of the sketch related functions can (and will) be used for other purposes.

1. Build in the sketch directory:  This will make it easy to handle artifacts after the build.
2. Separate sketch related functions from IDE installation script.  This is the main commit.
3. Create a separate job for the Cmake check.  This check was part of one of the Linux build.  I believe that it's not the best place for such a check.
4. Checking for the skip landmarks and validity of the the sketch directory were already done by `count_sketches`.

## Impact
CI


Ouss4:
One thing I didn't mention and didn't create a separate commit for is that the sketches built for MacOS and Windows are now the same.  Before, MacOS builds 2 more sketches than Windows, I'm not sure why was it this way (The build in Windows is that slow?). I run the workflow in my fork and they all build successfully, but if there is a need to separate them again I can bring the old behavior back.

me-no-dev:
> One thing I didn't mention and didn't create a separate commit for is that the sketches built for MacOS and Windows are now the same. Before, MacOS builds 2 more sketches than Windows, I'm not sure why was it this way (The build in Windows is that slow?). 

Yes. I have limited the number of sketches on Windows because it's really slow. It is more or less there to make sure that our tools are working correctly on Windows and that generally compiling passes. Same on Mac (just faster so a couple extra sketches). I would suggest either limiting the sketches back on windows or limiting them for both windows and mac (to stay the same)

Ouss4:
> Yes. I have limited the number of sketches on Windows because it's really slow. It is more or less there to make sure that our tools are working correctly on Windows and that generally compiling passes. Same on Mac (just faster so a couple extra sketches). I would suggest either limiting the sketches back on windows or limiting them for both windows and mac (to stay the same)

Alright, I have limited them both for Windows and maOS to make the code a bit simpler.

PR #6202, opened 2022-01-28
Fixes UART1 and UART2 default pins for ESP32-S3

SuGlider:
## Summary
This PR fixes an issue with UART1 default pins.
When using pins RX_1 = 18 and TX_1 = 17, UART1 will display a Break Error on those pins when they are floting (not connected).

It also defines RX_2 = 19 and TX_2 = 20 as default pins for UART2.

## Impact
The deaulf pins may look different from the original pinout diagram.

## Related links
None

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -64,7 +64,7 @@ void serialEvent1(void) {}
 #ifndef RX2
 #if CONFIG_IDF_TARGET_ESP32
 #define RX2 16
-#else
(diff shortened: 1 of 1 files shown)


me-no-dev:
Ahh @SuGlider you got me! Why did you make this PR to the master branch and not to the S3 support one? I forgot to double-check and merged it... duh

SuGlider:
@me-no-dev , FYI:
I also added this commit with a missing change to the S3 branch:
https://github.com/espressif/arduino-esp32/commit/b75a08c9556b3971d2da310a2e83f56ac76a5177

PR #6203, opened 2022-01-28
Add Github action for uploading arduino-esp32 component. 

magicarm22:
## Summary
Add upload-idf-componen.yml file for Github Workflow. After each push on the master branch, the component will be uploaded to the component registry

## Impact
After that PR users can use the Arduino-ESP32 component from the component registry.

### DIFF
--- a/.github/workflows/upload-idf-component.yml
+++ b/.github/workflows/upload-idf-component.yml
@@ -1,0 +1,19 @@
+name: Push components to https://components.espressif.com
+on:
+  push:
+    tags:
+      - v*
+jobs:
+  upload_components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: "recursive"
+
+      - name: Upload components to the component registry
+        uses: espressif/github-actions/upload_components@master
+        with:
+          name: arduino-esp32
+          namespace: espressif
+          api_token: ${{ secrets.IDF_COMPONENT_API_TOKEN }}
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -1,0 +1,20 @@
+description: "Arduino core for ESP32, ESP32-S and ESP32-C series of SoCs"
+url: "https://github.com/espressif/arduino-esp32"
+targets:
+  - esp32
+  - esp32s2
+  - esp32c3
+tags:
+  - arduino
+files:
+  include:
+    - "cores/**/*"
+    - "variants/esp32/**/*"
+    - "variants/esp32s2/**/*"
+    - "variants/esp32s3/**/*"
+    - "variants/esp32c3/**/*"
+    - "libraries/**/*"
+    - "CMakeLists.txt"
+    - "Kconfig.projbuild"
+  exclude:
+    - "**/*"


kumekay (changes requested):
Please leave only one commit in this PR

idf_component.yml
```diff
@@ -0,0 +1,2 @@
+description: "Arduino Wiring-based Framework for the Espressif ESP32, ESP32-S and ESP32-C series of SoCs"
+url: "https://github.com/espressif/arduino-esp32"
```
kumekay, line 2: @magicarm22 Please add targets:  ESP32, ESP32-S2 and ESP32-C3
and maybe some tags: like `arduino`

kumekay:
Thank you @magicarm22 , LGTM

@me-no-dev Could you please take a look? Are there any files that should be excluded from the component in the registry?

idf_component.yml
```diff
@@ -0,0 +1,8 @@
+description: "Arduino Wiring-based Framework for the Espressif ESP32, ESP32-S and ESP32-C series of SoCs"
```
igrr, line 1:
```suggestion
description: "Arduino core for ESP32, ESP32-S and ESP32-C series of SoCs"
```

.github/workflows/upload-idf-component.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          submodules: "recursive"
+
+      - name: Upload components to the component registry
+        uses: espressif/github-actions/upload_components@master
+        with:
+          name: arduino-esp32
+          namespace: espressif
+          api_token: ${{ secrets.IDF_COMPONENT_API_TOKEN }}
+          service_url: https://api.staging.components.espressif.tools/
```
igrr, line 20: Is this supposed to be replaced with production URL before merging?

kumekay: @magicarm22 Please remove the URL from here, if it's missing a default one will be used

me-no-dev:
> Thank you @magicarm22 , LGTM
> 
> @me-no-dev Could you please take a look? Are there any files that should be excluded from the component in the registry?

You actually do not need a large portion of the files. For example the `tools` folder which includes all headers and prebuilt libs for all chip (500MB+). We can skip the majority of files in the root as well. I will check, but you really only need CMakeLists.txt, core and variants folders.

.github/workflows/upload-idf-component.yml
```diff
@@ -0,0 +1,19 @@
+name: Push components to https://components.espressif.com
+on:
+  push:
+    tags:
+      - v*
+jobs:
+  upload_components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
```
Ouss4, line 10: Why is master used an not the latest version?

kumekay: Nice catch, `v2` will be a better choice,

@Ouss4  Could you please take a look, I made a PR to github-actions to update the readme: https://github.com/espressif/github-actions/pull/27

me-no-dev:
Merged @magicarm22 and @kumekay :)

grongor:
Hi! I was just looking for this component and I've seen that this new CI job was never run yet.

First, I think there is an error that the job is only run on tags with `v*` prefix, but all the tags in this repository don't have this prefix (see https://github.com/espressif/arduino-esp32/blob/master/.github/workflows/upload-idf-component.yml#L5).

Secondly: would you mind running the job manually for at least the latest (and preferably stable) tags, so that I can pull the component using the registry?

Thanks!

me-no-dev:
@grongor we will release the next "stable" very soon and it will be in the component manager.

PR #6216, opened 2022-01-31
HttpClient: Add cookie support (cookie jar)

mattsches1:
This is a proposal to add cookie support in `HttpClient`. With this modification, the client is able to receive and store cookies and automatically send them in a subsequent request, if applicable as per `Domain` and `Path` attributes. Receiving multiple cookies in the same response (multiple `set-cookie` headers) is supported.

Cookie expiry (`Expires`and `Max-Age` attributes) is respected, but relies on RTC to be set properly. 

`Secure` is respected, `HttpOnly` is stored, but not used anywhere.

See [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie](url) for a description of cookie attributes.

This proposal probably has room for additions, code style improvements whatsoever. I created this because I needed to handle cookies when dealing with AWS requests and thought others might have use for it. I tried my best to avoid memory leaks or other negative side effects. But I'm not a professional, and it would for sure be useful if others checked a bit upon possible weak spots.

Usage in addition to standard `HttpClient` handling):

Declare a cookie jar in the calling instance: 
`CookieJar cookie_jar;`

Before beginning the HTTP request, set 

```
    WiFiClientSecure *client = new WiFiClientSecure;

    if(client) {
        client->setCACert(cert);
        {
            // Add a scoping block for HTTPClient https to make sure it is destroyed before WiFiClientSecure *client is 
            HTTPClient https;

            https.setCookieJar(&cookie_jar);

            }
            if (https.begin(*client, url)) {
            ...

``` 
That's it.

The cookie jar is a vector of structs that holds all the cookies of the client and can be used for all requests.

The pull request includes concatenation of received headers, as already proposed in [https://github.com/espressif/arduino-esp32/pull/6183](url). The latter is not required for the cookie jar to work and has only been included since I had mistakenly been working on the master branch of my fork for the former. Sorry for this.

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -39,6 +39,9 @@
 
 #include "HTTPClient.h"
 
+/// Cookie jar support
+#include <time.h>
+
 #ifdef HTTPCLIENT_1_1_COMPATIBLE
 class TransportTraits
 {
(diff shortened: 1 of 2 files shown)


me-no-dev:
Very nice implementation. I will only suggest that you comment out the code for multiple headers with a note to be uncommented in the future if requests for it come.

mattsches1:
Done.

PR #6219, opened 2022-02-01
Update of CameraWebServer example

renebohne:
## Summary
If board has no PSRAM, we need to set the framebuffer location to DRAM:
config.fb_location = CAMERA_FB_IN_DRAM;

## Impact
It is just a minor change in an example. The "M5 esp32cam" board without PSRAM can run the example with this modification. Before, the example crashed.

### DIFF
--- a/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
+++ b/libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
@@ -62,6 +62,7 @@ void setup() {
     config.frame_size = FRAMESIZE_SVGA;
     config.jpeg_quality = 12;
     config.fb_count = 1;
+    config.fb_location = CAMERA_FB_IN_DRAM;
   }
 
 #if defined(CAMERA_MODEL_ESP_EYE)


libraries/ESP32/examples/Camera/CameraWebServer/CameraWebServer.ino
```diff
@@ -62,6 +62,7 @@ void setup() {
     config.frame_size = FRAMESIZE_SVGA;
     config.jpeg_quality = 12;
     config.fb_count = 1;
+    config.fb_location = CAMERA_FB_IN_DRAM;
```
me-no-dev, line 65: this should be set to DRAM only if the board does not have PSRAM

renebohne: yes, that's why I added it into the else branch of  if(psramFound()). In hope that the psramFound() function returns false only if the board does not have PSRAM.

me-no-dev: ahh sorry I did not expand the code and did not see the if statement :)

me-no-dev:
Merged, Thanks!

PR #6223, opened 2022-02-02
add feather esp32 v2 and qtpy c3 board def

ladyada:
update some pin names for adafruit boards with built in hardware
add variant.cpp's to auto-enable i2c, tft, neopixels on boot for some boards
add more fun adafruit boards :)


me-no-dev:
Merged :)

PR #6190, opened 2022-01-25
Resolve WString TODO

mrengineer7777:
Resolve TODO (XXX) by logging warning message.

*By completing this PR sufficiently, you help us to improve the quality of Release Notes*

-----------
## Summary
Replaces XXX comment with warning message.

## Impact
Logs warning if there isn't enough buffer space for a replace.

## Related links
NA

PR #6224, opened 2022-02-02
Fix String.replace() broken by #6190

s-hadinger:
## Summary
This PR fixes a bug introduced by #6190

There was no braces after `if` so the `return` statement is taken every time, not just in case of failure.

## Impact
String.replace is not doing an actual replacement most of the time.

### DIFF
--- a/cores/esp32/WString.cpp
+++ b/cores/esp32/WString.cpp
@@ -774,9 +774,10 @@ void String::replace(const String& find, const String& replace) {
         }
         if(size == len())
             return;
-        if(size > capacity() && !changeBuffer(size))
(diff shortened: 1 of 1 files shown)


tablatronix:
@me-no-dev ping
Fatal string replace bug, merge PR or revert https://github.com/espressif/arduino-esp32/pull/6190

me-no-dev:
Thanks @s-hadinger

mrengineer7777:
@s-hadinger Wow that's embarrassing. Thanks for the fix.

s-hadinger:
Thanks for the great work at Arduino core. Tasmota is a heavy user

PR #6226, opened 2022-02-03
Added dual antenna for WiFi (based on the ESP32-WROOM-DA module)

pedrominatel:
## Summary

Added the dual antenna configuration for the ESP32-WROOM-DA module.

## Impact
None

## Related links

[ESP32-WROOM-DA](https://www.espressif.com/sites/default/files/documentation/esp32-wroom-da_datasheet_en.pdf)

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.h
+++ b/libraries/WiFi/src/WiFiGeneric.h
@@ -174,6 +174,8 @@ class WiFiGenericClass
 
     bool initiateFTM(uint8_t frm_count=16, uint16_t burst_period=2, uint8_t channel=1, const uint8_t * mac=NULL);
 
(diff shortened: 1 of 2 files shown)


libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -174,6 +174,8 @@ class WiFiGenericClass
 
     bool initiateFTM(uint8_t frm_count=16, uint16_t burst_period=2, uint8_t channel=1, const uint8_t * mac=NULL);
 
+	bool setAntennaConfig(uint8_t gpio_ant1, uint8_t gpio_ant2, uint8_t default_rx_ant);
```
me-no-dev, line 177: indentation here seems wrong :)

me-no-dev: I would suggest that you rename the method to `setDualAntennaConfig` to make it more clearer that it should be used only when dual antenna is available.

PR #6235, opened 2022-02-05
Fixes RMT examples and adds ESP32-S3 config

SuGlider:
## Summary
Adds and fixes RMT examples to run with ESP32-S3 

## Impact
None

## Related links
None


me-no-dev:
@SuGlider please start targeting the s3 branch and not master.

PR #6239, opened 2022-02-06
large spiffs partitions for feathers2

stooged:
large spiffs partitions for the UM FeatherS2 board

the 16mb UM FeatherS2 board has no partition scheme for a large spiffs partition.

i have created the following partition .csv files for the feathers2.

16M Flash (1MB APP/13MB SPIFFS)
16M Flash (2MB APP/11MB SPIFFS)
16M Flash (3MB APP/9MB SPIFFS)

16M Flash No OTA (1MB APP/14MB SPIFFS)
16M Flash No OTA (2MB APP/13MB SPIFFS)
16M Flash No OTA (3MB APP/12MB SPIFFS)


me-no-dev:
SPIFFS is not a very good idea on such large partitions. Have you considered to use FatFS maybe?

stooged:
I might look into fatfs for future projects.

I only created the push to share the modifications i am using but if it is not a good idea then dont add it.

PR #6255, opened 2022-02-09
Update httpUpdate.ino

Kon4534:
there are no keywords for httpUpdate.onStart(), onEnd(), onProgress() or onError() in httpUpdate.

This results in a exit status 1 in Arduino IDE
Example: 
'class HTTPUpdate' has no member named 'onStart'

### DIFF
--- a/libraries/HTTPUpdate/examples/httpUpdate/httpUpdate.ino
+++ b/libraries/HTTPUpdate/examples/httpUpdate/httpUpdate.ino
@@ -36,22 +36,6 @@ void setup() {
 
 }
 
-void update_started() {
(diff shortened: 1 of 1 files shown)


atanisoft:
Rather than remove the usages from the methods that *ARE* available, it would be better to update the keywords.txt file in the library directory to include all methods that are publicly exposed in the class.

Kon4534:
> Rather than remove the usages from the methods that _ARE_ available, it would be better to update the keywords.txt file in the library directory to include all methods that are publicly exposed in the class.

Yes, you are right. After digging a little bit deeper I found out that my Arduino IDE configuration is the main problem, and not the code. 

Sorry for the inconvenience

PR #6272, opened 2022-02-12
Fixes UART pin setting + adds CTS/RTS HW Flow Control

### LINKED ISSUE #6185: Enable UART Flow Control
### Related area

UART

### Hardware specification

ESP32

No, i just need to use uart flow control.

Enable support for flow control (RTS and CTS, perhaps it would be also usefull to enable DTR and DSR)

Add a parameter on the "begin" function that allows to select if the serial should use CTS/RTS and a pin assigment (might be just adding de pins, and if they are different than -1 the flow control would be enabled).
Also a "setRts" and "setCts" function could be implemented to enable one or both flow control pins

### Additional context

_No response_

SuGlider:
## Summary
This PR has two components:
1- Fixes Serial.begin() in order to keep RX/TX pins the same, as previouly set, when those parameter are not used.
2 - Adds Hardware Flow Control and adds RTS/CTS pins setting 

## Impact

Serial will keep RX/TX pins unmodified in case of several ```begin()``` calls:

``` dart
void setup() {
    Serial1.begin(115200, SERIAL_8N1, 18, 19);
    Serial1.println("HELLO WORLD!");
  
    Serial1.begin(115200);
    Serial1.begin(115200);
    Serial1.println("HELLO WORLD!  2nd Time!");
}
```

Modified and New functions - from HardwareSerial class:
``` dart
    // Negative Pin Number will keep it unmodified, thus this function can set individual pins
    // SetPins shall be called after Serial begin()
    void setPins(int8_t rxPin, int8_t txPin, int8_t ctsPin = -1, int8_t rtsPin = -1);
    // Enables or disables Hardware Flow Control using RTS and/or CTS pins (must use setAllPins() before)
    void setHwFlowCtrlMode(uint8_t mode = HW_FLOWCTRL_CTS_RTS, uint8_t threshold = 64); 
```

Note: ```setPins()``` accepts negative pin number that will keep it unmodified.

## Related links
This replaces PR #6176 
Fixes #6185


P-R-O-C-H-Y (commented):
LGTM

mrengineer7777:
Possible regression from #6176 .  _rxPin(-1), _txPin(-1) were removed.  How are the previous pin states tracked?

SuGlider:
@mrengineer7777 
No need to track.
It is all done by IDF instead.

Please test the PR to check that it works as intended and let me know.

mrengineer7777:
> Please test the PR to check that it works as intended and let me know.
@SuGlider Apologies, as this isn't my issue and I'm not an Espressif employee I won't be able to test this. 

> The "pin tracking" will get lost only when `Serial.end()` is called, but it sounds to be the correct expected behavior.
Seems reasonable.

cores/esp32/HardwareSerial.cpp
```diff
@@ -114,37 +114,46 @@ void serialEventRun(void)
...
+                rxPin = rxPin < 0 ? RX1 : rxPin;
+                txPin = txPin < 0 ? TX1 : txPin;
+            break;
+#endif
+#if SOC_UART_NUM > 2                   // may save some flash bytes...
+            case UART_NUM_2:
+                rxPin = rxPin < 0 ? RX2 : rxPin;
+                txPin = txPin < 0 ? TX2 : txPin;
+            break;
+#endif
```
me-no-dev, line 146: always add `default:` for switch statements. ESP-IDF will complain

cores/esp32/HardwareSerial.cpp
```diff
@@ -114,37 +114,46 @@ void serialEventRun(void)
 }
 #endif
 
-
 HardwareSerial::HardwareSerial(int uart_nr) : _uart_nr(uart_nr), _uart(NULL), _rxBufferSize(256) {}
 
+enum { UART_NUM_0, UART_NUM_1, UART_NUM_2 };
```
me-no-dev, line 119: If this enum will be used just for the switch statement, let's use int instead

cores/esp32/HardwareSerial.cpp
SuGlider: done.

cores/esp32/HardwareSerial.cpp
SuGlider: It is solved now by including "driver/uart.h" that has those defines.

PR #6280, opened 2022-02-14
Fix Check for _cookieJar in HTTPClient (#6266)

### LINKED ISSUE #6266: HTTPClient: CookieJar is initialized as nullptr but never checked
### Board

ESP32

### Device Description

Any

### Hardware Configuration

Any

### Version

latest master

### IDE Name

IDF

### Operating System

macOS

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

With the latest addition of _cookieJar for HTTPClient, the _cookieJar is initialized as nullptr, but no check before accessing it exists (for example in generateCookieString).
This means (unless it is explicetely set) we are running in a nullptr exception/panic at this point

### Sketch

```cpp
HTTPClient client;
client.begin("https://google.com/);

if (HTTP_CODE_OK == client.GET())
{
  // Do Stuff
}
```

### Debug Message

```plain
None
```

### Other Steps to Reproduce

_No response_

bbsan2k:

-----------
## Summary
Fix nullptr in HTTPClient with _cookieJar

## Impact
Fixes #6266

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -1543,122 +1543,125 @@ void HTTPClient::clearAllCookies()
 
 void HTTPClient::setCookie(String date, String headerValue)
 {
-    #define HTTP_TIME_PATTERN "%a, %d %b %Y %H:%M:%S"
+    if (_cookieJar)
+    {
+        #define HTTP_TIME_PATTERN "%a, %d %b %Y %H:%M:%S"
 
-    Cookie cookie;
-    String value;
-    int pos1, pos2;
+        Cookie cookie;
+        String value;
+        int pos1, pos2;
 
-    headerValue.toLowerCase();
+        headerValue.toLowerCase();
 
-    struct tm tm;
-    strptime(date.c_str(), HTTP_TIME_PATTERN, &tm);
-    cookie.date = mktime(&tm);
+        struct tm tm;
+        strptime(date.c_str(), HTTP_TIME_PATTERN, &tm);
+        cookie.date = mktime(&tm);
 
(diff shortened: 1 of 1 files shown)


mrengineer7777:
Question: does this code fix just indent the function and put brackets around it for "if (_cookieJar)"?
Suggestion: rather than indenting all the code, use this shorter form:
`if (!_cookieJar) return;`  Note the "!", which is equivalent to if(_cookieJar==NULL).
Feel free to break it into two lines and/or add braces as fits the coding style.
e.g.

`if (!_cookieJar)
   return;`
`if (!_cookieJar) {
   return;
}`
`if (!_cookieJar) { return; }`

This comment is not a review on the actual fix, just a way to minimize changes and make it easier to review.

bbsan2k:
> Question: does this code fix just indent the function and put brackets around it for "if (_cookieJar)"? Suggestion: rather than indenting all the code, use this shorter form: `if (!_cookieJar) return;` Note the "!", which is equivalent to if(_cookieJar==NULL). Feel free to break it into two lines and/or add braces as fits the coding style. e.g.
> 
> `if (!_cookieJar) return;` `if (!_cookieJar) { return; }` `if (!_cookieJar) { return; }`
> 
> This comment is not a review on the actual fix, just a way to minimize changes and make it easier to review. Edit: Github does not display the CR for the inline code.

I wanted to avoid having multiple return paths for the code for better maintainability / debugging. This way the code reaches the end of the functions' scope, no matter what.

But I also understand the approach to try to return as soon as possible. What is the general strategy and code style for this within arduino framework / component?

me-no-dev:
I will agree with @mrengineer7777 here. Exit as soon as possible.
```cpp
if (!_cookieJar) {
    return;
}
```

PR #6282, opened 2022-02-14
ADC esp32s2 attenuation fix for DAC pins

P-R-O-C-H-Y:
## Summary
Changed fixed DAC pins which are different for ESP32S2 to predefined DAC pins from SOC. 

## Impact
None.

### DIFF
--- a/cores/esp32/esp32-hal-adc.c
+++ b/cores/esp32/esp32-hal-adc.c
@@ -27,6 +27,7 @@
 #include "soc/rtc_io_reg.h"
 #include "esp32/rom/ets_sys.h"
 #include "esp_intr_alloc.h"
+#include "soc/dac_channel.h"
 #define DEFAULT_VREF    1100
 static esp_adc_cal_characteristics_t *__analogCharacteristics[2] = {NULL, NULL};
 static uint16_t __analogVRef = 0;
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y:
> Working with S3 too?

S3 does not support DAC.

Jason2866:
Ahh, yes! This https://github.com/espressif/arduino-esp32/blob/esp32-s3-support/cores/esp32/esp32-hal-adc.c#L38-L41 is probably a copy&paste thing

P-R-O-C-H-Y:
The S3 has 20 ADC channels working, but no DAC channel ;)

cores/esp32/esp32-hal-adc.c
```diff
@@ -142,9 +144,9 @@ bool __adcAttachPin(uint8_t pin){
 #endif
     }
 #if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
```
me-no-dev, line 146: while here, let's use the SOC defines te test wether DAC is supported and maybe how many channels

P-R-O-C-H-Y: SOC define used now.
According to how many channels there are always 2 channels when supported.

me-no-dev: It's not given that this will be true in the future. Why not have more?

Issue #6228, opened 2022-02-03
Please document the Arduino core Tools menu options


technoblogy:
### Related area

Arduino core documentation

### Hardware specification

All ESP32, ESP-S2, ESP-C3, and ESP-S3 boards

### Is your feature request related to a problem?

Many of the options on the Arduino IDE Tools menu for the ESP32 core aren't obvious, but I can't find them documented anywhere. If they are documented, please mention where to find this information in the main Arduino core ReadMe.

For example:

**Upload Speed:** Why might I want to choose 115200 rather than the default 921600?
**USB CDC on Boot/USB Firmware MSC on Boot/USB DFU on Boot:** What are these for, and recommended settings?
**CPU Frequency:** I assume the **(WiFi/BT)** options mean you must choose one of these to use WiFi or BT.
**Flash Frequency:** How do I know what setting to choose?
**Flash Mode:** Again, how do I know what setting to choose?
**Partition Scheme:** What's the difference between FAT and FATFS? What about LittleFS?
**Arduino Runs On/Events Run On:** When would I want to use these?

Perhaps a statement saying that the default option is usually the recommended choice would solve some of these.

Searching on [docs.espressif.com](https://docs.espressif.com/projects/arduino-esp32/en/) doesn't find answers to most of these.

### Describe the solution you'd like

A sentence expanding the menu option into a sentence, explaining the conditions under which you might want to choose other than the default option (eg lower power versus higher speed), with an explanation of the acronyms (CDC, MSC, DFU, etc), and perhaps a link to fuller information would be ideal.

This should ideally go as a separate section **Arduino core Tools menu** in the documentation between **Installing (Windows, Linux and macOS)** and **Libraries**.

I'm happy to draft something, with gaps where I don't know the answers.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

henrygab:
Not a solution, but offering my decoding of some of the acronyms:

USB CDC == Communications Device Class
USB MSC == Mass Storage Class
USB DFU == Device Firmware Update

Thus, it's possible that choosing ...
* CDC causes exposure of a COM-port (aka Serial Port)
* MSC causes exposure of a USB flash storage device
* DFU causes exposure of a device firmware update USB class (e.g., Base Class FEh, subclass 01h, protocol 01h)

See generally https://www.usb.org/defined-class-codes

VojtechBartoska:
[...]

PR #6284, opened 2022-02-14
Added documentation for the Arduino IDE tools menu

pedrominatel:
## Summary
Added documentation for the Arduino IDE tools menu

## Impact
None.

## Related links
Closes https://github.com/espressif/arduino-esp32/issues/6228


docs/source/guides/tools_menu.rst
```diff
@@ -0,0 +1,201 @@
...
+* **DIO** - Dual I/O Fast Read
+    * Two SPI pins are used to write to the flash and to read from flash.
+
+* **QOUT** - Quad Output Fast Read
+    * Four SPI pins are used to read the flash data.
+
+* **DOUT** - Dual Output Fast Read
+    * Two SPI pins are used to read flash data.
+
+If you don't know how the board flash is physically connected or the flash memory model, try the **QIO/QOUT** first and then **DIO/DOUT**.
```
SuGlider, line 93: I think that testing DIO/DOUT first makes more sense as it uses a simpler pinout schema.
Some boards may be configured to DIO only and may reset with QIO. 
Testing QIO first may lead to issues in a first try, although I think it is very rare.

docs/source/guides/tools_menu.rst
```diff
@@ -0,0 +1,201 @@
...
+.. note:: If you have issues while flashing the device at high speed, try to decrease this value. This could be due to the external serial-to-USB chip limitations.
+
+CPU Frequency
+*************
+
+On this option, you can select the CPU clock frequency. This option is critical and must be selected according to the high-frequency crystal present on the board and the radio usage (Wi-Fi and Bluetooth).
+
+In some application, reducing the CPU clock frequency is recommended in order to reduce the power consumption.
+
+If you don't know why you shuld change this frequency, leave the default option.
```
SuGlider, line 62: Typo: shuld  ---> should

docs/source/guides/tools_menu.rst
```diff
@@ -0,0 +1,201 @@
...
+* **Error** - Onle at error level.
+* **Warning** - Only at warning level and above.
+* **Info** - Only at info level and above.
+* **Debug** - Only at debug level and above.
+* **Verbose** - Prints everything.
+
+PSRAM
+*****
+
+The PSRAM is an internal or external extended RAM present on some boards, modules or SoC..
```
SuGlider, line 131: or SoC. .  (two periods)

docs/source/guides/tools_menu.rst
```diff
@@ -0,0 +1,201 @@
...
+
+The USB Communications Device Class, or USB CDC, is a class used for basic communication to be used as a regular serial controller (like RS-232).
+
+This class is used for flashing the device without any other external device attached to the SoC.
+
+This option can be used to ``Enable`` or ``Disable`` this function at the boot. If this option is ``Enabled``, once the device is connected via USB, one new serial port will appear in the list of the serial ports.
+Use this new serial port for flashing the device.
+
+This option can be used as well for debugging via the ``Serial Monitor``. 
+
```
SuGlider, line 178: Maybe here it's worth to mention that ```Serial``` will direct all the output to the CDC device instead of UART0.
Thus a Serial.print("Hello!") will send the string to the USB, not to the UART.
In order to use the UART0, it must use ```Serial0``` instead of ```Serial```

docs/source/guides/tools_menu.rst
```diff
@@ -0,0 +1,201 @@
...
+USB Options
+-----------
+
+Some ESP32 families have a USB peripheral. This peripheral can be used for flashing and debugging.
+
+Currently, the SoC's with USB supported are:
+
+* ESP32-C3 (CDC only)
+* ESP32-S2
+* ESP32-S3 (in development mode, not stable yet)
```
VojtechBartoska, line 163: Maybe we can take a "data" for this from Supported SoC's table so when support changes. All documentation pages will be updated automatically.

OR

Just share here a link for [supported SoCs table](https://docs.espressif.com/projects/arduino-esp32/en/latest/getting_started.html#supported-soc-s) an update only that table.

docs/source/index.rst
```diff
@@ -14,6 +14,7 @@ Here you will find all the relevant information about the project.
    Getting Started <getting_started>
    Libraries <libraries>
    Tutorials <tutorials/tutorials>
+   Guides <guides/guides>
```
VojtechBartoska, line 17: Maybe this can be above Tutorials? I can also imagine this section under getting started.

VojtechBartoska (commented):
Few small notes. Otherwise it looks pretty good!

docs/source/guides/tools_menu.rst
pedrominatel: I think the opposite. If you have QIO/QOUT support on your HW, you should try it first and take the advantage of it.
It's better to fail first sometimes :)

me-no-dev: Chances are that DIO/DOUT will work anyways. Let's ask them to try QIO 80MHz first, so that they can get the best performance

me-no-dev: Actually there are very few flash chips that might need DIO and I know of only one board that requires DOUT. On other hand we will not get issues caused by performance degradation

pedrominatel: I've tried some alternative flash memories on ESP32 and only one worked out in DOUT mode. I think it was some Cypress S25 something.

Issue #5765, opened 2021-10-14
Esp32-c3 Bluetooth communication


dfrobotjw:
I want to get an Arduino example of esp32-c3. The function is that one esp32-c3 module sends data through Bluetooth, and the other module receives and prints data through Bluetooth

chegewara:
What is wrong with ALL those examples?
https://github.com/espressif/arduino-esp32/tree/master/libraries/BLE/examples

VojtechBartoska:
@dfrobotjw Can I consider this solved?

SuGlider:
@dfrobotjw 
Please use this 2 examples (BLE client and server):
https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/examples/BLE_client/BLE_client.ino
https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/examples/BLE_server/BLE_server.ino

They are marked not to be visible when ESP32-C3 board is selected, but they actually work fine with the C3.
You will need 2 C3 boards, one to run the client and another to run the server. 
Turn on menu option "Core Debug Level" to "Verbose" in order to get BLE log messages and to see what is going on in the boards.

PR #6285, opened 2022-02-14
Adds BLE examples to ESP32-C3 CI cycle

SuGlider:
-----------
## Summary
BLE examples have ```.skip.esp32c3``` file, when actually all those examples run perfectly fine with the C3.
This PR removes this restiction and adds those examples to the CI process.

## Impact
None.

## Related links
Related issue: https://github.com/espressif/arduino-esp32/issues/5765


VojtechBartoska:
Related issue: #5765

PR #6310, opened 2022-02-19
(boards.txt) Add partition scheme menu for AI Thinker ESP-CAM boards

HACKER-3000:
## Summary
Adds the Partition Scheme menu to the AI Thinker board

## Impact
Makes different partition schemes available in case someone needs OTA for example


me-no-dev:
thanks @HACKER-3000

PR #6323, opened 2022-02-21
ESP32-S3 SDMMC support

igrr:
## Summary

This PR adds SDMMC support for ESP32-S3.

Unlike ESP32, ESP32-S3 allows all SDMMC pins to be configured flexibly using GPIO matrix. A new function is added to SDMMC class to allow setting the pins.

## Related links

SDMMC docs for S3: https://docs.espressif.com/projects/esp-idf/en/latest/esp32s3/api-reference/peripherals/sdmmc_host.html

### DIFF
--- a/libraries/SD_MMC/src/SD_MMC.cpp
+++ b/libraries/SD_MMC/src/SD_MMC.cpp
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 Espressif Systems (Shanghai) PTE LTD
+// Copyright 2015-2021 Espressif Systems (Shanghai) PTE LTD
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


libraries/SD_MMC/src/SD_MMC.cpp
```diff
@@ -12,29 +12,76 @@
...
     : FS(impl), _card(NULL)
-{}
+{
+#ifdef SOC_SDMMC_USE_GPIO_MATRIX
+    _pin_clk = 1;
+    _pin_cmd = 2;
+    _pin_d0 = 3;
+    _pin_d0 = 4;
+    _pin_d0 = 5;
+    _pin_d0 = 6;
```
igrr, line 40: These will ideally have to come from pins_arduino.h, if the board variant specifies the pins.

me-no-dev: if not defined in pins_arduino.h, they should be `GPIO_NUM_NC` by default and the driver should not start if it has missing pins. That way we handle both boards with predefined pins and custom user configs.

me-no-dev: maybe something like:
```c
#ifdef BOARD_HAS_SDMMC
    _pin_clk = SDMMC_CLK;
    _pin_cmd = SDMMC_CMD;
    _pin_d0 = SDMMC_D0;
  #ifdef BOARD_HAS_1BIT_SDMMC
    _pin_d1 = -1;
    _pin_d2 = -1;
    _pin_d3 = -1;
  #else
    _pin_d1 = SDMMC_D1;
    _pin_d2 = SDMMC_D2;
    _pin_d3 = SDMMC_D3;
  #endif
#elif CONFIG_IDF_TARGET_ESP32
    _pin_clk = SDMMC_SLOT1_IOMUX_PIN_NUM_CLK;
    _pin_cmd = SDMMC_SLOT1_IOMUX_PIN_NUM_CMD;
    _pin_d0 = SDMMC_SLOT1_IOMUX_PIN_NUM_D0;
    _pin_d1 = SDMMC_SLOT1_IOMUX_PIN_NUM_D1;
    _pin_d2 = SDMMC_SLOT1_IOMUX_PIN_NUM_D2;
    _pin_d3 = SDMMC_SLOT1_IOMUX_PIN_NUM_D3;
#else
    _pin_clk = -1;
    _pin_cmd = -1;
    _pin_d0 = -1;
    _pin_d1 = -1;
    _pin_d2 = -1;
    _pin_d3 = -1;
#endif
```

variants/esp32s3box/pins_arduino.h
```diff
@@ -62,12 +62,14 @@ static const uint8_t T14 = 14;
...
 #define I2S_DOUT   15
 
+#define BOARD_HAS_SDMMC
 #define SDMMC_CLK  13
 #define SDMMC_CMD  11
 #define SDMMC_D0   14
 #define SDMMC_D1   12
 #define SDMMC_D2   10
 #define SDMMC_D3    9
+#define BOARD_MAX_SDMMC_FREQ SDMMC_FREQ_DEFAULT
```
igrr, line 72: To be removed after the above mentioned IDF issues are fixed and fixes are included into Arduino.

igrr:
To test on ESP32-S3-USB-OTG, apply the patch:
```patch
diff --git a/variants/esp32s3/pins_arduino.h b/variants/esp32s3/pins_arduino.h
index 5cd22fa5..f906bd3f 100644
--- a/variants/esp32s3/pins_arduino.h
+++ b/variants/esp32s3/pins_arduino.h
@@ -64,4 +64,15 @@ static const uint8_t T14 = 14;
 static const uint8_t DAC1 = 17;
 static const uint8_t DAC2 = 18;
 
+
+#define BOARD_HAS_SDMMC
+#define SDMMC_CLK  36
+#define SDMMC_CMD  35
+#define SDMMC_D0   37
+#define SDMMC_D1   38
+#define SDMMC_D2   33
+#define SDMMC_D3   34
+
+#define BOARD_MAX_SDMMC_FREQ SDMMC_FREQ_DEFAULT
+
 #endif /* Pins_Arduino_h */
```

me-no-dev:
I will actually add a whole board definition for ESP32-S3-USB-OTG, because it's an interesting case to have as example for board support.

PR #6341, opened 2022-02-23
Esp32 s3 support

me-no-dev:


Ouss4:
> @Ouss4 seems that S3 is not yet supported in hardware: https://github.com/espressif/arduino-esp32/runs/5498592665?check_suite_focus=true#step:7:48

It's `pytest-embedded` that's missing an S3 chip.  I'll submit a PR.

VojtechBartoska:
In this PR is included commit which will Close #5864.

chatfrog:
Have you really tested your code? I tested almost all examples, they could not pass compiling process. I tried Arduino IDE 1.8 and 2.0 separately with latest arduino ESP32 core 2.07  and 2.04.

ConstantinCreations:
I don't see how this code can possibly work with the ESP32-S3. You need to specify I2S pins for the S3 since there are no defaults. You can use almost any pin you want for I2S on an ESP32-S3.

PR #6357, opened 2022-02-28
WiFi DA: Added Dual Antenna to the docs and example created

pedrominatel:
## Summary

* Added the Dual Antenna documentation.
* Added the DA example.

## Impact

None.


me-no-dev:
you should revisit the example once you add the board definition :)

PR #6358, opened 2022-02-28
[esp32-hal-log.h] add #ifdef for LOG_LOCAL_LEVEL

cyberman54:
#ifdef added, to avoid compiler redefinition warnings for `LOG_LOCAL_LEVEL` if defined by application, and we `USE_ESP_IDF_LOG`


me-no-dev:
nice catch :)

PR #6392, opened 2022-03-08
Fixed the antenna GPIO argument on the DA example

pedrominatel:
## Summary

Fix the GPIO number argument for the dual antenna selection function in the example.

## Impact

The GPIO is the same for ANT1 and ANT2. This should not work as expected on the DA example.

## Related links

N/A

### DIFF
--- a/libraries/WiFi/examples/WiFiScanDualAntenna/WiFiScanDualAntenna.ino
+++ b/libraries/WiFi/examples/WiFiScanDualAntenna/WiFiScanDualAntenna.ino
@@ -30,7 +30,7 @@ void setup()
      * 
      * Set WiFi dual antenna configuration by passing the GPIO and antenna mode for RX ant TX
      */ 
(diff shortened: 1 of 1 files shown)


libraries/WiFi/examples/WiFiScanDualAntenna/WiFiScanDualAntenna.ino
```diff
@@ -30,7 +30,7 @@ void setup()
      * 
      * Set WiFi dual antenna configuration by passing the GPIO and antenna mode for RX ant TX
      */ 
-    err = WiFi.setDualAntennaConfig(GPIO_ANT1, GPIO_ANT1, WIFI_RX_ANT_AUTO, WIFI_TX_ANT_AUTO);
+    err = WiFi.setDualAntennaConfig(GPIO_ANT1, GPIO_ANT2, WIFI_RX_ANT_AUTO, WIFI_TX_ANT_AUTO);
```
mrengineer7777, line 33: Good catch.  setDualAntennaConfig appears to be a new function.

mrengineer7777 (approved):

PR #6394, opened 2022-03-08
Allows setting only one pin (rx or tx) in the first begin()

### LINKED ISSUE #6395: Tasmota crashes on ESP32-PICO-V3 with recent core
### Discussed in https://github.com/espressif/arduino-esp32/discussions/6356

<div type='discussions-op-text'>

<sup>Originally posted by **barbudor** February 27, 2022</sup>
Hi
I can't report this as an issue because I unfortunately can't provide more qualified information but I'm facing a problem with Tasmota on ESP32-PICO-V3.
I'm using Tasmota with a driver for the french national electricity meter called Linky. The driver is meant to receive frames on a UART, decode them and feed the energy information to Tasmota's internals for publication.
The driver has been unchanged for months and until recently my Tasmota build with this driver was working fine on either ESP32-D0WDQ6 (mini-D1-ESP32) and on ESP32-PICO-V2.
The last version that was working fine on the PICO was using what @Jason2866 is calling `Core/SDK 2_0_2_2/v4.4-3-g6afb23d90a`. This frozen Core/SDK was in Tasmota since early February.
2 days ago, Jason's upgraded the Tasmota Core/SDK to `2_0_2_3/v4.4-2.0.2-420-g4ff9ceb19a` which he describe as "based on latest IDF44 commit and latest commit of Arduino branch esp32-s3-support".

The problem occurs as soon as I enable the Teleinfo driver which uses either UART1 or UART2 and the ESP crash with a Guru Meditation, Load Prohibited.
It doesn't seem to matter which GPIO are used for the UART. It doesn't even need to actually receive serial data.
The same binary works fine on ESP32-D0WDQ6.
Previous binaries were working well on both boards.
I tested a few other drivers that are using serial and they don't create such problem.

I'm sorry for not being able to provide more details but I thought it was worth to drop a note there.
Thanks for the hard work.
</div>

SuGlider:
## Summary
This PR allow users to start HardwareSerial setting just one pin (Rx or Tx) or both of them.
It allows these possible initial configurations:

```cpp
// In this case, Serial will configure rxPin and only be able to receive data
Serial1.begin(9600, SERIAL_8N1, rxPin);
```
```cpp
// In this case, Serial will configure txPin and only be able to send data
Serial1.begin(9600, SERIAL_8N1, -1, txPin);
```

```cpp
// In this case, Serial will configure rxPin and txPin with default pins for the chip (as in HardwareSerial.cpp)
Serial1.begin(9600, SERIAL_8N1);
```

```cpp
// In this case, Serial will configure rxPin and txPin as defined by the user
Serial1.begin(9600, SERIAL_8N1, rxPin, txPin);
```

## Impact
None

## Related links

This PR solves a discussion at https://github.com/espressif/arduino-esp32/discussions/6356
Fixes #6395

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -261,19 +261,25 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
     if (!uartIsDriverInstalled(_uart)) {
         switch (_uart_nr) {
             case UART_NUM_0:
-                rxPin = rxPin < 0 ? SOC_RX0 : rxPin;
-                txPin = txPin < 0 ? SOC_TX0 : txPin;
+                if (rxPin < 0 && txPin < 0) {
+                    rxPin = SOC_RX0;
+                    txPin = SOC_TX0;
+                }
             break;
 #if SOC_UART_NUM > 1                   // may save some flash bytes...
             case UART_NUM_1:
-                rxPin = rxPin < 0 ? RX1 : rxPin;
-                txPin = txPin < 0 ? TX1 : txPin;
+               if (rxPin < 0 && txPin < 0) {
+                    rxPin = RX1;
+                    txPin = TX1;
+                }
             break;
 #endif
 #if SOC_UART_NUM > 2                   // may save some flash bytes...
             case UART_NUM_2:
-                rxPin = rxPin < 0 ? RX2 : rxPin;
-                txPin = txPin < 0 ? TX2 : txPin;
+               if (rxPin < 0 && txPin < 0) {
+                    rxPin = RX2;
+                    txPin = TX2;
+                }
             break;
 #endif
             default:


cores/esp32/HardwareSerial.cpp
```diff
@@ -261,19 +261,25 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
     if (!uartIsDriverInstalled(_uart)) {
         switch (_uart_nr) {
             case UART_NUM_0:
-                rxPin = rxPin < 0 ? SOC_RX0 : rxPin;
-                txPin = txPin < 0 ? SOC_TX0 : txPin;
+                if (rxPin < 0 && txPin < 0) {
```
mrengineer7777, line 264: This doesn't make sense to me.  The previous code 
```
rxPin = rxPin < 0 ? SOC_RX0 : rxPin;
txPin = txPin < 0 ? SOC_TX0 : txPin;
```
allowed changing pins individually.  With this code if rxPin is set then txPin won't be set to default.

Jason2866: @mrengineer7777  See https://github.com/espressif/arduino-esp32/discussions/6356#discussioncomment-2311636

SuGlider: Yes, exactly. Previous code was forcing any undefined pin to go to the default IO Pin, only in the first ```Serial.begin()``` call.

Now it only forces default pins when no pin is defined.

When only defining rxPin, tx IOMUX will be unset. 
Thus, ```Serial.write()``` would not have any effect, but it also doesn't crash ESP32. 
The same for txPin alone -> ```Serial.read()``` will always return -1.

SuGlider: > This doesn't make sense to me. The previous code
> allowed changing pins individually. With this code if rxPin is set then txPin won't be set to default.

Please note that this code only applies to the first time Serial.begin() is called.
Subsequent calls of Serial.begin() will change rxPin and/or TxPin individually.

For example:
```cpp
Serial1.begin(9600, SERIAL_8N1, 12, 14);   // first time calling begin will set rxPin and txPin
Serial1.begin(115200);                     // doesn't change any pin (rx/tx), but only baudrate
Serial1.begin(115200, SERIAL_8N1, -1, 27); // only changes txPin to 27, keeping rxPin in 12
Serial1.end();                             // no seeting is kept from the setup above
Serial1.begin(115200);                     // new begin() from an end() will configure rxPin and txPin 
                                           // to DEFAULT pins (Serial1 9 and 10 on ESP32)
```

mrengineer7777: Ok, so this is a deliberate rollback to previous behavior.  Looks like the crash was being caused by the TX pin defaulting to a pin that's not available on the user's hardware.  One question: does passing TX=-1 to the underlying driver uartBegin() cause any issues?  I'm guessing not based on this comment:
>>Thus, Serial.write() would not have any effect, but it also doesn't crash ESP32.

mrengineer7777 (approved):

cores/esp32/HardwareSerial.cpp
SuGlider: No issues with TX or RX as -1. ```uartBegin()``` will just not associate its digital pad to any IO_MUX signal.
The consequence is that the function related to the pin won't work (sending, receiving, etc). 
Anyway, this feature must be used with caution.

TS6000:
Hello
I have big problems on UART2.
Since the change from 1.xx to 2.02 data is lost.
I work with Arduino Studio.

I use UART2 with the pins (GPIO) 12 and 13

Torsten

PR #6364, opened 2022-03-01
Adds C++ std::function to Serial.onReceive()

SuGlider:
## Summary
This PR adds C++ std::funcion to ```HardwareSerial::onReceive()```
It also adds a way to get notification of UART errors using ```HardwareSerial::onReceiveError()```

if Fixes FIFO and buffer Overflow issues.

Example of labda function with ```Serial.onReceive()```:

``` cpp
  Serial.onReceive([]() {
    Serial.println("Got data!");
    while (Serial.available()) {
      Serial.write(Serial.read());
    }
    Serial.println();
  });
```

## Impact
Modernizes Arduino Serial in order to use C++ 11.

## Related links

Fixes #6326 
Fixes #6336 
Resolves #6302

PR #6397, opened 2022-03-08
Adds HardwareSerial::setRxTimeout()

gonzabrusco:
-----------
## Summary
Adds onReceiveTimeout() to HardwareSerial. Default value of the timeout is 10 symbols. This method allows the user to change the timeout on which the onReceive callback is called. 
Before this change, the callback was called on 120 bytes always and then on timeout (when receiving more than 120 bytes). Now the callback gets called ONLY on timeout. 

## Impact
No impact. Everything should work as always.

## Related links
No issue related. I did this modifications myself because on my modbus library I was seeing that on long messages (more than 120 bytes), the callbacks was getting called twice. Now it only calls for timeouts (of no receiving data).


gonzabrusco:
More information about this.

I've tested this timeout on the oscilloscope and it's working as expected! 🎉🎉🎉

One consideration. If an overflow occurs, and then you don't process the data on the buffer with onReceiveError callback, then onReceive will never get called again. Maybe I should add a warning on the comments.

SuGlider:
Thanks @gonzabrusco !
I'll review it soon.

TS6000:
Hello
I have big problems on UART2.
Since the change from 1.xx to 2.02 data is lost.
I work with Arduino Studio.

I use UART2 with the pins (GPIO) 12 and 13

Torsten

gonzabrusco:
> Hello I have big problems on UART2. Since the change from 1.xx to 2.02 data is lost. I work with Arduino Studio.
> 
> I use UART2 with the pins (GPIO) 12 and 13
> 
> Torsten

Hello @TS6000 . This is a Pull Request. Please create a new issue if you have one. But before creating one, take note that the UART code have changed a lot in the latest master. So please try again using master or wait for 2.0.3 version.

cores/esp32/HardwareSerial.cpp
```diff
@@ -210,15 +228,17 @@ void HardwareSerial::_uartEventTask(void *args)
             if(xQueueReceive(uartEventQueue, (void * )&event, (portTickType)portMAX_DELAY)) {
                 switch(event.type) {
                     case UART_DATA:
-                        if(uart->_onReceiveCB && uart->available() > 0) uart->_onReceiveCB();
+                        if(uart->_onReceiveCB && uart->available() > 0  && event.timeout_flag) uart->_onReceiveCB();
                         break;
                     case UART_FIFO_OVF:
                         log_w("UART%d FIFO Overflow. Consider adding Hardware Flow Control to your Application.", uart->_uart_nr);
                         if(uart->_onReceiveErrorCB) uart->_onReceiveErrorCB(UART_FIFO_OVF_ERROR);
+                        xQueueReset(uartEventQueue);
```
me-no-dev, line 236: why lose all previous data?

cores/esp32/HardwareSerial.cpp
```diff
@@ -210,15 +228,17 @@ void HardwareSerial::_uartEventTask(void *args)
...
                         break;
                     case UART_BUFFER_FULL:
                         log_w("UART%d Buffer Full. Consider encreasing your buffer size of your Application.", uart->_uart_nr);
                         if(uart->_onReceiveErrorCB) uart->_onReceiveErrorCB(UART_BUFFER_FULL_ERROR);
+                        xQueueReset(uartEventQueue);
```
me-no-dev, line 241: same question as above.

cores/esp32/HardwareSerial.cpp
gonzabrusco: I did this because it was like this on the previous UART task (before latest SuGuiller commit). But maybe this could be ommited. The reason is that when you have a big overflow (overflowing multiple times the FIFO, the event queue gets filled with one UART_BUFFER_FULL and multiple UART_FIFO_OVF. I though it was redundant to be called multiple times for the same event.

cores/esp32/HardwareSerial.cpp
gonzabrusco: Same answer.

cores/esp32/HardwareSerial.cpp
gonzabrusco: This is what I was talking about: https://github.com/espressif/arduino-esp32/blob/c014eaf3524dd45049a812a393c8fdfc92feb886/cores/esp32/esp32-hal-uart.c#L117

If you think this should be deleted, I have no opposition.

me-no-dev: I would rather discuss it, than just delete it :) @SuGlider want to pitch in an idea too?

gonzabrusco: Well, I guess this is a bit relative. It could happen that you have a very fast overflow and you will be called multiple times with the UART_FIFO_OVF event. But thinking about this more thoroughly, maybe that is not a "normal" or "frequent" situation. I added it back because I saw it on the original esp32-hal-uart code and when I was testing the overflow with the serial monitor, I noticed that behaviour. But as I said, probably is not a frequent situation and you could eventually lose something important on the event queue? (maybe an UART_DATA event after an overflow event?). Probably this should be deleted. But I will let @SuGlider give this opinion about this. I would love to have this on 2.0.3 because I will be using that version as the new core for my project (2.0.2 has many bugs with UART).

SuGlider:
> Before this change, the callback was called on 120 bytes always and then on timeout (when receiving more than 120 bytes). Now the callback gets called ONLY on timeout.

Yes, this is correct, onReceive Callback will be called right after 120 bytes or less.  
This PR creates an alternative for getting called back after Timeout in Symbol baudrate timing.

I got confused with the Function Name ```onReceiveTimeout()``` makes me think about setting a callback function and not a timeout paramter. ****I think it can be changed to ```setRxTimeout()```.****
 
 I will review the code and reseach about the implications, given that it may affect basic functions such as ```read()``` for instance.
 Thus, it may take a bit longer.

gonzabrusco:
> > Before this change, the callback was called on 120 bytes always and then on timeout (when receiving more than 120 bytes). Now the callback gets called ONLY on timeout.
> 
> Yes, this is correct, onReceive Callback will be called right after 120 bytes or less. This PR creates an alternative for getting called back after Timeout in Symbol baudrate timing.

Hi @SuGlider. About this, Now the onReceive Callback gets called on 10 symbol time UART silence by default (because it's IDF default) or 120 bytes received. This callback just allows you to set the callback timeout (because for example modbus has an interframe delay of 3.5 symbols) and disables the redundant call every 120 bytes (is that really needed?). That way you can be called only when all the data was already received (and not in the middle of the reception for messages longer than 120 bytes). A Modbus frame could be as long as 255 bytes for example.

The other alternative is to inform the onReceive callback if the callback was produced because of a timeout or because 120 bytes were received. That way the callback can be able to discern between the two and take different actions. 

> I got confused with the Function Name `onReceiveTimeout()` makes me think about setting a callback function and not a timeout parameter. ****I think it can be changed to `setRxTimeout()`.****
> 
> I will review the code and reseach about the implications, given that it may affect basic functions such as `read()` for instance. Thus, it may take a bit longer.

No problem about the name change. I just gave it that name because it was the timeout on which onReceive callback will be called. Take you time if you need to (although I would love this to be available on 2.0.3 which I will use for my project). About the implications on other functions, I'm not so sure if there is any. This just sets the timeout for the event UART_DATA. Not sure how this could affect other functions like read().

cores/esp32/HardwareSerial.cpp
SuGlider: Last suggestion is about changing the name of the function ```void HardwareSerial::onReceiveTimeout(uint8_t symbols_timeout)``` to ```void HardwareSerial::setOnReceiveTimeout(uint8_t symbols_timeout)``` or some name like ```setRxTimeout()```.


Timeout mode shall not be forced in the code... it shall be an option. check 

``` cpp
//  from void HardwareSerial::onReceiveTimeout(uint8_t symbols_timeout) :::

    if(symbols_timeout == 0) {
        _onReceiveTimeout = 1; // Never disable timeout
    }
```

``` cpp
// from void HardwareSerial::begin(...) :::

    // Set UART RX timeout
    if (_uart != NULL && _onReceiveCB != NULL) {
        uart_set_rx_timeout(_uart_nr, _onReceiveTimeout);
    }
```

``` cpp
// from void HardwareSerial::_uartEventTask(void *args)) ::: "&& event.timeout_flag"

                    case UART_DATA:
                        if(uart->_onReceiveCB && uart->available() > 0  && event.timeout_flag) uart->_onReceiveCB();
                        break;
```

cores/esp32/HardwareSerial.h
```diff
@@ -73,9 +73,21 @@ class HardwareSerial: public Stream
     HardwareSerial(int uart_nr);
     ~HardwareSerial();
 
-    // onReceive will setup a callback for whenever UART data is received
+    // onReceive will setup a callback for whenever UART data is received and a later timeout occurs (no receiving data after specified time):
+    // param function        is the callback to be called after reception timeout
+    
     // it will work as UART Rx interrupt -- Using C++ 11 std::fuction 
     void onReceive(OnReceiveCb function);
+
```
SuGlider, line 81: Please review the name of this function. 
Arduino uses function names such as **onSomething()** only when setting a callback function.
Please use **setSomething()** instead.

cores/esp32/HardwareSerial.h
```diff
@@ -138,6 +150,7 @@ class HardwareSerial: public Stream
     uart_t* _uart;
     size_t _rxBufferSize;
     OnReceiveCb _onReceiveCB;
+    uint8_t _onReceiveTimeout;
```
SuGlider, line 153: this may not be necessary to exist anymore as per the suggestions already made here.

SuGlider (changes requested):
Please review all the suggestions. Thanks!

cores/esp32/HardwareSerial.cpp
```diff
@@ -191,10 +192,27 @@ void HardwareSerial::onReceive(OnReceiveCb function)
...
+    }
+    HSERIAL_MUTEX_UNLOCK();
+}
+
+void HardwareSerial::onReceiveTimeout(uint8_t symbols_timeout)
+{
+    HSERIAL_MUTEX_LOCK();
+    
+    if(symbols_timeout == 0) {
+        _onReceiveTimeout = 1; // Never disable timeout
```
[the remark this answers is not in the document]
gonzabrusco: Will do.

cores/esp32/HardwareSerial.cpp
```diff
@@ -191,10 +192,27 @@ void HardwareSerial::onReceive(OnReceiveCb function)
...
+
     // this can be called after Serial.begin(), therefore it shall create the event task
     if (function != NULL && _uart != NULL && _eventTask == NULL) {
-        _createEventTask(this);
+        _createEventTask(this); // Create event task
+    }
+    HSERIAL_MUTEX_UNLOCK();
+}
+
+void HardwareSerial::onReceiveTimeout(uint8_t symbols_timeout)
```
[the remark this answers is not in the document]
gonzabrusco: Ok.

cores/esp32/HardwareSerial.cpp
```diff
@@ -128,7 +128,8 @@ HardwareSerial::HardwareSerial(int uart_nr) :
 _uart_nr(uart_nr), 
 _uart(NULL), 
 _rxBufferSize(256), 
-_onReceiveCB(NULL), 
+_onReceiveCB(NULL),
+_onReceiveTimeout(10),
```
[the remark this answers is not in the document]
gonzabrusco: This is the only thing I disagree. By default now (as before my pull request), onReceive callback is being called by timeout (10 symbos by default) or by rx fifo full (120 bytes). That is the standard behaviour by IDF. I don't think that Arduino core should use another default. I would prefer to let the timeout on because if you receive 119 bytes, you will never get called (unless you receive on more byte that will trigger UART_INTR_RXFIFO_FULL interrupt)

cores/esp32/HardwareSerial.cpp
```diff
@@ -210,15 +228,17 @@ void HardwareSerial::_uartEventTask(void *args)
             if(xQueueReceive(uartEventQueue, (void * )&event, (portTickType)portMAX_DELAY)) {
                 switch(event.type) {
                     case UART_DATA:
-                        if(uart->_onReceiveCB && uart->available() > 0) uart->_onReceiveCB();
+                        if(uart->_onReceiveCB && uart->available() > 0  && event.timeout_flag) uart->_onReceiveCB();
```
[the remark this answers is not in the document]
gonzabrusco: Ok. I will pass the timeout flag as a parameter to the callback.

cores/esp32/HardwareSerial.cpp
```diff
@@ -320,6 +340,12 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
     if (_uart != NULL && (_onReceiveCB != NULL || _onReceiveErrorCB != NULL) && _eventTask == NULL) {
         _createEventTask(this);
     }
+
+    // Set UART RX timeout
+    if (_uart != NULL && _onReceiveCB != NULL) {
+        uart_set_rx_timeout(_uart_nr, _onReceiveTimeout);
+    }
+
```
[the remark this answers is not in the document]
gonzabrusco: As I said before. The timeout behaviour is the default as per IDF (10 symbols defined in UART_TOUT_THRESH_DEFAULT). I think this is needed because the user could call setRxTimeout before or after the begin() . This is the same principle as you took with onReceive(). This is because uart_set_rx_timeout should be called once the uart is initialized.

cores/esp32/HardwareSerial.h
gonzabrusco: Will do.

cores/esp32/HardwareSerial.h
gonzabrusco: I will stress my previous comments here. Because you need to call uart_set_rx_timeout after the uart initialization, then I need a local variable to store my timeout config in case setRxTimeout was called before begin().

gonzabrusco:
I commited the requested changes but I'm still testing it. Please wait before reviewing again. @SuGlider

gonzabrusco:
This is ready for review again. I'm not sure if I should mark "Resolved" all previous conversations. @SuGlider

Here´s a simple sketch to test this. 

```
#include <Arduino.h>

void receiveFnc(bool timeout) {
    if(timeout) {
        Serial.printf("Received Data UART RX TIMEOUT. Available For Reading = %u\n", Serial.available());
    }
    else {
        Serial.printf("Received Data UART RX FIFO FULL. Available For Reading = %u\n", Serial.available());
    }
    while(Serial.available()) Serial.read(); // Discard received data.
}

void receiveErrorFnc(hardwareSerial_error_t error){
    Serial.print("UART Reception Error: ");
    switch(error) {
        case UART_BREAK_ERROR:
            Serial.println("UART_BREAK_ERROR");
            break;
        case UART_BUFFER_FULL_ERROR:
            Serial.println("UART_BUFFER_FULL_ERROR");
            break;
        case UART_FIFO_OVF_ERROR:
            Serial.println("UART_FIFO_OVF_ERROR");
            break;
        case UART_FRAME_ERROR:
            Serial.println("UART_FRAME_ERROR");
            break;
        case UART_PARITY_ERROR:
            Serial.println("UART_PARITY_ERROR");
            break;
    }
}

void setup() {
    // Setting UART Ring Buffer to 512 bytes
    Serial.setRxBufferSize(512);

    // Define UART Reception Callback
    Serial.onReceive(receiveFnc);
    
    // Define UART Error Callback
    Serial.onReceiveError(receiveErrorFnc);
    
    // Set RX Timeout. 
    // If set to 0, onReceive callback will only be called every 120 bytes.
    // If set to more then 0, onReceive callback will be called every 120 bytes or timeout (with less than 120 bytes)
    Serial.setRxTimeout(2);
    
    // Serial speed 9600 symbos per second
    Serial.begin(9600);
}

void loop() {
    // Nothing to do
}
```

cores/esp32/HardwareSerial.cpp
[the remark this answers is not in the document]
SuGlider: OK. I agree. Let's set the default always.

SuGlider (approved):
Thanks @gonzabrusco !
I just added a NULL test to ```eventQueueReset()```, in case it is called before ```Serial.begin()```, which can reset the board.

gonzabrusco:
> Thanks @gonzabrusco ! I just added a NULL test to `eventQueueReset()`, in case it is called before `Serial.begin()`, which can reset the board.

Nice catch!

What is the etiquete here in Espressif repositories about "conversations"? Should I mark them resolved or wait for @me-no-dev  or you to resolve them? Don't want to make a mess...

SuGlider:
> > Thanks @gonzabrusco ! I just added a NULL test to `eventQueueReset()`, in case it is called before `Serial.begin()`, which can reset the board.
> 
> Nice catch!

Actually, that change was not necessary because the function to get the queue already tests it. So I reverted the commit.

> What is the etiquete here in Espressif repositories about "conversations"? Should I mark them resolved or wait for @me-no-dev or you to resolve them? Don't want to make a mess...

I'll run a final testing sequence here and then I'll post a message to me-no-dev asking for his final review.

Thanks for the PR!

SuGlider:
@gonzabrusco 

I have tested this PR and also the last PR #6364 
I see no difference now... 

Running the example sketch, but removing the ```bool timeout``` from ```void receiveFnc(bool timeout)``` and also removing the new function ```Serial.setRxTimeout(5);```, I got exactly the same results, but it just can't know if the data came because of timeout or FIFO Full. 

I think that your intention was to only get called back when timeout happens, right?
This way the RingBuffer can hold all data at once and the callback can process it on a single call.

I have changed this PR to reflect this intention, which in fact is better for the user, as you have suggested.


New way of doing it:

``` cpp
#include <Arduino.h>

void receiveFnc() {
    Serial.printf("Received UART Data. Available For Reading = %u\n", Serial.available());
    while(Serial.available()) Serial.read(); // Discard received data.
}

void receiveErrorFnc(hardwareSerial_error_t error){
    Serial.print("UART Reception Error: ");
    switch(error) {
        case UART_BREAK_ERROR:
            Serial.println("UART_BREAK_ERROR");
            break;
        case UART_BUFFER_FULL_ERROR:
            Serial.println("UART_BUFFER_FULL_ERROR");
            break;
        case UART_FIFO_OVF_ERROR:
            Serial.println("UART_FIFO_OVF_ERROR");
            break;
        case UART_FRAME_ERROR:
            Serial.println("UART_FRAME_ERROR");
            break;
        case UART_PARITY_ERROR:
            Serial.println("UART_PARITY_ERROR");
            break;
    }
}

void setup() {
    // Setting UART Ring Buffer to 512 bytes
    Serial.setRxBufferSize(512);

    // Define UART Reception Callback
    Serial.onReceive(receiveFnc);   // default is being called on UART RX timeout
    // in order to be called when RX FIFO reaches 120 bytes or when RX timeout(with less than 120 bytes), 
    // set the callback in this way:
    // Serial.onReceive(receiveFnc, false); 
    
    // Define UART Error Callback
    Serial.onReceiveError(receiveErrorFnc);
    
    // Serial speed 9600 symbos per second
    Serial.begin(9600);
}

void loop() {
    // Nothing to do
}
```

Please let me know if you agree to these changes I have commited to your PR.
Thanks again!

SuGlider:
@gonzabrusco 
I think that removing most funtion and only adding a parameter to tell to onReceive() if it should consider only timeout or the combinations of timeout and FIFO with 120 bytes gets back to what was the initial intention of this PR.

No need to setRxTimeout because it already works as intended. This may be confusing for the regular Arduino user.

eventQueueReset() may be useful, so let's keep it.

Testing the example I have added here above, the user can decide in an easy way if the callback shall work only on timeout or when blocks of 120 bytes or timeout happens. It is flexible and simple.

Please let me know your thoughts.

gonzabrusco:
Hi @SuGlider . Thanks for the changes in the PR but unfortunately I have to disagree on the decision to remove setRXTimeout which was the primary goal of this PR.

The reason I needed this method is to be able to set correctly the timeout (in symbols) in which my callback was going to be called. This is CRITICAL on modbus over RS485 because you HAVE TO process any message in the bus in exactly 3.5 symbols time. If you process the buffer in 10 symbols time (like IDF default), you are arriving too late and you could end up seeing one master request and one slave response joined as a single message.

You can see that LIBMODBUS on IDF sets this timeout to 3 for example (using the same function I was using on the first place uart_set_rx_timeout). 

I agree with your changes but please revert the removal of setRXTimeout.

SuGlider:
@gonzabrusco 
I have added ```setRxTimeout()``` to the PR as you have suggested and done before.
I just added some checks to keep ```onReceive()``` safe and working even when user sets timeout to zero or uses this function when timeout was set to zero, in any order that both are called.

Please let me know if you also agree to this changes in order to start merging process.

gonzabrusco:
> @gonzabrusco I have added `setRxTimeout()` to the PR as you have suggested and done before. I just added some checks to keep `onReceive()` safe and working even when user sets timeout to zero or uses this function when timeout was set to zero, in any order that both are called.
> 
> Please let me know if you also agree to this changes in order to start merging process.

I loved the changes. Thanks!

SuGlider:
@me-no-dev - This PR is ready for merging, just waiting for your final review.

me-no-dev:
@gonzabrusco @SuGlider here seems to be a conflict now. Please fix

SuGlider:
Conflict solved. Thanks.

PR #6398, opened 2022-03-09
Added another overloaded WiFiSTAClass::begin() function that provides…

jpswensen:
… an easy way of creating a WPA2 Enterprise connections.

-----------
## Summary
The examples demonstrate how to create a WPA2 Enterprise connection, but it requires using various direct esp_idf functions. This patch is intended to create another overloaded version of the WiFi.begin() function that allows a user to create a WPA2 Enterprise connection in much the same way as different kinds of connections.

My only question for the core maintainers is whether I should leave those #ifdef's in there. I added them so that it was easy to disable all the code I added via defines from my platformio.ini file, but they technically aren't necessary.

## Impact
This should make it easier for novice users to create WPA2 Enterprise connections. For my university, I didn't need a root certificate or the client certificate or client key, so I haven't been able to debug those scenarios, but I built the begin functions to allow any one of those to be used, if needed.

I can confirm that eduroam-style WPA2 Enterprise networks that only require authentication with a username and password works as expected.

## Related links
N/A

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.h
+++ b/libraries/WiFi/src/WiFiGeneric.h
@@ -31,12 +31,6 @@
 #include "esp_smartconfig.h"
 #include "wifi_provisioning/manager.h"
 
-#ifdef ENABLE_WPA2_AUTHENTICATION
-#include <WiFiClient.h>
-#include <WiFiClientSecure.h>
-#include "esp_wpa2.h"
-#endif
-
 ESP_EVENT_DECLARE_BASE(ARDUINO_EVENTS);
 
 typedef enum {
--- a/libraries/WiFi/src/WiFiGeneric.h
+++ b/libraries/WiFi/src/WiFiGeneric.h
@@ -31,6 +31,12 @@
 #include "esp_smartconfig.h"
 #include "wifi_provisioning/manager.h"
 
+#ifdef ENABLE_WPA2_AUTHENTICATION
+#include <WiFiClient.h>
+#include <WiFiClientSecure.h>
+#include "esp_wpa2.h"
+#endif
+
 ESP_EVENT_DECLARE_BASE(ARDUINO_EVENTS);
 
 typedef enum {
--- a/libraries/WiFi/examples/WiFiClientEnterprise/WiFiClientEnterprise.ino
+++ b/libraries/WiFi/examples/WiFiClientEnterprise/WiFiClientEnterprise.ino
@@ -1,10 +1,22 @@
 #include <WiFi.h> //Wifi library
 #include "esp_wpa2.h" //wpa2 library for connections to Enterprise networks
 #define EAP_IDENTITY "login" //if connecting from another corporation, use identity@organisation.domain in Eduroam 
+#define EAP_USERNAME "login" //oftentimes just a repeat of the identity
 #define EAP_PASSWORD "password" //your Eduroam password
 const char* ssid = "eduroam"; // Eduroam SSID
 const char* host = "arduino.php5.sk"; //external server domain for HTTP connection after authentification
 int counter = 0;
+
+// NOTE: For some systems, various certification keys are required to connect to the wifi system.
+//       Usually you are provided these by the IT department of your organization when certs are required
+//       and you can't connect with just an identity and password.
+//       Most eduroam setups we have seen do not require this level of authentication, but you should contact
+//       your IT department to verify.
+//       You should uncomment these and populate with the contents of the files if this is required for your scenario (See Example 2 and Example 3 below).
+//const char *ca_pem = "insert your CA cert from your .pem file here";
+//const char *client_cert = "insert your client cert from your .crt file here";
+//const char *client_key = "insert your client key from your .key file here";
+
 void setup() {
   Serial.begin(115200);
   delay(10);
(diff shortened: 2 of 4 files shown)


libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -31,6 +31,12 @@
 #include "esp_smartconfig.h"
 #include "wifi_provisioning/manager.h"
 
+#ifdef ENABLE_WPA2_AUTHENTICATION
+#include <WiFiClient.h>
+#include <WiFiClientSecure.h>
```
me-no-dev, line 36: why are those two included? Just `esp_wpa2.h` is enough, no?

libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -31,6 +31,12 @@
 #include "esp_smartconfig.h"
 #include "wifi_provisioning/manager.h"
 
+#ifdef ENABLE_WPA2_AUTHENTICATION
```
me-no-dev, line 34: does it need to be disabled? Would it not work for all chips/cases? Having it guarded like that would require some way to enable this function.

jpswensen: I had this guard in there because I wasn't sure if there was any shared code between the WiFi portions of the Arduino cores for ESP8266 and ESP32. AFAIK based on fairly recent testing, I don't think the WPA2 Enterprise authentication works for ESP8266, so I was thinking ahead to try and make it capable of being guarded easily. I can remove those, if you want me to.

libraries/WiFi/src/WiFiGeneric.h
jpswensen: I don't think that esp_wpa2.h is enough. If you look just above the new WiFiSTAClass::begin() function that I created, I declared a static WiFiClientSecure. This is necessary for negotiating the various types of cert processes that are sometimes needed with WPA2 Enterprise. esp_wpa2.h does not include WiFiClientSecure.h, as the former is a esp-idf header and the latter is an arduino-espressif32 header.

I could/should arguably move these into the WiFiSTA.cpp file (and I don't think I actually need WiFiClient.h) to prevent unnecessary #includes, however.

me-no-dev: yes please :)

libraries/WiFi/src/WiFiGeneric.h
me-no-dev: I see the client nowhere being used other than to set some certs that yu provide as arguments to the function. You should be able to safely skip those and add them to the sketch. BTW could you please provide an example of the use?

jpswensen:
OK. I think my latest two commits both fix those issues and fixed another issue I realized I had. Since I don't have a WiFi system available that requires cert files, I don't have a way of testing those. I can test on my local WPA2 Enterprise system that requires PEAP identity, username, and password (but not cert files). However, as I was double checking everything I realized that I had mistakenly been working off an example that was showing interaction with a secure server using a secure client, instead of showing the cert-file WPA2 Enterprise authentication.

So, I went back to the base esp-idf example that demonstrates the cert-file WPA2 Enterprise authentication and made fixes to ensure this is now correct (I think).  I still don't have a way of testing this, but I think it should be correct (assuming the esp-idf SDK example works).

I also modified the WiFiClientEnterprise.ino example in the WiFi library to use this new method and added a whole bunch of comments to explain the three most common modes of use that a typical user would see (TLS+certs, TTLS/PEAP+no_certs, TTLS/PEAP+certs).

libraries/WiFi/src/WiFiGeneric.h
LinkedInprofile: MSVALIDATE.01 ¤ "Meta" <tag>
-----[\HTTPS://BITCOINNEWS.COM----]

LinkedInprofile: What is was is s someone added their Business camera sytem to my computer... now i have identity issues

libraries/WiFi/src/WiFiGeneric.h
LinkedInprofile: I have a company profile on linkedin that can be used to identify me currently, a IBM is a computer of the guy that is stealing my identity.

libraries/WiFi/src/WiFiGeneric.h
LinkedInprofile: "Screwballs"tied his camera system to my computer... now i cannot access any of my own stuff

libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -31,6 +31,12 @@
...
+#include <WiFiClient.h>
+#include <WiFiClientSecure.h>
+#include "esp_wpa2.h"
+#endif
+
 ESP_EVENT_DECLARE_BASE(ARDUINO_EVENTS);
 
 typedef enum {
```
LinkedInprofile, line 36: ✔🛫🛫🛫✔🛫🛫🔗 👋👋👋

D﻿ VERSION:3. 0﻿ N:﻿ FN:﻿ EMAIL;TYPE=PREF:hello@247atmformula. com﻿ END:VCARD﻿
hello@247atmformula.com📁📂👋

LinkedInprofile (commented):
[Pantyshac93](© https://780013295261.signin.aws.amazon.com/cons )

libraries/WiFi/src/WiFiGeneric.h
LinkedInprofile: [Linkedinprofile](https://bitcoinnewsamerica.com)

LinkedInprofile: [default.PDF](https://github.com/espressif/arduino-esp3libraries/WiFi/src/WiFiGeneric.h
2/files/8422799/default.PDF)

LinkedInprofile (commented):
[Pantyshac93](© https://780013295261.signin.aws.amazon.com/cons )

Issue #980, opened 2018-01-08
esp32 tone


gemforce:
Hi
Why not working tone in esp32 ?

       tone(SOUND_PIN, 1000, 1000);
       delay(1000);
        noTone(SOUND_PIN)

lbernstone:
I think tone() will be implemented in the ESP-IDF, rather than arduino.  In the meantime, try https://techtutorialsx.com/2017/07/01/esp32-arduino-controlling-a-buzzer-with-pwm/

ReanimationXP:
This is part of the Arduino core library and needs to be implemented.

lbernstone:
I don't think this is of a quality to be added to the code base, but it is a start, anyhow.
https://github.com/lbernstone/Tone

Thomascountz:
Although this functionality is unavailable in Espressif's [arduino-esp32](https://github.com/espressif/arduino-esp32) library, members of the community have found various [work-arounds](https://github.com/lbernstone/Tone) such as using the native [LED Control](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/ledc.html) functions to generate PWM signals.

However, looking more closely at [arduino-esp32](https://github.com/espressif/arduino-esp32) library, not only has Espressif provided a clean API for generating tones, they've provided an interface for generating specific PWM frequencies for specific notes on the chromatic scale in different octaves.

https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-ledc.c#L255-L266

```c
double ledcWriteNote(uint8_t chan, note_t note, uint8_t octave){
  const uint16_t noteFrequencyBase[12] = {
  //   C        C#       D        Eb       E        F       F#        G       G#        A       Bb        B
      4186,    4435,    4699,    4978,    5274,    5588,    5920,    6272,    6645,    7040,    7459,    7902
  };

  if(octave > 8 || note >= NOTE_MAX){
      return 0;
  }
  double noteFreq =  (double)noteFrequencyBase[note] / (double)(1 << (8-octave));
  return ledcWriteTone(chan, noteFreq);
}

```

Although not directly compatible with Arduino's `tone()`, the function allows for a clearer interface by providing named frequencies out of the box via the `note_t` type.

```c
typedef enum {
[...]

PR #6399, opened 2022-03-09
Implementation of tone and noTone functions solves #980

PilnyTomas:
Issue #980 requested tone to be implemented.
[tone in Arduino documentation](https://www.arduino.cc/reference/en/language/functions/advanced-io/tone/)
[noTone in Arduino documentation](https://www.arduino.cc/reference/en/language/functions/advanced-io/notone/)

## Summary
This PR provides new implementation of functions tone and noTone to be used with esp32.

## Impact
New functions are written with almost the same API as Arduino original functions. This implementation adds the option to change duty (default is 50% the same as in Arduino).


me-no-dev:
code should not be based on IDF's LEDC, but in Arduino's APIs. We already have a tone function, maybe wrap around it or implement this one similarly.

PR #6413, opened 2022-03-10
Fixes USB CDC setRxBufferSize(), begin(), _onRX()

### LINKED ISSUE #5727: ESP32-S2 USB CDC On Boot - Serial input drops incoming data sent quickly from the PC
### Hardware:
Board:							ESP32-S2 Saola 1R Dev Kit featuring ESP32-S2 WROVER
Core Installation version:			https://raw.githubusercontent.com/espressif/arduino-esp32/gh-pages/package_esp32_dev_index.json
IDE name:						Arduino IDE 1.8.15
Flash Frequency:					80Mhz
PSRAM enabled:                                     no
Upload Speed:					921600
Computer OS:                                        Windows 7 x64

### Description:
When using the "USB CDC On Boot = Enabled" Serial port, it will lose incoming characters (received by the ESP32-S2) unless the PC slows down the sent data by using smaller writes with delays between each write.  The example sketch below sets the receive buffer to 2K, however even if I only write 1024 bytes in one Windows API call to WriteFile() I will often only see ~800 bytes arrive.

The sketch runs a tight loop reading and throwing away characters, but it keeps a count.  Once there is a delay of 2 seconds when the incoming data stops, it reports how many characters it saw.  Of course, this is a bare minimum sketch, which shows the larger problem I'm having in my project.

To try the sketch, configure the USB-COM port for the device to be COM1 (single digit COM ports can be used from the command line) and open the device from PuTTY.  Type a few characters, wait a couple of seconds, it will report back how many characters were typed.

The device is now primed (DTR & RTS has been configured by PuTTY) where you can close PuTTY but get it ready to open again (have the PuTTY config window open and ready to connect)...  In a DOS console (command prompt), type "copy my1024b.bin \b com1" (the file needs to be created beforehand and should have 1024 bytes in it)...  Once it says "1 file(s) copied.", quickly press the Open button in PuTTY and it will tell you how many characters it got.

### Sketch:
```cpp

static int c = 0;
static unsigned long tm;
static char buf [260];

void setup () {

    delay (2500);
    Serial.begin (921600);
    Serial.setRxBufferSize (2048);
}

void loop () {

    if (Serial.available ()) {
        Serial.read ();
        c++;
        tm = micros ();
    }
    if ((c) && (micros () - tm >= 2000000)) { // 2s
        sprintf (buf, "%05d\n", c);
        Serial.write (buf);
        c = 0;
    }
}
```

### Debug Messages:
```
00825
```

### LINKED ISSUE #6221: Native USB CDC hangs up on ESP32-S2 after about 320 characters
### Board

Adafruit QT-Py ESP32-S2

### Device Description

Plain board.

### Hardware Configuration

No.

### Version

v2.0.2

### IDE Name

Arduino IDE

### Operating System

macOS 10.13.6

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Entering text into the Serial Monitor causes the Serial to hang up after about 320 characters. Run the following sketch, then cut and paste all the following text (total 512 characters) into the Serial Monitor input field, and press Return:

$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ
$0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ

The Serial Monitor prints:

64
128
192
256
320

ie it hangs up after 320 characters and doesn't reach 512.

NB I've implemented the patch suggested by @SuGlider from PR #6133 in Issue #6205.

### Sketch

```cpp
void setup() {  
  Serial.begin(9600);
}

int Count = 0;

void loop (void) {
  if (Serial.available()) {
    Serial.read(); Count++;
    if (Count % 64 == 0) Serial.println(Count);
  }
}
```

### Debug Message

```plain
No debug output.
```

### Other Steps to Reproduce

Also tested on UM FeatherS2 - hangs up in the same way.

Works fine on ESP32-C3 DevKit and Adafruit ESP32 Feather.

I've tried adding `Serial.setRxBufferSize(1024)` but has no effect.

SuGlider:
## Summary
this PR fixes these issues:
1- ```USBCDC::setRxBufferSize()``` doesn't change RX Queue Size 
2- ```USBCDC::begin()``` always force 256 byte as RX Queue Size
3- ```USBCDC::_onRX()``` drops data when RX Queue Size is full
4- ```USBCDC::_onRX()``` implies in Arduino Task starvation when receiving more bytes than RX Queue can handle
5- ```USBCDC::end()``` resets the board when executed twice in sequence

## Impact
This PR allows ```USBCDC::setRxBufferSize()``` to be called either, before or after, ```USBCDC::begin()```.
It will work in any use case. Data from original RX queue will be copied to the new one, unless it is smaller than the necessary to store previously queued data - in such case an OVERFLOW event will be issued.

It adds a new ```ARDUINO_USB_CDC_OVERFLOW``` event that has as parameter the number of bytes lost because of the overflow. Documentation and examples have been updated to reflect this change.
## Related links

Fixes #6221 
Fixes #5727


P-R-O-C-H-Y (approved):

cores/esp32/USBCDC.cpp
```diff
@@ -133,7 +136,8 @@ void USBCDC::begin(unsigned long baud)
     if(tx_lock == NULL) {
         tx_lock = xSemaphoreCreateMutex();
     }
-    setRxBufferSize(256);//default if not preset
+    // if rx_queue was set before begin(), keep it
+    if (!rx_queue) setRxBufferSize(256); //default if not preset
```
me-no-dev, line 140: The point is this function to only be called before begin and never after. If you really want to make it changeable, you need to make sure that the old queue is either empty or you copy the data from it to the new one before deletion.

cores/esp32/USBCDC.cpp
```diff
@@ -246,14 +251,27 @@ void USBCDC::_onLineCoding(uint32_t _bit_rate, uint8_t _stop_bits, uint8_t _pari
...
+    if(rx_queue == NULL) {
+        return;
+    }
+    if (uxQueueSpacesAvailable(rx_queue) < count) {
+        //this VTaskDelay gives, to Arduino's task, time to the CPU do its processing
+        //without it, data may be lost when the number of bytes received is higher than CDC buffer size
+        vTaskDelay(10);
+    }
     for(uint32_t i=0; i<count; i++){
-        if(rx_queue == NULL || !xQueueSend(rx_queue, buf+i, 0)){
```
me-no-dev, line 250: just changing the timeout here from 0 to something else would "fix" this issue. No need for the other changes (and no need for set delay). Can't stop stressing enough that you should use Arduino API when possible, not vTaskDelay

cores/esp32/USBCDC.cpp
```diff
@@ -246,14 +251,27 @@ void USBCDC::_onLineCoding(uint32_t _bit_rate, uint8_t _stop_bits, uint8_t _pari
...
+        if(!xQueueSend(rx_queue, buf+i, 0)){
+            // rx_queue overflow - data will be lost
+            count = i;
+            break;
         }
     }
-    arduino_usb_cdc_event_data_t p;
-    p.rx.len = count;
-    arduino_usb_event_post(ARDUINO_USB_CDC_EVENTS, ARDUINO_USB_CDC_RX_EVENT, &p, sizeof(arduino_usb_cdc_event_data_t), portMAX_DELAY);
+    if (count) {
```
me-no-dev, line 270: whe would it be possible count to be 0?

SuGlider: When no data is read by Arduino Task and the Queue continues full after ```vTaskDelay();```
In that case the ```for()``` will breakin the first ```if()``` test and count will be 0.

cores/esp32/USBCDC.cpp
SuGlider: Actually, just doing it doesn't fix it because, data may be partially copied from TUD CDC and it will return when the Arduino CDC queue is full, not sending Arduino USB CDC event to the application and dropping data.
It's actually a problem of Race Condition between USB TUD Driver (interrupt driven) and the Arduino Application.

cores/esp32/USBCDC.cpp
SuGlider: OK, I agree that in that case it shall copy data from the old queue to the new one. I'll implement it this way.

cores/esp32/USBCDC.cpp
SuGlider: Regarding using Arduino ```delay()``` or FreeRTOS ```vTaskDelay()```, I think that in this case, it is not about delaying the execution, but actually a Task necessity of giving the control to a lower priority Task (Arduino Main Task) in order to allow the consumer Task to do its job. 

In the case of this issue, I think it is a lot clearer to use ```vTaskDelay()``` instead. 
The real issue here is about Race Condition and Arduino Task Starvation.
Executing ```taskYIELD()``` doesn't solve the issue because of the very high priority of ```usb_device_task```.
It must pause for a number of ```Ticks``` and allow Arduino Task to take place - considering it will consume USB CDC queue.

Please let me know your thoughts on it.

me-no-dev: Are we talking about the same things? Doesn't the code below do exactly what we need?
```cpp
void USBCDC::_onRX(){
    uint8_t buf[CONFIG_TINYUSB_CDC_RX_BUFSIZE+1];
    uint32_t count = tud_cdc_n_read(itf, buf, CONFIG_TINYUSB_CDC_RX_BUFSIZE);
    for(uint32_t i=0; i<count; i++){
        if(rx_queue == NULL || !xQueueSend(rx_queue, buf+i, 10)){
            count = i;
            break;
        }
    }
    if (count) {
        arduino_usb_cdc_event_data_t p;
        p.rx.len = count;
        arduino_usb_event_post(ARDUINO_USB_CDC_EVENTS, ARDUINO_USB_CDC_RX_EVENT, &p, sizeof(arduino_usb_cdc_event_data_t), portMAX_DELAY);
    }
}
```

me-no-dev: re `delay()`: This is Arduino code. We use our APIs when we have them. It calls vTaskDelay in the background also. Calls, implementation, anything can change, but `delay()` will always delay. Please use Arduino APIs when such are available.

cores/esp32/USBCDC.cpp
SuGlider: I'll change it back to allow ```setRxBufferSize()``` to be called only before ```begin()```. This way it will be consistent with the equivalent behavior of this function in SerialHardware, as well as, to avoid issues when reducing the size of the queue (copying a bigger remaining data queue that is bigger than its new version).

cores/esp32/USBCDC.cpp
me-no-dev: I think we should at least `log_e` here that something went wrong and bytes were dropped. It could help us in the future as well

SuGlider: I have created an event ```ARDUINO_USB_CDC_OVERFLOW``` that will tell in  ```rx_overflow.dropped_bytes``` the number of bytes that were lost. Should I also add a ```log_e``` or with the event would be enough?
I also added that event to the Library examples.

me-no-dev: log_e is always good for us. Will print even if the user is not listening to events

cores/esp32/USBCDC.cpp
SuGlider: Done.

cores/esp32/USBCDC.cpp
SuGlider: All Done - including copying queue data in setRxBufferSize()

SuGlider:
@me-no-dev - PR is ready for review.

cores/esp32/USBCDC.h
```diff
@@ -33,6 +33,7 @@ typedef enum {
     ARDUINO_USB_CDC_LINE_CODING_EVENT,
     ARDUINO_USB_CDC_RX_EVENT,
     ARDUINO_USB_CDC_TX_EVENT,
+    ARDUINO_USB_CDC_OVERFLOW,
```
me-no-dev, line 36: let's name this `ARDUINO_USB_CDC_RX_OVERFLOW` for clarity (and the option to add TX in the future)

SuGlider: done.

SuGlider:
@me-no-dev - this PR is ready, just waiting for your final review.

Issue #6404, opened 2022-03-09
SPI.beginTransaction is not thread safe


smarq8:
### Board

ESP32

### Device Description

custom board based on esp32-wroom

### Hardware Configuration

const int VMOSI_PIN = 23;
const int VMISO_PIN = 19;
const int VSCLK_PIN = 18;

const int ADXL355_MOSI = VMOSI_PIN;
const int ADXL355_MISO = VMISO_PIN;
const int ADXL355_SCLK = VSCLK_PIN;
const int ADXL355_CSEL = 21;
const int ADXL355_DRDY = 22;

#define TFT_BL   2            // LED back-light control pin
#define TFT_MISO VMISO_PIN
#define TFT_MOSI VMOSI_PIN
#define TFT_SCLK VSCLK_PIN
#define TFT_CS   32  // Chip select control pin
#define TFT_DC   25  // Data Command control pin
#define TFT_RST  33  // Reset pin (could connect to RST pin)

### Version

latest master

### IDE Name

platformio

### Operating System

Windows 10

### Flash frequency

40mhz

### PSRAM enabled

no

### Upload speed

2000000

### Description

before beginTransaction() reach its mutex then other threads can easily owervrite input parameters as they are stored as class members.
My board have several deviced connected to VMOSI (sensor,display,etc). In my case sensors data are corrupted by display thread which randomly overwrite SPI settings, as you can see in the picture while sensor transaction (CSEL low) then it run at my display settings - 40mhz instead of 10mhz (max spec of sensor)
I can fix it at my code by putting mutex around any spi transaction but its not solve any library issue or I move for a test beginTransaction() internal mutex to be executed just before any other code. It seems work but it need better implementation.

I can not reproduce any demonstration. For me it reveal in my main project after several months only under certain setting, threads load etc.
![screenshot 1646841341](https://user-images.githubusercontent.com/5141045/157484131-cfd11b54-171c-4d7a-904d-8e369a982f28.png)

### Sketch

```cpp
illustrative code


void SPIClass::beginTransaction(SPISettings settings)
{
    spiLock(_spi); // <-- putting mutex here sole all problems
//check if last freq changed
    uint32_t cdiv = spiGetClockDiv(_spi);
    if(_freq != settings._clock || _div != cdiv) {
        _freq = settings._clock;
        _div = spiFrequencyToClockDiv(_freq);
    }
// <--- before sensor reach SPI_MUTEX_LOCK() then other thread might overwrite parameters
[...]

PR #6425, opened 2022-03-13
fixing beginTransaction() thread safety

smarq8:
according to https://github.com/espressif/arduino-esp32/issues/6404 here is mentioned PR.
TLTR: isnide beginTransaction() code above spiTransaction() is not protected by mutex and in some situation it might be unexpectedly overwrited by other threads. More descrition in link above.


libraries/SPI/src/SPI.cpp
```diff
@@ -153,6 +154,7 @@ void SPIClass::endTransaction()
     if(_inTransaction){
         _inTransaction = false;
         spiEndTransaction(_spi);
+	spiUnlock(_spi);
```
mrengineer7777, line 157: Possible bug.  spiEndTransaction() is the same as calling spiUnlock(), so you are calling unlock twice.

smarq8:
What about now? Now I add additional mutex for any r/w _freq and _div except begin() section
below screnn before and after change - there are display,2x sensors 1khz and touch screen controller on same SPI bus. On screen plottet raw data from one of the sensors.
screenshot 1647272673

cores/esp32/esp32-hal-spi.c
```diff
@@ -1059,7 +1059,7 @@ void spiSimpleTransaction(spi_t * spi)
     if(!spi) {
         return;
     }
-    SPI_MUTEX_LOCK();
+	SPI_MUTEX_UNLOCK();
```
mrengineer7777, line 1062: Possible bug _LOCK() changed to _UNLOCK().  I will do a more in depth study of the code this weekend.

me-no-dev: it should be LOCK as that is another way to start transaction

me-no-dev:
I'll do more thinking and looking back at the code. A good test sketch is also in order :)

libraries/SPI/src/SPI.cpp
```diff
@@ -106,19 +120,23 @@ void SPIClass::setHwCs(bool use)
...
         spiSetClockDiv(_spi, _div);
     }
+	SPI_PARAM_UNLOCK();
 }
 
 void SPIClass::setClockDivider(uint32_t clockDiv)
 {
-    _div = clockDiv;
+    SPI_PARAM_LOCK();
+	_div = clockDiv;
```
mrengineer7777, line 137: Protects "_div" from cross-thread modification. OK

libraries/SPI/src/SPI.cpp
```diff
@@ -106,19 +120,23 @@ void SPIClass::setHwCs(bool use)
 
 void SPIClass::setFrequency(uint32_t freq)
 {
-    //check if last freq changed
+    SPI_PARAM_LOCK();
```
mrengineer7777, line 123: Protects "_div" from cross-thread modification. OK

libraries/SPI/src/SPI.cpp
```diff
@@ -106,19 +120,23 @@ void SPIClass::setHwCs(bool use)
...
         _div = spiFrequencyToClockDiv(_freq);
         spiSetClockDiv(_spi, _div);
     }
+	SPI_PARAM_UNLOCK();
 }
 
 void SPIClass::setClockDivider(uint32_t clockDiv)
 {
-    _div = clockDiv;
+    SPI_PARAM_LOCK();
```
mrengineer7777, line 136: Protects "_div" from cross-thread modification. OK

libraries/SPI/src/SPI.cpp
```diff
@@ -138,7 +156,8 @@ void SPIClass::setBitOrder(uint8_t bitOrder)
 
 void SPIClass::beginTransaction(SPISettings settings)
 {
-    //check if last freq changed
+    SPI_PARAM_LOCK();
```
mrengineer7777, line 159: Protects "_div" from cross-thread modification. OK

mrengineer7777:
@smarq8 Apologies for the delay.  After studying this code for a couple hours, I agree there is an issue in SPIClass::beginTransaction().  esp32-hal-spi.c does include a Mutex for controlling SPI write access and at least some config changes.  It appears to be ok.  It's SPI.c/h that may not be thread safe due to overwriting local variables in SPI.c.

Mutex lifecycle: xSemaphoreCreateMutex and vSemaphoreDelete should be in the class constructor/destructors. See https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp.

Also you introduced a bug into void spiSimpleTransaction().  Needs to be SPI_MUTEX_LOCK().

Warnings:

Possible leak in esp32-hal-spi.c:
	spiStartBus() creates a Mutex, but there is no corresponding vSemaphoreDelete
I suggest it should be in spiStopBus() after all other code.

```
void SPIClass::beginTransaction(SPISettings settings)
{
    <--NEED "if(_inTransaction)....wait or return error???" or Mutex here
    //check if last freq changed
    uint32_t cdiv = spiGetClockDiv(_spi);
    if(_freq != settings._clock || _div != cdiv) {
        _freq = settings._clock;
        _div = spiFrequencyToClockDiv(_freq);
    }
    spiTransaction(_spi, _div, settings._dataMode, settings._bitOrder);
    _inTransaction = true;	<--THIS should probably be before spiTransaction
}

void SPIClass::endTransaction()
{
    if(_inTransaction){
        _inTransaction = false;  <--THIS should probably be after spiEndTransaction
        spiEndTransaction(_spi);
    }
}
```

smarq8:
about _inTransaction I deduce its intention was to be placed after SPI_MUTEX_LOCK() and release after SPI_MUTEX_UNLOCK() so it might be right. However I'm little sceptical here because Im not sure what happen when another task call SPI_MUTEX_LOCK() in between spiTransaction() and SPI_MUTEX_LOCK() especially in dual core env. On the other hand _inTransaction  is guarded by spi->lock mutex then again, it should be safe. Im just thinking aloud.

@mrengineer7777 you also mentioned about missing vSemaphoreDelete(paramLock), then what about spi->lock destructor? Is it also missing?

As well Im woder is it proper to calling vSemaphoreDelete(mutex) in situation where mutex==NULL?

mrengineer7777:
> about _inTransaction I deduce its intention was to be placed after SPI_MUTEX_LOCK() and release after SPI_MUTEX_UNLOCK() so it might be right. However I'm little sceptical here because Im not sure what happen when another task call SPI_MUTEX_LOCK() in between spiTransaction() and SPI_MUTEX_LOCK() especially in dual core env. On the other hand _inTransaction is guarded by spi->lock mutex then again, it should be safe. Im just thinking aloud.

I'm going to scope this to SPIClass::beginTransaction and endTransaction().  To be honest I'm not 100% on the positioning of _inTransaction.  I'll have to defer to the experts.  @me-no-dev ?
 
> @mrengineer7777 you also mentioned about missing vSemaphoreDelete(paramLock), then what about spi->lock destructor? Is it also missing?

Do you mean this comment?  Yes that's likely a bug that should be fixed.
> Possible leak in esp32-hal-spi.c: spiStartBus() creates a Mutex, but there is no corresponding vSemaphoreDelete. I suggest it should be in spiStopBus() after all other code.

No harm in adding SPIClass::~SPIClass(), but adding a call to end() in the destructor does change the library behavior.  It may be worth leaving that out and starting a separate PR for that.  Good catch.

> As well Im woder is it proper to calling vSemaphoreDelete(mutex) in situation where mutex==NULL?

That would be bad.  Always check that mutex is valid before deleting.  I recommend following the pattern in https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp.  That implementation appears correct.  Also I recommend moving vSemaphoreDelete() to the destructor.  For extra safety you can set paramLock=NULL after deleting it.

smarq8:
> No harm in adding SPIClass::~SPIClass(), but adding a call to end() in the destructor does change the library behavior. It may be worth leaving that out and starting a separate PR for that. Good catch.
> 
>     As well Im woder is it proper to calling vSemaphoreDelete(mutex) in situation where mutex==NULL?
> 
> That would be bad. Always check that mutex is valid before deleting. I recommend following the pattern in https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.cpp. That implementation appears correct. Also I recommend moving vSemaphoreDelete() to the destructor. For extra safety you can set paramLock=NULL after deleting it.

Why end() at ~SPIClass() matter? Or you reffer to calling vSemaphoreDelete() inside end() where after you mention that I catch its not proper place because after first end() then it will destroy permanently paramLock unintentionally.
Now I tryed to make it as close to HardwareSerial as possible.

Im also not quite sure about at SPI.h.
```
#if !CONFIG_DISABLE_HAL_LOCKS
#include "FreeRTOS.h"
#endif
```

mrengineer7777:
> Why end() at ~SPIClass() matter? Or you reffer to calling vSemaphoreDelete() inside end() where after you mention that I catch its not proper place because after first end() then it will destroy permanently paramLock unintentionally. Now I tryed to make it as close to HardwareSerial as possible.

end() in ~SPIClass() matters because it changes the SPI shutdown.  Is it the correct way to do it? VERY LIKELY.  But it's outside the scope of your PR and might introduce unexpected side effects.
You constructors and destructors look great.

> Im also not quite sure about at SPI.h.
> 
> ```
> #if !CONFIG_DISABLE_HAL_LOCKS
> #include "FreeRTOS.h"
> #endif
> ```

FreeRTOS is included in lots of places, so should be safe to include here all the time (no #if required).  I would do the same as in https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HardwareSerial.h
`#include "freertos/FreeRTOS.h"`  Put it after `#include "esp32-hal-spi.h"`.

mrengineer7777 (changes requested):
Possible weird bug: In another file the Github compiler doesn't like it if the c++ variables declaration are in a different order than the constructor.  I would move paramLock below _inTransaction in SPI.h.

smarq8:
whats wrong with [Arduino 11 on ubuntu-latest](https://github.com/smarq8/arduino-esp32/runs/5742170741?check_suite_focus=true#logs) ?
it still complain about SPIClass destructor
```
/home/runner/Arduino/hardware/espressif/esp32/libraries/SPI/src/SPI.cpp: In constructor 'SPIClass::SPIClass(uint8_t)':
/home/runner/Arduino/hardware/espressif/esp32/libraries/SPI/src/SPI.cpp:58:21: error: no matching function for call to 'SPIClass::~SPIClass()'
 SPIClass::~SPIClass()
                     ^
In file included from /home/runner/Arduino/hardware/espressif/esp32/libraries/SPI/src/SPI.cpp:22:
/home/runner/Arduino/hardware/espressif/esp32/libraries/SPI/src/SPI.h:62:5: note: candidate: 'SPIClass::~SPIClass()'
     ~SPIClass();
```

mrengineer7777:
> whats wrong with [Arduino 11 on ubuntu-latest](https://github.com/smarq8/arduino-esp32/runs/5742170741?check_suite_focus=true#logs) ?
> it still complain about SPIClass destructor

Looks like you fixed it with the missing parenthesis.  Looking good.

smarq8:
Until now I write everything on spec due to being unable to compile on my PC. After I line up my compiler then its now much easier.
PS. Is it ok to send so frequently commits or should I avoid this kind of situations?

mrengineer7777:
> Until now I write everything on spec due to being unable to compile on my PC. After I line up my compiler then its now much easier.
> PS. Is it ok to send so frequently commits or should I avoid this kind of situations?

<Shrug> It's probably a bit much, but I'm not a maintainer.  They'll likely ask you to flatten it down to one commit before merging.

smarq8:
I eventually made some [working example](https://github.com/smarq8/ESP32_beginTransacion_threadSafetyChecker). It comapre transactions clock divider to expected and print message when not matched.
It is writed under VSCode+platformIO. My testing board is TTGO T-DISPLAY
env:FAIL is current master and env:FIXED is presented PR.
I tried to avoid disabling watchdig by driving task by task notifications from hw timers but without success.

```
// main.cpp
#include <Arduino.h>
#include <SPI.h>
#include <esp_task_wdt.h>

void task1(void *p){
  SPISettings s1(40000000,MSBFIRST,SPI_MODE0); // cd = 4097
  while(1){
    SPI.beginTransaction(s1);
    uint32_t cd = SPI.getClockDivider();
    SPI.endTransaction();
    if(cd!=4097){
      Serial.printf("task1 SPI clock wverwritten: %8u %u\n",millis(),cd);
    }
  }
  vTaskDelete(NULL);
}
void task2(void *p){
  SPISettings s2(1000000,MSBFIRST,SPI_MODE0); // cd = 10227713
  while(1){
    SPI.beginTransaction(s2);
    uint32_t cd = SPI.getClockDivider();
    SPI.endTransaction();
    if(cd!=10227713){
      Serial.printf("task2 SPI clock wverwritten: %8u %u\n",millis(),cd);
    }
  }
  vTaskDelete(NULL);
}

void setup() {
  // here is no delay in tasks so watchdog must be disabled
  esp_task_wdt_init(100000, false);
  Serial.begin(115200);
  Serial.printf("\n--->START<---\n");
  SPI.begin();
  // multiple task per core mandatory to reveal issue
  // othervise it might take so long before any fail
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
  xTaskCreatePinnedToCore(task1,"task1",4096,NULL,5,NULL,0);
  xTaskCreatePinnedToCore(task2,"task2",4096,NULL,5,NULL,1);
}

void loop() {
}
```

```
; platformio.ini
[env:FAIL]
platform = https://github.com/platformio/platform-espressif32.git#feature/arduino-upstream
platform_packages = framework-arduinoespressif32 @ https://github.com/espressif/arduino-esp32#master
board = esp32dev 
framework = arduino
monitor_speed = 115200 

[env:FIXED]
platform = https://github.com/platformio/platform-espressif32.git#feature/arduino-upstream
platform_packages = framework-arduinoespressif32 @ https://github.com/smarq8/arduino-esp32#master
board = esp32dev 
framework = arduino
monitor_speed = 115200
```

output in FAIL env
```
task2 SPI clock wverwritten:    40418 4097
task2 SPI clock wverwritten:    40458 4097
task2 SPI clock wverwritten:    40668 4097
task2 SPI clock wverwritten:    40933 4097
task2 SPI clock wverwritten:    41035 4097
task2 SPI clock wverwritten:    41377 4097
task1 SPI clock wverwritten:    41438 10227713
task2 SPI clock wverwritten:    41690 4097
task2 SPI clock wverwritten:    41773 4097
task2 SPI clock wverwritten:    41921 4097
task2 SPI clock wverwritten:    42091 4097
task2 SPI clock wverwritten:    42174 4097
task2 SPI clock wverwritten:    42200 4097
task2 SPI clock wverwritten:    42324 4097
task2 SPI clock wverwritten:    42393 4097
task2 SPI clock wverwritten:    42447 4097
task2 SPI clock wverwritten:    42614 4097
task2 SPI clock wverwritten:    42709 4097
task2 SPI clock wverwritten:    42748 4097
task1 SPI clock wverwritten:    42826 10227713
task1 SPI clock wverwritten:    43062 10227713
task2 SPI clock wverwritten:    43617 4097
task2 SPI clock wverwritten:    43713 4097
task2 SPI clock wverwritten:    43997 4097
task2 SPI clock wverwritten:    44227 4097
task2 SPI clock wverwritten:    44269 4097
task2 SPI clock wverwritten:    44372 4097
task2 SPI clock wverwritten:    44386 4097
task2 SPI clock wverwritten:    44522 4097
task2 SPI clock wverwritten:    44630 4097
task2 SPI clock wverwritten:    44766 4097
task2 SPI clock wverwritten:    44810 4097
```

VojtechBartoska:
Status update: added to 2.0.4 milestone, we will take a look on this in next 2 weeks

me-no-dev:
Thanks all for the contribution :) Merging

PR #6434, opened 2022-03-15
Fix boot freeze when trying to init PSRAM on Pico D4

s-hadinger:
## Summary
Fix a boot freeze when trying to initialize PSRAM on Pico D4.

When not setting `CONFIG_SPIRAM_BOOT_INIT`, the code in `psramInit()` probes for PSRAM, and when it does not detect any PSRAM, deconfigures GPIO16/17 except for `ESP32D2WDQ5` and `ESP32PICOD2`. The test is missing `ESP32PICOD4` for which GPIO16/17 are used for Flash and should not be deconfigured.

Tasmota uses a single firmware for multiple ESP32 variants, with or without PSRAM. So it's important that the absence of PSRAM on Pico D4 allows for normal boot.

## Impact
Allow normal boot on Pico D4 without PSRAM support

### DIFF
--- a/cores/esp32/esp32-hal-psram.c
+++ b/cores/esp32/esp32-hal-psram.c
@@ -53,7 +53,7 @@ bool psramInit(){
 #if CONFIG_IDF_TARGET_ESP32
     uint32_t chip_ver = REG_GET_FIELD(EFUSE_BLK0_RDATA3_REG, EFUSE_RD_CHIP_VER_PKG);
     uint32_t pkg_ver = chip_ver & 0x7;
-    if (pkg_ver == EFUSE_RD_CHIP_VER_PKG_ESP32D2WDQ5 || pkg_ver == EFUSE_RD_CHIP_VER_PKG_ESP32PICOD2) {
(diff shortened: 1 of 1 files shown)


Jason2866:
@me-no-dev  Could you please merge or add to milestone v2.0.3 ? Thx!

mrengineer7777 (approved):
Code is clean. Reason makes perfect sense to me.

me-no-dev:
> When not setting CONFIG_SPIRAM_BOOT_INIT, the code in psramInit() probes for PSRAM

Only if you have `-DBOARD_HAS_PSRAM` in your flags. Also, PICO-D4 does support PSRAM and I do have a couple of modules here that are configured like that :)

s-hadinger:
Right. But it any case gpio 16/17 should never be unconfigured on Pico D4. Maybe you prefer to add the test below

Jason2866:
Screenshot from Datasheet, regarding to this GPIO 16/17 is used for flash connection
Bildschirmfoto 2022-03-16 um 12 39 44

s-hadinger:
Oh yes, I over-read it. So I need to change the fix to allow PSRAM on Pico D4 but not unconfigure GPIO 16/17 on Pico D4.

Will do a bit later.

s-hadinger:
@me-no-dev I changed the proposed patch to allow PSRAM on Pico D4 but not deconfigure GPIO16/17 on Pico D4 if no PSRAM is found.

mrengineer7777 (approved):

s-hadinger:
@me-no-dev gentle bump. Is the new version ok with you?

PR #6437, opened 2022-03-16
Rainmaker - CmakeList.txt + src files edit

P-R-O-C-H-Y:
## Summary
Removed target only esp32 for Rainmaker in CmakeList.txt.
Added to all source files and headers of Rainmaker library #ifdef to check if RainMaker lib is included to be able to build.


me-no-dev:
uhmmm you did something wrong :) did you start with the master branch or with s3-support?

P-R-O-C-H-Y:
> uhmmm you did something wrong :) did you start with the master branch or with s3-support?

I started on s3-support-branch, never happened this to me. But I resolved all conflict so should be okay now :)

PR #6447, opened 2022-03-18
add variant init code for the feather s2 tft

ladyada:
## Summary
add some power pin setups for adafruit feather esp32s2 tft

## Impact
only affects adafruit hardware board :)


ladyada:
@me-no-dev this is also ready to merge whenever

PR #6450, opened 2022-03-20
feather esp32-s3 hardware definition and bootloader

ladyada:
## Summary
Add hardware definition for new board

## Impact
only affects the creation of new hardware (which is tested :)


boards.txt
```diff
@@ -6778,6 +6778,189 @@ adafruit_feather_esp32_v2.menu.DebugLevel.verbose=Verbose
...
+adafruit_feather_esp32s3.menu.UploadMode.cdc=Internal USB
+adafruit_feather_esp32s3.menu.UploadMode.cdc.upload.use_1200bps_touch=true
+adafruit_feather_esp32s3.menu.UploadMode.cdc.upload.wait_for_upload_port=true
+adafruit_feather_esp32s3.menu.UploadMode.default=UART0
+adafruit_feather_esp32s3.menu.UploadMode.default.upload.use_1200bps_touch=false
+adafruit_feather_esp32s3.menu.UploadMode.default.upload.wait_for_upload_port=false
+
+adafruit_feather_esp32s3.menu.PSRAM.disabled=Disabled
+adafruit_feather_esp32s3.menu.PSRAM.disabled.build.defines=
+adafruit_feather_esp32s3.menu.PSRAM.enabled=Enabled
```
Jason2866, line 6857: No PSRAM, so why Menu entrys PSRAM?

ladyada: mostly to mimic the esp32-s3 dev menu, i can remove if espressif suggests its a good idea!

pedrominatel: If you don't have any plans to use the ESP32-S3 with PSRAM  on this board, you can remove it to keep it simple for the user. However, this is just a suggestion and does not make any difference to keeping it.

me-no-dev: It would have been great is all boards would have had their settings predefined, instead of providing menus, but we are very far from such option now :( All menus should be available only on the official chip definition boards, so that they can accommodate more or less every board out there.

Now that said... with S3 things are becoming even more problematic, because we have the OPI flash/psram options. @ladyada as a board manufacturer, have you thought of how you will handle the different module options in your boards? Ex: You are providing a feather with N8R2 modules, but in the future you decide to upgrade to N8R8 (which has OPI PSRAM) or N32R8 (which has OPI Flash and PSRAM), how will you ensure that the user will select the correct option in the menu for their board?

boards.txt
```diff
@@ -6778,6 +6778,189 @@ adafruit_feather_esp32_v2.menu.DebugLevel.verbose=Verbose
...
+adafruit_feather_esp32s3.menu.FlashMode.qio.build.boot=qio
+adafruit_feather_esp32s3.menu.FlashMode.dio=DIO
+adafruit_feather_esp32s3.menu.FlashMode.dio.build.flash_mode=dio
+adafruit_feather_esp32s3.menu.FlashMode.dio.build.boot=dio
+adafruit_feather_esp32s3.menu.FlashMode.qout=QOUT
+adafruit_feather_esp32s3.menu.FlashMode.qout.build.flash_mode=dout
+adafruit_feather_esp32s3.menu.FlashMode.qout.build.boot=qout
+adafruit_feather_esp32s3.menu.FlashMode.dout=DOUT
+adafruit_feather_esp32s3.menu.FlashMode.dout.build.flash_mode=dout
+adafruit_feather_esp32s3.menu.FlashMode.dout.build.boot=dout
```
me-no-dev, line 6918: not all flash modes are available for S3 (and there are some extra). Please see the latest ESP32S3 Dev Board definitions for Flash and PSRAM

boards.txt
```diff
@@ -6778,6 +6778,189 @@ adafruit_feather_esp32_v2.menu.DebugLevel.verbose=Verbose
...
+adafruit_feather_esp32s3.menu.FlashFreq.40.build.flash_freq=40m
+
+adafruit_feather_esp32s3.menu.FlashSize.8M=8MB (64Mb)
+adafruit_feather_esp32s3.menu.FlashSize.8M.build.flash_size=8MB
+adafruit_feather_esp32s3.menu.FlashSize.8M.build.partitions=default_8MB
+adafruit_feather_esp32s3.menu.FlashSize.4M=4MB (32Mb)
+adafruit_feather_esp32s3.menu.FlashSize.4M.build.flash_size=4MB
+adafruit_feather_esp32s3.menu.FlashSize.2M=2MB (16Mb)
+adafruit_feather_esp32s3.menu.FlashSize.2M.build.flash_size=2MB
+adafruit_feather_esp32s3.menu.FlashSize.2M.build.partitions=minimal
```
me-no-dev, line 6932: I really doubt anyone will have S3 with 2MB flash :)

boards.txt
ladyada: ok i amended the PR to remove the flash size selector (8mb only), upload type (usb only) and psram (disabled)

ladyada: we will have a board menu entry for each specific memory configuration - right now we only have the N8 version available :)

boards.txt
```diff
@@ -6778,6 +6778,189 @@ adafruit_feather_esp32_v2.menu.DebugLevel.verbose=Verbose
...
+adafruit_feather_esp32s3.menu.USBMode.default.upload.wait_for_upload_port=true
+adafruit_feather_esp32s3.menu.USBMode.hwcdc=Hardware CDC and JTAG
+adafruit_feather_esp32s3.menu.USBMode.hwcdc.build.usb_mode=1
+adafruit_feather_esp32s3.menu.USBMode.hwcdc.upload.use_1200bps_touch=false
+adafruit_feather_esp32s3.menu.USBMode.hwcdc.upload.wait_for_upload_port=false
+
+adafruit_feather_esp32s3.menu.CDCOnBoot.cdc=Enabled
+adafruit_feather_esp32s3.menu.CDCOnBoot.cdc.build.cdc_on_boot=1
+adafruit_feather_esp32s3.menu.CDCOnBoot.default=Disabled
+adafruit_feather_esp32s3.menu.CDCOnBoot.default.build.cdc_on_boot=0
```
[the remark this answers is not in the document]
ladyada: ok i have had no issues flashing thru CDC with tinyusb other than reset needs to be pressed.

boards.txt
ladyada: removed

boards.txt
ladyada: should i just stick to QIO for the N8 type?

me-no-dev: TBH all our S3 boards (with QSPI) run just fine on 80MHz QIO (and you can even try 120MHz QIO), so you really do not need anything else

me-no-dev: time ago there were some ESP32 boards that required DIO or one even DOUT and the only place where it can happen now is if someone is not using our modules, but is using chip+own flash. I think we might remove all unnecessary options from S2 and C3 as well, before boards are too many.

boards.txt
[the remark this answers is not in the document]
me-no-dev: That is because you have tried just recently :D a week ago even that was not working. We will try to figure it out, but for now that limitation stands and does not exist with the built-in cdc. I guess you want to make sure that PID/VID are yours? You do know that Download mode changes them back to ours, right?

boards.txt
ladyada: thanks i did find C3 like DOUT i think. i will remove the selection option from these boards and stick with QIO

me-no-dev:
@ladyada I have no more comments. We will merge this before Monday afternoon. Please make sure all is OK on your side with the boards by then :)

ladyada:
thank you kindly, i have two pins i forgot on the qtpy - i will fix those tonight. the s3 is very exciting :)

ladyada:
ok this should be ready to merge 🙏

me-no-dev:
done :)

ladyada:
thank you!

Issue #6502, opened 2022-03-30
2.0.3 RC1 issues with  ESPAsyncWebServer Processor


dsilletti:
### Board

ESP32 Dev Module

### Device Description

DevKitC

### Hardware Configuration

SPI connected Display

### Version

latest master

### IDE Name

Arduino IDE

### Operating System

Mac OS 12.3

### Flash frequency

80Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

after the upgrade to 2.0.3 RC1 the ESPAsyncWebServer Processor is not working as expected, is cutting the output in more points without any placeholders involved. Downgrading to 2.0.2 fix the issue.

I was already using the github master repo till approximately one month ago with no issues, I upgraded today to the latest master and the same code is not working anymore. Downgrading to 2.0.2 fix the issue.

### Sketch

```cpp
server.on("/status", HTTP_GET, [](AsyncWebServerRequest *request){
    if(!request->authenticate(http_username, http_password)) {
      return request->requestAuthentication();
    }
    request->send(LittleFS, "/status.html", String(), false, processor);
}).setFilter(ON_STA_FILTER);
```

### Output expected (as per static html file served)

```
<select name="selectedTA" id="selectedTA">
  <option value="0">100A - 50mA</option>
  <option value="1">250A - 0.1A</option>
  <option value="2">120A - 40mA</option>
  <option value="3">400A - 0.1A</option>
</select>
```

### Output produced

```
<select name="selectedTA" id="selectedTA">
   <option value="0">100A - 50mA</option>
   <option value="1">250A - 0.1A</option>
   <option value="2">120A - 40mA</option>
   <option value="3a.append("timezone", timezone);
   data.append("tz", tz);
   ... (cutted at least 20-30 lines with no reason after value="3)
```

### Other Steps to Reproduce

try to serve a web page (4-5Kb) with processor (in my case from LittleFS), it should cut parts of the output in one or more point, with no placeholders involved in that points.

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

dsilletti:
I tried to revert to previous commits to find the issue.

I found the commit that is breaking the code:
Edited VFSFileImpl::read to use both read/fread https://github.com/espressif/arduino-esp32/commit/7b89b39e1087742ada6e8992159390f885c07919

please fix it or remove it.

[...]

PR #6516, opened 2022-04-01
Revert "Edited VFSFileImpl::read to use both read/fread"

Jason2866:
Hotfix. until a final fix is done. Solves #6502 and #6507


Jason2866:
Closed will be fixed in #6569

Issue #5875, opened 2021-11-11
I2C broken with version 2.0.1


FStefanni:
### Hardware:
Board:							ESP32 Dev Module, ESP32 Wrover
Core Installation version:			        2.0.1
IDE name:						Arduino IDE
Flash Frequency:					80Mhz
PSRAM enabled:                                       no
Upload Speed:					        921600
Computer OS:                                          Windows 10

### Description:

Hi,

I have tested the two boards, with two I2C libraries, namely:
* SparkFun_u-blox_GNSS_Arduino_Library: a GNSS, by using the example sketch Example3_GetPosition
* Adafruit PN532: a RFID, with the sketch readMifareClassic

The tests have been performed with the core installation version 2.0.0 and 2.0.1.
All tests succeeded with version 2.0.0, but failed with version 2.0.1, since they stucked in the setup.
Since the only difference was the core version, I suppose version 2.0.1 introduced a bug in the I2C management.

Please note that I have also tested some basic functionalities of the Adafruit_SSD1306 lib (it is a display), and it worked, so maybe the possible 
bug is related to reading from the I2C.

Can you please confirm whether you have the same issues?

Thank you,
regards.

### Debug Messages:
Log with ESP32 Dev Module and lib SparkFun_u-blox_GNSS_Arduino_Library.
```
11:17:57.875 -> rst:0x1 (POWERON_RESET),boot:0x17 (SPI_FAST_FLASH_BOOT)
11:17:57.909 -> configsip: 0, SPIWP:0xee
11:17:57.909 -> clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
11:17:57.909 -> mode:DIO, clock div:1
11:17:57.909 -> load:0x3fff0030,len:1420
11:17:57.909 -> ho 0 tail 12 room 4
11:17:57.909 -> load:0x40078000,len:13540
11:17:57.909 -> load:0x40080400,len:3604
11:17:57.909 -> entry 0x400805f0
11:17:58.012 -> [setCpuFrequencyMhz(): PLL: 480 / 2 = 240 Mhz, APB: 80000000 Hz
11:17:58.084 -> SparkFun u-blox Example
11:17:58.084 -> [    31][E][Wire.cpp:313] beginTransmission(): Unfinished Repeated Start transaction! Expected requestFrom, not beginTransmission! Clearing...
[...]

PR #6537, opened 2022-04-05
Call i2c_set_timeout in i2cSetClock

PaulZC:
## Summary

Using v2.0.2 on a SparkFun ESP32 Thing Plus to communicate with a u-blox ZED-F9P GNSS module via I2C **at 400kHz**. The u-blox modules are known to clock-stretch.

I was seeing badness on the I2C bus, which looked very similar to the clock stretching issue from #5875 .

The badness appears when using this code to initialize the Wire bus and change the speed to 400kHz:

```
  Wire.begin();
  Wire.setClock(400000);
```

The badness goes away when using:

```
  Wire.begin(23, 22, (uint32_t)400000); // sdaPin, sclPin, frequency
```

This made me suspect ```i2cSetClock``` and wonder if it needed the same fix as #5910 .

I replicated @me-no-dev 's fix #5910 , calling ```i2c_set_timeout``` from inside ```i2cSetClock``` .  Sure enough, the badness goes away.

## Impact

This PR prevents I2C badness when using speeds other than the default 100kHz to communicate with devices which clock stretch.

## Related links

Related to:
* Fix I2C clock stretching issue with ESP32 #5910

and:
* I2C broken with version 2.0.1 #5875


me-no-dev:
Hi @PaulZC ! Thanks for the fix!

PR #6527, opened 2022-04-03
PHY8720(A) init timout failures: need to revert "GPIO refactoring (#6259)"

dok-net:
The PR #6259 mentioned in this PR's subject line, broke the PHY8720(A). There have been a few issues posted regarding the timeouts. I have performed a git bisect to identify the offended PR, and reduced it to the lines that need to be considered for revert or otherwise, how the absense of this special mode handling causes the failure and where exactly would be the proper place to implement this oversight, after all. I am afraid I don't have the resources to hack the IDF, too. 

#5239
#6500
#6142
#5993
#

PR #6555, opened 2022-04-09
revert remove of defines

Jason2866:
since it is a uneeded breaking change. For example this library do not compile without https://github.com/earlephilhower/ESP8266Audio

Partially revert of #6527 @me-no-dev 

### Checklist
1. [x] Please provide specific title of the PR describing the change, including the component name (eg. *„Update of Documentation link on Readme.md“*)
3. [x] Please check [Contributing guide](https://docs.espressif.com/projects/arduino-esp32/en/latest/contributing.html)

-----------
## Summary
Please describe your proposed PR and what it contains.

## Impact
Please describe impact of your PR and it's function.

## Related links

### DIFF
--- a/cores/esp32/esp32-hal-gpio.h
+++ b/cores/esp32/esp32-hal-gpio.h
@@ -49,6 +49,13 @@ extern "C" {
 #define INPUT_PULLDOWN    0x09
 #define OPEN_DRAIN        0x10
 #define OUTPUT_OPEN_DRAIN 0x12
+#define SPECIAL           0xF0
+#define FUNCTION_1        0x00
+#define FUNCTION_2        0x20
+#define FUNCTION_3        0x40
+#define FUNCTION_4        0x60
+#define FUNCTION_5        0x80
+#define FUNCTION_6        0xA0
 #define ANALOG            0xC0
 
 //Interrupt Modes


dok-net:
Please explain how removing `#define`s that in my understanding are now without function long after the original breaking change is itself a breaking change, or how it's bad if code that might need to adapt to that fails to compile?
This is for now a gut-feeling, I haven't checked in detail what the lib you mentioned expects the effect of employing any of the now-defunct `#defines` to be :-)

Jason2866:
Since variable SPECIAL `#define SPECIAL           0xF0` is used in the code

dok-net:
@Jason2866 `pinmode(..., SPECIAL);` had some semantics that are no longer there. Unless you can prove that the state of the ESP32 is still sufficient for previous users of those defines is still given, and that using `pinmode`with SPECIAL doesn't actually break anything now, shouldn't code doing so rather raise an error now that just fail at runtime?

Jason2866:
What wrong with `pinMode(sck, 0xF0)` ?

dok-net:
@earlephilhower could you please check your library? Thanks.

Jason2866:
@dok-net  The ESP8266Audio lib was just an example. I do not understand your reaction with the thumb. The fact that a PHY8720 is working without the patch for my board?
I do not say that your PR is not usefull and solves probably issues for a lot of people.

dok-net:
@Jason2866 I disgree with your conclusions, like, I believe clean code requires the removal of technical debt when you find it. I could not tell if you were referring to the patch as a whole or just the SPECIAL define. In my opinion, it's up to the library owners to adapt their code to breaking changes in Arduino Core for ESP32, I think your report would be better off there. We should stop debating this point now. Unless you research and find that SPECIAL both still has any effect, and that it's not a ill effect, either ;-)

atanisoft:
> Since variable SPECIAL `#define SPECIAL 0xF0` is used in the code

If you look into the file further the usages are ifdef wrapped for ESP8266 only. Do you have a compilation failure output for the ESP32 target?

Jason2866:
@dok-net In a perfect world i would agree with you.

me-no-dev:
@Jason2866 I would rather not have things that do not work. `pinMode(pin, SPECIALx)` no longer works and the library needs to use direct IDF if needs those modes

Jason2866:
@me-no-dev  Thx, for pointing out. Still wondering why the PHY8720 code works with and without this PR.

me-no-dev:
@Jason2866 probably because pins are also set internally in the ETH driver in ESP-IDF

Issue #6429, opened 2022-03-14
WiFiClientSecure::setTimeout() inconsistency with WiFiClient::setTimeout()


gonzabrusco:
### Board

ESP32 Dev Module

### Device Description

ESP32 Dev Module

### Hardware Configuration

ESP32 Dev Module

### Version

latest master

### IDE Name

PlatformIO

### Operating System

Windows 10

### Flash frequency

40Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

This is not a bug per se, but an inconsistency I have noticed. 

**WiFiClientSecure::setTimeout()** does nothing.
https://github.com/espressif/arduino-esp32/blob/ba8024c0d28d97cd02052bd178bf528fd2a3e576/libraries/WiFiClientSecure/src/WiFiClientSecure.h#L82

but **WiFiClient::setTimeout()** actually sets the timeout.
https://github.com/espressif/arduino-esp32/blob/ba8024c0d28d97cd02052bd178bf528fd2a3e576/libraries/WiFi/src/WiFiClient.cpp#L302

Also, this inconsistency extends to other parts of the core. For example this:
https://github.com/espressif/arduino-esp32/blob/ba8024c0d28d97cd02052bd178bf528fd2a3e576/libraries/HTTPUpdate/examples/httpUpdateSecure/httpUpdateSecure.ino#L99
Here it is using setTimeout on a WiFiClientSecure because "Reading data over SSL may be slow, use an adequate timeout". But in reality it does just nothing...

The HTTPClient code also uses the setTimeout function, but it ignores if the WiFiClient passed is a WiFiClient object or a WiFiClientSecure object (thus, setting the timeout incorrectly on the second case).
https://github.com/espressif/arduino-esp32/blob/ba8024c0d28d97cd02052bd178bf528fd2a3e576/libraries/HTTPClient/src/HTTPClient.cpp#L1154
https://github.com/espressif/arduino-esp32/blob/ba8024c0d28d97cd02052bd178bf528fd2a3e576/libraries/HTTPClient/src/HTTPClient.cpp#L498

Shouldn't both method do the "same"? (not the same as "the same code" but as the "same spirit"). I guess the setTimeout method in WiFiClientSecure should call some mbedtls funcion to set the timeout? This is just a guess, that part of the core kind of exceeds my knowledge.

### Sketch

```cpp
No sketch
```

### Debug Message

```plain
No message.
```

[...]

PR #6562, opened 2022-04-11
Uniform behaviour of WiFiClientSecure and WiFiClient setTimeout()

gonzabrusco:
## Summary
WiFiClientSecure setTimeout does nothing (literally nothing), while WiFiClient setTimeout really modifies the socket send and receive timeouts. This is not an uniform behaviour and I think this should be fixed. This is my proposal but I would like you to review it very **thoroughly** in case I'm ommiting something.

Now both classes have a class variable called `_timeout`. These variables are setted to the original timeout defaults in the constructors of each class (3000 ms for WiFiClient and 30000ms for WiFiClientSecure). 

`setTimeout()` now sets this variable but also tries to configure the socket in case the connection was previously established (this is the original WiFiClient behaviour except for the class variable setting). This variable (`_timeout`) will be used for each new connection WiFiClient and WiFiClientSecure make. 

In case of WiFiClientSecure there's also `setHandshakeTimeout()`, this is kept as is, but any interaction between `_timeout` and `sslclient->handshake_timeout` was removed because now they should remain independent timeouts. 

Previous to this PR, you could not set the send and receive timeouts of WiFiClientSecure. They were always forced to 30000 ms. Now the behaviour is similar to WiFiClient. 
Also previously with WiFiClient, if you set the timeout BEFORE the connection, it would end up ignored because the socket was not created and the configuration was not saved (there was no `_timeout` class variable).

Honestly I must admit that I felt a bit surprised to see these inconsistencies on such an important part of the arduino core. Please review it and let me know what you think. This is just a proposal but feel free to change everything you want. 

## Impact
Now you can call setTimeout() before a connection in WiFiClient.
Now you can call setTimeout() with a WiFiClientSecure and actually set something...
WiFiClientSecure send and receive timeouts now are independent of the handshake timeout.

## Related links
I created this bug report a few days ago (https://github.com/espressif/arduino-esp32/issues/6429#issuecomment-1095040366)


dexion:
Why did you not set `Stream`'s timeout in `WifiClientSecure` like `Client::setTimeout(seconds * 1000); `?
`_timeout` variable is overloaded in ancestor and it wouldn't be set for `Stream`'s functions like `Stream::readStringUntil` which uses `Stream::timedRead`? BTW `_timeout` should be `int`, it can't be less zero. I would prefer use `unsigned long` for clarity )

gonzabrusco:
> Why did you not set `Stream`'s timeout in `WifiClientSecure` like `Client::setTimeout(seconds * 1000); `? `_timeout` variable is overloaded in ancestor and it wouldn't be set for `Stream`'s functions like `Stream::readStringUntil` which uses `Stream::timedRead`? BTW `_timeout` should not be `int`, it can't be less zero. I would prefer use `unsigned int` for clarity )

About the type, I used `int` following what WiFiClientSecure used originally, to be consistent. Also, all the prototypes of the class use "int32_t timeout" as the passing type for the timeout. So I respected that too. Nevertheless, I agree with you that maybe all should be an unsigned type.

About the `Client::setTimeout(seconds * 1000);`, actually I didn't do It because I wasn't sure if WiFiClient should do it in the first place. What they are modifying calling the Client::Timeout is the timeout in the Stream Class as you said. But I think that with WiFiClient and WiFiClientSecure that timeout is meaningless because everything is handled at the LWIP level. But I could be wrong, I wasn't really sure about that so I decided to "not change it".

me-no-dev:
Hi @gonzabrusco ! Your implementation looks great. Thanks!

PR #6567, opened 2022-04-12
Salvage ESP_IDF TAGs when ArduinoCore logging is enabled

ankostis:
## Summary

When ArduinoCore loggining system is enabled (`DUSE_ESP_IDF_LOG=0`)
the tags of any `ESP_LOGx()` pre-existing log-statements are ignored,
losing thus significant contextual info when developers are looking at the log dumps.

This PR joins the tag into the printf format-string of the log-statement,
like: `[tag] rest of format string`

## Impact
With this PR, the following log-statement:
```c
ESP_LOGD("NET", "state %ix", state);
```
...instead of producing a log-row lacking completely the tag, like:
```
[ 73419][D][telelogger.cpp:915] telemetry(): state: 33C
```
...it now generates this:
```
[ 73419][D][telelogger.cpp:915] telemetry(): [NET] state: 33C
```

## Issues
* Have not checked whether this PR breaks test-cases.
* I provide the code here instead of a pure issue, to facilitate interacting with the resultant log-rows.


me-no-dev:
Thanks @ankostis :)

PR #6586, opened 2022-04-16
Minor changes in WiFiClientEnterprise.ino

martinius96:
## Summary
**This pull request contains changes in [WiFiClientEnterprise.ino](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/examples/WiFiClientEnterprise/WiFiClientEnterprise.ino) file in main branch:**

- added PROGMEM for certs for storing them in flash memory of ESP32
- changed host (arduino.php5.sk) that is not working for more than year
- new host (arduino.clanweb.eu) will provide JSON output for client after successful connection
- added F macro to constant Strings
- changed Eduroam to eduroam in comments (eduroam network is still with small e character)

**Expected output after pull request:**
eduroam - WiFiClientEnterprise on ESP32

## Impact
**Purpose of this Pull Request is to:**
- Imporoved User Experience
- User can now connect to host after gets connectivity from eduroam network

### DIFF
--- a/libraries/WiFi/examples/WiFiClientEnterprise/WiFiClientEnterprise.ino
+++ b/libraries/WiFi/examples/WiFiClientEnterprise/WiFiClientEnterprise.ino
@@ -1,86 +1,89 @@
 #include <WiFi.h> //Wifi library
 #include "esp_wpa2.h" //wpa2 library for connections to Enterprise networks
-#define EAP_IDENTITY "login" //if connecting from another corporation, use identity@organisation.domain in Eduroam 
+#define EAP_IDENTITY "login" //if connecting from another corporation, use identity@organisation.domain in eduroam or anonymous@organisation.domain
 #define EAP_USERNAME "login" //oftentimes just a repeat of the identity
-#define EAP_PASSWORD "password" //your Eduroam password
-const char* ssid = "eduroam"; // Eduroam SSID
-const char* host = "arduino.php5.sk"; //external server domain for HTTP connection after authentification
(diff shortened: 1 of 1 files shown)


atanisoft:
FYI, [PROGMEM](https://github.com/espressif/arduino-esp32/blob/6cfe4613e4b4846e1ab08c7f78b7ea241f52c7da/cores/esp32/pgmspace.h#L32) and [F(..)](https://github.com/espressif/arduino-esp32/blob/6cfe4613e4b4846e1ab08c7f78b7ea241f52c7da/cores/esp32/WString.h#L38-L40) are effectively no-op on ESP32.

mrengineer7777 (commented):
Note this changes the target host from "arduino.php5.sk" to "arduino.clanweb.eu".  The old site appears dead.

Xylopyrographer:
@atanisoft Mike: Re: your comment on PROGMEM and F(...)

Interesting & certainly news to me.

What is the proper arduino-esp32 way of forcing things into flash?

martinius96:
I am surprised for these things about PROGMEM and F macro for me..
It looks like anything that is const is automatically saved in flash.
And Strings also too without F macro I think.

mrengineer7777:
> In case of Strings I am not sure.

String literals `Serial.println("Hello World");` are stored in flash on ESP32.  The only thing I'm not sure on is:
`char buf[] = "Hello";`  Should be in RAM
`const char buf[] = "Hello";`  Should be in FLASH

PR #6602, opened 2022-04-21
Change pinMode OUTPUT to INPUT_OUTPUT

### LINKED ISSUE #5552: ESP32 C3 digitalRead() returns always 0, should return the pin level
When I do a digitalWrite() to set the GPIO pin (e.g. 2) to HIGH, a digitalRead(2) returns always 0
It should return the pin level as it does not the regular ESP32 or the S2
Arduino-esp32 version: 2.0.0-alpha1
Regards Helmut

P-R-O-C-H-Y:
## Summary
This change is to match the official Arduino API, where you can use digitalRead(pin) on OUTPUT pin to read its level.

## Impact
None.

## Related links
Closes #5552

### DIFF
--- a/cores/esp32/esp32-hal-gpio.h
+++ b/cores/esp32/esp32-hal-gpio.h
@@ -42,7 +42,7 @@ extern "C" {
 
 //GPIO FUNCTIONS
 #define INPUT             0x01
-#define OUTPUT            0x02
+#define OUTPUT            0x03 //Changed from 0x02 to behave the same as Arduino pinMode(pin,OUTPUT) where yo ucan read the state of pin even when it is set as OUTPUT
 #define PULLUP            0x04
 #define INPUT_PULLUP      0x05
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-gpio.h
```diff
@@ -42,7 +42,7 @@ extern "C" {
 
 //GPIO FUNCTIONS
 #define INPUT             0x01
-#define OUTPUT            0x02
+#define OUTPUT            0x03 //Changed from 0x02 to behave the same as Arduino pinMode(pin,OUTPUT) where yo ucan read the state of pin even when it is set as OUTPUT
```
atanisoft, line 45: `where yo ucan` typo... It may also make sense to put the comment on it's own line.

P-R-O-C-H-Y: Thanks for pointing that out. Fixed :)

Issue #6508, opened 2022-03-31
Face Detection in CameraWebServer Example Not Working in 2.0.3


HaydenHu:
### Board

ESP32-CAM

### Device Description

CAMERA_MODEL_AI_THINKER  ESP32-CAM

### Hardware Configuration

#define SIOD_GPIO_NUM     26
#define SIOC_GPIO_NUM     27

### Version

latest master

### IDE Name

arduino IDE

### Operating System

win10

### Flash frequency

40MHZ

### PSRAM enabled

yes

### Upload speed

115200

### Description

![image](https://user-images.githubusercontent.com/30103277/161090119-24d90fa6-7548-4578-a591-9694cc52cf70.png)

### Sketch

```cpp
#define CAMERA_MODEL_AI_THINKER // Has PSRAM
```

### Debug Message

```plain
rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1344
load:0x40078000,len:13516
load:0x40080400,len:3604

entry 0x400805f0
[mum2-hal-cpu.c:214] setCpuFrequencyMhz(): PLL: 480 / 2 = 240 Mhz, APB: 80000000 Hz
[   842][I][esp32-hal-psram.c:96] psramInit(): PSRAM enabled



[  1122][D][WiFiGeneric.cpp:852] _eventCallback(): Arduino Event: 0 - WIFI_READY
[  1223][D][WiFiGeneric.cpp:852] _eventCallback(): Arduino Event: 2 - STA_START
...
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

me-no-dev:
yes, the API for image recognition changed and we need to update the examples. It will be done.

Lupo-Grigio:
@me-no-dev Thank you. I can add a little bit of info to this issue. Under Arduino IDE 2.0.0.rc5 and Arduino IDE 1.8.19
If you enable face detection or recognition by either one or both of the following #define
```
#define CONFIG_ESP_FACE_DETECT_ENABLED 1
#define CONFIG_ESP_FACE_RECOGNITION_ENABLED 1
```
[...]

PR #6603, opened 2022-04-21
Update camera example to support face detection and recognition

me-no-dev:
Fixes: #6508


HaydenHu:
The enroll face button cannot register faces

me-no-dev:
recognition does not work at all on chips other than ESP32-S3. Is that what you are running?

HaydenHu:
image
I'm using this development board, which uses esp32, not esp32s3

me-no-dev:
The updated AI library for image recognition does work on ESP32, but it would take 20 seconds to run a single 320x240 frame (compared to a few hundred milliseconds on ESP32-S3), so I have disabled it on ESP32 and ESP32-S2 to save flash space and memory. If you really want face recognition, maybe go back to 1.0.6.

libraries/ESP32/examples/Camera/CameraWebServer/partitions.csv
```diff
@@ -0,0 +1,5 @@
+# Name,   Type,  SubType, Offset,   Size,    Flags
+nvs,      data,  nvs,     0x9000,   0x5000,
+otadata,  data,  ota,     0xe000,   0x2000,
+app0,     app,   ota_0,   0x10000,  0x3d0000,
+fr,       data,        ,  0x3e0000, 0x20000,
```
knopkakhua, line 5: coredump, data, coredump,0x3F0000,0x10000,

PR #6668, opened 2022-04-30
Added new "Non TinyUF2 bootloader only" capability for boards that support TinyUF2 bootloaders

UnexpectedMaker:
## Description of Change
Implemented a new method for allowing folks to preserve their TinyUF2 bootloader setup when flashing their boards via Arduino IDE, without being locked out of not using it if they want to use a different partition.

Adafruit had already added support for keeping (reflashing) the TinyUF2 bootloader when flashing via Arduino IDE (thanks @ladyada  ), but the issue with it is it doesn't allow users to `not choose` to use it. Even if they select a specific partition scheme from the partition drop down list, it ignores that selection and only does the TinyUF2 partitioning/bootloader thing.

We wanted to let users choose between keeping their TinyUF2 partitioning and bootloader, or choose another partition scheme and have the TinyUF2 support disabled and just operate like they would expect.

We've implemented this by adding the TinyUF2 support as a partition scheme option, and using these build options in `platform.txt` to choose to use teh UF2 path or use the standard path.

tinyuf2_support

## Tests scenarios
This has been tested in 2.0.3RC1 and latest master on all UM ESP32-S3 boards, with and without TinyUF2 support, and also tested on the Adafruit QT PY to ensure it still retains the UF2 and on a TinyS2 with a standard 4MB partition, and all boards and options works correctly!  

## Related links
 N/A


UnexpectedMaker:
@me-no-dev I don't understand why CLA assistant is wanting me to sign by 2 identities when I only use the `Unexpected Maker` profile. I don't even know how to change to this other `Seon Rozenblum` profile :(

UnexpectedMaker:
> @me-no-dev I don't understand why CLA assistant is wanting me to sign by 2 identities when I only use the `Unexpected Maker` profile. I don't even know how to change to this other `Seon Rozenblum` profile :(

Don't worry, @prplz helped me fix the commit and now it's on my signed user.

VojtechBartoska:
> @me-no-dev I don't understand why CLA assistant is wanting me to sign by 2 identities when I only use the `Unexpected Maker` profile. I don't even know how to change to this other `Seon Rozenblum` profile :(

That's weird. I have no idea why this happened.

UnexpectedMaker:
> > @me-no-dev I don't understand why CLA assistant is wanting me to sign by 2 identities when I only use the `Unexpected Maker` profile. I don't even know how to change to this other `Seon Rozenblum` profile :(
> 
> That's weird. I have no idea why this happened.

It was actually from my side. I sometimes use SmartGit and sometimes use command line, and my git credentials must have been messed in SmartGit, so it used my email instead of my getup login. It's all good now :)

PR #6676, opened 2022-05-02
ESPLwIPClient::setTimeout conflict fix with Stream::setTimeout

### LINKED ISSUE #5558: ESPLwIPClient setTimeout conflicts with Stream setTimeout
ESPLwIPClient::setTimeout conflicts with Stream::setTimeout

the classes hierarchy is Print <- Steam <- Client <- ESPLwIPClient <- WiFiClient 

https://github.com/espressif/arduino-esp32/blob/b580bb23fdcce0d3063b0b0f766a74e32bd8ea67/cores/esp32/Stream.h#L61

https://github.com/espressif/arduino-esp32/blob/b580bb23fdcce0d3063b0b0f766a74e32bd8ea67/libraries/WiFi/src/WiFiClient.h#L36

P-R-O-C-H-Y:
## Description of Change
Arduino Stream::settimeout() is not virtual so it cannot be virtual in ESPLwIPClient.
This change should fix this conflict.

## Related links
Closes #5558

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -495,7 +495,7 @@ void HTTPClient::setTimeout(uint16_t timeout)
 {
     _tcpTimeout = timeout;
     if(connected()) {
-        _client->setTimeout(timeout + 500); /* / 1000 removed, WifiClient setTimeout changed to ms */
+        _client->setTimeout(timeout);
     }
 }
 
@@ -1165,7 +1165,7 @@ bool HTTPClient::connect(void)
     }
 
     // set Timeout for WiFiClient and for Stream::readBytesUntil() and Stream::readStringUntil()
-    _client->setTimeout(_tcpTimeout + 500); /* / 1000 removed, WifiClient setTimeout changed to ms */	
+    _client->setTimeout(_tcpTimeout);	
 
     log_d(" connected to %s:%u", _host.c_str(), _port);
 
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -495,7 +495,7 @@ void HTTPClient::setTimeout(uint16_t timeout)
 {
     _tcpTimeout = timeout;
     if(connected()) {
-        _client->setTimeout((timeout + 500) / 1000);
+        _client->setTimeout(timeout + 500); /* / 1000 removed, WifiClient setTimeout changed to ms */
     }
 }
 
@@ -1165,7 +1165,7 @@ bool HTTPClient::connect(void)
     }
 
     // set Timeout for WiFiClient and for Stream::readBytesUntil() and Stream::readStringUntil()
-    _client->setTimeout((_tcpTimeout + 500) / 1000);	
+    _client->setTimeout(_tcpTimeout + 500); /* / 1000 removed, WifiClient setTimeout changed to ms */	
 
     log_d(" connected to %s:%u", _host.c_str(), _port);
 
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -404,8 +404,8 @@ size_t WiFiClient::write(const uint8_t *buf, size_t size)
         if(_lastWriteTimeout != _timeout){
             if(fd() >= 0){
                 struct timeval timeout_tv;
-                timeout_tv.tv_sec = _timeout/1000;
-                timeout_tv.tv_usec = 0;
+                timeout_tv.tv_sec = _timeout / 1000;
+                timeout_tv.tv_usec = (_timeout % 1000) * 1000;
                 if(setSocketOption(SO_SNDTIMEO, (char *)&timeout_tv, sizeof(struct timeval)) >= 0)
                 {
                     _lastWriteTimeout = _timeout;
@@ -475,8 +475,8 @@ int WiFiClient::read(uint8_t *buf, size_t size)
     if(_lastReadTimeout != _timeout){
         if(fd() >= 0){
             struct timeval timeout_tv;
-            timeout_tv.tv_sec = _timeout/1000;
-            timeout_tv.tv_usec = 0;
+            timeout_tv.tv_sec = _timeout / 1000;
+            timeout_tv.tv_usec = (_timeout % 1000) * 1000;
             if(setSocketOption(SO_RCVTIMEO, (char *)&timeout_tv, sizeof(struct timeval)) >= 0)
             {
                 _lastReadTimeout = _timeout;
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -211,6 +211,8 @@ void WiFiClient::stop()
     clientSocketHandle = NULL;
     _rxBuffer = NULL;
     _connected = false;
+    _lastReadTimeout = 0;
+    _lastWriteTimeout = 0;
 }
 
 int WiFiClient::connect(IPAddress ip, uint16_t port)
@@ -329,13 +331,6 @@ int WiFiClient::getSocketOption(int level, int option, const void* value, size_t
         log_e("fail on fd %d, errno: %d, \"%s\"", fd(), errno, strerror(errno));
     }
     return res;
-}
-
-void WiFiClient::setTimeout(uint32_t seconds)
-{
-    _lastReadTimeout = _timeout;
-    _lastWriteTimeout = _timeout;
-    _timeout = seconds * 1000;
 }
 
 int WiFiClient::setOption(int option, int *value)
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -331,23 +331,25 @@ int WiFiClient::getSocketOption(int level, int option, const void* value, size_t
     return res;
 }
 
-
-int WiFiClient::setTimeout(uint32_t seconds)
+void WiFiClient::setTimeout(uint32_t seconds)
 {
+    _lastReadTimeout = Client::getTimeout();
+    _lastWriteTimeout = _lastReadTimeout;
     Client::setTimeout(seconds * 1000); // This should be here?
     _timeout = seconds * 1000;
-    if(fd() >= 0) {
-        struct timeval tv;
-        tv.tv_sec = seconds;
-        tv.tv_usec = 0;
-        if(setSocketOption(SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval)) < 0) {
-            return -1;
-        }
-        return setSocketOption(SO_SNDTIMEO, (char *)&tv, sizeof(struct timeval));
-    }
-    else {
-        return 0;
-    }
+    
+    // if(fd() >= 0) {
+    //     struct timeval tv;
+    //     tv.tv_sec = seconds;
+    //     tv.tv_usec = 0;
+    //     if(setSocketOption(SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval)) < 0) {
+    //         return -1;
+    //     }
+    //     return setSocketOption(SO_SNDTIMEO, (char *)&tv, sizeof(struct timeval));
+    // }
+    // else {
+    //     return 0;
+    // }
 }
 
 int WiFiClient::setOption(int option, int *value)
@@ -418,6 +420,18 @@ size_t WiFiClient::write(const uint8_t *buf, size_t size)
         tv.tv_usec = WIFI_CLIENT_SELECT_TIMEOUT_US;
         retry--;
 
+        if(_lastWriteTimeout != _timeout){
+            if(fd() >= 0){
+                struct timeval tv;
+                tv.tv_sec = _timeout/1000;
+                tv.tv_usec = 0;
+                if(setSocketOption(SO_SNDTIMEO, (char *)&tv, sizeof(struct timeval)) >= 0)
+                {
+                    _lastWriteTimeout = _timeout;
+                }
+            }
+        }
+
         if(select(socketFileDescriptor + 1, NULL, &set, NULL, &tv) < 0) {
             return 0;
         }
@@ -477,6 +491,18 @@ size_t WiFiClient::write(Stream &stream)
 
 int WiFiClient::read(uint8_t *buf, size_t size)
 {
+    if(_lastReadTimeout != _timeout){
+        if(fd() >= 0){
+            struct timeval tv;
+            tv.tv_sec = _timeout/1000;
+            tv.tv_usec = 0;
+            if(setSocketOption(SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval)) >= 0)
+            {
+                _lastReadTimeout = _timeout;
+            }
+        }
+    }
+
     int res = -1;
     if (_rxBuffer) {
         res = _rxBuffer->read(buf, size);
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -335,21 +335,7 @@ void WiFiClient::setTimeout(uint32_t seconds)
 {
     _lastReadTimeout = Client::getTimeout();
     _lastWriteTimeout = _lastReadTimeout;
-    Client::setTimeout(seconds * 1000); // This should be here?
     _timeout = seconds * 1000;
-    
-    // if(fd() >= 0) {
-    //     struct timeval tv;
-    //     tv.tv_sec = seconds;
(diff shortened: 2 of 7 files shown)


P-R-O-C-H-Y:
@me-no-dev @igrr PTAL on these changes.

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -459,6 +455,18 @@ size_t WiFiClient::write(Stream &stream)
 
 int WiFiClient::read(uint8_t *buf, size_t size)
 {
+    if(_lastReadTimeout != _timeout){
+        if(fd() >= 0){
+            struct timeval timeout_tv;
+            timeout_tv.tv_sec = _timeout/1000;
+            timeout_tv.tv_usec = 0;
```
igrr, line 462: should this be `_timeout % 1000`? (To support timeouts less than one second)

igrr (approved):
Just one note, LGTM otherwise.

Also please check if any examples used WifiClient::setTimeout and see if they should be updated to take milliseconds instead of seconds.

libraries/WiFi/src/WiFiClient.cpp
P-R-O-C-H-Y: Should it be implemented this way to use `tv_sec` and `tv_usec` at the same time?

```
timeout_tv.tv_sec = _timeout / 1000;
timeout_tv.tv_usec = (_timeout % 1000) * 1000;
```

For 1500ms `_timeout` that will be `timeout_tv.tv_sec = 1` and `timeout_tv.tv_usec = 500000`

P-R-O-C-H-Y:
> Just one note, LGTM otherwise.
> 
> Also please check if any examples used WifiClient::setTimeout and see if they should be updated to take milliseconds instead of seconds.

Checked  and changed in this [commit](https://github.com/espressif/arduino-esp32/pull/6676/commits/25b2062c0a63666a0e99eb9bd23d6a319bf9c149)

libraries/WiFi/src/WiFiClient.cpp
P-R-O-C-H-Y: Edited, now the seconds are not rounded and timeout can be less than 1 second

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -495,7 +495,7 @@ void HTTPClient::setTimeout(uint16_t timeout)
 {
     _tcpTimeout = timeout;
     if(connected()) {
-        _client->setTimeout((timeout + 500) / 1000);
+        _client->setTimeout(timeout + 500); /* / 1000 removed, WifiClient setTimeout changed to ms */
```
igrr, line 498: @P-R-O-C-H-Y I would suggest to put notes such as this into commit message body, instead of adding them as comments. The reader of the code is probably not interested in every change which has been done to each line, they just need to understand the final version.

For example, I don't understand why there is timeout+500 here.

I think previously +500 in (timeout + 500)/1000 was related to rounding. But with /1000 removed, looks like +500 is not needed anymore?

igrr:
One more comment about leftover +500 and not using comments as a changelog. Otherwise still LGTM.

libraries/HTTPClient/src/HTTPClient.cpp
me-no-dev: Yes, this was there for rounding. `+500` should be removed

me-no-dev:
So the titles says that you are fixing a conflict inheriting timeout between Stream and Client, but in the code I see `setTimeout(seconds)` being removed and "replaced" by `Stream::setTimeout(milliseconds)`. This is a rather significant change for anyone who uses seconds in their code. Chances are that "seconds" would be low value that would cause premature timeout for their clients after this change is applied. Even if we should go forward with this, it seems like rather big change for `2.0.3`

libraries/HTTPClient/src/HTTPClient.cpp
P-R-O-C-H-Y: The rounding was removed, missed that.
Thanks for suggestions about commenting the code / commits. Will have it in mind for future changes :)

gonzabrusco:
This is kind of destroying what I wanted to accomplish with this recent PR (https://github.com/espressif/arduino-esp32/pull/6562). 

If the timeout is in seconds, it should remain in seconds or change it everywhere it is used! (Wificlientsecure for example)

gonzabrusco:
Sorry I do not undestand something about this approach. What would be the correct way of setting the timeout now? I see you removed the setTimeout from WiFiClient and WiFiClientSecure... 

I'm currently using a MQTT Class that takes a Client inherited class (WiFiClient, WiFiClientSecure, etc) and when you start a MQTT Class object (mqttObject.connect()), it internally calls Client.connect() but without a timeout parameter (this is out of my control unless the author modifies this). If I loose the ability to set the timeout of my Client beforehand it can be problematic because the MQTT class does not call connect passing a timeout parameter. How this situation can be handled with your proposal?

P-R-O-C-H-Y:
> I would like to understand SO_RCVTIMEO and SO_SNDTIMEO. Does it make WiFiClient read() and write() blocking? Is it good to have the same timeout for read() (which shouldn't block) as for Stream class's blocking functions (readBytes, parseInt, find) which internally use read()?

Just want to mention, that the implementation is the same as it was before. This PR just resolves the conflict between the setTimeout functions. 
I don't understand, why the read shouldn't block? You may need to wait some time to get the data to read, if data do not come, do timeout.

JAndrassy:
> I don't understand, why the read shouldn't block? You may need to wait some time to get the data to read, if data do not come, do timeout.

read() should only read what is immediately available in buffer. in embedded systems blocking can be a problem.

VojtechBartoska:
We are postponing this Pull Request as more investigation and testing is needed. Added to 2.1.0 milestone.

podaen:
Could you please set Client::setTimeout(_timeout) in Client Connect to support lower versions dat doesn't have that function when changing the socket on a new connection.

podaen:
> The change of timeout from seconds to ms is a breaking change.

Out of my tests setting the client, stream and/or SSL timeouts in ms is an easier way to set all timeouts to the same value.

me-no-dev:
> Out of my tests setting the client, stream and/or SSL timeouts in ms is an easier way to set all timeouts to the same value.

Such change will break ANY current library/sketch that sets the timeout in seconds. This API has been such for long time (probably coming from 8266 even). Even worst, it will break it in such way that probably none of the sketched will be able to even connect to anything (imagine having set a timeout of 5 seconds and that turns into 5ms after the change)

podaen:
Your right, but by add an procedure with argument true/false...

TD-er:
> > Out of my tests setting the client, stream and/or SSL timeouts in ms is an easier way to set all timeouts to the same value.
> 
> Such change will break ANY current library/sketch that sets the timeout in seconds. This API has been such for long time (probably coming from 8266 even). Even worst, it will break it in such way that probably none of the sketched will be able to even connect to anything (imagine having set a timeout of 5 seconds and that turns into 5ms after the change)

ESP8266 implementation does call Client::setTimeout, which uses msec.

On ESP32 it is IMHO done incorrect as it is effectively overriding the existing `setTimeout` function in `Stream`.
So it makes a difference if you call setTimeout on a point of type `Stream` or `WiFiClient` which is very confusing to say the least.

I'm not sure it will actually break existing implementations as I do expect almost anyone expects this to be in msec.
So changing it may probably fix those implementations.

TD-er:
Another argument to change it into msec is that `WiFiUDP` uses msec for its `setTimeout()` function.

everslick:
Just an observation from my side: I have a firmware which builds for ESP32 and ESP8266 which calls `setTimeout()` in some places (`HTTPClient`/`WiFiClient` and `WiFIClientSecure`) and I never actually realized, that the API wants me to use *seconds* on ESP32 (the ESP8266 code came first). Now reading this thread I went through my code and tried to *fix* all occurrences of `setTimeout()` (ms on ESP8266 and seconds on ESP32), which, you guessed it, broke it completely.

For me `HTTPClient::setTimeout(5)` sets the timeout to 5ms, which makes it unable to connect, setting it to *5000* seems to let it default to something reasonable like 8 seconds, which at least 'works'.

I think switching to `ms` might as well fix more programs then break 'em.

mrengineer7777:
I recently ran into this problem as well (Arduino/ESP32 2.03).  setTimeout() here is unclear if it should be seconds or ms:

`void HTTPClient::setTimeout(uint16_t timeout)
{
    _tcpTimeout = timeout;
    if(connected()) {
        _client->setTimeout((timeout + 500) / 1000);
    }
}`

_client call is `WiFiClient::setTimeout(uint32_t seconds)`, so I assumed seconds but the connection immediately times out.  It only works if I call HTTPClient::setTimeout(30000) for 30 seconds.

JAndrassy:
@mrengineer7777 and @everslick  HTTPClient has milliseconds and is not a part of this issue

SuGlider:
I was reviewing the PR #7686 and I made a suggestion to add time in milliseconds, but still keeping the compatibility to the API with seconds.

The idea would be use time as float.
I can tell Arduino to wait 500ms by using 0.5, 2 seconds by using 2 or 2.0, 1250ms by using 1.25 as timeout parameter.

``` cpp
int WiFiClient::setTimeout(float seconds)
```

I started a discussion about it in https://github.com/espressif/arduino-esp32/pull/7686#issuecomment-1407480299
It would cover the overload with `uint32_t seconds` given  that C already promotes `int` to `float` automatically. 

Please let me know your opinion about it.

podaen:
Form define -> int32_t -> unsigned float -> unsigned long. You might change the name from timeout to a unit. :-)
			     			                  -> int

mrengineer7777:
> The idea would be use time as `float`.

@SuGlider I cringe at the idea of `float` overhead.  My preference would be to specify everything in ms.  To avoid a breaking change perhaps we could introduce a new function setTimeout_ms()?  This is a complicated PR so it needs more study.  At the moment I would prefer to see some of our other open PRs merged.

podaen:
I don't know if this is allready been sad and can be off topic, but why there is a timeout function in WiFiCient? Currently here are two moments where the timeout is set... by creating a WiFiClient (set the default [ms]) and in connecting before sending the request (custum timeout[sec]). 

We are going in bridges. At the end it will be always in ms. I will comeback on this, but for now lets continue on thinking when the timeout need to be set. I think the solution of Garcia will work, if the promoted from uint32 to an foat works like they say.

I once changed the timeout in a huge lib working with streams and cam to the conclusion that set socket timeout must be able to set at any time.
The stream timeout is to read, find etc, but can it go up? Stream is connected to Client, ESPLwIPClient, WiFiClient. The default timeout in Stream is 1ms, that was to small for finding a word, for cutting the streams in to chapters. I already had set the timeout in to WiFiClient, but the Stream timeout was not set. 
I remember now what the problem was... Running in a timeout with a stream and than auto connecting to the next stream, than the stream broke more often, when it else did not. Was it because that the socket timeout was set to default or was it that a new socket needed to be created. The solution was starting back from the beginning.

Continuing on the second or ms.
WiFiClient::connect [ms]
WiFiClient::setTimeout [sec] 

Because the setsockopt from the LwIp is in seconds. So what I sad before that it always ends up in milliseconds is not true. But it doesn't make it easier and using the word 'timeout' doesn't helps now certainly when it gets more complicated. That is everything I got form working with streams. I hope it helps!

VojtechBartoska:
relates https://github.com/espressif/arduino-esp32/issues/6835

me-no-dev:
@P-R-O-C-H-Y can you rebase this please

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -393,6 +419,8 @@ int WiFiClientSecure::setTimeout(uint32_t seconds)
 }
 
 int WiFiClientSecure::fd() const
+
+int WiFiClientSecure::setSocketOption(int option, char* value, size_t len)
```
me-no-dev, line 423: something must have happened here? This line belongs elsewhere IMHO

P-R-O-C-H-Y: Should be good now.

libraries/WiFiClientSecure/src/WiFiClientSecure.h
```diff
@@ -82,6 +82,7 @@ class WiFiClientSecure : public WiFiClient
     bool getFingerprintSHA256(uint8_t sha256_result[32]) { return get_peer_fingerprint(sslclient, sha256_result); };
     int setTimeout(uint32_t seconds);
     int fd() const;
+    int setSocketOption(int option, char* value, size_t len);
```
me-no-dev: this should not be here, if it's not implemented in the ClientSecure class (it's coming from WiFiClient)

JAndrassy:
@me-no-dev I recommend my PR https://github.com/espressif/arduino-esp32/pull/8863 instead of this one

me-no-dev:
@JAndrassy this PR solves a different thing, which is inheritance of `setTimeout`. Your PR should be rebased on top of this one, as it will not require some of your changes anymore.

JAndrassy:
so my PR renamed setTimeout to setConnectionTimeout. this one now removed setTimeout, and my PR adds it back as setConnectionTimeout. Except of WiFiClientSecure where this PR didn't remove the shadowing setTimeout so my PR still renames it to setConnectionTimeout. 
the rest of the PR was same except a comment like 
/* /1000 removed, WifiClient setTimeout changed to ms */
(which doesn't make sense for a future reader)
<end of rant>

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -477,6 +472,18 @@ size_t WiFiClient::write(Stream &stream)
 
 int WiFiClient::read(uint8_t *buf, size_t size)
 {
+    if(_lastReadTimeout != _timeout){
+        if(fd() >= 0){
+            struct timeval timeout_tv;
+            timeout_tv.tv_sec = _timeout / 1000;
+            timeout_tv.tv_usec = (_timeout % 1000) * 1000;
+            if(setSocketOption(SO_RCVTIMEO, (char *)&timeout_tv, sizeof(struct timeval)) >= 0)
```
cziter15, line 480: Cosmethic change, but I suggest to use sizeof(timeout_tv)

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -199,6 +203,16 @@ size_t WiFiClientSecure::write(const uint8_t *buf, size_t size)
     if (!_connected) {
         return 0;
     }
+    if(_lastWriteTimeout != _timeout){
+        struct timeval timeout_tv;
+        timeout_tv.tv_sec = _timeout / 1000;
+        timeout_tv.tv_usec = (_timeout % 1000) * 1000;
+        if(setSocketOption(SO_SNDTIMEO, (char *)&timeout_tv, sizeof(struct timeval)) >= 0)
```
cziter15, line 210: Cosmethic change, but I suggest to use sizeof(timeout_tv) instead of sizeof(struct timeval)

libraries/WiFiClientSecure/src/WiFiClientSecure.cpp
```diff
@@ -209,6 +223,18 @@ size_t WiFiClientSecure::write(const uint8_t *buf, size_t size)
 
 int WiFiClientSecure::read(uint8_t *buf, size_t size)
 {
+    if(_lastReadTimeout != _timeout){
+        if(fd() >= 0){
+            struct timeval timeout_tv;
+            timeout_tv.tv_sec = _timeout / 1000;
+            timeout_tv.tv_usec = (_timeout % 1000) * 1000;
+            if(setSocketOption(SO_RCVTIMEO, (char *)&timeout_tv, sizeof(struct timeval)) >= 0)
```
cziter15, line 231: Cosmethic change, but I suggest to use sizeof(timeout_tv) instead of sizeof(struct timeval)

libraries/WiFi/src/WiFiClient.cpp
TD-er: I disagree, as it is good practice to actually use the struct/class for the `sizeof()` instead of the object.
This way you are less likely to end up with the value `4` as result of your `sizeof()` call when working with pointers instead of reference.
Thus this will reduce the chance of programming errors.

cziter15: Imagine you change the type of `timeout_tv` somewhere earlier and forget to replace all `sizeof (struct timeval)` later in the code. Maybe a better thing is to declare const variable/define/constexpr timeval_size just after the `timeout_tv` declaration and then use this variable later in the code.

TD-er: Sure, or maybe there is a compiler flag which can warn when `sizeof()` is given a pointer as that's the actual problem.

PR #6685, opened 2022-05-04
Adjustable Serial Event Task Stack Size And Priority

gonzabrusco:
This is just a proposal so we can change on compile time the stack size and priority of the Serial Event Task.

On my current software, the serial part does not fit the current 2048 bytes of stack that the Serial Event Task offers me. So I'm forced to implement a binary semaphore for synchronization (the serial event task just gives a semaphore). This is a waste of resources because now I have two tasks (one for the serial event task and one for my actual routine) and a semaphore. Even worse, it's slower.

Thanks!

### DIFF
--- a/Kconfig.projbuild
+++ b/Kconfig.projbuild
@@ -31,6 +31,7 @@ choice ARDUINO_RUNNING_CORE
         bool "CORE 1"
     config ARDUINO_RUN_NO_AFFINITY
         bool "BOTH"
+        depends on !CONFIG_FREERTOS_UNICORE
 
 endchoice
 
@@ -58,6 +59,7 @@ choice ARDUINO_EVENT_RUNNING_CORE
         bool "CORE 1"
     config ARDUINO_EVENT_RUN_NO_AFFINITY
         bool "BOTH"
+        depends on !CONFIG_FREERTOS_UNICORE
 
 endchoice
 
@@ -66,21 +68,6 @@ config ARDUINO_EVENT_RUNNING_CORE
     default 0 if ARDUINO_EVENT_RUN_CORE0
     default 1 if ARDUINO_EVENT_RUN_CORE1
     default -1 if ARDUINO_EVENT_RUN_NO_AFFINITY
-
-choice ARDUINO_UDP_RUNNING_CORE
-    bool "Core on which Arduino's UDP is running"
-    default ARDUINO_UDP_RUN_CORE1
-    help
-        Select on which core Arduino's UDP run
-
-    config ARDUINO_UDP_RUN_CORE0
-        bool "CORE 0"
-    config ARDUINO_UDP_RUN_CORE1
-        bool "CORE 1"
-    config ARDUINO_UDP_RUN_NO_AFFINITY
-        bool "BOTH"
-
-endchoice
 
 choice ARDUINO_SERIAL_EVENT_TASK_RUNNING_CORE
     bool "Core on which Arduino's Serial Event task is running"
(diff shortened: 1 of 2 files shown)


me-no-dev:
I will suggest that this PR is coupled with change to the component options [here](https://github.com/espressif/arduino-esp32/blob/master/Kconfig.projbuild) so that idf-component users are also able to configure this.

gonzabrusco:
> I will suggest that this PR is coupled with change to the component options [here](https://github.com/espressif/arduino-esp32/blob/master/Kconfig.projbuild) so that idf-component users are also able to configure this.

Done!

SuGlider (approved):
@gonzabrusco - Thanks for the contribution!
LGTM

Kconfig.projbuild
```diff
@@ -82,6 +82,39 @@ choice ARDUINO_UDP_RUNNING_CORE
 
 endchoice
 
+choice ARDUINO_SERIAL_EVENT_TASK_RUNNING_CORE
```
me-no-dev, line 85: this option should depend on wether the chip has more than one core. Example implementation: https://github.com/espressif/esp-idf/blob/master/components/lwip/Kconfig#L675

me-no-dev:
Very nice indeed. Happy to see you use `xTaskCreateUniversal`

Kconfig.projbuild
gonzabrusco: Hello @me-no-dev , I'm a bit confused about this suggestion. In the other config entries (about running cores) there's no dependency on this define, should this be added to every entry aswell?

Also, I did a search about that specific define, but I've found CONFIG_FREERTOS_UNICORE and not FREERTOS_UNICORE, is that a bug in the line you linked to me? 

I guess I could answer myself some of this questions, but I don't really know how to test this. I don't know what's the procedure to use this kconfig file. Could you point me to a guide?

SuGlider: `CONFIG_` prefix is added to KConfig Options in its created header file automaticaly.
So Kconfig Menu option FREERTOS_UNICORE will be seen as a `#define CONFIG_FREERTOS_UNICORE` in the source code.

SuGlider: I think we may need to review it in the whole Kconfig.projectbuild.

I also got the same doubt as @gonzabrusco ...
Shall we include this dependency for all not valid options when there is just one Core (S2/C3)?
Should the default option go to Core0 when there is just one Core?

gonzabrusco: I think that if we use `xTaskCreateUniversal` it does not matter the default option. 

https://github.com/espressif/arduino-esp32/blob/0b3f1a9fa9bb5cc4df9169fe9130f324cbcba9db/cores/esp32/esp32-hal-misc.c#L148

me-no-dev: True that `xTaskCreateUniversal` will take care regardless of how many cores, but this is a safe-net inside the function itself. It still makes sense to hide options that are invalid for the current target in ESP-IDF. This is how it is done for all IDF components. Arduino is no different :)

> I think we may need to review it in the whole Kconfig.projectbuild.

Probably. It has not really been touched since ESP32, which has 2 cores always :D At least what we support for it.

gonzabrusco: I'm on vacation now, so feel free to change what you want to move this forward.

me-no-dev: We are not in a hurry, so no worries :) Have fun at your vacation

mrengineer7777:
Conflicts with #6718, which seems to implement some of the same fixes.  Recommend merging 6718 and then updating this PR as necessary.  Great work on this @gonzabrusco !

SuGlider:
I've made it aligned to #6718 and eliminated merging conflicts.
Ready for merging along with #6718

SuGlider (approved):

PR #6688, opened 2022-05-04
Upload component to the registry on any tag

kumekay:
## Description of Change

`upload-idf-component.yml` action was only running for tags like `v*`, while in this project tags like `2.0.3` are used. With this change, the workflow will run for all tags.


VojtechBartoska (approved):
Did you also handle a manual upload of [2.0.3](https://github.com/espressif/arduino-esp32/releases/tag/2.0.3) version?

PR #6695, opened 2022-05-05
Changed pinMode() default interrupt type DISABLED to previously set

### LINKED ISSUE #6669: `pinMode` disables interrupts as a side effect
### Board

ESP32-DevKitC

### Device Description

DevKitC

### Hardware Configuration

For the demo code below, connect a switch or square wave to pin 19.

### Version

v2.0.2

### IDE Name

PlatformIO

### Operating System

macOS

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

The `pinMode` call, in addition to changing pullups/whether a pin is driving the output or not/etc, appears to disable pin interrupts as a side effect. In the ESP-IDF version of the code, `intr_type` is explicitly set to `GPIO_INTR_DISABLE`. In the older register code, the field is accidentally overwritten by the write to `GPIO.pin[x].val` when it's really just trying to set `GPIO.pin[x].pad_driver`.

This is a problem because while `pinMode` is `IRAM_ATTR` and safe to call from interrupt handlers, `attachInterrupt` is not (and it's also rather heavyweight), so it's not possible to re-enable the interrupt if `pinMode` is being used in an ISR.

`pinMode` should just preserve whatever previous state of interrupts existed; if the user has an interrupt configured on a pin and wants to change the pin mode, they probably know what they are doing. If they want to disable the interrupt they can call `detachInterrupt` directly.

### Sketch

```cpp
#include "Arduino.h"

// This sketch should print the number of times pin 19 changes.

uint32_t interrupt_count = 0;

void IRAM_ATTR pin_interrupt(void *arg) {
  interrupt_count++;
  pinMode(19, INPUT);
}

void setup() {
  pinMode(19, INPUT);
  attachInterruptArg(19, pin_interrupt, 0, CHANGE);
}

void loop() {
  ESP_LOGD("loop", "interrupt count %d", interrupt_count);
  delay(500);
}
```

### Debug Message

```plain
n/a
```

### Other Steps to Reproduce

_No response_

P-R-O-C-H-Y:
## Description of Change
pins default configuration has `intr_type =  GPIO_INTR_DISABLE`
With this implementation, it will set the intr_type to previously set intr_type. It will no longer disable interrupt, when pinmode is called multiple times on same pin with interrupt enabled.

There is no option to get pin config from ESP-IDF API, that's why I made this implementation.

## Tests scenarios

```
//To trigger ISR, connect GPIO 19 to GND.

uint32_t interrupt_count = 0;

void IRAM_ATTR pin_interrupt(void *arg) {
  interrupt_count++;
  pinMode(19, INPUT);
  //attachInterruptArg(19, pin_interrupt, 0, CHANGE);
}

void setup() {
  pinMode(19, INPUT);
  attachInterruptArg(19, pin_interrupt, 0, CHANGE);
}

void loop() {
  ESP_LOGD("loop", "interrupt count %d", interrupt_count);
  delay(500);
}
```

## Related links
Closes #6669

### DIFF
--- a/cores/esp32/esp32-hal-gpio.c
+++ b/cores/esp32/esp32-hal-gpio.c
@@ -100,11 +100,11 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
     gpiohal.dev = GPIO_LL_GET_HW(GPIO_PORT_0);
 
     gpio_config_t conf = {
-        .pin_bit_mask = (1ULL<<pin),			        /*!< GPIO pin: set with bit mask, each bit maps to a GPIO */
-        .mode = GPIO_MODE_DISABLE,                      /*!< GPIO mode: set input/output mode                     */
-        .pull_up_en = GPIO_PULLUP_DISABLE,              /*!< GPIO pull-up                                         */
-        .pull_down_en = GPIO_PULLDOWN_DISABLE,          /*!< GPIO pull-down                                       */
-        .intr_type = gpiohal.dev->pin[pin].int_type     /*!< GPIO interrupt type - previously set                 */
+        .pin_bit_mask = (1ULL<<pin),                 /*!< GPIO pin: set with bit mask, each bit maps to a GPIO */
+        .mode = GPIO_MODE_DISABLE,                   /*!< GPIO mode: set input/output mode                     */
+        .pull_up_en = GPIO_PULLUP_DISABLE,           /*!< GPIO pull-up                                         */
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,       /*!< GPIO pull-down                                       */
+        .intr_type = gpiohal.dev->pin[pin].int_type  /*!< GPIO interrupt type - previously set                 */
     };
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-gpio.c
```diff
@@ -95,12 +95,16 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
         log_e("Invalid pin selected");
 		return;
 	}
+
+    gpio_hal_context_t gpiohal;
+    gpiohal.dev = GPIO_LL_GET_HW(GPIO_PORT_0);
```
me-no-dev, line 100: Indentation issue here :)

P-R-O-C-H-Y: Wierd, I dont see that in GitHub Deskto / VS Code or GitHub integrated editor.

P-R-O-C-H-Y: REplaced tabs with spaces, should be fine now :)

PR #6702, opened 2022-05-06
Add new board XIAO_WIFI

0hotpotman0:
**Summary**
Added board definitions for my new XIAO_WIFI base on the ESP32-C3

**Impact**
Should have no impact on the development

**link**

This is my board pic.
image


lbernstone:
Wow. Tiny!!

0hotpotman0:
> Wow. Tiny!!

yes, it is a small size ESP32 C3, like a cion.

me-no-dev:
where can we get the board? looks amazing :)

boards.txt
```diff
@@ -14375,3 +14375,153 @@ AirM2M_CORE_ESP32C3.menu.DebugLevel.verbose=Verbose
...
+
+SEEED_XIAO_WIFI.build.tarch=riscv32
+SEEED_XIAO_WIFI.build.target=esp
+SEEED_XIAO_WIFI.build.mcu=esp32c3
+SEEED_XIAO_WIFI.build.core=esp32
+SEEED_XIAO_WIFI.build.variant=SEEED_XIAO_WIFI
+SEEED_XIAO_WIFI.build.board=SEEED_XIAO_WIFI
+SEEED_XIAO_WIFI.build.bootloader_addr=0x0
+
+SEEED_XIAO_WIFI.build.cdc_on_boot=0
```
me-no-dev, line 14409: Your board seems to be using the native USB port for flashing and communication. Are you sure that you do not want this option to be on by default? If you decide to do so, you also need to switch the order of the menu items for CDCOnBoot, because the first option is taken by default

0hotpotman0:
> where can we get the board? looks amazing :)

Soon、          but I still have some pins may need to correct， it will be a nice board.

boards.txt
0hotpotman0: image
Do you mean like here set to 1?      but I saw other ESP32 C3 boards that all use this parameter.

0hotpotman0: done!   I changed it.

me-no-dev: You should also set it to 1 in the line above: `SEEED_XIAO_WIFI.build.cdc_on_boot=1`

0hotpotman0:
Hi @me-no-dev   
when my board info can release on the Arduino IDE?  I redownload the board info but still cannot find my board info.

VojtechBartoska:
in Arduino IDE, it will be available when we release a new version (Release candidate or stable). It's planned around mid of June.

PR #6705, opened 2022-05-06
fix: compiler warnings about printf-format for pointers in `esp32-hal-cpu.c`

ankostis:
the compiler complains about these 2 places in `esp32-hal-cpu.cpp` of miss-matching %X to function & void pointers,
and x2 in `Tone.cpp` miss-matching `uint` with `long uint.

Sample warning messages:

```
.../framework-arduinoespressif32/cores/esp32/esp32-hal-cpu.c:132:9: note: in expansion of macro 'log_e'
         log_e("not found func=%08X arg=%08X",cb,arg);
         ^~~~~
.../framework-arduinoespressif32/tools/sdk/esp32/include/log/include/esp_log.h:276:27: warning: format '%X' expects argument of type 'unsigned int', but argument 7 has type 'void *' [-Wformat=]
```

I believe the correct should be `%p`, but i preserved the 8-width in my PR.
(cannot really test the code, this fix just stops the compiler warnings).

## Steps to Reproduce

To get those warnings, i had to enable `-Wformat` build-flag and define `DUSE_ESP_IDF_LOG 1`,
so the _esp_idf_ log-routines to apply, that they are appropriately marked with `__attribute__ (( format(...) ))`).

### DIFF
--- a/cores/esp32/Tone.cpp
+++ b/cores/esp32/Tone.cpp
@@ -28,7 +28,7 @@ static void tone_task(void*){
     xQueueReceive(_tone_queue, &tone_msg, portMAX_DELAY);
     switch(tone_msg.tone_cmd){
       case TONE_START:
-        log_d("Task received from queue TONE_START: _pin=%d, frequency=%u Hz, duration=%u ms", tone_msg.pin, tone_msg.frequency, tone_msg.duration);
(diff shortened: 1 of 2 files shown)


ankostis:
Discovered another x2 issues  in `Tone.cpp`, 
- [x] commit appended,
- [x] OP above amended.

me-no-dev:
Thanks @ankostis :)

PR #6731, opened 2022-05-12
Adds DHCP Range Setup to APMode

### LINKED ISSUE #6271: Setting DHCP IP-range start for softAP
### Related area

WiFi softAP DHCP

### Hardware specification

ESP32

When several Wi-Fi stations are connected to an ESP's softAP and the ESP reboots, it can happen that some stations hold on to their old DHCP lease. However, the DHCP server does not know and re-assigns such still-in-use IP addresses. As a workaround for the problem, the DHCP server could use a different IP range after each restart (i.e. cycling through several ranges), but this requires the possibility of configuring the DHCP server.

Possibility of configuring the DHCP server:

*) Adding IPAddress argument to softAPConfig:
bool softAPConfig(IPAddress local_ip, IPAddress gateway, IPAddress subnet, IPAddress dhcpstart);

(can be an optional argument using a default value, or the current 3-argument version softAPConfig(local_ip,gateway,subnet) calls the extended 4-argument version)

*) This requires adding an argument to set_esp_interface_ip:
esp_err_t set_esp_interface_ip(esp_interface_t interface, IPAddress local_ip=IPAddress(), IPAddress gateway=IPAddress(), IPAddress subnet=IPAddress(), IPAddress dhcpstart=IPAddress());

An implementation could look like this:

  dhcps_lease_t lease;
  lease.enable = true;
  lease.start_ip.addr = static_cast<uint32_t>(dhcpstart);
  lease.end_ip.addr   = static_cast<uint32_t>(dhcpstart) + (10 << 24);

instead of
  lease.start_ip.addr = static_cast<uint32_t>(local_ip) +  (1 << 24);
  lease.end_ip.addr   = static_cast<uint32_t>(local_ip) + (11 << 24);
  
This extension would help me (and probably some others) a lot.
Thanks for all your great work,
Christian

_No response_

### Additional context

_No response_

SuGlider:
-----------
## Description of Change
This PR adds an optional parameter to `WiFi.softAPConfig()` in order to set the initial address of the DHCP Leasing Range.
It has no impact over sketches that do not set this parameter because it will work as the default previous way. 

The PR also improves IP verification for errors with subnet, collision with DHCP range and setting IPs out of the subnet range.

## Tests scenarios
Tested on ESP32, ESP32-S2, ESP32-C3 and ESP32-S3.
Tested with different configurations for IPs, including using different net masks.

## Related links
Fixes #6271

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -82,7 +82,7 @@ esp_err_t set_esp_interface_hostname(esp_interface_t interface, const char * hos
 	return ESP_FAIL;
 }
 
-esp_err_t set_esp_interface_ip(esp_interface_t interface, IPAddress local_ip=IPAddress(), IPAddress gateway=IPAddress(), IPAddress subnet=IPAddress()){
+esp_err_t set_esp_interface_ip(esp_interface_t interface, IPAddress local_ip=IPAddress(), IPAddress gateway=IPAddress(), IPAddress subnet=IPAddress(), IPAddress dhcp_lease_start=INADDR_NONE){
 	esp_netif_t *esp_netif = esp_netifs[interface];
 	esp_netif_dhcp_status_t status = ESP_NETIF_DHCP_INIT;
 	esp_netif_ip_info_t info;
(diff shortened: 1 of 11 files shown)


libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -138,19 +138,49 @@ esp_err_t set_esp_interface_ip(esp_interface_t interface, IPAddress local_ip=IPA
...
-            lease.end_ip.addr = dhcp_ipaddr + (11 << 24);
-        } else {
-            // make range stay in the begining of the netmask range
-            dhcp_ipaddr = (dhcp_ipaddr & 0x00FFFFFF);
-            lease.start_ip.addr = dhcp_ipaddr + (1 << 24);
-            lease.end_ip.addr = dhcp_ipaddr + (11 << 24);
+        uint8_t CIDR = WiFiGenericClass::calculateSubnetCIDR(subnet);
+        // netmask must have room for at least 12 IP addresses (AP + GW + 10 DHCP Leasing addresses)
+        // netmask also must be limited to the last 8 bits of IPv4, otherwise this function won't work
+        if (CIDR > 28 || CIDR < 24) {
```
me-no-dev, line 144: why does the netmask have to be the last 8 bits? What is wrong with being 16 bits instead?

mrengineer7777 (commented):
Parameter changes to softAPConfig() look fine.

Can set_esp_interface_ip() be declared as a protected function in WiFiGeneric.h instead of declaring in each class that calls it?

Since this adds functionality rather than a bugfix, should this be targeted for release 2.1.0?

I'll leave it up to the maintainers to review changes in WiFiGeneric.cpp.

libraries/WiFi/src/WiFiGeneric.cpp
SuGlider: In order to make it work correctly for netmask with more than 8 bits, it would be better to invert the byte order when IPAddress is cast to uint32_t.
I'm not sure that changing it would not break the code in other places or for other users.
Given that this issue has been already there since the very beginning, I just added code to prevent it from happening, limiting the netmask to 8 bits.

Example: 
1.2.3.255 goes to 0xFF030201 in the current IPAddress Class casting method.
Adding 1, means to add (1 << 24) and it goes to 0x00030201 = IP 1.2.3.0 (it is worst when adding 11 << 24 to create a DHCP range).
If it were inverted, it would go to 0x010203FF, instead. 
So adding 1 would result in 0x01020400 = IP 1.2.4.0, respectively, adding 11 would result in 0x01020410 = IP1.2.4.10 (a lot better way to make the increment!)

It can be achieved by forcing this byte order inversion/reversion in the WiFiGenericClass code... is it worth it?

SuGlider:
> Can set_esp_interface_ip() be declared as a protected function in WiFiGeneric.h instead of declaring in each class that calls it?

Yes, nice! :-)

> Since this adds functionality rather than a bugfix, should this be targeted for release 2.1.0?

We are reserving breaking changes for the 2.1.0. It doesn't break anything, therefore it is targeted for the 2.0.4

libraries/WiFi/src/WiFiGeneric.cpp
SuGlider: Inverting the bytes also makes it way easier to compare  `(uint32_t) IPAddress` (greater, less than) in order to check if it belongs to an IP range or if it could break netmask limits and so on.

me-no-dev: But inverting it will make it otherwise incompatible with everything else. We need to have the byte order that computers and IDF use. I do not see a reason why not add a method/methods to internally flip the order and do the operations agains another IP :)

SuGlider: OK, I'll add the code for it.

SuGlider: @me-no-dev - I tested it with netmask 255.255.0.0 for this cases:
1- AP = 192.168.1.250 and DHCP Range = 192.168.1.251 to 192.168.2.4
2- AP = 192.168.1.1 and DHCP Range = 192.168.2.1 to 192.168.2.10

In both cases the IDF call `tcpip_adapter_dhcps_option()` fails and returns error 0x5001.
It only works if AP IP address and DHCP range are in the same network with netmask 255.255.255.0 to 255.255.255.240
So, it seems that IDF only works with 8 bits subnet.

Therefore, this PR is ready.

SuGlider: Confirmed by looking into the esp_netif_lwip.c code:
https://github.com/espressif/esp-idf/blob/master/components/esp_netif/lwip/esp_netif_lwip.c#L1857-L1862

it only uses the last 8 bits of the AP address, limiting netmask effect to also 8 bits.

``` cpp
                    /*config ip information must be in the same segment as the local ip*/
                    softap_ip >>= 8;
                    if ((start_ip >> 8 != softap_ip)
                        || (end_ip >> 8 != softap_ip)) {
                        return ESP_ERR_ESP_NETIF_INVALID_PARAMS;
                    }
```

SuGlider: This PR is ready for merging.

SuGlider (commented):
All necessary changes are done.

libraries/WiFi/src/WiFiGeneric.cpp
SuGlider: I also found out an IDF limit for the size of DHCP range: DHCPS_MAX_LEASE, default is 0x64 = 100 addresses (lwip/include/apps/dhcpserver/dhcpserver.h)

SuGlider: Commited changes that prepare the code for any netmask (for the future, when IDF supports it as well).

PR #6752, opened 2022-05-16
Fillbuffer Fix In WiFiClient::read

podaen:
## Description of Change
In rare occasions when condition pos==fill is false in WiFiClient buffer.read, the fillbuffer is not called. The AND can never can be true and the buffer will not be filled. By place it on top, it will fill the buffer first before it continues.

## Test senario
I have tested my Pull Request on Arduino-esp32 core v2.0.2 with ESP32 and ESP32-S2 Board with this scenario.

Kind regards
DP

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -109,12 +109,11 @@ public:
     }
 
     int read(uint8_t * dst, size_t len){
-        size_t fb = fillBuffer();
-        if(!dst || !len || (_pos == _fill && !fb)){
+        if(!dst || !len || (_pos == _fill && !fillBuffer())){
             return _failed ? -1 : 0;
         }
         size_t a = _fill - _pos;
-        if(len <= a || ((len - a) <= (_size - _fill) && fb >= (len - a))){
+        if(len <= a || ((len - a) <= (_size - _fill) && fillBuffer() >= (len - a))){
             if(len == 1){
                 *dst = _buffer[_pos];
             } else {
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -109,11 +109,12 @@ public:
     }
 
     int read(uint8_t * dst, size_t len){
-        if(!dst || !len || (_pos == _fill && !fillBuffer())){
+        size_t fb = fillBuffer();
+        if(!dst || !len || (_pos == _fill && !fb)){
             return _failed ? -1 : 0;
         }
         size_t a = _fill - _pos;
-        if(len <= a || ((len - a) <= (_size - _fill) && fillBuffer() >= (len - a))){
+        if(len <= a || ((len - a) <= (_size - _fill) && fb >= (len - a))){
             if(len == 1){
                 *dst = _buffer[_pos];
             } else {
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -18,7 +18,7 @@ jobs:
   gen_chunks:
     if: |
       contains(github.event.pull_request.labels.*.name, 'hil_test') ||
-      github.event_name == 'schedule'
+      (github.event_name == 'schedule' && github.repository == 'espressif/arduino-esp32')
     name: Generate Chunks matrix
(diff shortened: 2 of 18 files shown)


libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -109,11 +109,12 @@ class WiFiClientRxBuffer {
...
 
     int read(uint8_t * dst, size_t len){
-        if(!dst || !len || (_pos == _fill && !fillBuffer())){
+        size_t fb = fillBuffer();
+        if(!dst || !len || (_pos == _fill && !fb)){
             return _failed ? -1 : 0;
         }
         size_t a = _fill - _pos;
-        if(len <= a || ((len - a) <= (_size - _fill) && fillBuffer() >= (len - a))){
+        if(len <= a || ((len - a) <= (_size - _fill) && fb >= (len - a))){
```
sweetlilmre, line 117: The original code would call fillBuffer() at line 112 and 116. Your change now only calls the function once.
Should there not be a call `fb = fillBuffer();` before the check at line 117?

podaen: It doesn't do any difference. (len - a) will tell us that there is engough room to copy it in to the allocated memory. So I don't see the advatage why call it back. a is allready been determed and by calling fill buffer a will not be up to date anymore.

sweetlilmre: I see what you mean. If the call to `fillBuffer()` is sufficient to fill the request then there is no need to call it again.
If not, the subsequent calls in the while loop will either sort this out or exit if not enough data is available.

So then, in the original case for your PR: If `pos != fill` then `(_pos == _fill && !fillBuffer())` will short circuit and `fillBuffer()` will not be called. As `pos != fill`, there is data available. If there is enough data in the buffer it will be copied out and returned to the caller. If not, the code will call `fillBuffer()` in the while loop. 

What is the condition that causes the failure?

podaen: > I see what you mean. If the call to `fillBuffer()` is sufficient to fill the request then there is no need to call it again.

You can't call fillbuffer engough. There just need to be space to store.

The problem was that _pos != _fill , so evryting was done. and therefore diden't fill the buffer but still there was engough room to store.

> If `pos != fill` then `(_pos == _fill && !fillBuffer())` will short circuit and `fillBuffer()` will not be called.

Exacly

> As `pos != fill`, there is data available to copy in the buffer but it doesn't continue

If pos!=fill it doesn't mean there is no more space in the buffer. it will never overflow the buffer.

podaen: I don't have the log anymore, but I think it was after checking left and that go over again to raid.

```
while(left){
            if(!fillBuffer()){
                return len - left;
            }
            a = _fill - _pos;
            toRead = (a > left)?left:a;
            memcpy(buf, _buffer + _pos, toRead);
            _pos += toRead;
            left -= toRead;
            buf += toRead;
        }
        return len;
```

podaen: I checked evry rare situation where it could been in. I can't find any where it can't handle it. Maybe I am mistaken.

Things I do remember is 
len=512
_size = _fil. 
slow stream and sometimes fast.
buffer didn't fill

sweetlilmre: This is what was confusing me about the code :) it **looks** like your issue shouldn't happen with sufficient calls to `read()` i.e. either you'll have enough bytes or the next `fillBuffer()` call in the while loop will work or eventually a packet will arrive on the wire and then you can proceed. 

That said if calling `fillBuffer()` has no side effects then your change would have no negative impact. The best bet I think would be to try and replicate.

podaen: I did find an issue, I just followed the code and I think I diden't understand it than well. What you said is something verry importent. Sufficient reads is importent to make it work good.I did some more test, nothing found. I will role back my changes. Thanks for your  invested time.

sweetlilmre: No problem, glad to help out :) Can you close this PR so that it's off the list?

PR #6770, opened 2022-05-18
update adafruit boards tinyuf2 and partition scheme

hathach:
## Description of Change
- Follow up to #6668, this PR update partition scheme for Adafruit Boards. ~It address one of the walk-around @UnexpectedMaker need to do is renaming the tinyuf2.bin/bootloader.bin in the variant folder.~
- Also take this chance to update tinyuf2 for Adafruit from 0.6.0 to 0.9.2 version
- Fix ssue with atmegazero_esp32s2 board (also use tinyuf2), which has 16MB flash on board but incorrectly use 4MB partition.csv. Not sure if I am overstepping here, if yes, I could revert changes to this board.

## Tests scenarios

- Arduino IDE 1.8.19 on Ubuntu 22.04
- arduino-esp32 core at v2.0.4 commmit ed33e157529082fd8942b7bd9561a9bea9cbe703
- Adafruit Boards: Feather S2, Feather S2 TFT, Metro S2, Funhouse, Magtag, qtpy S2, S3 devkitC (as prototyping for Adafruti Feather S3).

## Related links
- #6668
- https://github.com/adafruit/tinyuf2/releases/tag/0.9.2

@ladyada


platform.txt
```diff
@@ -127,8 +127,8 @@ build.flash_freq=80m
 build.boot=qio
 build.boot_freq={build.flash_freq}
 build.bootloader_addr=0x1000
-build.custom_bootloader=bootloader
-build.custom_partitions=partitions
+build.custom_bootloader=
+build.custom_partitions=
```
prplz, line 131: This makes the build search for `variant/.bin`, expanded from `"{build.variant.path}"/{build.custom_bootloader}.bin`. Maybe it should be left as is?

hathach: yeah, this is meant to expanded to non-existed files by default in the variant/board folder so that one of the espressif default parition csv can be used if an TinyUF2 scheme isn't selected.

me-no-dev: it will still expand to non-existing if you name yours to something else. Default values are there to allow variants to add bootloader.bin and partitions.csv and be done, without having to specify anything in boards.txt. Please rename your bootloaders to something else and the defaults here will work as expected :)

hathach: update PR to revert this changes and rename the bootloader/partition in the variant folder as suggested.

me-no-dev:
Thanks @hathach

hathach:
> Thanks @hathach

I am the one to say thanks :)

PR #6798, opened 2022-05-25
Remove nonexisting folder from include path

maxgerhardt:
## Description of Change

The PlatformIO builder scripts attempt to include `tools/sdk/esp32/include/config` to the include path (and similiarly for the other MCU types), but this folder does not exist anymore.

VSCode emits a "Problem" warning for non-existing folders declared in the therefrom created `.vscode/c_cpp_properties.json`.

grafik

The fix trivially removes these folders from the include path. If the `include/config` folder was renamed and is not yet in the PlatformIO builder script, let me know.

## Tests scenarios
Tested with PlatformIO using

```ini
[env:esp32dev]
platform = espressif32
board = esp32dev
framework = arduino
platform_packages = 
   framework-arduinoespressif32@https://github.com/maxgerhardt/arduino-esp32.git#patch-2
```

and the "Problems" view cleared out.

grafik

## Related links
None.

### DIFF
--- a/tools/platformio-build-esp32.py
+++ b/tools/platformio-build-esp32.py
@@ -117,7 +117,6 @@ env.Append(
     ],
 
     CPPPATH=[
-        join(FRAMEWORK_DIR, "tools", "sdk", "esp32", "include", "config"),
         join(FRAMEWORK_DIR, "tools", "sdk", "esp32", "include", "newlib", "platform_include"),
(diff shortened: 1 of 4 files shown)


me-no-dev:
This actually needs fixing in the lib-builder

me-no-dev:
OK. Thanks :) Merging here and already fixed in the lib-builder for the future.

Issue #4941, opened 2021-03-13
analogReadMilliVolts returns wrong values on ESP32-S2


newHeiko:
### Hardware:
Board:							ESP32-S2-WROOM
Core Installation version:			idf-release/v4.2 (current HEAD: beedeea)
IDE name:						Arduino IDE
Flash Frequency:					40Mhz
PSRAM enabled:                                    no
Upload Speed:					921600 via USB
Computer OS:                                        Linux

### Description:

#4386 is still open in current HEAD of idf-release/v4.2 branch.

I have my own PCB with an ESP32-S2-WROOM and a 10k-potentiometer connected to pin 9. My multimeter shows ~990mV and the ESP32-S2 will read between 1411mV and 1418mV.

When I change the potentiometer, the readout changes, with a roughly constant factor between the measurement and the readout.

Please let me know if there is anything I can do to debug / test.

Heiko

### Sketch:
```cpp
void setup() {
  Serial.begin(115200);
  delay(5000);
  uint32_t analogValue = analogReadMilliVolts(9);
  Serial.println(String("Read ") + analogValue + " mV at time " + millis());
  delay(5000);
  analogValue = analogReadMilliVolts(9);
  Serial.println(String("Read ") + analogValue + " mV at time " + millis());
  delay(5000);
  analogValue = analogReadMilliVolts(9);
  Serial.println(String("Read ") + analogValue + " mV at time " + millis());
}

void loop() {
}
```
### Debug Messages:
```
ESP-ROM:esp32s2-rc4-20191025
Build:Oct 25 2019
rst:0x1 (POWERON),boot:0xa (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:2
load:0x3ffe6100,len:0x8
load:0x3ffe6108,len:0x620
load:0x4004c000,len:0xa40
load:0x40050000,len:0x284c
entry 0x4004c190
[   648][D][esp32-hal-cpu.c:204] setCpuFrequencyMhz(): PLL: 480 / 3 = 160 Mhz, APB: 80000000 Hz
Read 1415 mV at time 5661
Read 1413 mV at time 10661
Read 1415 mV at time 15661
```

lbernstone:
Can you turn on verbose debugging and post the relevant section, particularly about the vref calibration.

newHeiko:
[...]

PR #6799, opened 2022-05-25
ADC fixes and improvements

P-R-O-C-H-Y:
## Description of Change

- Fixed channel / unit selection for ESP32C3 (has only 5 channels on 1 adc unit).
- Removed unnecessary includes.
- Fixed different attenuation on channels to be taken in care in __analogReadMilliVolt.
- All chips in __analogReadMilliVolts now use adc characteristics + calculation from raw to voltage as ESP32 does.

## Tests scenarios
Tested on all chips.

## Related links
#Closes 6775

### DIFF
--- a/cores/esp32/esp32-hal-adc.c
+++ b/cores/esp32/esp32-hal-adc.c
@@ -13,49 +13,27 @@
 // limitations under the License.
 
 #include "esp32-hal-adc.h"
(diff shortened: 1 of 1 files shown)


newHeiko:
JFTR: This (probably) also fixes #4941

Heiko

PR #6802, opened 2022-05-25
Rainmaker: Added missing standard ui types and standard parameter types and functions

sivar2311:
## Description of Change
The Arduino implementation is missing some standard UI types and standard parameter types.
Also, the `addValidStrList` function is missing to be able to implement a mode controller parameter.
Additionally, a new overload of the `value` function with `const char*` parameter type has been added.

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.3 with ESP32 

## Related links
Missing types: [Standard Predefined Types](https://rainmaker.espressif.com/docs/standard-types.html)
Missing function: `addValidStrList`: [3rd Party Integrations](https://rainmaker.espressif.com/docs/3rd-party.html) example code for "Generic Mode Parameter"

### DIFF
--- a/libraries/RainMaker/src/RMakerParam.cpp
+++ b/libraries/RainMaker/src/RMakerParam.cpp
@@ -30,4 +30,13 @@ esp_err_t Param::updateAndReport(param_val_t val)
     }
     return err;
 }
+
+esp_err_t Param::addValidStrList(const char** str_list)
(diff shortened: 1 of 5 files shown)


sivar2311:
Sorry, this was my first attempt. I will do another one.

PR #6808, opened 2022-05-26
Implement simple RGB driver via digitalWrite; solving #6783

### LINKED ISSUE #6783: Implement LED_BUILTIN constant + digitalWrite() overload for RGB LED
### Related area

RGB LED, Blink example

### Hardware specification

C3, S3, dev boards with RGB LED

No issue found

Implement `LED_BUILTIN` constant for dev boards with RGB LED. Where not applicable print a helpful error message - for example, `This board does not have builtin LED` instead of default `error: 'LED_BUILTIN' was not declared in this scope`
For boards with RGB LED overload function `void digitalWrite(uint8_t pin, uint8_t val)` to activate the RGB LED via the appropriate driver.
Sample use of overloaded function:
```
digitalWrite(LED_BUILTIN, HIGH); // RGB turns full white
digitalWrite(LED_BUILTIN, LOW); // RGB turns off

// Nice to have (1) - ability to control each channel
digitalWrite(RED_LED_BUILTIN, HIGH); // Red channel full brightness
digitalWrite(GREEN_LED_BUILTIN, HIGH); // Green channel full brightness
digitalWrite(BLUE_LED_BUILTIN, HIGH); // Blue channel full brightness
// Similar for LOW -> channel is off

// Nice to have (2) - ability to control brightness
digitalWrite(RED_LED_BUILTIN, 128); // Red channel turns on to 50% brightness
```

_No response_

### Additional context

_No response_

PilnyTomas:
## Description of Change
This solution enables users to use Blink example with RGB LED on board.
When using function `digitalWrite` with constant `LED_BUILTIN` as pin parameter an RGB LED driver in the background will set white / off.
Added new example BlinkRGB extending the original Blink example with a demonstration of separate RGB channel setup.

## Tests scenarios
Tested on
S3 DevkitC
C3 DevkitC
S2 Kaluga

## Related links
Closes #6783


cores/esp32/esp32-hal-gpio.c
```diff
@@ -91,6 +91,13 @@ static InterruptHandle_t __pinInterruptHandlers[SOC_GPIO_PIN_COUNT] = {0,};
 
 extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
 {
+#ifdef BOARD_HAS_NEOPIXEL
```
igrr, line 94: Could we consider moving the board-specific code to the `variants` folder?

I'm worried that the more similar features we add for specific boards, the more complex the core code will become. It would be better to handle this in the board variant folder. It would also demonstrate authors of 3rd party board packages how to add such custom functionality.

To avoid duplication of code, you can still define `RGBLedWrite` somewhere in the core, but preferably in a separate source file (esp32-hal-rgb-led.c?)

PilnyTomas: I'm a little bit confused about what should go where.
`pinMode` -> `variants/`  ... new file?
`RGBLedWrite` -> `esp32-hal-rgb-led.c` ?

PilnyTomas: Should be solved in https://github.com/espressif/arduino-esp32/pull/6808/commits/583e0301427b6e98a87c3fa94fa100d2fc19cb7f

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -0,0 +1,22 @@
...
+#endif
+
+#include "esp32-hal.h"
+/*
+#include "soc/soc_caps.h"
+#include "pins_arduino.h"
+*/
+
+//#ifdef BOARD_HAS_NEOPIXEL
+  void RGBLedWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val);
```
me-no-dev, line 15: please cleanup the commented includes and ifdefs. Arduino starts function names with lower case. I would suggest that you rename the function to `neopixelWrite`. `rgbLed` is very general and should be attributed to actual RGB Led with separate leads for each color (and controlled through PWM).

cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -0,0 +1,51 @@
+#include "esp32-hal-rgb-led.h"
+
+#ifdef BOARD_HAS_NEOPIXEL
```
me-no-dev, line 3: seems you have the guard here but it is commented in the header?

cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -0,0 +1,51 @@
...
+
+void RGBLedWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val){
+  log_d("RGB: %d %d %d", red_val, green_val, blue_val);
+  rmt_data_t led_data[24];
+  static rmt_obj_t* rmt_send = NULL;
+  static bool initialized = false;
+
+  uint8_t _pin;
+  if(pin == LED_BUILTIN){
```
me-no-dev, line 12: this should be evaluated only if the board has neopixel, else LED_BUILTIN would be lower than SOC_GPIO_PIN_COUNT

cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -0,0 +1,51 @@
+#include "esp32-hal-rgb-led.h"
+
+#ifdef BOARD_HAS_NEOPIXEL
+
+void RGBLedWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val){
+  log_d("RGB: %d %d %d", red_val, green_val, blue_val);
```
me-no-dev, line 6: such logs should be with level "Verbose" instead. Blinking led will print a lot of things :)

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -0,0 +1,22 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "esp32-hal.h"
+/*
+#include "soc/soc_caps.h"
+#include "pins_arduino.h"
+*/
+
```
me-no-dev, line 13: You should have safeguard against LED_BRIGHTNESS not being defined:
```cpp
#ifndef LED_BRIGHTNESS
#define LED_BRIGHTNESS 64
#endif
```

variants/esp32s2/variant.cpp
```diff
@@ -0,0 +1,13 @@
+#include "esp32-hal-rgb-led.h"
+
+extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
+{
+  log_d("foo");
```
me-no-dev, line 5: forgotten?

variants/esp32c3/variant.cpp
```diff
@@ -0,0 +1,13 @@
+#include "esp32-hal-rgb-led.h"
+
+extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
+{
+  log_d("foo");
```
me-no-dev, line 5: forgotten?

variants/esp32c3/variant.cpp
```diff
@@ -0,0 +1,13 @@
+#include "esp32-hal-rgb-led.h"
+
+extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
```
me-no-dev, line 3: why is this necessary? pinMode in the core should just return if pin is not valid.

cores/esp32/esp32-hal-rgb-led.c
PilnyTomas: this was also a forgotten debug output. I will remove it in the next commit.

cores/esp32/esp32-hal-rgb-led.c
PilnyTomas: The entire function was safeguarded `#ifdef BOARD_HAS_NEOPIXEL`
But I will remove safeguard for the entire function a wrap only this.

variants/esp32c3/variant.cpp
PilnyTomas: I was trying to implement this https://github.com/espressif/arduino-esp32/pull/6808#pullrequestreview-986096611

me-no-dev: I understand, I just question the necessity to have such code in the variant just to get something as common as neopixel on the board working.
@igrr any comments?

Something like this in the core's `__pinMode` would skip execution:
```cpp
if(pin > SOC_GPIO_PIN_COUNT && pin == LED_BUILTIN){
    return; // or pin -= SOC_GPIO_PIN_COUNT; to allow setting the mode.
}
```

me-no-dev: @PilnyTomas this is still outstanding. Requiring variant.cpp to blink a led is not desired situation.

PilnyTomas: @igrr Could you please give us more info about why should we have board-specific code in the variants folder instead of having in the common file?
Should I keep it like this, or revert back to the common file?

igrr: I was thinking that this is behavior we want only on specific boards which have a WS2812B, and where we'd like to provide such emulation.

Makers of other boards should be able implement some different logic, or opt out of such emulation. In the latter case, the code for emulating an LED ideally shouldn't be compiled or linked.

me-no-dev: if you do not have that LED defined, code will not be compiled anyway. requiring all boards with RGB leds to include a variant cpp for two lines is a bit too much IMO. There are plenty boards with RGB leds

igrr: Although, given that you are introducing a BOARD_HAS_NEOPIXEL macro, maybe we can indeed move this functionality into the core.

PR #6817, opened 2022-05-27
Improve support for hand-written asm source files

ivankravets:
Improve support for hand-written asm source files and document flag scopes

### DIFF
--- a/tools/platformio-build-esp32.py
+++ b/tools/platformio-build-esp32.py
@@ -36,7 +36,11 @@ assert isdir(FRAMEWORK_DIR)
 
 env.Append(
     ASFLAGS=[
-        "-x", "assembler-with-cpp"
+        "-mlongcalls",
+    ],
+    
+    ASPPFLAGS=[
+        "-x", "assembler-with-cpp",
     ],
 
     CFLAGS=[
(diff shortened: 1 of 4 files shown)


me-no-dev:
these parts of the files are automatically generated by the lib-builder. Values are extracted from IDF's build system.

ivankravets:
How about with `ASFLAGS` and `ASPPFLAGS`? There is a bug in the current scripts with `"-x", "assembler-with-cpp"` which is set for `ASFLAGS` instead of `ASPPFLAGS`.

me-no-dev:
[This](https://github.com/espressif/esp32-arduino-lib-builder/blob/master/tools/copy-libs.sh#L247-L313) is the part of the script that generates your configs. Above it is how they are obtained. I suggest you look there for resolution :)

ivankravets:
Could you adapt my fixes for the ASM? I'm not so good in the Bash. I'll close this PR. Thanks.

me-no-dev:
@ivankravets will the change for "ASPPFLAGS" be enough? Can I just rename "ASFLAGS" to "ASPPFLAGS"?

ivankravets:
If you don't use bare ASM (without C preprocessor, small `.s`), then just rename `ASFLAGS` -> `ASPPFLAGS`.

The `ASFLAGS` are common for `.s` and `.S`, whereas `ASPPFLAGS` only for `.S`. The `.s` very often require machine-related flags. However, it seems they are not required for ESP toolchains.

me-no-dev:
Will be part of 2.0.4 :)

PR #6844, opened 2022-06-08
WireMaster example clarity

Luuccc:
- Created temporary variable, removing confusion with the reuse of "error". Where Wire.requestFrom() doesn't return an error itself.
- Added a cast to help clarify when and why bytes are being read.


me-no-dev:
Hi @Luuccc please sign the CLA, so that we can merge your changes

PR #6887, opened 2022-06-20
Refactor PlatformIO build scripts

valeros:
This pull request includes the following two major improvements: 

- Refactored structure of the build scripts. Currently, the build scripts for each ESP32 family include a lot of duplicated code. Updating this code in four different places is tedious and error-prone. In this PR, the common code is moved to the generic `platformio-build.py` script.
- Updated process of searching for a proper partition table and bootloader image. It now behaves similarly to the Arduino IDE by searching in the variant directory in the first place.

I believe this PR shouldn't affect the autogeneration process implemented in the `esp32-arduino-lib-builder` script as only the common parts of the code were moved.


me-no-dev:
merged! Will apply this to the lib-builder and see the result later :)

PR #6907, opened 2022-06-24
[WiFi] Clarify arduino wifi debug output

Rotzbua:
## Description of Change

`PSK` can concern different ciphers. Make it clear that it belongs to `WPA`.

Renamed output `PSK` to `WPA_PSK`.

## Tests scenarios

Enable debugging and check output.

## Related links

None.


me-no-dev:
Makes sense :) thanks!

Issue #6832, opened 2022-06-03
The HiFreq_ADC Example Code cannot run.


Globefishp:
### Board

WEMOS LOLIN32 LITE

### Device Description

just on breadboard.

### Hardware Configuration

a Potentiometer is connected to my broad and used as Analogue input.

### Version

v2.0.3

### IDE Name

Arduino IDE

### Operating System

Windows 10 1809

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

The example code <HiFreq_ADC.ino> can compile and upload to board, but Serial output says it is continuously rebooting. 

Here is my serial output, the backtrace tool cannot decode that(probably my usage is wrong, i just paste that to the tool.)
```
Rebooting...
ets Jun  8 2016 00:22:57

rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1184
load:0x40078000,len:12812
load:0x40080400,len:3032
entry 0x400805e4
Output frequency: 78125

assert failed: xQueueGenericSend queue.c:832 (pxQueue->pcHead != ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == ((void *)0) || pxQueue->u.xSemaphore.xMutexHolder == xTaskGetCurrentTaskHandle())


Backtrace:0x4008306d:0x3ffb2dc00x40087ebd:0x3ffb2de0 0x4008cd1d:0x3ffb2e00 0x400886b2:0x3ffb2f30 0x4008465a:0x3ffb2f70 0x40084715:0x3ffb2fa0 0x400d6f1e:0x3ffb2fc0 0x400d2e42:0x3ffb2ff0 0x400d131f:0x3ffb3020 




ELF file SHA256: 0000000000000000
```

Beside, when compile the code, there is a warning says:
```
C:\Users\Globefish\AppData\Local\Arduino15\packages\esp32\hardware\esp32\2.0.2\libraries\ESP32\examples\I2S\HiFreq_ADC\HiFreq_ADC.ino: In function 'void i2sInit()':
[...]

PR #6917, opened 2022-06-27
WIP; Reworked HiFreq_ADC example; Closing #6832

PilnyTomas:
## Description of Change
Original code does not work (crashing) - related issue #6832 - now fixed.
Extended description to be more helpful.
Added options to modify the setup at the top of the code via constants.
Added option do plot with Arduino Serial plotter.
Even if the crashing was solved alone the ledc PWM would not output any signal.

## Tests scenarios
Tested on ESP32 with oscilloscope and signal generator.

## Related links
Closing #6832


PilnyTomas:
> A comment at the top that it is esp32 only will probably prevent some issues.

I don't understand, what do you mean?

me-no-dev:
@PilnyTomas can this example be based on our new Arduino I2S lib?

PilnyTomas:
> @PilnyTomas can this example be based on our new Arduino I2S lib?

It can, but our Arduino I2S is using an extra set of buffers and I'm afraid it could have a bad influence on speed.

PR #6964, opened 2022-07-10
Make USBHIDKeyboard work at boot

RefactorFactory:
Add a boot_protocol parameter to the USBHIDKeyboard constructor. When true:

1. The USB interface descriptor will have an interface subclass of boot and
   and interface protocol of keyboard. This will cause some PC BIOS to send a
   a SET PROTOCOL BOOT request to the device.

2. The USB report descriptor will not have a report ID because boot protocol
   does not use report IDs.

3. When the host sends reports to the device, USBHIDKeyboard will accept a
   report ID of 0 instead of checking for HID_REPORT_ID_KEYBOARD(1).

4. When the device sends reports to the host, it will not send a report ID
   because boot protocol does not use report IDs.

5. Use endpoint address of 1 for input and output because some PC BIOS require
   it.

### Checklist

-----------
## Description of Change

USBHIDKeyboard didn't work at boot on my devices, so I've prototyped changes that make it work. I'm looking for feedback.

This adds a parameter to the USBHIDKeyboard constructor and when it is true, behavior is changed. I considered removing the parameter and changing the behavior all the time, but it seemed too risky.

This also adds a new parameter to the USBHID constructor, but it defaults to the old behavior.

I considered only using the new behavior if the host sends the SET PROTOCOL BOOT request, but it turns out that some BIOS do not send that request.

It is strange that some PC BIOS require specific endpoint addresses of 1 for input and output. I tried to do the reservation of the endpoint addresses similar to how USBCDC does it.

Some of the code seems ugly, so I'm looking for feedback.

## Tests scenarios

I tested this with Arduino-esp32 core v2.0.4 with an ESP32 S2 board, with one old laptop and one new desktop, both from major PC manufacturers.


cores/esp32/esp32-hal-tinyusb.c
```diff
@@ -661,6 +665,13 @@ esp_err_t tinyusb_enable_interface(tinyusb_interface_t interface, uint16_t descr
         log_e("Interface %s invalid or already enabled", (interface >= USB_INTERFACE_MAX)?"":tinyusb_interface_names[interface]);
         return ESP_FAIL;
     }
+    if(interface == USB_INTERFACE_HID && reserve_endpoints){
+        // Some simple PC BIOS requires specific endpoint addresses for keyboard at boot
+        if(!tinyusb_reserve_out_endpoint(1) ||!tinyusb_reserve_in_endpoint(1)){
```
RefactorFactory, line 670: Without this reservation of endpoints IN 1, OUT 1, USBCDC will reserve OUT 3, IN 4, IN 5 (line 665), then when `USBHID.cpp:tusb_hid_load_descriptor()` calls `tinyusb_get_free_in_endpoint()`/`tinyusb_get_free_out_endpoint()`, those functions will return IN 3, OUT 4 because *those functions prefer returning endpoints whose opposite is already used*.

Some alternatives to calling `tinyusb_reserve_out_endpoint()`/`tinyusb_reserve_in_endpoint()` here to reserve IN 1, OUT 1:

1. `USBHID.cpp:tusb_hid_load_descriptor()` could call `tinyusb_get_free_duplex_endpoint()` which would likely get endpoints IN 1, OUT 1 because `USBHID.cpp:tusb_hid_load_descriptor()` runs early enough.
2. Change `tinyusb_get_free_in_endpoint()`/`tinyusb_get_free_out_endpoint()` not to prefer returning endpoints whose opposite is already used. Or add a flag parameter.

RefactorFactory:
Because USBHIDKeyboard is simple, these changes could be greatly simplified by doing some or all of the following:

1. Remove all the new flags and new parameters.
2. In the USB interface descriptor, **always** use an interface subclass of boot protocol.
3. **Never** use a report ID in the USB report descriptor or when receiving or sending reports.
4. **Always** use endpoints 1 IN, 1 OUT for USBHIDKeyboard.

I'm looking for feedback on whether some of these simplifications should be done.

RefactorFactory:
Any feedback on what I can do to get these changes in? I'm open to suggestions. Thanks.

RefactorFactory:
This has a problem on one of my PCs at boot where the PC does not send "SET PROTOCOL BOOT" to the device and it expects the device to send reports prefixed with a report ID. I'll investigate further.

me-no-dev:
USB Keyboards actually provide two HID devices, one for boot and another for all other cases. The non-boot device can have extra functions, buttons, etc.
I know tinyusb can have two devices, so looking into implementing that instead, seems like the better option here :)

RefactorFactory:
Ok, I've reworked this with some improvements.

I investigated a real keyboard from a major PC manufacturer and found that it makes the main HID device a boot keyboard. The real keyboard that I investigated has another HID device, but it is for special keys. So, like a real keyboard, I made `USBHIDKeyboard` use a descriptor that makes it a boot device, but it doesn't speak boot protocol (namely, omitting report IDs from reports) unless the host requests it.

This major PC manufacturer of a 2021 PC (which happens to not request boot protocol) needed two other work-arounds:

1. Use the endpoints 1 IN and 1 OUT. Without this, the device would never become ready at boot. Maybe there is some other deeper problem going on, but this work-around worked.
2. For reports using the shift key, break out the shift key state changes into separate reports. Without this, shifted characters would not be recognized at all (not even as non-shifted). I found this hard to believe, but this work-around worked. Because this work-around may not be necessary for everyone, I added an API, `USBHIDKeyboard::setShiftKeyReports()`, that can be called to disable this behavior, but left the default behavior enabled so that users don't have to spend hours rediscovering this problem.

Let me know any feedback, thanks.

RefactorFactory:
Any feedback on what I can do to get these changes in? I'm open to suggestions. Thanks.

RefactorFactory:
Any feedback on what I can do to get these changes in? I'm open to suggestions. Thanks.

me-no-dev:
sorry for the delay @RefactorFactory . We will look into this very soon and have it in what will become version 3.0.0 (based on IDF 5.x)

VojtechBartoska:
Hello @RefactorFactory, please sign CLA. Thanks

P-R-O-C-H-Y (approved):
Looks good. Can you please sign the CLA as requested? 
Thanks @RefactorFactory for the PR!

lucasssvaz (approved):
Works as expected. Tested on my ASRock motherboard (that didn't work without this PR).

RefactorFactory:
> Hello @RefactorFactory, please sign CLA. Thanks

> Looks good. Can you please sign the CLA as requested? Thanks @RefactorFactory for the PR!

I signed the CLA. Sorry it took so long. Thanks.

PR #6969, opened 2022-07-11
allow HTTPClient to handle large payloads, in case the client (such as WifiClientSecure) does not

r-downing:
## Description of Change

WiFiClient handles large write() by breaking it down and sending the whole thing in pieces, but WiFiClientSecure may just do a partial send and return the partial size, causing this to fail and return HTTPC_ERROR_SEND_PAYLOAD_FAILED

## Tests scenarios

I have tested this successfully on an ESP32-S2 by posting a large jpeg file to an https server successfully.

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -618,8 +618,13 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
         }
 
         // send Payload if needed
-        if(payload && size > 0) {
-            if(_client->write(&payload[0], size) != size) {
+        if(payload && (size > 0)) {
+            size_t totalSent = 0;
+            while(_client->connected() && (totalSent < size))
+            {
+                totalSent += _client->write(&payload[totalSent], (size - totalSent));
+            }
+            if(totalSent != size) {
                 return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
             }
         }


libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -618,8 +618,13 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
         }
 
         // send Payload if needed
-        if(payload && size > 0) {
-            if(_client->write(&payload[0], size) != size) {
+        if(payload && (size > 0)) {
+            size_t totalSent = 0;
+            while(_client->connected() && (totalSent < size))
+            {
+                totalSent += _client->write(&payload[totalSent], (size - totalSent));
```
me-no-dev, line 625: if `write` returns `0` and client is still connected, that would mean that the socket is not ready to be written to. I suggest you catch this situation and delay a bit before retrying. Currently the code will try in a busy loop and on single core chips (or when code runs on the same core and WiFi) this is not a good idea.

PilnyTomas:
@r-downing thank you for your contribution!
Could you please provide a simple example demonstrating the fixed and unfixed state?

libraries/HTTPClient/src/HTTPClient.cpp
r-downing: Something like this?
```suggestion
                const size_t thisSend = _client->write(&payload[totalSent], (size - totalSent));
                totalSent += thisSend;
                if(thisSend == 0) {
                    delay(10);
                }
```

Should we add a delay? Or just bail out if `write` returns 0? I can't remember, but maybe I thought `client->write` was blocking?

r-downing:
> @r-downing thank you for your contribution! Could you please provide a simple example demonstrating the fixed and unfixed state?

I don't currently have access to the code or hardware I was working on when I encountered the issue originally, but I could try to generate a quick example if needed.

Basically in any case where you send a large payload (1000's of bytes) with HTTPClient and WifiClientSecure it will get cut off.

PilnyTomas:
> if needed

Yes, please.

me-no-dev:
Closing because already implemented here: https://github.com/espressif/arduino-esp32/pull/8006

PR #6990, opened 2022-07-15
Added #include <vector> into BLEAdvertisedDevice.h

PilnyTomas:
## Description of Change
Added `#include <vector>` into `BLEAdvertisedDevice.h`
On Windows when including `BLEAdvertisedDevice.h` there was an error `'vector' in namespace 'std' does not name a template type`

## Tests scenarios
Tested compilation of almost empty/default sketch, only added `#include <BLEAdvertisedDevice.h>` on Windows in Arduino IDE 1.8.19
Without the fix appear error mentioned [here](https://github.com/espressif/arduino-esp32/issues/6593#issuecomment-1179544904).
With the fix successfully compiles.
## Related links
Partially related to #6593 (original problem is different)


SuGlider (approved):
LGTM

PR #6992, opened 2022-07-15
Allow access to the PWM channel assigned to the Pin. Opportunity to change PWM configurations

VasilyRakche:
## Description of Change
In current code, after calling analogWrite(), pin is automatically assigned a channel. It is also configured. But:

- There is **no way to change the channel configurations** since one does not have access to the channel assigned to the pin - [arduino-esp32/cores/esp32/esp32-hal-ledc.c:212](https://github.com/VasilyRakche/arduino-esp32/blob/cac3e8b3e35229979cc27cb1914efd1f5ffaf120/cores/esp32/esp32-hal-ledc.c#L212)
- Of course one can just use bare *ledc* functions, but there is no way to make it compatible with analogWrite(). If user calls the analogWrite() until it reaches the channel limit, original *ledc* configuration will be overwritten

### Example. Better compatibility between *ledc* and native Arduino
```
#define PIN 20

// simple usage
analogWrite(PIN, 0);

// advanced user configuration update
channel = analogGetChannel(PIN);
int new_freq = 20000;
int new_bit_num = 9;
ledcSetup(channel, new_freq, new_bit_num);  

```


cores/esp32/esp32-hal-ledc.c
```diff
@@ -226,3 +226,7 @@ void analogWrite(uint8_t pin, int value) {
     ledcWrite(pin_to_channel[pin] - 1, value);
   }
 }
+
+int8_t analogGetChannel(uint8_t pin) {
+    return pin_to_channel[pin] - 1;
+}
```
P-R-O-C-H-Y, line 232: Can you fix the missing new line at the end of the file please?

P-R-O-C-H-Y (commented):
Thanks @VasilyRakche for the PR, seems reasonable to have this option availible :)
Just fix the 1 comment I have, else its good and we can merge 👍

cores/esp32/esp32-hal-ledc.c
VasilyRakche: Please check if its ok now :)

P-R-O-C-H-Y: Looks good now :)

PilnyTomas (approved):
LGTM

PR #7004, opened 2022-07-18
[WIFI] ifdef check should match log_x function call

mazgch:
-----------
## Description of Change
compiler will complain with an error
`esp32/2.0.4/libraries/WiFi/src/WiFiGeneric.cpp:499:48: error: unused variable 'event' [-Werror=unused-variable]`
about unused variable event if compiled when ARDUHAL_LOG_LEVEL was set < ARDUHAL_LOG_LEVEL_VERBOSE.

## Tests scenarios
When compiled in Arduino it throws an error, check it will no longer throw an error.


PilnyTomas (approved):
LGTM

P-R-O-C-H-Y (approved):
LGTM! Thanks for PR @mazgch 👍

PR #7016, opened 2022-07-20
Adds Wire::setBufferSize function

### LINKED ISSUE #6967: Wire.h not able to change buffer length from .ino file
### Board

ESP32 DEV Module

### Device Description

Using generic ESP32 dev board

### Hardware Configuration

non

### Version

v2.0.3

### IDE Name

Arduino IDE

### Operating System

Windows 11

### Flash frequency

80

### PSRAM enabled

no

### Upload speed

921600

### Description

I am using a [vl53l3cx ](https://www.st.com/resource/en/datasheet/vl53l3cx.pdf)TOF sensor with an ESP32 board. 

I am using this [library and its example](https://github.com/pololu/vl53l1x-st-api-arduino) for testing and it works perfectly but I need to change inside the ESP32 package to make it work.  To be specific inside the Wire.h file of the ESP32 package. 

Initially, the example was not working directly but I read somewhere in the form that it needs to increase the I2C buffer size as the received packet size from the sensor is big therefore default size will not work.

> #define I2C_BUFFER_LENGTH 128

First I tested it by changing the I2C_BUFFER_LENGTH  to 256 directly in the Wire.h file and the example code is working perfectly.  This test was done in the old ESP32 package release V1.0.6

But In the latest releases that I can see after V2.0.3 in Wire.h file it's defined like this

> #ifndef I2C_BUFFER_LENGTH
>     #define I2C_BUFFER_LENGTH 128
> #endif

So I thought I can increase the size of the buffer externally directly from the Arduino .ino file itself like this before including the Wire.h

> #define I2C_BUFFER_LENGTH 256
> #include<Wire.h> 

But on uploading in this way the sensor data is not read properly like previously, So I checked by printing the I2C_BUFFER_LENGTH  in Arduino code as well as inside the Wire.begin(); library function.  
And shockingly it printed 256 in the Arduino code and when begin to function is called it printed 128. 
So it seems like it not making any change to the I2C_BUFFER_LENGTH  even it's defined before calling the library. 

Is there any way to change the I2C_BUFFER_LENGTH  without making any modifications to the ESP32 package itself? 

As I think the Arduino compile is somehow compiling all ESP32 packagess before even we include them. As I have done the same test with the blank sketch with only Wire.h library included and I2C_BUFFER_LENGTH  same results were seen.

Thank you,
Dharmik Patel

### Sketch

```cpp
#define I2C_BUFFER_LENGTH 256
#include<Wire.h>

void setup() {
  Serial.begin(9600);
  delay(2000);
  Serial.println(I2C_BUFFER_LENGTH);  // here it will print 256 that what we need 
  Wire.begin();  // but in actually its not changed here it will print 128
}

void loop() {

}
```

### Debug Message

```plain
None
```

### Other Steps to Reproduce

To debug properly add this line inside your Wire.cpp file function 
> bool TwoWire::begin(int sdaPin, int sclPin, uint32_t frequency)

add this inside that 
> Serial.println(I2C_BUFFER_LENGTH);

Finally, it will look like this

> // Master Begin
> bool TwoWire::begin(int sdaPin, int sclPin, uint32_t frequency)
> {
> 	Serial.println(I2C_BUFFER_LENGTH);

### LINKED ISSUE #7042: Setting I2C_BUFFER_LENGTH does not set underlying space requirements for larger receive buffer size
### Board

Custom board

### Device Description

ESP-WROOM-32 with I2C device connected

### Hardware Configuration

Pins 21 and 22 are used for I2C and are working fine

### Version

v2.0.4

### IDE Name

Arduino 1.8.19

### Operating System

Windows 10

### Flash frequency

80 MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

I needed to increase the size of the I2C message that is received from a peripheral and this works within the Arduino libraries but the underlying hal/idf code does not make use of the I2C_BUFFER_LENGTH being defined. I found that 384 bytes is the maximum size allowed for receipt of messages. If a larger value than this is provided, I get an assertion failure:

`
0x40093d1d: __assert_func at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/newlib/assert.c line 85
0x3ffb5900: Unknown Method at Unknown File line unknown
0x4008ed3a: xQueueGenericSend at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/freertos/queue.c line 821 (discriminator 2)
0x3ffb5a30: Unknown Method at Unknown File line unknown
0x400e92e5: TwoWire::requestFrom(unsigned short, unsigned int, bool) at E:\Program Files\Programming\Arduino_IDE\Arduino15\packages\esp32\hardware\esp32\2.0.4\libraries\Wire\src/Wire.cpp line 397
`

I would argue that the the two sizes should be linked and a readable error be generated when the size limit is hit. Setting the core logging  level results for debug output results in no additional detail.

I would also like to know how to increase this allowed size of the items used in the underlying code.

### Sketch

```cpp
Will be provided if requested
```

### Debug Message

```plain
No debug error is being provided, but would be helpful if one was.
```

### Other Steps to Reproduce

_No response_

SuGlider:
## Description of Change
Adds a new Wire (I2C) functionality to set the RX/TX buffer size and change it any time in the sketch.
Some I2C devices and sensors may require a bigger buffer for its transactions.

## Tests scenarios

``` cpp
#include "Wire.h"

void setup() {
  Wire::setBufferSize(512);
  Wire::begin();  // as I2C Master
}

void loop() {
}

```

## Related links
Fixes #6967 
Fixes #7042


r-downing:
If the size doesn't need to be changed at runtime, perhaps this could be done with a template size, so you don't have to use up more heap memory?

```
template <size_t desiredCapacity = DEFAULT_SIZE>
class TwoWire: public Stream
...
    uint8_t txBuffer[desiredCapacity];
```

SuGlider:
Arduino APIs use to be "user proof" also... So creating an API with all necessary checks is safer.

libraries/Wire/src/Wire.cpp
```diff
@@ -205,6 +287,10 @@ bool TwoWire::begin(int sdaPin, int sclPin, uint32_t frequency)
         started = true;
         goto end;
     }
+    if (!allocateWireBuffer()) {
+        // failed! Error Message already issued
+        goto end;
+    }
     if(!initPins(sdaPin, sclPin)){
         goto end;
     }
```
P-R-O-C-H-Y, line 296: What if `allocateWireBuffer()` succeed, but `initPins(sdaPin, sclPin)` fails and proceed to `goto end`, we should `freeWireBuffer()` in the `end:`

P-R-O-C-H-Y: Same in SLAVE `TwoWire:begin()`

libraries/Wire/src/Wire.h
```diff
@@ -34,8 +34,9 @@
 #endif
 #include "Stream.h"
 
+#define WIRE_HAS_BUFFER_SIZE    1
```
P-R-O-C-H-Y, line 37: Just for info, what is this define used for? Thanks

P-R-O-C-H-Y (commented):
Looks good, will test tomorrow :) take a look on my comments @SuGlider

libraries/Wire/src/Wire.cpp
SuGlider: If you look at `allocateWireBuffer()`, you will see that this function has already taken care of it and calls `freeWireBuffer()` before returning.

 `allocateWireBuffer()` will make sure that or it successfully allocated Rx and Tx together or otherwise will will not allocate/free both of them.

libraries/Wire/src/Wire.h
SuGlider: This `#define` is common in Arduino Mainstream to allow other implementations to add or not a specific functionality in their code. 

Example: 
https://github.com/arduino/ArduinoCore-avr/blob/master/libraries/Wire/src/Wire.h#L31-L32
``` cpp
// WIRE_HAS_END means Wire has end()
#define WIRE_HAS_END 1
```

libraries/Wire/src/Wire.cpp
r-downing: I think he's saying, if `allocateWireBuffer()` succeeds, but then `initPins()` fails and jumps to end, the buffers will still be allocated and there's no call to free them there

P-R-O-C-H-Y: Exactly

libraries/Wire/src/Wire.h
P-R-O-C-H-Y: If this is common in Arduino why not to have it :) maybe we can add `WIRE_HAS_END` as well :)

libraries/Wire/src/Wire.cpp
P-R-O-C-H-Y: And we should add the `freeWireBuffer()` to the `TwoWire:end()` function as well.

libraries/Wire/src/Wire.h
P-R-O-C-H-Y: If its common in Arduino mainstream why not to have it :) Maybe we can add the `#define WIRE_HAS_END 1` as well in this PR. What do you think?

libraries/Wire/src/Wire.cpp
SuGlider: Ok, Thanks! Yes, it is right. I'll add the ´freeWireBuffer()`.

libraries/Wire/src/Wire.h
SuGlider: done

libraries/Wire/src/Wire.cpp
SuGlider: done

libraries/Wire/src/Wire.cpp
```diff
@@ -132,6 +135,81 @@ bool TwoWire::setPins(int sdaPin, int sclPin)
...
+    }
+
+#if !CONFIG_DISABLE_HAL_LOCKS
+    if(lock != NULL){
+        //acquire lock
+        if(xSemaphoreTake(lock, portMAX_DELAY) != pdTRUE){
+            log_e("could not acquire lock");
+            return 0;
+        }
+#endif
```
P-R-O-C-H-Y, line 187: If you call `Wire.setBufferSize()` before calling `Wire.begin()`, nothing will happen, because the `lock` is not created yet. WE need to be the same as it is in `TwoWire::begin()` function.

```
#if !CONFIG_DISABLE_HAL_LOCKS
    if(lock == NULL){
        lock = xSemaphoreCreateMutex();
        if(lock == NULL){
            log_e("xSemaphoreCreateMutex failed");
            return false;
        }
    }
    //acquire lock
    if(xSemaphoreTake(lock, portMAX_DELAY) != pdTRUE){
        log_e("could not acquire lock");
        return false;
    }
```

libraries/Wire/src/Wire.cpp
```diff
@@ -132,6 +135,81 @@ bool TwoWire::setPins(int sdaPin, int sclPin)
...
+    } else {
+        // no memory allocated yet, just change the size value - allocation in begin()
+        BufferSize = bSize;
+    }
+#if !CONFIG_DISABLE_HAL_LOCKS
+        //release lock
+        xSemaphoreGive(lock);
+    }
+#endif
+    return bSize;
```
P-R-O-C-H-Y, line 209: It makes more sense for the return to be `BufferSize` to be able to check, if the function made the change :)

P-R-O-C-H-Y (changes requested):

libraries/Wire/src/Wire.cpp
```diff
@@ -205,13 +288,18 @@ bool TwoWire::begin(int sdaPin, int sclPin, uint32_t frequency)
...
+        goto end;
+    }
     if(!initPins(sdaPin, sclPin)){
         goto end;
     }
     err = i2cInit(num, sda, scl, frequency);
     started = (err == ESP_OK);
 
 end:
+    freeWireBuffer();
```
P-R-O-C-H-Y, line 302: With this change its not working now. I did not see that it always run to `end:` even if all initialisations succeeds. We need to add a check to make it work correctly
```
if(!started) {
freeWireBuffer();
}
```

libraries/Wire/src/Wire.cpp
SuGlider: In this case, we just must set bufferSize (I changed the captilization for keeping style) before returning, because `begin()` will `malloc()` the memory.

libraries/Wire/src/Wire.cpp
SuGlider: In case it fails by `malloc()` problem, `bufferSize` will keep the requested size and `bSize` will go to zero, as return value. So, with a `begin()` it will still be able to try to allocate the memory based on `bufferSize`.

libraries/Wire/src/Wire.cpp
SuGlider: yes, right, it falls through and causes the issue. :-)

SuGlider: I don't like this `goto` way of doing it... but let's keep "the style" ;-)

libraries/Wire/src/Wire.cpp
SuGlider: I added a `else` for the `if(lock != NULL) { ...} else { bufferSize = bSize; }` to solve it.

libraries/Wire/src/Wire.cpp
SuGlider: Lines 207 ... 210 shall solve the issue of `setBufferSize()` called before `begin()`

libraries/Wire/src/Wire.cpp
P-R-O-C-H-Y: Why not to create the lock like its in `begin()` or `setPin()`? Then there is no need to `bufferSize = bSize` twice so the code will be simpler :)

```
size_t TwoWire::setBufferSize(size_t bSize)
{
    // Maximum size .... HEAP limited ;-)
    if (bSize < 32) {    // 32 bytes is the I2C FIFO Len for ESP32/S2/S3/C3
        log_e("Minimum Wire Buffer size is 32 bytes");
        return 0;
    }

#if !CONFIG_DISABLE_HAL_LOCKS
    if(lock == NULL){
        lock = xSemaphoreCreateMutex();
        if(lock == NULL){
            log_e("xSemaphoreCreateMutex failed");
            return false;
        }
        //acquire lock
        if(xSemaphoreTake(lock, portMAX_DELAY) != pdTRUE){
            log_e("could not acquire lock");
            return 0;
        }
    }
#endif
    // allocateWireBuffer allocates memory for both pointers or just free them
    if (rxBuffer != NULL || txBuffer != NULL) {
        // if begin() has been already executed, memory size changes... data may be lost. We don't care! :^)
        if (bSize != BufferSize) {
            // we want a new buffer size ... just reset buffer pointers and allocate new ones
            freeWireBuffer();
            BufferSize = bSize;
            if (!allocateWireBuffer()) {
                // failed! Error message already issued
                bSize = 0; // returns error
                log_e("Buffer allocation failed");
            }
            log_d("buff size = %d", BufferSize);
        } // else nothing changes, all set!
    } else {
        // no memory allocated yet, just change the size value - allocation in begin()
        BufferSize = bSize;
        log_d("buff size = %d", BufferSize);
    }
#if !CONFIG_DISABLE_HAL_LOCKS
    //release lock
    xSemaphoreGive(lock);
    
#endif
    return bSize;
}
```

libraries/Wire/src/Wire.cpp
P-R-O-C-H-Y: Ok I got it, we can leave the `bSize` now :) It was not making sense before the lock change. It was returning inputed `bSize` even when `bufferSize` did not changed :) So marking as solved

libraries/Wire/src/Wire.cpp
SuGlider: Adding the extra lock code is more expensive in terms of flash consumption, but, that's ok.  
Done.

SuGlider: Opps... there is an error in the lock code... if it is not NULL, it is never acquired.... fixing it.

P-R-O-C-H-Y (approved):
LGTM :) Good job @SuGlider

PR #7019, opened 2022-07-21
Enable configuring target variants

kohlschuetter:
This change enables picking the right board configuration from the
variants/ folder. Previously, we would always pick the default
configuration (e.g., "esp32" instead of "heltec_wifi_lora_32_V2").


VojtechBartoska:
Same here @kohlschuetter, we will review this at the end of August when @me-no-dev is able to take a look on this.

me-no-dev (approved):

PR #7020, opened 2022-07-21
WiFiClientSecure: Fix warning about missing PSK-based ciphersuites

kohlschuetter:
Make the warning actionable by pointing at the right setting in Kconfig.


SuGlider (approved):
Thanks @kohlschuetter for the contribution!

PR #7034, opened 2022-07-25
Adds default Wire1 pins auto-setup

### LINKED ISSUE #6999: Have Wire1 pins autoset just like Wire
### Related area

I2C Peripheral

### Hardware specification

Any hardware based on ESP32

Wire is auto-setup for any ESP variant to the SDA/SCL pins
https://github.com/espressif/arduino-esp32/blob/master/variants/adafruit_qtpy_esp32s2/pins_arduino.h#L29
but SDA1/SCL1 pins do not do anything... it would be nice if setPins() was not required if these are #defined!
https://github.com/espressif/arduino-esp32/blob/master/variants/adafruit_qtpy_esp32s2/pins_arduino.h#L32

we can submit a PR, but want to know whether it would be rejected before spending the time.

_No response_

### Additional context

_No response_

SuGlider:
## Description of Change
Some boards define SDA1 and SCL1 pins for Wire1 peripheral.
This information isn't used in `Wire1.begin()` as pins default.
The PR adds this functionality and set default pins when they are correctly created in `pins_arduino.h`

These pins shall be set in `variants/<board_name>/pins_arduino.h` like this example:

``` cpp
#define WIRE1_PIN_DEFINED 1             // See Wire.cpp at bool TwoWire::initPins(int sdaPin, int sclPin)
static const uint8_t SDA1 = 2;
static const uint8_t SCL1 = 4;
```

## Tests scenarios
Tested with ESP32/S2/S3.

Note ESP32-C3 has just 1 I2C peripheral

## Related links
Closes #6999


P-R-O-C-H-Y:
@SuGlider Do we need to change the example to Wire1? Not all boards will have Wire1 pins defined. :)

libraries/Wire/examples/WireMaster/WireMaster.ino
```diff
@@ -7,16 +7,16 @@ uint32_t i = 0;
...
 void loop() {
   delay(5000);
 
   //Write message to the slave
-  Wire.beginTransmission(I2C_DEV_ADDR);
-  Wire.printf("Hello World! %u", i++);
-  uint8_t error = Wire.endTransmission(true);
+  Wire1.beginTransmission(I2C_DEV_ADDR);
+  Wire1.printf("Hello World! %u", i++);
+  uint8_t error = Wire1.endTransmission(true);
```
P-R-O-C-H-Y, line 19: @SuGlider Do we need to change the example to Wire1? Not all boards will have Wire1 pins defined. :)
And the example won't work correctly now, when `Wire1.begin()` called, but requesting bytes from **Wire** `Wire.requestFrom(I2C_DEV_ADDR, 16);`. Take a look :)

P-R-O-C-H-Y (commented):
@SuGlider Would be good to add to our official boards SDA1 and SCL1 defines too. (eg. esp32, esp32s2 and esp32s3 variants).

libraries/Wire/examples/WireMaster/WireMaster.ino
SuGlider: Actually, it was commited by mistake... I'll revert it.

ladyada:
Tested a QT Py ESP32 (where the QT connector is Wire1) and it worked great :)

ladyada (approved):

P-R-O-C-H-Y (approved):

PR #7037, opened 2022-07-25
./.github/scripts: Restore the build directory

Ouss4:
## Description of Change
The default build directory was changed during the last PR to accomodate the multi-configurations build.  However, this also changed it for a single sketch build.  This had a side effect of requiring to pass it as a parameter to `pytest` because `pytest-emebdded` default to `build` if none is provided.  So the command became:`pytest tests --build-dir build0 -k hello_world`.  This PR brings the old behavior for single sketch builds so one can test with just `pytest tests -k hello_world`.

## Tests scenarios
Locally with the scripts and the test sketches present in-tree.  The CI will test more.
## Related links
N/A


P-R-O-C-H-Y (approved):
LGMT, tested locally and everything works.

PR #7044, opened 2022-07-27
Update WiFiGeneric.cpp

zhangyanjiaoesp:
## Description of Change
Enlarge the cache_tx_buf num

## Related links
Related PR in Arduino Lib: https://github.com/espressif/arduino-esp32/pull/5791

close https://github.com/espressif/esp-idf/issues/8992

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -663,7 +663,7 @@ bool wifiLowLevelInit(bool persistent){
         }
 
         wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
-	
+
 	if(!WiFiGenericClass::useStaticBuffers()) {
 	    cfg.static_tx_buf_num = 0;
             cfg.dynamic_tx_buf_num = 32;
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -663,6 +663,15 @@ bool wifiLowLevelInit(bool persistent){
         }
 
         wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
+	
+	if(!WiFiGenericClass::useStaticBuffers()) {
+	    cfg.static_tx_buf_num = 0;
+            cfg.dynamic_tx_buf_num = 32;
+	    cfg.tx_buf_type = 1;
+            cfg.cache_tx_buf_num = 4;  // can't be zero!
+	    cfg.static_rx_buf_num = 4;
+            cfg.dynamic_rx_buf_num = 32;
+        }
 
         esp_err_t err = esp_wifi_init(&cfg);
         if(err){
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -664,15 +664,6 @@ bool wifiLowLevelInit(bool persistent){
 
         wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
 
-	if(!WiFiGenericClass::useStaticBuffers()) {
-	    cfg.static_tx_buf_num = 0;
-            cfg.dynamic_tx_buf_num = 32;
-	    cfg.tx_buf_type = 1;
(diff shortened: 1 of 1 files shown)


libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -664,15 +664,6 @@ bool wifiLowLevelInit(bool persistent){
 
         wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
 
-	if(!WiFiGenericClass::useStaticBuffers()) {
```
atanisoft, line 667: this was added to reduce the memory footprint of the WiFi driver as it increased considerably from IDF v3 to IDF v4.

Refer to https://github.com/espressif/arduino-esp32/pull/5791 for this. @SuGlider fyi...

SuGlider: Yes... Default now is WiFi Dynamic Allocation in order to reduce HEAP consumption.
If the user wants to go with Static Allocation, the application shall use `WiFiGenericClass::useStaticBuffers(true);` before setting the WiFi Mode with `WiFiGenericClass::mode()`

SuGlider:
@zhangyanjiaoesp - 
Please try using the API that sets Static WiFi allocation:

``` cpp
void setup() {
  Serial.begin(115200);
  WiFi.useStaticBuffers(true);
  WiFi.mode(WIFI_STA);
  
  // rest of the necessary code ...
}
```

``` WiFi.useStaticBuffers(true);``` will force Arduino to use deafult sdkconfig WiFi setup:

```
CONFIG_ESP32_WIFI_STATIC_RX_BUFFER_NUM=8
CONFIG_ESP32_WIFI_DYNAMIC_RX_BUFFER_NUM=32
CONFIG_ESP32_WIFI_STATIC_TX_BUFFER=y
CONFIG_ESP32_WIFI_TX_BUFFER_TYPE=0
CONFIG_ESP32_WIFI_STATIC_TX_BUFFER_NUM=8
CONFIG_ESP32_WIFI_CACHE_TX_BUFFER_NUM=16
```

File: libraries/WiFi/src/WiFiGeneric.cpp [hunk 2]
```diff
@@ -664,15 +664,6 @@ bool wifiLowLevelInit(bool persistent){
...
-	    cfg.static_tx_buf_num = 0;
-            cfg.dynamic_tx_buf_num = 32;
-	    cfg.tx_buf_type = 1;
-            cfg.cache_tx_buf_num = 1;  // can't be zero!
-	    cfg.static_rx_buf_num = 4;
-            cfg.dynamic_rx_buf_num = 32;
-        }
-
```
SuGlider, line 675: I reject this PR and changes.... as explained in 
https://github.com/espressif/esp-idf/issues/8992#issuecomment-1197443125
https://github.com/espressif/esp-idf/issues/8992#issuecomment-1197446232

SuGlider:
@zhangyanjiaoesp - Some background about the PR #5791:

At the time of this PR #5791, the idea was to free some heap space when WiFi was activated.

One possible solution, for the lack of heap, was to disable malloc over psram, as done in IDF 3.3, and then enable dynamic buffers in sdkconfig.

The PR #5791 was intended to find a way to keep psram malloc working and, at the same time, enable dynamic WiFi buffer allocation to free some heap memory. This combination was impossible due to a limitation imposed by KConfig.

This work around used in the PR was a nice suggestion from @igrr  to try to override WiFi driver and kconfig limitation. 
We implemented it and it is working for a number of Arduino users as of today. I don't see a real benefit in disabling it.

libraries/WiFi/src/WiFiGeneric.cpp
zhangyanjiaoesp: @SuGlider Although the customers can call `WiFiGenericClass::useStaticBuffers(true)` to solve the problem, but I think we also need to change the value `cfg.cache_tx_buf_num = 1` to at least `4`, this will ensure the customers can send several packets at the same time.

SuGlider: OK, Agree. Please commit this change. This will cost 5K RAM in the heap. It is worth it.

SuGlider (approved):
Thanks @zhangyanjiaoesp !

PR #7060, opened 2022-07-29
Removes -Werror=all from cflags

### LINKED ISSUE #7024: sprintf statement only gives compile error if compiled with compiler warning level "All" and "More" (-Wall -Werror=all)
### Board

compiler problem

### Device Description

compiler problem

### Hardware Configuration

compiler problem

### Version

v2.0.4

### IDE Name

Arduino, Sloeber

### Operating System

%

### Flash frequency

compiler problem

### PSRAM enabled

no

### Upload speed

compiler problem

### Description

The sketch below throws the error
```
E:\WORKSPACE_ARDUINO\ArduinoSketches\sketch_jul21a\sketch_jul21a.ino: In function 'void setup()':
sketch_jul21a:7:26: error: '%s' directive writing up to 19 bytes into a region of size 14 [-Werror=format-overflow=]
   sprintf(sStringBuffer, "Acc.X %s", &sStringBuffer[10]);
   ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
E:\WORKSPACE_ARDUINO\ArduinoSketches\sketch_jul21a\sketch_jul21a.ino:7:10: note: 'sprintf' output between 7 and 26 bytes into a destination of size 20
   sprintf(sStringBuffer, "Acc.X %s", &sStringBuffer[10]);
   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus.exe: some warnings being treated as errors
exit status 1
'%s' directive writing up to 19 bytes into a region of size 14 [-Werror=format-overflow=]
```
 if compiled with -Wall -Werror=all

**For all non ESP32 platforms it compiles without any error!** - see https://github.com/ArminJo/Arduino-BlueDisplay/actions/runs/2714739545

### Sketch

```cpp
char sStringBuffer[20];

void setup() {
  Serial.begin(115200);
  sprintf(&sStringBuffer[10], "123.4567"); // edited: original code was: dtostrf(123.45, 4, 1, &sStringBuffer[10]);
  sprintf(sStringBuffer, "Acc.X %s", &sStringBuffer[10]);
  Serial.println(sStringBuffer);
}

void loop() {
}
```

### Debug Message

```plain
E:\WORKSPACE_ARDUINO\ArduinoSketches\sketch_jul21a\sketch_jul21a.ino: In function 'void setup()':
sketch_jul21a:7:26: error: '%s' directive writing up to 19 bytes into a region of size 14 [-Werror=format-overflow=]
   sprintf(sStringBuffer, "Acc.X %s", &sStringBuffer[10]);
   ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
E:\WORKSPACE_ARDUINO\ArduinoSketches\sketch_jul21a\sketch_jul21a.ino:7:10: note: 'sprintf' output between 7 and 26 bytes into a destination of size 20
   sprintf(sStringBuffer, "Acc.X %s", &sStringBuffer[10]);
   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus.exe: some warnings being treated as errors
exit status 1
'%s' directive writing up to 19 bytes into a region of size 14 [-Werror=format-overflow=]
```
```


### Other Steps to Reproduce

_No response_

### PR DESCRIPTION
**Pull request by SuGlider**
-----------
## Description of Change
`-Werror=all` creates some issues when used, which are not desirable. 
Erroring out those warning-errors sometimes is too hard and not really important for Arduino users.
Most other Arduino platform do not set this cflag.

## Tests scenarios
See issue #7024

## Related links
Closes #7024

### CODE REVIEW

**Review by P-R-O-C-H-Y — APPROVED** [review]
I agree with this change. Some errors should be marked as warning instead.

**Review by igrr — CHANGES_REQUESTED** [review]
All the examples should be compile-able without warnings. Currently CI detects this due to this -Werror=all flag. If you remove it, you need to add it back somewhere during CI build to make sure that a new warning results in CI build failure.

**Comment by igrr** [comment]
For example, you can try adding this flag back during CI build by creating a platform.local.txt file before building the sketches (https://arduino.github.io/arduino-cli/0.21/platform-specification/#platformlocaltxt)

**Comment by SuGlider** [comment]
> For example, you can try adding this flag back during CI build by creating a platform.local.txt file before building the sketches (https://arduino.github.io/arduino-cli/0.21/platform-specification/#platformlocaltxt)

Changed the PR to keep `platform.txt` intact for CI and modify `warning cflags` in `platform.local.txt` only - by adding the file to the project. 

I tested it and this works with Arduino IDE that doesn't considers warnings as an error when building the sketch.

**Review by PilnyTomas — APPROVED** [review]
LGTM

**Comment by igrr** [comment]
@SuGlider as far as i understand the platform.local.txt file is supposed to be local — I.e. this is the file which the developer can create locally to adjust some build settings.
Im'n not sure we are allowed to include a platform.local.txt file with the installation.
Would you consider keeping your previous change to platform.txt and instead creating a temporary platform.local.txt in CI before the build job starts?

**Comment by igrr** [comment]
One more request, please also check that a warning in one of the examples still causes CI to fail.

**Comment by SuGlider** [comment]
> @SuGlider as far as i understand the platform.local.txt file is supposed to be local — I.e. this is the file which the developer can create locally to adjust some build settings. Im'n not sure we are allowed to include a platform.local.txt file with the installation. Would you consider keeping your previous change to platform.txt and instead creating a temporary platform.local.txt in CI before the build job starts?

I'd need to understand how the CI scripts work first. 
I'm not familiar with those CI building scripts. 
Not sure if they read `platform.local.txt` when building.

Another way to move forward, would be to change `platform.txt` to remove `´Werror=all` from "IDE cflags" and then create another `CI.cflags.xxxx` that would be picked up by CI to build it. 
This way we would have separated sets of `cflags` for each process.

Do you see a potential issue if we create separated `CI.xxx.yyy` setup in platform.txt?

**Comment by SuGlider** [comment]
> One more request, please also check that a warning in one of the examples still causes CI to fail.

I found warnings from `libraries/BLE/examples/BLE5_periodic_sync/BLE5_periodic_sync.ino` and `I2S example`.
I'll check all the CI runners log output.

**Comment by SuGlider** [comment]
I talked to @Ouss4 and I'll rework it further. 
Run some CI testing within this PR and so on.

**Comment by SuGlider** [comment]
OK. @igrr @Ouss4 

It is now working as desired:
1- platform.txt (user IDE) has -Werror=all removed.
2- Reorganized platform.txt to separate clearly Warning flags and allow those to be easily managed (created .w_flags)
3- Created tools/platform.local.txt that is only used (copied) in CI time by install-arduino-core-esp32.sh script
4- All warning setup/cflags can now be changed in tools/platform.local.txt - it is only used by CI building process
5- I have removed all the warnings waiving (-Wnoxxxxx) and now we have many errors in CI, but this means it worked fine.

We shall decide which waring flags should stay in CI. Suggestions are welcome.

File: .github/scripts/install-arduino-core-esp32.sh [hunk 1]
```diff
@@ -27,6 +27,9 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
     cd esp32
     #git submodule update --init --recursive > /dev/null 2>&1
 
+    echo "Copying CI restrictive warning settings from tools/platform.local.txt"
+    cp tools/platform.local.txt .
+
```
Ouss4, line 32: Wasn't this inside the `if else` statement above? It didn't work?
It should be added after line 20 (inside the `if`).

SuGlider: it didn't work in side the `if`

SuGlider: I had to move it outside in order to really copy the file

.github/workflows/push.yml
```diff
@@ -42,7 +42,9 @@ jobs:
           ./tools/dist
           ~/arduino_ide
         key: ${{ runner.os }}-${{ hashFiles('package/package_esp32_index.template.json',
+                                            'platform.txt',
                                             'tools/get.py',
+                                            'tools/platform.local.txt',
```
Ouss4, line 47: Is this one required?

.github/scripts/install-arduino-core-esp32.sh
Ouss4: I think it was inside the `else` not the `if` where it should be. Also there is a `cd` missing (or the path needs to be updated accordingly by just adding esp32/)

.github/workflows/push.yml
SuGlider: If this file changes, it means that cflags (any) have changed in CI building processs... really not sure if this is necessary.
Let me know.

.github/scripts/install-arduino-core-esp32.sh
Ouss4: You didn't force push, so we it's still here: https://github.com/espressif/arduino-esp32/pull/7060/commits/1cabb075155a6264f8a2d730c8c4cdc0a231988a

.github/workflows/push.yml
Ouss4: The cache includes the toolchains and the Arduino IDE.  The toolchains are unrelated to this.
As for the IDE, the question to ask is if `platform.local.txt` will introduce new options that affect it.  I think the answer is no, unless `platform.txt` does so and we override them with `platform.local.txt`. So I believe it's redundant.
That said, it doesn't really hurt to keep it, I'm just trying to understand if we really need it.  A concern would be if we keep customizing it and invalidating the cache everytime for each PR that does so.

.github/scripts/install-arduino-core-esp32.sh
SuGlider: I tested it inside `else` exactly as in https://github.com/espressif/arduino-esp32/commit/1cabb075155a6264f8a2d730c8c4cdc0a231988a
But it didn't work, CI was passing because `platform.local.txt` was not being copied.

After moving it ouside, it worked fine...

.github/workflows/push.yml
SuGlider: I believe that we will change it for finishing this PR and we will not change it for a long while after that.
There may be a very small chance of need for changing something in the `platform.local.txt` and not in the `platform.txt` in the future, but as I said, I think it will be very rare.

.github/scripts/install-arduino-core-esp32.sh
Ouss4: In the CI the `else` path is not taken, this is why it wasn't being copied.  It's in the `if` where it should be.
`if [ ! -z "$GITHUB_REPOSITORY" ];  then`: if `GITHUB_REPOSITORY` is defined, i.e., this is a Github repo.

Ouss4: If we copied it unconditionally, I think w will end up with a case of https://github.com/espressif/arduino-esp32/pull/7060#issuecomment-1204875043

SuGlider: No, this is only copied in the CI docker space. It will never affect the end user, as suggested in https://github.com/espressif/arduino-esp32/pull/7060#issuecomment-1204875043

@igrr was actually referring to the first commit I made that added `platform.local.txt` to the main Arduino ESP32 Github folder, which would go to all end users and therefore affect them.

SuGlider: I understand that the script can copy `tools/platform.local.txt` unconditionally, which is necessary for CI to consider desired Warning flags when compiling the examples.

.github/workflows/push.yml
SuGlider: `platform.txt` is mainly intended to the Arduino end user and as base for CI.
`tools/platform.local.txt` is only intended for CI, not used at all by the Arduino IDE.

.github/scripts/install-arduino-core-esp32.sh
igrr: My assumption is that users won't run `.github/scripts/install-arduino-core-esp32.sh`, so -Werror is not going to affect them.

(Unless they want to replicate the behavior of CI; in which case they can run this script and should get the same result as in CI.)

SuGlider:
> One note, platform.txt file is generated by arduino-lib-builder: https://github.com/espressif/esp32-arduino-lib-builder/blob/29387059e8a15b7afb526bdfca9bb9f1c7d65b28/tools/copy-libs.sh#L459-L477.
> 
> I think we should update the template used in arduino-lib-builder, otherwise the changes in platform.txt will be lost upon the next update.
> 
> Edit: maybe what the script is doing is actually _editing_ the previous platform.txt file; in which case it's probably fine to make the changes here.

I'll also check it and verify it.

.github/scripts/install-arduino-core-esp32.sh
Ouss4: Test scripts do run `install-arduino-core-esp32.sh` to get some environment variables.  Running unit tests, as found under `tests` using `tests_build.sh` for instance will have this file copied.  Which is something users are supposed to do.
Other than this, this is a logic for Github, it should be under the Github path, regardless if it introduces issues or not.

SuGlider: Ok, you are talking about HIL testing, right? Those PRs marked with HIL tag.
In that case it is still CI only, isn't it? 
So we will have -Werror=all and no -Wnoxxxx in HIL testing.
Do you see a problem with it?

Ouss4: I was referring to running those tests locally using the `tests_build.sh` script.  It works similar to the CI.

SuGlider: OK, thanks for clarifying. 
The absolute most of Arduino Users will never use `tests_build.sh`. 
Therefore, I'd say that if it happens, the user that does use it, is an expert one, and that person will know what to do.

I still see no problem with the script copying the file to the Docker file system.

Ouss4: Let's take one step back please for me to understand the issue.  What's the problem with the below diff? It doesn't work?
```diff
diff --git a/.github/scripts/install-arduino-core-esp32.sh b/.github/scripts/install-arduino-core-esp32.sh
index 8584da5b..02f29c6a 100755
--- a/.github/scripts/install-arduino-core-esp32.sh
+++ b/.github/scripts/install-arduino-core-esp32.sh
@@ -18,6 +18,8 @@ if [ ! -d "$ARDUINO_ESP32_PATH" ]; then
     if [ ! -z "$GITHUB_REPOSITORY" ];  then
         echo "Linking Core..."
         ln -s $GITHUB_WORKSPACE esp32
+        echo "Copying CI restrictive warning settings from tools/platform.local.txt"
+        cp esp32/tools/platform.local.txt .
     else
         echo "Cloning Core Repository..."
         git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
```

> I still see no problem with the script copying the file to the Docker file system.

The issues, as I see them, are:
1. This is a logic for Github but it's mixed with the common path and not with the Github one.
2. The script copies build altering files to the home directory unbeknownst to the user and also doesn't clean them.

As for who uses `tests_build.sh` I hope anyone writing drivers for Arduino would.  That means tests are written and validated locally.

SuGlider: > Let's take one step back please for me to understand the issue. What's the problem with the below diff? It doesn't work?

It didn't work when it is executed by Github Action started by "On Push".
I see that when it uses GitHub-hosted runners with Ubuntu Image,  which is created fresh by the runner and it has nothing installed. So, I understand that the test `[ ! -z $GITHUB_REPOSITORY" ]` will always fail. 

>This is a logic for Github but it's mixed with the common path and not with the Github one.

This is just a Shell Script executed by a Runner that will use some VM with empty Ubuntu/Mac/Windows files system.
Then the script clones Arduino-IDE, ESP32 Arduino Core into it in order to excute the Tasks.
Right after cloning or linking (which I think never happens with Github Actions), it will copy `platform.local.txt` from `esp32/tools` to `esp32`. 

Looking into `install-arduino-core-esp32.sh` I can see other places that use Common Path - I just used it in the right place.
```
    if [ ! -z "$GITHUB_REPOSITORY" ];  then
        echo "Linking Core..."
        ln -s $GITHUB_WORKSPACE esp32
    else
        echo "Cloning Core Repository..."
        git clone https://github.com/espressif/arduino-esp32.git esp32 > /dev/null 2>&1
    fi

    #echo "Updating Submodules ..."
    cd esp32                                                   <<<<<<<<<<<<<<<<<<<<<<<
    #git submodule update --init --recursive > /dev/null 2>&1

    echo "Copying CI restrictive warning settings from tools/platform.local.txt"
    cp tools/platform.local.txt .
    
    echo "Installing Platform Tools ..."
    cd tools && python get.py    <<<<<<<<<<<<<<<<<<<<<<<<
    cd $script_init_path

    echo "ESP32 Arduino has been installed in '$ARDUINO_ESP32_PATH'"
    echo ""
```

> The script copies build altering files to the home directory unbeknownst to the user and also doesn't clean them.

Please clarify when does it happen. It copies files to the VM File System that only exists in the CI and Github Action Runner Time. Everything is destroyed after the end of VM execution. Nothing goes to the User File System, unless the script is executes directly in a Shell Command Line - which I understand is not the purpose of this script. 

### Anyway,  it works with Ubuntu/Mac/Windows and Arduino-Builder. It doesn't work with PlatformIO.

Ouss4: > It didn't work when it is executed by Github Action started by "On Push".

Was this ever tried?

> Right after cloning or linking (which I think never happens with Github Actions),

You can check every on-push workflow, since the creation of this workflow, and you can see that it actually links the repo and does not clone it.  Cloning in Github Action is done with the Checkout Action.
image

image

Ouss4:
BTW, is the plan to then fix all the warnings? Will this be part of this PR?

VojtechBartoska:
Issue update: This PR needs more investigation and proper testing so it's not visible for 2.0.5 milestone.

SuGlider:
> BTW, is the plan to then fix all the warnings? Will this be part of this PR?

Yes, we shall fix all the warning, maybe in separated PRs along time, before merging this one.

.github/scripts/install-platformio-esp32.sh
```diff
@@ -48,6 +48,9 @@ else
     git clone --recursive https://github.com/espressif/arduino-esp32.git "$PLATFORMIO_ESP32_PATH" > /dev/null 2>&1
 fi
 
+echo "Copying CI restrictive warning settings from tools/platform.local.txt"
```
me-no-dev, line 51: PIO does not use any Arduino configuration file.

platform.txt
```diff
@@ -25,13 +25,21 @@ compiler.sdk.path={runtime.platform.path}/tools/sdk/{build.mcu}
 compiler.prefix={build.tarch}-{build.target}-elf-
 
 #
-# ESP32 Support Start
+# Common Warning ignoring flags
 #
+compiler.common.w_flags=-Wno-error=unused-function -Wno-error=unused-variable -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-sign-compare -Wno-error=unused-but-set-variable
+
+#
+# ESP32S2 Support Start
```
me-no-dev, line 33: ESP32, not ESP32S2

me-no-dev:
what is the point of splitting the flags in such way? They are taken from ESP-IDF and filled in `platform.txt` and overwriting them could cause issues in headers, etc. All you need to do is edit two lines (removing `-Werror=all` from `platform.txt`) and create in CI `platform.local.txt` with `-Werror=all` for `compiler.warning_flags.all`, like `echo "compiler.warning_flags.all=-Wall -Werror=all -Wextra" > platform.local.txt`

gfvalvo:
> what is the point of splitting the flags in such way? They are taken from ESP-IDF and filled in `platform.txt` and overwriting them could cause issues in headers, etc. All you need to do is edit two lines (removing `-Werror=all` from `platform.txt`) and create in CI `platform.local.txt` with `-Werror=all` for `compiler.warning_flags.all`, like `echo "compiler.warning_flags.all=-Wall -Werror=all -Wextra" > platform.local.txt`
What is "CI"?

RoCorbera:
> What is "CI"?

Continuous integration (CI) automatically builds, tests, and integrates code changes within a shared repository.
https://resources.github.com/ci-cd/

mrengineer7777:
What is keeping us from making progress on this PR? What examples need to be fixed?

me-no-dev:
@SuGlider closing this for now. we can loo into better implementation on the current platform later

Issue #5624, opened 2021-09-02
Enabling IPv6 SLAAC option for arduino-esp32 by default


sanmira:
### Hardware:
| | |
| ------------- | ------------- |
| **Board** | ESP32-POE |
| **Core Installation version**  | 2.0.0 |
| **IDE name** | Platform.io |
| **Flash Frequency**  | 40Mhz |
| **PSRAM enabled** | no |
| **Upload Speed** | 115200 |
| **Computer OS** |  Windows 10 |                              

### Description:
I was trying to configure IPv6 client on my ESP32 device and faced an issue with device's inability to obtain a global IPv6 address.
After investigation I've found that for the brebuilded ESP-IDF SDK  **"CONFIG_LWIP_IPV6_AUTOCONFIG"** flag was not set (**SLAAC** option in the menuconfig).

I was able to rebuild the LwIP library manually with correct configuration. After that IPv6 started working as expected and I was able to get global IPv6 address using **tcpip_adapter_get_ip6_global()** function.

Can we enable the **SLAAC** option before building the ESP-IDF SDK for Arduino by default? 
Because, without this option IPv6 client is unusable with Arduino if I understand it correctly.

VojtechBartoska:
Hello @sanmira, thanks for your feature request. It needs more investigation from our side. Are you able to provide a code for testing this?

VojtechBartoska:
Please check merged PR which can solve your issue. 

https://github.com/espressif/esp32-arduino-lib-builder/pull/67

It will be available in next release 2.0.4.

VojtechBartoska:
@sanmira Did you retest your code under 2.0.4 release? Thanks

fryefryefrye:
With 2.0.4 still can not get public IPv6 address,only local link address.

```
[337040][D][WiFiGeneric.cpp:929] _eventCallback(): Arduino Event: 8 - STA_GOT_IP6
STA IPv6: fe80:0000:0000:0000:a6cf:12ff:fe99:eabc
```

Under same WiFi AP. ESP8266 can got public IPv6 address and communication is OK.
```
Waiting for ipv6 address
IPv6 address: 2409:8920:270:3b4f:a220:a6ff:fe21:af77
```

fryefryefrye:
> I was able to rebuild the LwIP library manually with correct configuration. After that IPv6 started working as expected and I was able to get global IPv6 address using **tcpip_adapter_get_ip6_global()** function.

[...]

PR #6667, opened 2022-04-29
New Class MacAddress similar to IPAddress

mrengineer7777:
## Description of Change
This adds new classes MacAddress and MacAddress8.  MacAddress allows for better handling of MAC and BSSID addresses.  MacAddress8 is for 8-byte EUI-64 as noted in esp_mac.h.

## Test scenarios
Tested on Arduino Feather ESP32.
```
[env]
monitor_speed = 115200
build_flags = 

[env:fthr-esp32ard]
platform = https://github.com/tasmota/platform-espressif32/releases/download/v2.0.2.3/platform-espressif32-2.0.2.3.zip
framework = arduino
board = featheresp32
```

## RC1
No new changes. Bugfixes accepted.

## TESTING
See test project here: https://github.com/mrengineer7777/MA_Test.

## Related links
Closes #6658.

PR #7065, opened 2022-07-30
Add "Get Global IPv6 Address" and update the example.

fryefryefrye:
https://github.com/espressif/arduino-esp32/issues/5624#issuecomment-1200153422

-----------
## Description of Change
Help to Get "Global IPv6 Address"

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.4 with ESP32.

log out put:
```
STA Local IPv6: fe80:0000:0000:0000:a6cf:12ff:fe99:eabc
STA Global IPv6: 2409:8920:0270:3d67:a6cf:12ff:fe99:eabc
```


lbernstone:
I think this needs a short conversation about what "localIP" really means here.  If it is just the assigned address, then scope should be a parameter in localIPv6 rather than making a new method.

fryefryefrye:
> I think this needs a short conversation about what "localIP" really means here. If it is just the assigned address, then scope should be a parameter in localIPv6 rather than making a new method.

In IPv4 , only have remote address and local address.

In IPv6, address which call "Local Link address" which is start with "fe80", is calculated by MAC address ,and can only used in LAN.    Global IPv6 Address   can be access on the internet. But they are all "local address".

Use parameter is an optional.

VojtechBartoska:
Thanks for the PR @fryefryefrye.

I'm adding this to 2.1.0 milestone because this needs more investigation and work. We will evaluate this PR soon.

VojtechBartoska:
Related PRs: #6712 & #6667

sgryphon:
I suggest not trying to patch in support for just one additional address, but recognise that with IPv6 a device will usually have varying numbers of multiple addresses.

e.g. Have a signature like `IPAddress localIP(int8_t index = 0)` that returns as many addresses as are available (or NONE once they have run out).  Maybe also have a signature like `int8_t getLocalIPs(IPAddress ipAddresses[], int8_t max)` that will fill an array (up to max), and return the number of addresses.

Note that ArduinoCore-API now has IPAddress that supports both v4 and v6. Generally the client code doesn't want to care about which version of IP is being used.

Note that local means this end of the connection (as opposed to the remote end), and is different from link-local.

As discussed above, link-local are addresses starting with "fe80:". The equivalent in IPv4 is a "169.254" address, which is only assigned when there is no network, whereas with IPv6 they are always assigned.

In IPv4 and endpoint usually only has one address, e.g. either a global address, private network address, or if no network then a link-local address (e.g. used in point-to-point connections).

With IPv6 an endpoint will usually have multiple local addresses -- a link-local address (autogenerated), as well as one or more global scope addresses, including both globally routeable addresses and maybe unique local addresses ("fd.."), both stateless/EUI-64 addresses and maybe DHCP assigned addresses, and with privacy extensions possibly also some additional rotating addresses.

For example, my local machine has 1x IPv4, and 7x IPv6 addresses.

Although Arduino/lwIP may not yet support all these types of addresses, then interface should prepare for varying numbers of addresses.

Generally when getting a single address, i.e. `localIP()` it is probably to display to the user (or in a log) as the address, useful for checking connections at the other end, etc.

Without any knowledge of the intended destination, I suggest ordering local IP addresses as if the destination is a global address, if possible, i.e. preference global addresses (globally routable or unique local addresses) over link-local, and IPv6 over IPv4 (see RFC 6724).

This won't always be what the user wants (e.g. in a dual stack network maybe they want to check the address in the logs of an IPv4 only destination), but if you only return a single address then you can't always know what will be needed.

me-no-dev:
@sgryphon gave very good proposal. IPv6Address class will go into the history, now that IPAddress will support both v4 and v6. I will leave this open, just to reming me to add proper integration to the interfaces to get all assigned IPAddresses (both v4 and v6 flavors)

sgryphon:

PR #7080, opened 2022-08-03
Changed Neopixel related static constants to #defines

ChuckVanzant:
To be consistent with other Adafruit ESP32 variants having Neopixels, changed the Neopixel related static constants to #defines in pins_arduino.h for the adafruit_qtpy_esp32 variant. This also resolves build issues where Neopixel related defines are used to include specific related code for the Qt Py ESP32 variant.


hathach:
look good, thanks for the PR to update adafruit defines

SuGlider:
@ChuckVanzant - please sign the CLA in order to move forward.

PR #7094, opened 2022-08-07
add lolin s3 board def

wemos:
add lolin s3 board def


davy39:
Shouldn't it be `lolin_s3.upload.maximum_size=16777216 ` for 16MB ?

SuGlider:
> Shouldn't it be `lolin_s3.upload.maximum_size=16777216 ` for 16MB ?

No. ESP32 works with partitions. The current used size is the default partition schema.

Issue #7076, opened 2022-08-02
HttpClient.cpp#setCookie()


fvaryu:
### Board

ESP32Dev Kit

### Device Description

DevKit

### Hardware Configuration

none

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

macOS

### Flash frequency

40Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

 HttpClient.cpp#setCookie() 

```
void HTTPClient::setCookie(String date, String headerValue)
{
    if (!_cookieJar)
    {
        return;
    }
    #define HTTP_TIME_PATTERN "%a, %d %b %Y %H:%M:%S"

    Cookie cookie;
    String value;
    int pos1, pos2;

    headerValue.toLowerCase(); // ? why toLoerCase()
  // other code
}
```
Why turn the value into lowercase, if the server returns uppercase? i don't understand。thx

```
  CookieJar cookies;
  http.setCookieJar(&cookies);
```
The server returns a mixed case token, but I get lowercase by CookieJar

### Sketch

```cpp
void HTTPClient::setCookie(String date, String headerValue)
{
    if (!_cookieJar)
    {
        return;
    }
    #define HTTP_TIME_PATTERN "%a, %d %b %Y %H:%M:%S"

    Cookie cookie;
    String value;
    int pos1, pos2;

   headerValue.toLowerCase();

    struct tm tm;
    strptime(date.c_str(), HTTP_TIME_PATTERN, &tm);
    cookie.date = mktime(&tm);
}
```

### Debug Message

```plain
None
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

SuGlider:
Hummm..... that seems a bug. Cookies are case sensitive.

SuGlider:
@fvaryu - If possible, please test the PR #7112 and let me know if it fixes the issue. Thanks.

PR #7134, opened 2022-08-16
Cookie(set-cookie) Value are case sensitive

fvaryu:
Cookie(set-cookie) Value are case sensitive, fix HTTPClient.setCookie() , remove headerValue.toLowCase(). (#7112)(#7076)


P-R-O-C-H-Y:
Hi @fvaryu, PR #7112 from @SuGlider is already fixing this issue. No need to create another PR with the same change.
Closing this.

PR #7171, opened 2022-08-23
Fixed switch rename issue in examples espressif/esp-rainmaker#152

sanketwadekar:
Fixed issue wherein changing the name of switches was not getting reflected in the phone app after the device reboots.
I have tested this on ESP32 using Arduino 2.0.3 and Rainmaker Android app
Closes espressif/esp-rainmaker#152


me-no-dev:
Thanks @sanketwadekar :) looks good!

Issue #7114, opened 2022-08-11
Make IPv6 support compatible with ESP8266 implementation


sgryphon:
### Related area

Networking support

### Hardware specification

n/a

### Is your feature request related to a problem?

The implementation of IPv6 as a separate type of address adds complexity.

Particularly for an Arduino-level user with limited experience, having to know which type of address to use complicates things. It would make a lot more sense if something like a DNS lookup simply returned an IP address, or checking the local address of a network connection, without knowing the type.

The implementation is also incompatible with the Arduino ESP8266 libraries, which are just a wrapper around lwIP ip_addr_t from the core Espressif libraries. https://github.com/esp8266/Arduino/blob/master/cores/esp8266/IPAddress.h

For example, a simple example that displays the current IP address for ESP8266 will work in IPv4-only, IPv6-only, and dual stack  environments, with a single reference to IPAddress.

The example code would not work for ESP32, and ESP32 would require more complicated code to check both the IPAddress and IPv6Address, and decide which to show.

### Describe the solution you'd like

Use an implementation similar (or the same as) ESP8266: https://github.com/esp8266/Arduino/blob/master/cores/esp8266/IPAddress.h

This has full backwards compatibility with IPv4 addresses, adds support for IPv6, does not complicate APIs by adding duplicate methods everywhere.  It even preserves the smaller memory footprint (and implementation detail) when compiled without IPv6.

### Describe alternatives you've considered

The existing IPv6Address adds complexity and is incompatible. It is not yet widely used, so would be easy to deprecated and change before merging the planned wider IPv6 support.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

VojtechBartoska:
Hello, adding this to 2.1.0 milestone. We will investigate in general IPv6 soon and consider your proposal.

Issue #6242, opened 2022-02-07
IPv6 readiness and support "by default"


nuclearcat:
### Related area

tcp/ip stack

### Hardware specification

any

### Is your feature request related to a problem?

IPv6 support for IoT

### Describe the solution you'd like

I want to create issue where i can track esp32 ipv6 readiness, ask questions regarding code and propose patches.
First question i have:
libraries/WiFi/src/WiFiGeneric.cpp
function
esp_err_t WiFiGenericClass::_eventCallback(arduino_event_t *event)
event ARDUINO_EVENT_WIFI_STA_CONNECTED
any particular reason why this line commented?
`
        esp_netif_create_ip6_linklocal(esp_netifs[ESP_IF_WIFI_STA]);
`
Having it uncommented brings at least local IPv6 support.

I think if it might bring compatibility issues, i can send PR that will add some flag in wifi struct, which user can set to enable ipv6.
For user it will look something like:
<pre>
  Serial.begin(115200);
+ WiFi.IPV6(true);
  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, password);
</pre>

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

VojtechBartoska:
@nuclearcat Any updates?

nuclearcat:
Yes might work on it at free time, probably this week

VojtechBartoska:
thanks @nuclearcat!

nuclearcat:
I added series of patches in PR, that provide necessary support for IPv6, but it needs serious testing, especially as i dont have yet devboards on s2, s3 and c3.
Also i will test in more scenarios in upcoming days and probably will provide additional test/demo sketches.

nuclearcat:
Any feedback? :)

leonelsr:
So... Another year has passed. Any progress on IPv6 support for ESP32 on Arduino Environment?

nuclearcat:
Latest PR is https://github.com/espressif/arduino-esp32/pull/7722 , it's not mine anymore, but i believe can call myself as co-author and i'm totally happy with it (it is used also in Tasmota).
I suggest to support author and ping Espressif maintainers, as it's important matter to merge it.

anuragrao04:
What's the status of this PR? I have been trying to get an ipv6 address for my esp32 and it doesn't work. Here's the code:

```c
#include <WiFi.h>
[...]

PR #7174, opened 2022-08-23
Add v6 support to IPAddress to match ArduinoCore-API

### LINKED ISSUE #6600: Add localIPv6 IPv6Address.toShortString()
### Related area

WIFI IPv6Address

### Hardware specification

ESP32

**Feature request**

Most OLED/LCD/TFT have small width, as IPv6 can be shorted/abreviated, so its usefull to display it short as possible.

Normal IPv6 Address: 
    fe80:0000:0000:0000:32ae:a4ff:fe02:e488
    
It can be short as:
    fe80::32ae:a4ff:fe02:e488

I have implemented it at IPv6Address.h/cpp but it is overrited every time i update sdk, so i will apreciate if this feature can be part of SDK.

I have change IPv6Address class, includding this function, that work as expected, im a beginner, so this function can be better writed.
```
String IPv6Address::toShortString() const {
	unsigned short val;
	std::stringstream stream;
	bool skip;
		
	for(int i = 0; i < 16; i+=2) {
		val = ((_address.bytes[i] << 8) | _address.bytes[i+1]);
		
		if ( val == 0 ) {
			if ( ! skip ) {
				skip = true; 
				stream << ":"; 
			}
		} else {
			if ( skip ) skip = false;
			
			stream << std::right << std::setfill('0') << std::setw(4) << std::hex << static_cast<int>(val);
			
			if ( i < 14 ) stream << ":";
		}
	}
	
	return String( stream.str().c_str() );
}
```

### Additional context

_No response_

sgryphon:

### Checklist
1. [X] Please provide specific title of the PR describing the change, including the component name (*eg. „Update of Documentation link on Readme.md“*)
2. [X] Please provide related links (*eg. Issue which will be closed by this Pull Request*)
3. [X] Please **update relevant Documentation** if applicable
4. [X] Please check [Contributing guide](https://docs.espressif.com/projects/arduino-esp32/en/latest/contributing.html)

-----------
## Description of Change

Update the IPAddress class to support v6 addresses, using the same interface changes as ArduinoCore-API (and thus compatible with other Arduino based platforms).

Note that the actual API changes are minimal:
* add a type() getter and corresponding enum
* add three new constructors: one for 16 octets, empty with type specifier, and byte pointer with type specifier
* add constant for IP6ADDR_ANY (similar function as IPADDR_NONE)

The changes are backward compatible, i.e. any IPv4 code will continue to get the same outputs for a given inputs. (There are extensive unit test cases in ArduinoCore-API).

Note: Whilst backwards compatible, this change does increase the memory footprint from 4 bytes to 20 bytes (assuming the enum compiles as an underlying int32_t). 

## Tests scenarios

**NOTE:** I am not sure how to build and test the project as I can't find instructions in the readme or contributing files. I can see the github actions scripts, but they are a bit more complicated that the ArduinoCore-API project, so I haven't reverse engineered out the steps to run all the tests yet (or the dependencies what is needed).

I am putting this code up for review while I figure this out (and maybe GitHub can run the tests).

Note that the code is mostly cut and paste from ArduinoCore-API project, which has unit tests, so should work. (The only bit not directly copied was the toString() function changes).

## Related links

Related to #7114, although it doesn't make it directly compatible with ESP8266 but with ArduinoCore, but does close off the issue.

Essentially fixes #6600 as it uses the RFC 5952 canonical format, which includes shortening of the left-most largest group of two or more zero fields. (The simple example in #6600 simply did the first zero fields). Note that the fromString() function parses all IPv6 representations, not just the canonical one. (There are extensive unit tests in the ArduinoCore-API project)

It is related to #6242 as it provides core support for the IPAddress class to have both v4 and v6 addresses, enabling support to be added (internally) to other classes (such as WiFi, Ethernet, etc) to handle IPv6 (without having to complicate their interfaces).

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -20,60 +20,197 @@
 #include <Arduino.h>
 #include <IPAddress.h>
 #include <Print.h>
+#include <StreamString.h>
 
-IPAddress::IPAddress()
+IPAddress::IPAddress() : IPAddress(IPv4) {}
+
+IPAddress::IPAddress(IPType ip_type)
 {
-    _address.dword = 0;
+    _type = ip_type;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
 }
 
 IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet)
 {
-    _address.bytes[0] = first_octet;
-    _address.bytes[1] = second_octet;
-    _address.bytes[2] = third_octet;
-    _address.bytes[3] = fourth_octet;
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX] = first_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 1] = second_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 2] = third_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 3] = fourth_octet;
+}
+
+IPAddress::IPAddress(uint8_t o1, uint8_t o2, uint8_t o3, uint8_t o4, uint8_t o5, uint8_t o6, uint8_t o7, uint8_t o8, uint8_t o9, uint8_t o10, uint8_t o11, uint8_t o12, uint8_t o13, uint8_t o14, uint8_t o15, uint8_t o16) {
+    _type = IPv6;
+    _address.bytes[0] = o1;
+    _address.bytes[1] = o2;
+    _address.bytes[2] = o3;
+    _address.bytes[3] = o4;
+    _address.bytes[4] = o5;
+    _address.bytes[5] = o6;
+    _address.bytes[6] = o7;
+    _address.bytes[7] = o8;
+    _address.bytes[8] = o9;
+    _address.bytes[9] = o10;
+    _address.bytes[10] = o11;
+    _address.bytes[11] = o12;
+    _address.bytes[12] = o13;
+    _address.bytes[13] = o14;
+    _address.bytes[14] = o15;
+    _address.bytes[15] = o16;
 }
 
 IPAddress::IPAddress(uint32_t address)
 {
-    _address.dword = address;
+    // IPv4 only
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    _address.dword[IPADDRESS_V4_DWORD_INDEX] = address;
+
+    // NOTE on conversion/comparison and uint32_t:
+    // These conversions are host platform dependent.
+    // There is a defined integer representation of IPv4 addresses,
+    // based on network byte order (will be the value on big endian systems),
+    // e.g. http://2398766798 is the same as http://142.250.70.206,
+    // However on little endian systems the octets 0x83, 0xFA, 0x46, 0xCE,
+    // in that order, will form the integer (uint32_t) 3460758158 .
 }
 
-IPAddress::IPAddress(const uint8_t *address)
+IPAddress::IPAddress(const uint8_t *address) : IPAddress(IPv4, address) {}
+
+IPAddress::IPAddress(IPType ip_type, const uint8_t *address)
 {
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+    _type = ip_type;
+    if (ip_type == IPv4) {
+        memset(_address.bytes, 0, sizeof(_address.bytes));
+        memcpy(&_address.bytes[IPADDRESS_V4_BYTES_INDEX], address, sizeof(uint32_t));
+    } else {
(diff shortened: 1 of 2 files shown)


sgryphon:
I have not been able to run the tests locally, but did get them running (and now all successful / green) in my GitHub repo. Link to the successful run is here: https://github.com/sgryphon/arduino-esp32/actions/runs/2931290749

This PR says it needs a contributor to approve first-time running of the workflows.

sgryphon:
Concern: Regarding the != operator. 

* Usually the == and != operators should be specified together, as often they are used interchangeably, e.g. I would expect "!(a == b)" to give the same result as "a != b" (although not always the case, e.g. nullability in SQL).
* The != operator does not exist in the current IPAddress.
* When I add it, it fails some tests where the is ambiguity of "ip != 0" between implicit conversion of IPAddress to int, and then compare two int, vs conversion of 0 to IPAddress, and then comparison of IP address.
* To fix this, I think that the conversions could be marked explicit.

While adding the != operator is backwards compatible (it adds more), changing the existing conversion to be explicit would be a (fairly small) change in existing functionality.

Personally I think the missing != is a bug, that needs to be fixed, even if it breaks existing implicit conversion.

mrengineer7777:
I see ArduinoCore does not have an IPv6 library.  Opinion: If we are added v6 support to IPAddress, we should delete IPv6Address library.

sgryphon:
> Opinion: If we are adding v6 support to IPAddress, we should delete IPv6Address

That would be a breaking change. It is okay to do this for a major version, but usually you would want to have the replacement already added (i.e. add this backwards compatible change first), then foreshadow & let people know (mark the old code deprecated, if the language supports this, and tell them to use the new). 

Then a future version would remove the code.

e.g. a plan could be to add the IPv6 support first; if that is going to be release v3.0.0, then add a comment that IPv6Address will be removed in v4.0.0 and to use IPAddress instead (and then remove it in v4).

me-no-dev:
@sgryphon are you around to rebase this? And if possible, update it to the latest API from Arduino? We would like to merge it now :)

sgryphon:
> @sgryphon are you around to rebase this? And if possible, update it to the latest API from Arduino? We would like to merge it now :)

Thanks for the ping. I will have a look, but shouldn't take too long.

me-no-dev:
>  I will have a look, but shouldn't take too long.

Just copying the latest IPAddress from ArduinoCore-API should be sufficient :) there are a couple changes since you submitted this PR.

sgryphon:
I have rebased, and added the recent (minor) changes from ArduinoCore... but I see you now have a commit message linter, so will also need to clean them up (I may get time over the weekend, let me see).

me-no-dev:
> but I see you now have a commit message linter

Don't worry, we are still experimenting with it :) Thanks for updating the files!

sgryphon:
I've updated the commit messages to match the linter format.
BTW when I did the initial rebase I also squashed.
Before merging I did do a three-way compare and nothing had changed in arduino-esp32 since; and then I applied the changes from arduinocore api (some, e.g. toString) already existed in -esp32

me-no-dev:
@sgryphon could you please sign the CLA (second message in the thread)

sgryphon:
> @sgryphon could you please sign the CLA (second message in the thread)

CLA is signed.

sgryphon:
If anyone wants to have a discussion on wider IPv6 (not just IPAddress, which should be aligned with ArduinoCore), I created a general IPv6 discussion that can be used: https://github.com/espressif/arduino-esp32/discussions/9009

s-hadinger:
Let me explain more for those who are not familiar with link-local IPv6. Link-local addressed are used extensively by Matter protocol.

If you have multiple network interfaces, there is no way to for sure which network interface is used by which link-local addresses. Hence an additional information `zone` is used to track which interface must be used.

In text representation, they are separate with `%`, example: `fe80::86cc:a8ff:fe64:b768%st1` is a link-local address on wifi `st1`, `fe80::86cc:a8ff:fe64:c768%en2` is on Ethernet

me-no-dev:
@s-hadinger check and comment https://github.com/espressif/arduino-esp32/pull/9016 please

TD-er:
There is something still wrong with the implementation of this code.

When I try to get the client IP (`webserver.client().remoteIP()`)  I now only get the last byte.
Thus 192.168.10.129 is now shown as 129.0.0.0

Looks like an alignment issue, so maybe on another build with only minor changes it might be shifted differently?

sgryphon:
> PR that doesn't take into account

Small steps.

That PR was mostly written 1 year ago, to align with ArduinoCore (which does not have zone ID). ArduinoCore in turn was actually focussed for mobile (4G) devices, with a single interface assigned a global address, because carriers are moving to IPv6 only networks... so quite a different case that Matter / Thread.

The new PR (linked) integrates in the zone ID ... and I suggesting contributing it back to ArduinoCore, for consistency between platforms.

I also added a section to the discussion I created https://github.com/espressif/arduino-esp32/discussions/9009

Because yes, people misunderstand things like link-local, e.g. confusing it with local, as opposed to remote (the ends of a connection) and link-local vs global (different scopes).

PR #7207, opened 2022-08-31
[NORA-W1] Fix SPI pin assignments to match Arduino pinout.

mazgch:
## Description of Change
- fixing the default pin assignment for the SPI peripheral for target added Yesterday.  
 
## Tests scenarios
- tested on EVK-NORA-W10

### DIFF
--- a/variants/nina_w10/pins_arduino.h
+++ b/variants/nina_w10/pins_arduino.h
@@ -23,10 +23,10 @@ static const uint8_t RX = 3;
 static const uint8_t SDA = 12;
 static const uint8_t SCL = 13;
 
-static const uint8_t SS    = 5;
-static const uint8_t MOSI  = 23;
-static const uint8_t MISO  = 19;
-static const uint8_t SCK   = 18;
+static const uint8_t SS    = 34;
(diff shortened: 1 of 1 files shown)


variants/nina_w10/pins_arduino.h
```diff
@@ -23,10 +23,10 @@ static const uint8_t RX = 3;
 static const uint8_t SDA = 12;
 static const uint8_t SCL = 13;
 
-static const uint8_t SS    = 5;
-static const uint8_t MOSI  = 23;
-static const uint8_t MISO  = 19;
-static const uint8_t SCK   = 18;
+static const uint8_t SS    = 34;
```
atanisoft, line 26: shouldn't this change be being made in variants/nora_w10/pins_arduino.h that was added via https://github.com/espressif/arduino-esp32/pull/7191/files?

mazgch:
sorry I apologise my fault.

variants/nina_w10/pins_arduino.h
mazgch: Thanks for letting me know, my mistake, here is the correct fix. 
#7208

PR #7208, opened 2022-08-31
[NORA-W1] Fix SPI pin assignments to match Arduino pinout.

mazgch:
## Description of Change
- fixing the default pin assignment for the SPI peripheral for new NORA-W10 added recently.  So that they match the GPIOs connected to D10-D13 of the Arduino compatible pin headers on the EVK.
 
## Tests scenarios
- tested on EVK-NORA-W10


mazgch:
This is a fix to #7191 that was merged by @me-no-dev . would be nice to add it in next release so that target if fixed from beginning.

P-R-O-C-H-Y (approved):

PR #7214, opened 2022-09-02
Update Arduino as ESP-IDF documentation: How to use USBHID classes with ESP32-S2 and ESP32-S3

RefactorFactory:
## Description of Change
When using Arduino as an ESP-IDF component, one needs to do special setup if you want to use arduino-esp32 USBHID classes. This explains how.

These instructions are based on esp32-arduino-lib-builder's build process, including https://github.com/espressif/esp32-arduino-lib-builder/blob/master/tools/update-components.sh which explains how to clone tinyusb.

## Tests scenarios
I have tested my Pull Request on ESP-IDF v4.4.1 and Arduino-esp32 core v2.0.4 with ESP32-S2 Board with this scenario.


VojtechBartoska:
Thanks for the PR @RefactorFactory.

@P-R-O-C-H-Y Can you please help with the review as you contributed to this part of Documentation in the past? Thanks

me-no-dev (approved):

P-R-O-C-H-Y (approved):

PR #7221, opened 2022-09-05
Add new ULP example for ESP32

SuGlider:
## Description of Change
It adds a new ULP example for ESP32 that demonstrates how to create ULP code with Arduino.
The Code executes a smooth blink on the GPIO2 which is executes in background by the ESP32 ULP.
There is a RTC MEM parameter used to tell the ULP code the Blinking Speed.

## Tests scenarios
Tested on the ESP32 with a builtin LED in the GPIO2

## Related links
None.


P-R-O-C-H-Y (approved):
Example works great, good job @SuGlider.

Issue #5002, opened 2021-04-01
(SD/FS) Batch deletion slowed by openNextFile()


brunojoyal:
### Hardware:
Board:							ESP32 DEVKIT V1
Core Installation version:		1.0.6
IDE name:						PlatformIO
Peripherals:                    SD Card (SPI)
Partition Scheme:               FAT32

### Description:

Batch deletion of files is very slow due to having to rely on openNextFile().

Consider the script below which:

(1) populates the SD card with 500 2kb gibberish files when 'a' is pressed;
(2) removes 500 files when 'b' is pressed.

The result when pressing 'a', and then 'b' when that is finished, is: 

```
...
496: Removed /TEST/724092.TXT 48172 ms
497: Removed /TEST/4778.TXT 48345 ms
498: Removed /TEST/669317.TXT 48518 ms
499: Removed /TEST/533569.TXT 48691 ms
Using openNextFile
Total time 78663 ms, 48691 ms of which were spent removing files.
```

Much of the time spent is not spent removing files but just opening them...

This was partly addressed around 2018-2019 in the discussion following issue #1394 (https://github.com/espressif/arduino-esp32/issues/1394). The patch provided by driftregion at the time (https://github.com/espressif/arduino-esp32/issues/1394#issuecomment-511379408) significantly improves this process and it seems to have unfortunately fallen through the cracks, with none of these changes ever being implemented. 
I have cloned the arduino-esp32 framework into /brunojoyal/arduino-esp32-sdfix in which I replaced the FS library with the one patched by driftregion. Putting

```
[env]
framework = arduino
platform_packages = framework-arduinoespressif32 @ https://www.github.com/brunojoyal/arduino-esp32-sdfix.git
```

in platformio.ini to use this framework, and uncommenting the #def SDFIX line in my test script, we get this result:

```
494: Removed /TEST/241878.TXT 48015 ms
495: Removed /TEST/332820.TXT 48189 ms
496: Removed /TEST/795255.TXT 48362 ms
497: Removed /TEST/348514.TXT 48536 ms
498: Removed /TEST/772893.TXT 48710 ms
499: Removed /TEST/188217.TXT 48884 ms
Using getNextFileName
Total time 49922 ms, 48884 ms of which were spent removing files.
```

Much better.

Question/Request: Would it be possible to include some or all of the changes made by driftregion to the FS library? Most importantly, the getNextFileName() method, or something similar.

Many thanks.

### Sketch:  
```cpp

[...]

PR #7229, opened 2022-09-06
Implements seekDir and getNextFileName on FS Lib to improve performance

Lucasczm:
## Description of Change
Implemented two functions: 
 - getNextFileName:  Get the next file name with better performance than getNextFile function. To improve file listing.
 - seekDir: Set the position of directory for improve search.

Examples: 
```cpp
 File root = SD.open("/data");
 String filename;
 //Return true if have next file, false if not
 if (root.getNextFileName(filename)) {
     Serial.println(filename); //print "file1"
 }
```

## Tests scenarios
Tested on PlatformIO and Arduino core v2.0.4 with  ESP32 LILYGO PoE Board

## Related links

#5002


me-no-dev:
For `getNextFileName` why don't you return the filename directly (or empty String if not available) instead?

libraries/FS/src/FS.cpp
```diff
@@ -187,6 +187,22 @@ File File::openNextFile(const char* mode)
...
+    if(!_p){
+        return false;
+    }
+    return _p->seekDir(position);
+}
+
+boolean File::getNextFileName(String &filename)
+{
+    if (!_p) {
+        return NULL;
```
SuGlider, line 200: I see an inconsistency here given that the function `getNextFileName()` shall return a `boolean`.
it is better to `return false` instead.

SuGlider (changes requested):

libraries/FS/src/FS.cpp
me-no-dev: @SuGlider see my comment above. It's better to return `String` directly, not false or NULL.

SuGlider: Ok, it make the API easier returning a String as you proposed.

PilnyTomas:
Tested using [this sketch](https://github.com/PilnyTomas/issues/blob/main/PR_7729/PR_7729.ino) confirming that the new function is way much faster:
new function ~0.24ms per folder read
old function `openNextFile` ~142ms per folder read

PilnyTomas (approved):
Apart from the requested changes looks good.

libraries/FS/src/FS.cpp
Lucasczm: i maked the changes, now its return `String` name directly or String empty

Lucasczm (commented):
Now its return String name directly or String empty

tueddy:
The speed increase is amazing! My webserver based SD fileexplorer needs 6.8 seconds with getNextFile(), now it's just 100ms.

But how can we decide wether it's a file or a (sub-)directory? My explorer JavaScript needs the information "isDir=false/true" 
I would expect the name has a trailing slash when it's a directory.

Or can it enhanced with a similar function getNextPathName() ?

tueddy:
The speed improvement is amazing! For a webserver SD filexplorer the time for listing is currently about 2 seconds (very blocking), with this PR only 20 msecs (hundreds of mp3 files on SD card). Works fine here!

But for the Javascript based file explorer tree i also need the information wether it's a file or directory. Is there a way to get it?
Suggestion: add a trailing slash in filename if it's a directory or add an overload function with out-param isDirectory.
Some other information like size, creation date,  last mod-date would also be helpful for sorting the tree. 

Thank's!

tueddy:
I made a speedtest example for PIO here https://github.com/tueddy/FS/tree/main/example
Tested my SD-cards with a lot of files in root directory:


Samsung Evo 32GB:
getNextFileName(), done reading root-directory 36 elemts in 18 ms = 0.50 ms/folder
openNextFile(), done reading root-directory 36 elements in 492 ms = 13.67 ms/elements

Kingston 8GB, a bit older card:
getNextFileName(), done reading root-directory 42 elemts in 29 ms = 0.69 ms/folder
openNextFile(), done reading root-directory 42 elements in 1119 ms = 26.64 ms/elements

Transcend 16GB:
getNextFileName(), done reading root-directory 84 elemts in 83 ms = 0.99 ms/folder
openNextFile(), done reading root-directory 84 elements in 6960 ms = 82.86 ms/elements

Scandisk 16GB:
getNextFileName(), done reading root-directory 151 elemts in 68 ms = 0.45 ms/folder
openNextFile(), done reading root-directory 151 elements in 6872 ms = 45.51 ms/elements

me-no-dev (approved):

me-no-dev:
@SuGlider waiting on your approval here :)

P-R-O-C-H-Y (approved):

libraries/FS/src/vfs_api.cpp
```diff
@@ -476,6 +476,36 @@ FileImplPtr VFSFileImpl::openNextFile(const char* mode)
...
+    if (file->d_type != DT_REG && file->d_type != DT_DIR) {
+        return "";
+    }
+    String fname = String(file->d_name);
+    String name = String(_path);
+    if (!fname.startsWith("/") && !name.endsWith("/")) {
+        name += "/";
+    }
+    name += fname;
+    return name;
```
vortigont, line 506: just a small hint, this could be reduced to avoid a couple of extra copy's

```c++
    String fname(file->d_name);
    if (!fname.startsWith("/") && !name.endsWith("/")) {
        name += "/";
    }
    return String(_path)+fname;
```

vortigont (commented):
a small hint for String's

libraries/FS/src/vfs_api.cpp
mrengineer7777: @vortigont Your change loses `String name` declaration, and then doesn't use name in return. BUG.

vortigont: right, it's the `name` in second condition. my bad, I should have my glasses on :) tnx!

SuGlider (approved):
@Lucasczm - Thanks for the PR! Very good.

SuGlider:
@Lucasczm - I can't rebase it from here. Please do it in your end in order to be able to merge it. Thanks.

tueddy:
Can't test it just at the moment, but can I distinguish whether in the listing is a file or directory with this commit?
Made a comment [above](https://github.com/espressif/arduino-esp32/pull/7229#issuecomment-1257071883)..

Thank's for your work!

BlueAndi:
@tueddy The seekDir() implementation will only be useful if the other functions are available in FS and VFS too, see
https://github.com/espressif/esp-idf/blob/16a4ee7c36a848ca155791677ce011f3ca75c519/components/newlib/platform_include/sys/dirent.h#L51-L57

tueddy:
getNextFileName() works fine in 2.0.6, but how can i detect now if return value is a file or directory?

libraries/FS/src/vfs_api.cpp
```diff
@@ -476,6 +476,36 @@ FileImplPtr VFSFileImpl::openNextFile(const char* mode)
...
+    if (!_isDirectory || !_d) {
+        return "";
+    }
+    struct dirent *file = readdir(_d);
+    if (file == NULL) {
+        return "";
+    }
+    if (file->d_type != DT_REG && file->d_type != DT_DIR) {
+        return "";
+    }
```
nomis, line 499: This behaviour is inconsistent with `openNextFile()` which skips to the next file.

It's not possible to tell the difference between "end of file list" and "something that is not a file or directory".

libraries/FS/src/vfs_api.cpp
```diff
@@ -476,6 +476,36 @@ FileImplPtr VFSFileImpl::openNextFile(const char* mode)
...
+    String fname = String(file->d_name);
+    String name = String(_path);
+    if (!fname.startsWith("/") && !name.endsWith("/")) {
+        name += "/";
+    }
+    name += fname;
```
nomis, line 505: This is the "path", not the "name" of the file.

This results in more effort by users of this function to remove the directory name (which they already have) from every filename.

(The `openNextFile()` function returns a `File` so it's possible to use `name()` or `path()`)

PR #7258, opened 2022-09-15
Update pins_arduino.h for D-Duino32

alessandromrc:
changed static const uint8_t to define macro as it may lead to issues when compiling like here:
https://github.com/filoconnesso/Tweakly/actions/runs/3060172324/jobs/4938372368

### Checklist

-----------
## Description of Change

## Tests scenarios

## Related links

### DIFF
--- a/variants/d-duino-32/pins_arduino.h
+++ b/variants/d-duino-32/pins_arduino.h
@@ -14,8 +14,8 @@
 static const uint8_t TX = 1;
 static const uint8_t RX = 3;
 
-#define SDA = 5
-#define SCL = 4
+#define SDA 5
+#define SCL 4
 
 static const uint8_t SS    = 15;
 static const uint8_t MOSI  = 13;
--- a/variants/d-duino-32/pins_arduino.h
+++ b/variants/d-duino-32/pins_arduino.h
@@ -14,8 +14,8 @@
 static const uint8_t TX = 1;
 static const uint8_t RX = 3;
 
-static const uint8_t SDA = 5;
-static const uint8_t SCL = 4;
+#define SDA = 5
+#define SCL = 4
 
 static const uint8_t SS    = 15;
 static const uint8_t MOSI  = 13;


me-no-dev:
first issue I see is that the library is building agains version 1.0.6, which is really old at this point. Should build against master or similar. Also I see no reason why it will give this error, if both are `static const uin8_t`

alessandromrc:
> first issue I see is that the library is building agains version 1.0.6, which is really old at this point. Should build against master or similar. Also I see no reason why it will give this error, if both are `static const uin8_t`

the fact about using define is probably because the value with static const uin8_t is assigned in runtime while the macro define is in compile time as it is a compiler macro

me-no-dev:
> Ps. seems like that https://dl.espressif.com/dl/package_esp32_index.json returns the 1.0.6 version of the core

That is a very old link for json. Please see https://docs.espressif.com/projects/arduino-esp32/en/latest/installing.html#installing-using-arduino-ide

alessandromrc:
> > Ps. seems like that https://dl.espressif.com/dl/package_esp32_index.json returns the 1.0.6 version of the core
> 
> That is a very old link for json. Please see https://docs.espressif.com/projects/arduino-esp32/en/latest/installing.html#installing-using-arduino-ide

I'm actually going to update it but shouldn't it be updated there too?

me-no-dev:
> but shouldn't it be updated there too?

Not that simple anymore. JSONs are updated by CI and current link is our best option. DL server has been changed.

alessandromrc:
> 

Isn't the best solution just redirecting that link to the current json configuration? It would prevent errors like this forever

alessandromrc:
> first issue I see is that the library is building agains version 1.0.6, which is really old at this point. Should build against master or similar. Also I see no reason why it will give this error, if both are `static const uin8_t`

Anyway I'm going to close this PR as the issue is practically resolved but still I hope that the dl link will get updated for future users.

PR #7278, opened 2022-09-20
Fix call to HTTPClient::setTimeout

everslick:
Public facing network classes use seconds instead of ms for setTimeout().


bertmelis:
Also see #6676

PR #7284, opened 2022-09-21
Fixes ESP32-S2 CDC Debug Logging

### LINKED ISSUE #6766: Issue realted ESP32-S2 CDC to the Serial.setDebugOutput(true);
I am starting to think that the current cause of the problem (on v2.0.3 stable) is some kind of race/lockup between two or more tasks, all of which want to write to the USB CDC. In my case, this happens due to concurrent use of log_X statements in more than one task. This scenario works correctly when using native serial pins.

Consider the following sketch:

```c++
#include <Arduino.h>

void setup()
{
  Serial.begin(115200);
  // setDebugOutput() call commented out for test
  //Serial.setDebugOutput(true);
  delay(2000);
}

void loop()
{
  static uint32_t counter = 1;

  Serial.printf("Using Serial.printf(), this is line %u of output, at msec %u...\r\n", counter, millis());
  //log_i("Using log_i(), this is line %u of output, at msec %u...", counter, millis());
  counter++;

  // No delay at all, blast port at full speed!
}
```

This is a single task using the serial port. When using USB CDC as the serial port, no combination of use or non-use of setDebugOutput and Serial.printf/log_i manages to lockup the board.

However, my sketches use WiFi, which starts a couple of tasks, each outputting via log_X. Also, AsyncTCPSock, which starts its own task. With this, I have noticed that the startup process, where all of these tasks display debugging strings, is the more likely place where the sketch hangs on output. After this, only the main loop keeps outputting strings, and it runs more or less normally.

_Originally posted by @avillacis in https://github.com/espressif/arduino-esp32/issues/6221#issuecomment-1124215183_

I have been reviewing the code of USBCDC.cpp , which handles the USB serial port in ESP32-S2. So far I have noticed the following:
- The setDebugOutput() call installs a function, `cdc0_write_char` as a character output for debugging with `ets_install_putc1`.
- The `cdc0_write_char` function, in turn, calls `USBCDC::write(uint8_t)`, and through it, `USBCDC::write(const uint8_t *, size_t)`.
- The `USBCDC::write(const uint8_t *, size_t)` appears to acknowledge that it might be called from within an ISR context, as it checks with `xPortInIsrContext()`, and uses `xSemaphoreTakeFromISR()` or `xSemaphoreTake()` depending on the result.
- According to the Espressif documentation, `xSemaphoreTakeFromISR()` outputs a flag, `pxHigherPriorityTaskWoken`, that is supposed to inform whether a higher-priority task was awoken as a result of calling the ISR-specific function. And if so, the function is supposed to request a context switch by calling `portYIELD_FROM_ISR()`.
- HOWEVER... the current implementation of `USBCDC::write(const uint8_t *, size_t)` collects this flag, only to discard it without taking any action based on its value, and certainly without calling `portYIELD_FROM_ISR()` anywhere in the function body.

So, what possible consequences might arise from failing to invoke`portYIELD_FROM_ISR()` as indicated in the documentation?

https://github.com/espressif/arduino-esp32/issues/6221#issuecomment-1125187189

SuGlider:
## Description of Change
When using ESP32-S2 with USB CDC and debug output is set also to go to the CDC (`Serial.setDebugOutput(true)`), the S2 freezes after a few line of logging.

This issue only occurs with the ESP32-S2. It works fine with ESP32-S3 and CDC.

## Tests scenarios
Tested with this sketch on the ESP32-S2 and ESP32-S3. IDE must set `USB CDC On Boot: "Enabled"`

``` cpp
#include <Arduino.h>

static void test_taskfunc(void * p)
{
    uint32_t tasknum = (uint32_t)p;

    while (true) {
        log_i("This is task instance %u running...", tasknum);
        
        delay(10 * random(1, 10));
    }
}

void setup()
{
    Serial.begin(115200);
    Serial.setDebugOutput(true);
    delay(2000);
    Serial.println("\n==============\nStarting...\n");
    log_e("Starting with LOGs.");
    Serial.println("==============\n\n");
    delay(750);

    for (auto i = 0; i < 5; i++) {
        TaskHandle_t th;

        // Created tasks all need to have different priorities in order to trigger lockup
        UBaseType_t taskPrio = 1 + 2 + i;

        if (pdPASS == xTaskCreate(test_taskfunc, "test_taskfunc", 4096, (void *)(i + 1), taskPrio, &th)) {
            log_i("Task instance %u created with priority %u", i+1, 3 + i);
        } else {
            log_e("Task instance %u creation FAILED", i+1);
        }
    }
}

void loop()
{
    log_i("This is the main loop running...");

    delay(10 * random(1, 10));
}

```

## Related links
Fixes #6766


avillacis:
The good news: After applying this change, on my ESP32-S2 dev board, redirecting debug output to the USB CDC does not hang the sketch anymore. The sketch keeps running without interruption.

The (tolerably) bad news: It appears that some messages are not completely flushed to the opened serial terminal, so that the strings appear truncated. However, when more data is written with the log_X functions, the rest of the apparently-truncated string is emitted before the next message appears.

SuGlider:
> The (tolerably) bad news: It appears that some messages are not completely flushed to the opened serial terminal, so that the strings appear truncated. However, when more data is written with the log_X functions, the rest of the apparently-truncated string is emitted before the next message appears.

I guess that this is because the example has 2 tasks, writing to the CDC at the same time and one may interrupt the other while still writing the log, character by character, truncating the previous message by another task.

Maybe the best way to solve it is by creating a Message Queue within the sketch and a task that consumes this queue, printing one log message at a time, in order to make sure that it will not get truncated. In this case the tasks shall enqueue the messages using a specific API, which could also sort them by some critical ordering method, for instance.

VojtechBartoska:
**PR Update:** Investigate why function `tud_cdc_n_write_available` freeze?

brentru:
@SuGlider @VojtechBartoska  I am experiencing a similar lockup on the latest BSP (due to `setDebugOutput(true)` which I need for debugging). This PR resolves this issue.

What is the status of this PR in general, and the status of it making it into the next release?

paclema:
I can confirm too that this helps with ESP32-S2! When is it planned to be merged?

SuGlider:
@me-no-dev - Retested. The PR is necessary to use USB OTG CDC logging in ESP32-S2.

SuGlider:
@me-no-dev - This PR should be applied to branch master and to release/v2.x to make sure it goes to a future 2.0.15

PR #7301, opened 2022-09-26
[Termod S3] fix keep rebooting reading CHG

TAMCTec:
## Description of Change
Fix Termod S3 keep rebooting while reading CHG

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.5 with Termod S3 with this scenario

### DIFF
--- a/variants/tamc_termod_s3/variant.cpp
+++ b/variants/tamc_termod_s3/variant.cpp
@@ -15,24 +15,10 @@ float getBatteryCapacity() {
 }
 
 bool getChargingState() {
-  pinMode(CHG, INPUT_PULLUP);
   return !digitalRead(CHG);
 }
 
-void (*__onChargeStart__)();
-void (*__onChargeEnd__)();
-void setOnChargeStart(void (*func)()) { __onChargeStart__ = func; }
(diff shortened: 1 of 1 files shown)


variants/tamc_termod_s3/variant.cpp
```diff
@@ -15,24 +15,10 @@ float getBatteryCapacity() {
...
-  if (getChargingState()) {
-    __onChargeStart__();
-  } else {
-    __onChargeEnd__();
-  }
-}
-
 extern "C" void initVariant(void){
-  attachInterrupt(CHG, chargeIsr, CHANGE);
+  pinMode(CHG, INPUT_PULLUP);
```
me-no-dev, line 22: what would happen if you keep everything original, but you remove `pinMode(CHG, INPUT_PULLUP);` from `getChargingState()` and put it before `attachInterrupt(CHG, chargeIsr, CHANGE);` in `initVariant(void)`?

TAMCTec: Yes! It worked! Thank you. It keeps rebooting before and i didn't know why. And ISR is not that important, so i just remove it. Thanks a lot! Will commit another pull request.

PR #7351, opened 2022-10-14
Fix unnecessary DNS query in hostByName and deadlock in ssl_client

### LINKED ISSUE #7350: DNS query is issued in ssl_client's start_ssl_client method, even when IP address is provided
No details about the device required as whole ecosystem is affected.

Affected component: WiFiClientSecure, ssl_client and probably more...

### Description

When using WiFiClient secure and passing IPAdddress, unnecessary DNS query flow is issued. Why is it ever calling hostByName function? That's probably caused by confusion of mbedtls hostname with endpoint hostname. On Arduino-esp8266 NULL is passed as hostname in this situation.

Expected resoultion:
When providing raw IP address, no DNS query should be issued. This can be partially fixed by #7351.
Hostname should be NULL (like in esp8266 Arduino core) or should be able to be defined by the user.

### Sketch

```cpp
WiFiClientSecure clientSecure;
clientSecure->setInsecure(); // << -- to allow self signed and skip some verification

IPAddress serverIp;
if (serverIp.fromString("192.168.1.111"))
{
   clientSecure.connect(serverIp, 443);  // << -- this will trigger unnecessary DNS querying for IP address.
}
```

cziter15:
## Description of Change

Pull request to fix problems found when auditing WiFiClientSecure implementation used by ksIotFrameworkLib.

I have few IoT devices running 24/7 with MQTT. Recently added SSL support and identified that unit based on ESP32 is unable to reconnect after losing connection. I've been looking at WiFi code handling in my library, found some bugs, fixing them improved situation a little bit, but still sometimes device was unable to get back online.

After hours of testing by trial-and-error remotely, I have decided to get pcb on my desk and quickly identified few bugs related to WiFiClientSecure.

This PR aims to fix these bugs:
- Fixes possibility to stuck inside while loop in function send_ssl_data. When retry is triggered and then internet connection dies, buffer can be overloaded with data causing execution to block there. Adding timeout prevents from stucking forever.
- Removes unnecessary DNS query flow, when IP address is passed as a hostname inside hostByName function. Generally speaking there is a strange flow of coverting IPAddress (when using IPAddress variant of connect method) to a string, then passing it to hostByName function in WiFiClientSecure to... get IPAddress back, involving whole mechanism of waiting for DNS response.

## Tests scenarios
Tested on custom board with ESP-WROOM-32 module without PSRAM. Latest master with applied changes. Looks like ESP32 device is getting back online like other ESP8266 based units.

## Related links
Closes #7350 #7356


cziter15:
Anyway, it looks like handshake_timeout might not be required to be explicitly defined. Maybe handling it all under socket timeout is a good way?

me-no-dev (approved):

SuGlider (approved):
Nice PR, @cziter15 ! Thanks!

PR #7361, opened 2022-10-15
Add missing function getsocketoption

rtpmsys:
WiFimanager.cpp connect function truncated all connnect timeout values below 1000ms to 0ms and rounded timeouts to the full second, e.g. 2.5 seconds became 2 seconds. For timeouts below 1000ms,  the connect function failed. This fix handles timeouts correctly, by putting the fractional part of the timeout into the timeout_us variable.

Added function getsocketoptions, it was missing.

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -231,7 +231,7 @@ int WiFiClient::connect(IPAddress ip, uint16_t port, int32_t timeout)
     FD_ZERO(&fdset);
     FD_SET(sockfd, &fdset);
     tv.tv_sec = _timeout / 1000;
-    tv.tv_usec = 0;
+    tv.tv_usec = (_timeout  % 1000) * 1000;
 
 #ifdef ESP_IDF_VERSION_MAJOR
     int res = lwip_connect(sockfd, (struct sockaddr*)&serveraddr, sizeof(serveraddr));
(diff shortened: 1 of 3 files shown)


me-no-dev:
This PR seems to include two non-relevant changes. Could you please split the timeout fix and `getSocketOption` into separate PRs?

TD-er:
Why keep on insisting to have a `WiFiClient::setTimeout` function in this class using seconds while every other piece of code uses msec?
Cast it to `Client` and call that timeout will expect msec.

rtpmsys:
I submitted only the timeout correction and omitted the 

> This PR seems to include two non-relevant changes. Could you please split the timeout fix and `getSocketOption` into separate PRs?

I submitted just the timeout change which fixes the error with connect timeout values below 1 second.

rtpmsys:
> Any updates @rtpmsys?

I submitted a new pull request with just the timeout fix. Hope it makes it through the system, now.

rtpmsys:
> Why keep on insisting to have a `WiFiClient::setTimeout` function in this class using seconds while every other piece of code uses msec?
> Cast it to `Client` and call that timeout will expect msec., which is a clear indication this design choice in the past was wrong.

You are right that the setTimeout function only expects seconds, I guess for compatibility reasons? The setTimeout luckily only seems to affect the write and read from the socket, not the connect.
My proposed fix deals with the connect timeout, which comes before the write or read, therefore it is much more important that it can be made smaller than 1 second.

mrengineer7777:
@rtpmsys #7686 was merged.  Time to update this PR?

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -231,7 +231,7 @@ int WiFiClient::connect(IPAddress ip, uint16_t port, int32_t timeout)
     FD_ZERO(&fdset);
     FD_SET(sockfd, &fdset);
     tv.tv_sec = _timeout / 1000;
-    tv.tv_usec = 0;
```
rtpmsys, line 234: Moved into another PR

rtpmsys:
I will submit a new PR for adding the missing function getSocketOption.

PR #7372, opened 2022-10-19
Fix BluetoothSerial thinking it's disconnected

aridet:
Currently, if a second connection occur in Bluetooth, this second connection is automatically disconnected (line 284). This disconnection trigger the ESP_SPP_CLOSE_EVT, which clear the bit SPP_CONNECTED. But the first connection remain active, and this flag shouldn't be cleared.

This fix the issue, by clearing the flag only if the last connection is closed

### Checklist

-----------
## Description of Change

## Tests scenarios

## Related links


me-no-dev:
Shouldn't similar thing be necessary for line 286 and 287? They would also alter the bits, regardless if connection is first or second.

VojtechBartoska:
First of all, thank you for the PR. Any updates on this @aridet?

aridet:
Hi,

I don't think lines 286 and 287 are affected by this as we are in the case ESP_SPP_SRV_OPEN_EVT. It is fired when a connection successfully occur. This disable SPP_DISCONNECTED (xEventGroupClearBits disable the flag) and enable SPP_CONNECTED (xEventGroupSetBits set the flag). This seems correct to me.
If we are first or second connection, the result is always the same, there is an active connection, so the flags are correct for me.

Regards.

P-R-O-C-H-Y (approved):

P-R-O-C-H-Y:
> Hi,
> 
> I don't think lines 286 and 287 are affected by this as we are in the case ESP_SPP_SRV_OPEN_EVT. It is fired when a connection successfully occur. This disable SPP_DISCONNECTED (xEventGroupClearBits disable the flag) and enable SPP_CONNECTED (xEventGroupSetBits set the flag). This seems correct to me. If we are first or second connection, the result is always the same, there is an active connection, so the flags are correct for me.
> 
> Regards.

Make sense, thanks for this PR. Approved 👍

Issue #6921, opened 2022-06-28
No real-time serial.read() possible


OekoSolveMG:
### Board

M5 Stack Core 1 / 2

### Device Description

[M5 Stack Core 1 / 2](https://docs.m5stack.com/en/core/core2) with the [LAN Module with W5500](https://docs.m5stack.com/en/base/lan_v12)

### Hardware Configuration

[Pinmap M5 Core 2](https://docs.m5stack.com/en/core/core2)
[Pinmap Lan module](https://docs.m5stack.com/en/base/lan_v12)

### Version

v2.0.3

### IDE Name

PlatformIO

### Operating System

Windows 10

### Flash frequency

40 MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

Low baudrate (9600) causes the `serial.read()` command to fail and not read bytes on the Modbus. Might there be some interfering code that runs in timed cycles with a very high priority that could creates issues for the `serial.read()`.

![image](https://user-images.githubusercontent.com/94832869/176104052-3d884fa5-611a-4f39-9a60-9f4a46f6148e.png)

The failed read seen above happens every 6th read.

![image](https://user-images.githubusercontent.com/94832869/176105656-5c49660d-6777-422b-bc70-021a207f3bf2.png)

**Complete message:**

The third line is a pin that gets toggled to high before `serial.read()` and to low after  `seral.read()`.

![image](https://user-images.githubusercontent.com/94832869/176105885-50d8bb26-a384-4226-942f-db682f7e8ccb.png)

### Sketch

```cpp
// Allocate initial receive buffer size: 1 block of BUFBLOCKSIZE bytes
  const uint16_t BUFBLOCKSIZE(512);
  uint8_t *buffer = new uint8_t[BUFBLOCKSIZE];
  ModbusMessage rv;

  // Index into buffer
  register uint16_t bufferPtr = 0;
  // Byte read
  register int b; 

  // State machine states, RTU mode
  enum STATES : uint8_t { WAIT_DATA = 0, IN_PACKET, DATA_READ, FINISHED };

  // State machine states, ASCII mode
  enum ASTATES : uint8_t { A_WAIT_DATA = 0, A_DATA, A_WAIT_LEAD_OUT, A_FINISHED };

  register uint8_t state;

  // Timeout tracker
  unsigned long TimeOut = millis();

  // RTU mode?
  if (!ASCIImode) {
    // Yes.
    state = WAIT_DATA;
    // interval tracker 
[...]

Issue #6849, opened 2022-06-09
HardwareSerial not returning any read data if UART_BREAK_ERROR triggered


AcuarioCat:
### Board

ESP32 WROOM-32E

### Device Description

Plain module on pcb

### Hardware Configuration

GPIO21,22 connected to I2C

### Version

v2.0.3

### IDE Name

Visual Micro

### Operating System

Windows 11

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

I'm trying to read from an SDM120M modbus energy meter. The meter is returning data. The last byte is 00 but after the last bit the data line stays low causing a framing error. This triggers an RX break in the uartEventTask.

No data is then available (serialX.available() returns 0)
Finally after a few calls the UART FIFO fills up and the uart then returns 120 from serialX.available() and provides provides the 120 bytes (presumably as UART_FULL_THRESH_DEFAULT is triggered).

### Sketch

```cpp
Sketch is the sdm_live_page_esp32_hwserial.ino example from the SDM library :
https://github.com/reaper7/SDM_Energy_Meter

I added an onReceiveError callback to the SDM.cpp to trap the error.

void rxerr(int err) {
    Serial.printf("Receive error:%d\n", err);
}
```

### Debug Message

```plain
[1232656][W][HardwareSerial.cpp:256] _uartEventTask(): UART1 RX break.
Receive error:0
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

SuGlider:
>Finally after a few calls the UART FIFO fills up and the uart then returns 120 from serialX.available() and provides provides the 120 bytes (presumably as UART_FULL_THRESH_DEFAULT is triggered).

I understand that the BREAK event doesn't prevent the sketch from getting the data.
Could you also provide more details about the sketch used to read the SDM120M device?

Have you tried to used a library such as https://github.com/reaper7/SDM_Energy_Meter

AcuarioCat:
If it's of use - I just tried the same code but using SoftwareSerial instead of HardwareSerial, same GPIO pins (RX-26, TX-25) and it functions correctly. Change back to HardwareSerial and it fails.

SuGlider:
Let me try to reproduce this issue with 2 ESP32 (one receiving data and the other one simulating the SDM120M with a break at the end of the packet). I'll post here the findings.

SuGlider:
[...]

Issue #6689, opened 2022-05-04
UART Performance issues with 2.0.2 version


handmade0octopus:
### Board

Not board specific

### Device Description

Not board specific

### Hardware Configuration

Not board specific

### Version

v2.0.2

### IDE Name

PlatformIO 4.2.0

### Operating System

Windows 10

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

I have been using Platformio 3.5.0 without any issues for very long time.

I tried to moving on to version 4.2.0 based off 2.0.2 but it seems it has lots of underlying problems or it requires me to rewrite most of my code.

I can't share all of it as its quite advanced, instead I will provide you with few insights and code that had dramatic performance issues. I run all code on mostly tasks with priority 4-8 and less important ones being on 3 and 2.

Top is version 3.5.0 of Platformio (I don't know exactly which Arduino version it's based off - sorry).
Bottom is version 4.2.0
![Arduino Platformio 3.5.0 (top) to 4.2.0 (bottom) comparision](https://i.imgur.com/EpyfSfc.jpg)

Note - I left device to run for 5 minutes to make sure test is fair.

Displays are as follows (starting from left top corner):
* Log speed - is .CSV logging appending the file with ~30 parameters, speed is in rows/second - as you can see on 3.5.0 it was stable and flat and it's jittery on new version (but it does keep up.

* Max heap - I wouldn't worry about this one, but we can see change in max heap and it's much smaller than on previous version.

* RPS - it's "responses per second" - basically I use single wire serial connection (LIN type) to gather data from old typ ECU, here my performance is most important and as you can see it doesn't keep up with it. Maybe my code here is wrong as it's very complicated to provide great stability, but I have been using this one for few years and never has issues before.

It's worth mentioning I am using GPS at 5Hz data rate constantly sending data to my second Serial input.

* Free heap - here difference is huge and I can't believe it's only due to change from 3.5.0 to 4.2.0 - but I can't find any other culprit!

But the worst offender of all is the part of sketch below that I use to transfer data from/to SD card. Below is snippet of function I use to get those performance stats:

Before I managed to get constantly around 71 ms:
```
[E][Gauge.S.cpp:1536] handleCopyFile(): Time to read /displayConfig.json -     73 ms
```

But now it's not even funny:
```
[E][Gauge.S.cpp:1536] handleCopyFile(): Time to read /displayConfig.json -   1466 ms
```

[...]

PR #7412, opened 2022-10-31
Fixes inconsistencies and adds extended HardwareSerial examples

SuGlider:
## Description of Change
This PR adds some examples that use API extensions to HardwareSerial, implemented to better controls UART communication using ESP32 Arduino.

#### HardwareSerial changes:

UART FIFO Full will depend on the baud rate. 
For lower or equal to 57600 baud,  it will force to emulate byte-by-byte reading on UART ISR.
For higher baud rate, it will keep IDF default of 120 bytes for FIFO FULL Interrupt.
The FIFO Full parameter can also be changed by the application at any time using `HardwareSerial::setRxFIFOFull(uint8_t fifoBytes)`

It also fixes some inconsistencies found when using the examples. 

- When Rx timeout is Zero (disabled), there is only one possible option which is execute the callback when FIFO reaches 120 bytes
- In case that onReceive() shall work only with RX Timeout, FIFO shall be high. This is a work around for an IDF issue with events and low FIFO Full value (< 3)

## Tests scenarios
The examples have been tested using ESP32, ESP32S2, ESP32S3 and ESP32C3.

## Related links
#6921
#6689 
#6849


Jason2866:
The headline is missleading. It is not just adding examples.

SuGlider:
@me-no-dev @P-R-O-C-H-Y 
This PR is ready for your review. Thanks!

me-no-dev (approved):

P-R-O-C-H-Y (approved):

PR #7413, opened 2022-10-31
Added TOUCH test sketch + test script

P-R-O-C-H-Y:
## Description of Change
Added CI HW test for Touch peripheral.
Touch supported on ESP32, ESP32S2, ESP32S3.


tests/touch/touch.ino
```diff
@@ -0,0 +1,196 @@
...
+
+#if CONFIG_IDF_TARGET_ESP32S2
+  #define RELEASED_VALUE        10000 //10000- read value to pass test
+  #define PRESSED_VALUE         42000 //40000+ read value to pass test
+  #define INTERRUPT_THRESHOLD   30000
+#elif CONFIG_IDF_TARGET_ESP32S3
+  #define RELEASED_VALUE        25000 //25000- read value to pass test
+  #define PRESSED_VALUE         100000 //150000+ read value to pass test
+  #define INTERRUPT_THRESHOLD   80000
+#endif
```
me-no-dev, line 64: I would suggest to add another `#else` here and an error, in order to catch future chips with touch functionality.

P-R-O-C-H-Y: @me-no-dev Did you mean something like this? So the test will not compile or fail for new chip (so we now it needs to be adjusted)? All tests will fail with these values (chip is using TOUCH_V2).
```
#else //defaults for upcoming chips (needs to be adjusted)
  #define RELEASED_VALUE        -1     
  #define PRESSED_VALUE         0xFFFFFFFF 
  #define INTERRUPT_THRESHOLD   -1      
#endif
```

me-no-dev: could even be as easy as 
```cpp
#else
  #error Test not currently supported on this chip. Please adjust and try again
#endif
```

P-R-O-C-H-Y: Done :) best and easiest option. Thanks

PR #7415, opened 2022-10-31
[DOCS] Add ESP32S3 to libraries support list

P-R-O-C-H-Y:
## Description of Change
Added ESP32S3 to libraries support list.
Added SDMMC to the list.


P-R-O-C-H-Y:
@pedrominatel As you can see I have added the S3 to the table, but when is "Not Supported" in cell it it doesn't look good. Can we do something with that? :) Thanks for tips.

pedrominatel (approved):
LGTM

docs/source/libraries.rst
```diff
@@ -9,53 +9,53 @@ Supported Peripherals
...
-| USB           | Not Supported | Yes           | Yes           | ESP32-C3 only CDC/JTAG        |
-+---------------+---------------+---------------+---------------+-------------------------------+
-| Wi-Fi         | Yes           | Yes           | Yes           |                               |
-+---------------+---------------+---------------+---------------+-------------------------------+
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| Peripheral    | ESP32         | ESP32-S2      | ESP32-C3      | ESP32-S3      | Comments               |
++===============+===============+===============+===============+===============+========================+
+| ADC           | Yes           | Yes           | Yes           | Yes           |                        |
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| Bluetooth     | Yes           | Not Supported | Not Supported | Not Supported | Bluetooth Classic      |
```
SuGlider, line 17: ESP32-S3 supports BLE, the same for ESP32-C3
We can replace Yes and Not Supported by "BLE and Classic" or "BLE only".
Remove the comments - "Classic"

docs/source/libraries.rst
```diff
@@ -9,53 +9,53 @@ Supported Peripherals
...
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| Ethernet      | Yes           | Not Supported | Not Supported | Not Supported | (*)                    |
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| GPIO          | Yes           | Yes           | Yes           | Yes           |                        |
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| Hall Sensor   | Yes           | Not Supported | Not Supported | Not Supported |                        |
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| I2C           | Yes           | Yes           | Yes           | Yes           |                        |
++---------------+---------------+---------------+---------------+---------------+------------------------+
+| I2S           | Yes           | No            | No            | Yes           | WIP                    |
```
SuGlider, line 31: I think that ESP32-S2 has I2S support... please check with @PilnyTomas

SuGlider (commented):
LGTM! Please just check the I2S note that I left.

docs/source/libraries.rst
P-R-O-C-H-Y: Fixed :) All chips has I2S supported.

SuGlider: I saw that the I2S example excludes the C3 with the .skip.esp32c3 tag file. Maybe Pilny's library doesn't support this SoC.

P-R-O-C-H-Y: I got info from him that it supports :) but C3 has only 1 I2S. Maybe the example is made using both I2S peripherals :)

PR #7418, opened 2022-10-31
FreeRTOS.ino: fix usage of vTaskDelay

WereCatf:
The original code assumes 100Hz FreeRTOS tick rate and just supplies vTaskDelay with the assumed number of ticks required for the wanted delay. This patch simply fixes it to use portTICK_PERIOD_MS, thereby working correctly regardless of what tick rate FreeRTOS has been configured to run at.

Signed-off-by: Nita Vesa <werecatf@outlook.com>

### DIFF
--- a/libraries/ESP32/examples/FreeRTOS/FreeRTOS.ino
+++ b/libraries/ESP32/examples/FreeRTOS/FreeRTOS.ino
@@ -67,9 +67,11 @@ void TaskBlink(void *pvParameters)  // This is a task.
   for (;;) // A Task shall never return or exit.
   {
     digitalWrite(LED_BUILTIN, HIGH);   // turn the LED on (HIGH is the voltage level)
-    vTaskDelay(100);  // one tick delay (15ms) in between reads for stability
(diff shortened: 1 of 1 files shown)


me-no-dev:
I agree that comments are not correct, but do keep in mind that Arduino runs with 1000Hz tick rate, so you do not really need to divide by `portTICK_PERIOD_MS`. `vTaskDelay(1000)` would delay for 1000 ms in Arduino

me-no-dev (approved):

WereCatf:
> so you do not really need to divide by `portTICK_PERIOD_MS`. `vTaskDelay(1000)` would delay for 1000 ms in Arduino

No, you do not, but if someone is looking at these examples trying to learn something, it's IMHO best to show them a way of using vTaskDelay where it works correctly regardless of what the FreeRTOS tick rate has been configured at. It translates better to other microcontrollers and environments that way as well.

me-no-dev:
> it's IMHO best to show them a way of using vTaskDelay where it works correctly regardless of what the FreeRTOS tick rate has been configured at.

Agreed and approved :)

PR #7450, opened 2022-11-09
ESP-IDF v4.4.3

me-no-dev:


mrengineer7777:
Interesting.  Lots of documentation changes and some tweaks to IDF functions.  I see copyright headers have changed on some files, like "tools/sdk/esp32/include/esp_rom/esp32/esp_rom_caps.h".  Looks like a template.  Do we have a tool that automatically replaces those?

me-no-dev:
> Do we have a tool that automatically replaces those?

I suggest you ask the ESP_IDF folks, since those files come from there. I would guess that they at least used some one-off script :)

PR #7460, opened 2022-11-10
Added example scanning maximum ledc frequencies

PilnyTomas:
Added example scanning maximum ledc frequencies

### DIFF
--- a/libraries/ESP32/examples/AnalogOut/ledcFrequency/ledcFrequency.ino
+++ b/libraries/ESP32/examples/AnalogOut/ledcFrequency/ledcFrequency.ino
@@ -1,0 +1,52 @@
+/*
+ * This sketch will map the maximum frequency depending on the bit resolution for the current SoC.
+ * Run the sketch and wait for the Final report.
+ * Ignore the error messages from incorrect settings such as these:
+ *  "E (4190) ledc: requested frequency and duty resolution can not be achieved, try reducing freq_hz or duty_resolution. div_param=255"
+ *
+ * Date: 10 Nov 2022
+ * Author: Tomas Pilny
+ */
+
+#include "soc/soc_caps.h"
+
+#define PIN 2
+
+void setup() {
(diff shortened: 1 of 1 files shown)


libraries/ESP32/examples/AnalogOut/ledcFrequency/ledcFrequency.ino
```diff
@@ -0,0 +1,52 @@
...
+ * Date: 10 Nov 2022
+ * Author: Tomas Pilny
+ */
+
+#include "soc/soc_caps.h"
+
+#define PIN 2
+
+void setup() {
+  ledcAttachPin(PIN,analogGetChannel(PIN));
```
P-R-O-C-H-Y, line 16: This should be removed its not necessary, it called inside`analogWrite()`.

libraries/ESP32/examples/AnalogOut/ledcFrequency/ledcFrequency.ino
```diff
@@ -0,0 +1,52 @@
...
+      }
+    } // while not found the maximum
+    freq_array[resolution-1] = frequency;
+  } // for all resolutions
+
+  printf("Bit resolution | Max Frequency [Hz]\n");
+  for(uint8_t r = 1; r <= SOC_LEDC_TIMER_BIT_WIDE_NUM; ++r){
+    printf("             %d | %u\n", r, freq_array[r-1]);
+  }
+}
```
P-R-O-C-H-Y, line 47: Would be great to test and print even minimal frequency. It can be handy too. 
So if possible print **Bit resolution | Min Frequency [Hz] | Max Frequency [Hz]** :)

P-R-O-C-H-Y (changes requested):
PTAL on my comments. Test looks really good and can be handy for users 👍

P-R-O-C-H-Y (approved):

SuGlider (approved):
Nice!

PR #7472, opened 2022-11-14
Decreasing SPIFFS partition size by 4 bytes to workaround issue #7467

CoryCharlton:
## Description of Change
Decreasing SPIFFS partition size by 1KB to workaround issue #7467

## Tests scenarios
Tested with PIO using Arduino Core v2.0.5 using an Unexpected Maker FeatherS3

## Related links
Works around #7467


tools/partitions/default_16MB.csv
```diff
@@ -3,4 +3,4 @@ nvs,      data, nvs,     0x9000,  0x5000,
 otadata,  data, ota,     0xe000,  0x2000,
 app0,     app,  ota_0,   0x10000, 0x640000,
 app1,     app,  ota_1,   0x650000,0x640000,
-spiffs,   data, spiffs,  0xc90000,0x370000,
+spiffs,   data, spiffs,  0xc90000,0x36FC00,
```
SuGlider, line 6: @CoryCharlton - please test it with:

spiffs, data, spiffs, 0xc90000,0x36FFFC,

tools/partitions/large_spiffs_16MB.csv
```diff
@@ -3,4 +3,4 @@ nvs,      data, nvs,     0x9000,  0x5000,
 otadata,  data, ota,     0xe000,  0x2000,
 app0,     app,  ota_0,   0x10000, 0x480000,
 app1,     app,  ota_1,   0x490000,0x480000,
-spiffs,   data, spiffs,  0x910000,0x6F0000,
+spiffs,   data, spiffs,  0x910000,0x6EFC00,
```
SuGlider, line 6: @CoryCharlton - please test it with:

spiffs, data, spiffs, 0x910000,0x6EFFFC,

SuGlider (changes requested):

tools/partitions/large_spiffs_16MB.csv
CoryCharlton: Confirm they work and committed the updated files.

BlueAndi:
@SuGlider, @CoryCharlton The 4 byte less corresponds in a filesystem which is not 4 k aligned. Did you try an OTA update of the filesystem? 

https://github.com/espressif/arduino-esp32/blob/972c3bbd00e7294739ba9b5eea0db3a0298d0b53/libraries/Update/src/Updater.cpp#L153-L158
https://github.com/espressif/arduino-esp32/blob/972c3bbd00e7294739ba9b5eea0db3a0298d0b53/libraries/Update/src/Updater.cpp#L207-L215
https://github.com/espressif/arduino-esp32/blob/972c3bbd00e7294739ba9b5eea0db3a0298d0b53/libraries/Update/src/Update.h#L32

Means this must be considered:
```c
#define SPI_FLASH_SEC_SIZE  4096    /**< SPI Flash sector size */
```

Jason2866:
Is this still needed with the change/fix of esptool commit? https://github.com/espressif/esptool/tree/21e59148e73e384947327a0883a753685e03d143

SuGlider:
> Is this still needed with the change/fix of esptool commit? https://github.com/espressif/esptool/tree/21e59148e73e384947327a0883a753685e03d143

I think that it may have fixed the issue reported by @CoryCharlton 
@CoryCharlton - could you please test the latest version of esptool to check if the issue you reported is solved?
Thanks.

SuGlider:
@me-no-dev - not sure how this fix to esptool may be reflected into the Arduino Core tools. 
Would it be necessary to update some files?

me-no-dev:
@SuGlider this fix needs to be released from esptool team and then we can try to update it for Arduino

Jason2866:
@me-no-dev  Not using Arduino IDE so maybe a total dumb question. Why not using esptool.py directly? You are using in Arduino Lib Builder components from latest master branch to build the libs. Why not doing for esptool.py too?

CoryCharlton:
> I think that it may have fixed the issue reported by @CoryCharlton
@CoryCharlton - could you please test the latest version of esptool to check if the issue you reported is solved?
Thanks.

Confirmed that this change resolves the issue. Closing this PR and will close the issue.

me-no-dev:
> @me-no-dev Not using Arduino IDE so maybe a total dumb question. Why not using esptool.py directly? You are using in Arduino Lib Builder components from latest master branch to build the libs. Why not doing for esptool.py too?

we pull esptool when released (or when it's necessary), but in this case, the change that fixes this issue is not part of any released esptool yet. We can not use the one in ESP-IDF, because for Windows and Mac we need the compiled versions (which are part of the releases of esptool)

PR #7475, opened 2022-11-15
DNSServer fix custom code replies

vortigont:
custom code reply was sending garbage from a buffer instead of crafted DNS header


mrengineer7777:
I agree writing from buffer with sizeof(DNSHeader) in replyWithCustomCode() is a BUG.  Not sure what the intent was there.

Checking for null buffer in replyWithIP() is also pointless and should be removed.  buffer is not referenced in that function and should never be null when called from processNextRequest().

vortigont:
@VojtechBartoska,
this fix is ready for review. Tnx!

PR #7512, opened 2022-11-25
Fixed issue where esp32 won't reconnect to WiFi AP if the AP was restarted.

sanketwadekar:
-----------
## Description of Change
In the WiFi Provisioning example, the esp32 won't reconnect to the AP if the AP lost power and was restarted at some later point.

## Tests scenarios
I have tested this using WiFiProv example on ESP32. 
I connected the ESP32 to Mobile Hotspot and turned the Hotspot OFF/ON multiple times.

## Related links
Closes espressif/esp-rainmaker#194

Co-authored-by: Pedro Minatel <pedro.minatel@espressif.com>


P-R-O-C-H-Y:
Hi @sanketwadekar I have retest this issue and when I turn HotSpot on my phone OFF, the reason for disconnecting is `Reason: 3 - AUTH_LEAVE`, so when I turn the HotSpot ON again, ESP32 automatically reconnects.

sanketwadekar:
> Hi @sanketwadekar I have retest this issue and when I turn Hotspot on my phone OFF, the reason for disconnecting is `Reason: 3 - AUTH_LEAVE`, so when I turn the Hotspot ON again, ESP32 automatically reconnects.
> 

Some users(Rainmaker) have reported that their device won't reconnect and the reason in their case was UNSPECIFIED,  and I too was able to reproduce this. Would keeping this change cause issues for other users?

P-R-O-C-H-Y:
> 

I will discuss it with colleagues. I am not sure, if there can be any other reasons marked as "UNSPECIFIED", where we should not do auto-reconnect.

VojtechBartoska:
@me-no-dev and @pedrominatel we have discussed this PR a bit during today's meeting and we were not sure about proposed fix. Can you please help with review? Thanks.

pedrominatel:
Adding the `WIFI_REASON_UNSPECIFIED` to the `isReconnectableReason` could fix but I think the `WIFI_REASON_UNSPECIFIED` could be generic and cause some unexpected reconnection attempt when reconnecting is not necessary.

This fix could be acceptable but more investigation can be done to be sure that this fix will work in most cases.

Is there any other reason that could be considered to be a reconnectable reason in the [wifi_err_reason_t](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv417wifi_err_reason_t)?

PR #7520, opened 2022-11-27
Ipv6 support v2

nuclearcat:
This patch partially depends on:
https://github.com/espressif/esp32-arduino-lib-builder/pull/67
Without this patch we will get only Link local IPv6 (still useful for MDNS and etc).
With patch we will get also global IPv6 address by SLAAC.
By default IPv6 disabled, until it is properly tested.

Tested on BasicHttpClient by adding:
wifiMulti.IPv6(true);
before: wifiMulti.addAP() call
and fetching ipv6.google.com which is ipv6 only address.

Enabling Core Debug Level: verbose
If IP6 obtained, in logs will be visible:
[ 8028][V][WiFiGeneric.cpp:380] _arduino_event_cb(): IF[0] Got IPv6: IP Index: 0, Zone: 2, fe80:0000:0000:0000:xxxx:xxxx:xxxx:xxxx
[ 8028][D][WiFiGeneric.cpp:852] _eventCallback(): Arduino Event: 8 - STA_GOT_IP6
[ 11028][V][WiFiGeneric.cpp:380] _arduino_event_cb(): IF[0] Got IPv6: IP Index: 1, Zone: 0, 2a0d:yyyy:0000:4000:yyyy:yyyy:yyyy:yyyy
[ 11028][D][WiFiGeneric.cpp:852] _eventCallback(): Arduino Event: 8 - STA_GOT_IP6

This is linked to: https://github.com/espressif/arduino-esp32/issues/6242

This patch based on prior work , feedback and contributions of @sgryphon

Signed-off-by: Denys Fedoryshchenko <denys.f@collabora.com>

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -20,6 +20,7 @@
 #include <Arduino.h>
 #include <IPAddress.h>
 #include <Print.h>
+#include <StreamString.h>
 
 IPAddress::IPAddress() : IPAddress(IPv4) {}
 
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -113,6 +113,12 @@ bool IPAddress::operator==(const IPAddress& addr) const
         && (memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) == 0);
 }
 
+bool IPAddress::operator!=(const IPAddress& addr) const
+{
+    return (addr._type != _type)
+        || (memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) != 0);
+}
+
 bool IPAddress::operator==(const uint8_t* addr) const
 {
     // IPv4 only comparison to byte pointer
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -21,59 +21,195 @@
 #include <IPAddress.h>
 #include <Print.h>
 
-IPAddress::IPAddress()
+IPAddress::IPAddress() : IPAddress(IPv4) {}
+
+IPAddress::IPAddress(IPType ip_type)
 {
-    _address.dword = 0;
+    _type = ip_type;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
 }
 
 IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet)
 {
-    _address.bytes[0] = first_octet;
-    _address.bytes[1] = second_octet;
-    _address.bytes[2] = third_octet;
-    _address.bytes[3] = fourth_octet;
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX] = first_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 1] = second_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 2] = third_octet;
+    _address.bytes[IPADDRESS_V4_BYTES_INDEX + 3] = fourth_octet;
+}
+
+IPAddress::IPAddress(uint8_t o1, uint8_t o2, uint8_t o3, uint8_t o4, uint8_t o5, uint8_t o6, uint8_t o7, uint8_t o8, uint8_t o9, uint8_t o10, uint8_t o11, uint8_t o12, uint8_t o13, uint8_t o14, uint8_t o15, uint8_t o16) {
+    _type = IPv6;
+    _address.bytes[0] = o1;
+    _address.bytes[1] = o2;
+    _address.bytes[2] = o3;
+    _address.bytes[3] = o4;
+    _address.bytes[4] = o5;
+    _address.bytes[5] = o6;
+    _address.bytes[6] = o7;
+    _address.bytes[7] = o8;
+    _address.bytes[8] = o9;
+    _address.bytes[9] = o10;
+    _address.bytes[10] = o11;
+    _address.bytes[11] = o12;
+    _address.bytes[12] = o13;
+    _address.bytes[13] = o14;
+    _address.bytes[14] = o15;
+    _address.bytes[15] = o16;
 }
 
 IPAddress::IPAddress(uint32_t address)
 {
-    _address.dword = address;
+    // IPv4 only
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    _address.dword[IPADDRESS_V4_DWORD_INDEX] = address;
+
+    // NOTE on conversion/comparison and uint32_t:
+    // These conversions are host platform dependent.
+    // There is a defined integer representation of IPv4 addresses,
+    // based on network byte order (will be the value on big endian systems),
+    // e.g. http://2398766798 is the same as http://142.250.70.206,
+    // However on little endian systems the octets 0x83, 0xFA, 0x46, 0xCE,
+    // in that order, will form the integer (uint32_t) 3460758158 .
 }
 
-IPAddress::IPAddress(const uint8_t *address)
+IPAddress::IPAddress(const uint8_t *address) : IPAddress(IPv4, address) {}
+
+IPAddress::IPAddress(IPType ip_type, const uint8_t *address)
 {
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+    _type = ip_type;
+    if (ip_type == IPv4) {
+        memset(_address.bytes, 0, sizeof(_address.bytes));
+        memcpy(&_address.bytes[IPADDRESS_V4_BYTES_INDEX], address, sizeof(uint32_t));
+    } else {
+        memcpy(_address.bytes, address, sizeof(_address.bytes));
+    }
 }
 
 IPAddress& IPAddress::operator=(const uint8_t *address)
 {
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+    // IPv4 only conversion from byte pointer
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    memcpy(&_address.bytes[IPADDRESS_V4_BYTES_INDEX], address, sizeof(uint32_t));
     return *this;
 }
 
 IPAddress& IPAddress::operator=(uint32_t address)
 {
-    _address.dword = address;
+    // IPv4 conversion
+    // See note on conversion/comparison and uint32_t
+    _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
+    _address.dword[IPADDRESS_V4_DWORD_INDEX] = address;
     return *this;
 }
 
 bool IPAddress::operator==(const uint8_t* addr) const
 {
-    return memcmp(addr, _address.bytes, sizeof(_address.bytes)) == 0;
+    return (addr._type == _type)
+        && (memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) == 0);
+}
+
+bool IPAddress::operator==(const uint8_t* addr) const
+{
+    // IPv4 only comparison to byte pointer
+    // Can't support IPv6 as we know our type, but not the length of the pointer
+    return _type == IPv4 && memcmp(addr, &_address.bytes[IPADDRESS_V4_BYTES_INDEX], sizeof(uint32_t)) == 0;
+}
+
+uint8_t IPAddress::operator[](int index) const {
+    if (_type == IPv4) {
+        return _address.bytes[IPADDRESS_V4_BYTES_INDEX + index];
+    }
+    return _address.bytes[index];
+}
+
+uint8_t& IPAddress::operator[](int index) {
+    if (_type == IPv4) {
+        return _address.bytes[IPADDRESS_V4_BYTES_INDEX + index];
+    }
+    return _address.bytes[index];
 }
 
 size_t IPAddress::printTo(Print& p) const
 {
     size_t n = 0;
-    for(int i = 0; i < 3; i++) {
-        n += p.print(_address.bytes[i], DEC);
+
+    if (_type == IPv6) {
+        // IPv6 IETF canonical format: compress left-most longest run of two or more zero fields, lower case
+        int8_t longest_start = -1;
+        int8_t longest_length = 1;
+        int8_t current_start = -1;
+        int8_t current_length = 0;
+        for (int8_t f = 0; f < 8; f++) {
+            if (_address.bytes[f * 2] == 0 && _address.bytes[f * 2 + 1] == 0) {
+                if (current_start == -1) {
+                    current_start = f;
+                    current_length = 1;
+                } else {
+                    current_length++;
+                }
+                if (current_length > longest_length) {
+                    longest_start = current_start;
+                    longest_length = current_length;
+                }
+            } else {
+                current_start = -1;
+            }
+        }
+        for (int f = 0; f < 8; f++) {
+            if (f < longest_start || f >= longest_start + longest_length) {
+                uint8_t c1 = _address.bytes[f * 2] >> 4;
+                uint8_t c2 = _address.bytes[f * 2] & 0xf;
+                uint8_t c3 = _address.bytes[f * 2 + 1] >> 4;
+                uint8_t c4 = _address.bytes[f * 2 + 1] & 0xf;
+                if (c1 > 0) {
+                    n += p.print((char)(c1 < 10 ? '0' + c1 : 'a' + c1 - 10));
+                }
+                if (c1 > 0 || c2 > 0) {
+                    n += p.print((char)(c2 < 10 ? '0' + c2 : 'a' + c2 - 10));
+                }
+                if (c1 > 0 || c2 > 0 || c3 > 0) {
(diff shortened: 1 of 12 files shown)


nuclearcat:
@VojtechBartoska this is updated PR that is more standard compliant with upstream Arduino and easier(more transparent) to use.

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1473,6 +1492,37 @@ int WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult)
...
+int WiFiGenericClass::hostByName6(const char* aHostname, ip_addr_t& aResult)
+{
+    ip_addr_t addr;
+    struct dns_api_msg arg;
+
+    memset(&arg, 0x0, sizeof(arg));
+    waitStatusBits(WIFI_DNS_IDLE_BIT, 16000);
+    clearStatusBits(WIFI_DNS_IDLE_BIT | WIFI_DNS_DONE_BIT);
+
+    err_t err = dns_gethostbyname_addrtype(aHostname, &addr, &wifi_dns6_found_callback,
```
sgryphon, line 1511: Maybe add a comment above this to explain:
// Retrieves a single IP address, either from the cache or lookup
// The address type parameter means to try IPv6 first, and then fallback to IPv4 if that fails

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1473,6 +1492,37 @@ int WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult)
     return (uint32_t)aResult != 0;
 }
 
+/**
+ * Resolve the given hostname to an IP6 address.
+ * @param aHostname     Name to be resolved
+ * @param aResult       IPv6Address structure to store the returned IP address
```
sgryphon, line 1498: Return could be IPv6 or IPv4, so change comment:

* @param aResult       IP address structure to store the returned IP address, either IPv6 if available, otherwise IPv4

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1473,6 +1492,37 @@ int WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult)
...
+        if (arg.result == 1) {
+            aResult = arg.ip_addr;
+        }
+    }
+    setStatusBits(WIFI_DNS_IDLE_BIT);
+    return (uint32_t)err == ERR_OK || (err == ERR_INPROGRESS && arg.result == 1);
+}
+
 IPAddress WiFiGenericClass::calculateNetworkID(IPAddress ip, IPAddress subnet) {
 	IPAddress networkID;
```
sgryphon, line 1528: This function `calculateNetworkID` and the two below `calculateBroadcast` and `calculateSubnetCIDR` only work for IPv4.

They also have an assumption that the subnet mask is in the correct format (all 1's followed by all 0's) -- if the format is incorrect, they will give weird results (but I think returning an error would be better).  e.g. calculateSubnetCIDR("255.128.0.0") returns 9, whilst calculateSubnetCIDR("255.1.0.255") is invalid but returns 16.

IPv6 doesn't use the mask format and calls the CIDR value the Prefix Length. 

Networks still exist, but a more useful signature would be: calculateNetworkID(IPAddress ip, uint_8 prefixLength)

Broadcast addresses don't exist in IPv6; instead various multicast address ranges fulfil the same purpose.

Suggestion:
* Have calculateSubnetCIDR return 255 if invalid
* Have calculateSubnetCIDR work for IPv6 (return value will be 0-128, whilst IPv4 will be 0-32), but probably won't be used.
* Add an alternative for calculateNetwork(IPAddress, uint_8 prefixLength).
* Call to calculateNetworkID(IPAddress ip, IPAddress subnet) can call calculateSubnetCIDR() to get the length, then return an error value if invalid, otherwise chain to the one that takes length parameter
* Call to calculateBroadcast should check IPv4 only, and could also use calculateSubnetCIDR() to make sure it is valid.

libraries/WiFi/src/WiFiServer.cpp
```diff
@@ -76,14 +76,23 @@ void WiFiServer::begin(uint16_t port, int enable){
...
+  sockfd = socket(AF_INET6 , SOCK_STREAM, 0);
   if (sockfd < 0)
     return;
   setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(int));
-  server.sin_family = AF_INET;
-  server.sin_addr.s_addr = _addr;
-  server.sin_port = htons(_port);
+  server.sin6_family = AF_INET6;
+  if (_addr.type() == IPv4) {
+    log_e("---------------- IPv4");
```
sgryphon, line 86: Are these logs meant to still be here? why all the "-" at the beginning?

nuclearcat: Forgotten debug, will remove

libraries/WiFi/src/WiFiGeneric.cpp
nuclearcat: I will do additional commit to handle that, i will try to check example projects that use it, to make some test cases.

VojtechBartoska:
thanks @nuclearcat for your work and PR! will be reviewed for next major release 3.0.0. It will take us some time.

cores/esp32/IPAddress.h
```diff
@@ -83,14 +88,21 @@ class IPAddress: public Printable
     virtual size_t printTo(Print& p) const;
     String toString() const;
 
+    IPType type() { return _type; }
```
TD-er, line 91: Please make this a `const` function.

vortigont:
Just out of curiosity is there any reason not to reuse LWIP's [ip_addr_t](https://github.com/espressif/arduino-esp32/blob/fd72cf46ad6fc1a6de99c1d83ba8eba17d80a4ee/tools/sdk/esp32/include/lwip/lwip/src/include/lwip/ip_addr.h#L69) from [lwip/ip_addr.h](https://github.com/espressif/arduino-esp32/blob/master/tools/sdk/esp32/include/lwip/lwip/src/include/lwip/ip_addr.h) instead of `_address` union member in this class?
I mean, LWIP already have [many](https://www.nongnu.org/lwip/2_0_x/group__ipaddr.html) useful macro's and functions to reuse, like [ipaddr_ntoa()](https://www.nongnu.org/lwip/2_0_x/group__ipaddr.html#gaabe63fa43772222df917d4f57cbe416e) so not to reinvent `IPAddress::fromString6`.
ESP-IDF examples and code also rely on lwip's types, so exposing `ip_addr_t` as `IPAddress`'s class member could make life easier in many cases.

sgryphon:
> Just out of curiosity is there any reason not to reuse LWIP's ip_addr_t

Probably because the implementation is copied from the ArduinoCore-API (I was an author), which does not specify/depend on LWIP.

Seeing as you are using LWIP, then it makes sense to replace the implementation with LWIP, so this change sounds like a good idea, and will simplify things.  So long as you match the API that is what matters.

Using LWIP should make things easier, and I had a quick look and it seems compatible (it is what is being used underneath anyway), e.g. the conversion to string.

It would be nice to have unit tests for the arduino-esp32 library (the arduniocore-api has them) for some corner cases.

e.g. It looks like ntoa in LWIP only does a single pass, which means it doesn't produce canonical format if there is a second longer run of zeros (it always does the first, i.e. left-most, run of 2 or more).  It is still a valid string representation, and in the majority of cases, probably all of them in practice, it will be the same, but it would be nice to have tests to validate this.

Also, note that the fix would be upstream in LWIP, not in this code. I still support using LWIP ip_addr_t as a far simpler implementation (with IPAddress.h just being a wrapper that adapts it to the Arduino API).

vortigont:
> Also, note that the fix would be upstream in LWIP, not in this code. I still support using LWIP ip_addr_t as a far simpler implementation (with IPAddress.h just being a wrapper that adapts it to the Arduino API).

Indeed, LWIP's `ntoa` implementation is not longest greedy as recommended [rfc5952](https://www.rfc-editor.org/rfc/rfc5952#section-4) but still could a be better option to bridge with ESP-IDF code if `IPAddress` is supplied with c-tor overload to instantiate/copy from `ip_addr_t`. I like the idea of a wrapper!
Thanks for your opinion, @sgryphon!
Let's see for community's feedback here )

cores/esp32/IPAddress.h
```diff
@@ -26,55 +26,60 @@
...
 class IPAddress: public Printable
 {
 private:
     union {
-        uint8_t bytes[4];  // IPv4 address
-        uint32_t dword;
+        uint8_t bytes[16];
+        uint32_t dword[4];
     } _address;
+    IPType _type;
```
s-hadinger, line 45: Would it be possible to use `ip_addr_t` instead of this local structure. This is what is done in ESP8266 and makes it easier to call esp-idf functions, saving conversions from `IPAddress` and `ip_addr_t`

cores/esp32/IPAddress.h
s-hadinger, line 91: Can you please make it `const`:

`IPType type() const { return _type; }`

cores/esp32/IPAddress.h
TD-er: Please also initialize `_type` as it is not always initialized. (e.g. when using `fromString4` or `fromString6`)
So maybe the type could be extended to have a 3rd type: `NotSet` or something similar.
But then the checks for the `_type` should include both checks, for IPv4 and IPv6 as now it is more like a boolean check. It is either IPv4 or IPv6.

s-hadinger: Esp-idf has a 'not set' IP although I'm not a big fan because it creates confusion to users and needs additional tests. 

I'm more in favor of having IP ADDRESS always initialized to ANY to the widest supported length, I.e. v6 of `LWIP_IPV6` is defined

cores/esp32/IPAddress.cpp
```diff
@@ -117,7 +268,80 @@ bool IPAddress::fromString(const char *address)
...
+    if (double_colons != -1) {
+        for (int i = colons * 2 - double_colons * 2 - 1; i >= 0; i--)
+            _address.bytes[16 - colons * 2 + double_colons * 2 + i] = _address.bytes[double_colons * 2 + i];
+        for (int i = double_colons * 2; i < 16 - colons * 2 + double_colons * 2; i++)
+            _address.bytes[i] = 0;
+    }
+
+    _type = IPv6;
     return true;
 }
```
sgryphon, line 347: Need to also define the `IN6ADDR_ANY` constant

```
const IPAddress arduino::IN6ADDR_ANY(IPv6);
```

cores/esp32/IPAddress.h
sgryphon: > `_type` as it is not always initialized

fromString is an instance method, not static/factory. By default _type is initialized to IPv4, to be backwards compatible with existing code. Using xxx(IPv6) initialises to IN6ADDR_ANY.

(Which I noticed setting the constant was missing / not copied from ArduinoCore-API).

If you then call fromString() and it successfully parses as either 4 or 6 then it will be set accordingly.

If it fails to parse, then it will remain as is (usually IPv4), but the byte values will be meaningless (partially parsed).

Maybe if parse fails, then it should set to all zeros? (which will be IPv4 ANY or IPv6 ANY, depending on the original)

TD-er: @sgryphon I guess you're right.

Missed that one. I haven't run into situations where the _type variable was uninitialized, just looked through the code to see if all were initialized as I've been bitten by uninitialized variables myself quite a lot lately and thus was looking through the code to see if all were initialized.

> Maybe if parse fails, then it should set to all zeros? (which will be IPv4 ANY or IPv6 ANY, depending on the original)

I guess it should be set to some default constructed state, where it would be obvious to take the default IPv4 constructor for the `fromString4` and IPv6 for the `fromString6`.

Then calls like `isSet` (or whatever the call is for `IPAddress`) will return false.

TD-er: Ah yes, so the type itself should have an any type as well. (or maybe a bitmask, for IPv4 any and IPv6 any?, but this will break LWIP consistency, so maybe not...)
So I guess it should just not change the address in any way if the `fromStringX` call was not given a properly formatted string?

cores/esp32/IPAddress.cpp
sgryphon: Based on discussion of type, and compatibility with LWIP, we might want 3 constants for ANY, i.e. representing to listen on all IPv4, all IPv6, or all (dual stack) addresses. 

And corresponding third value value IPType (above):

```
enum IPType
{
    IPv4 = 0,
    IPv6 = 6, //numeric values for compatibility with LWIP
    ANY = 46 // only used for dual stack INADDR_ANY to listen on all addresses 
};

IPAddress INADDR_NONE(0, 0, 0, 0); // Represents IPv4 any, for backwards compatibility, i.e. listen on all IPv4 address.
IPAddress IN6ADDR_ANY(IPv6); // Represents IPv6 any, i.e. listen on all IPv6 address.
IPAddress INADDR_ANY(ANY); // Represents dual stack any, i.e. listen on all addresses, both IPv4 and IPv6
```

cores/esp32/IPAddress.h
sgryphon: I added a suggestion for this in the comments around the ANY constants.

cores/esp32/IPAddress.cpp
```diff
@@ -20,78 +20,225 @@
...
+    {
+        StreamString s;
+        s.reserve(40);
+        printTo(s);
+        return s;
+    }
+
+    // IPv4
     char szRet[16];
     sprintf(szRet,"%u.%u.%u.%u", _address.bytes[0], _address.bytes[1], _address.bytes[2], _address.bytes[3]);
```
s-hadinger, line 215: This line is wrong now. It should be

`    sprintf(szRet,"%u.%u.%u.%u", _address.bytes[IPADDRESS_V4_BYTES_INDEX], _address.bytes[IPADDRESS_V4_BYTES_INDEX+1], _address.bytes[IPADDRESS_V4_BYTES_INDEX+2], _address.bytes[IPADDRESS_V4_BYTES_INDEX+3]);`

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -210,22 +215,34 @@ int WiFiClient::connect(IPAddress ip, uint16_t port)
...
 int WiFiClient::connect(IPAddress ip, uint16_t port, int32_t timeout)
 {
+    struct sockaddr_storage serveraddr;
     _timeout = timeout;
-    int sockfd = socket(AF_INET, SOCK_STREAM, 0);
+    int sockfd = -1;
+
+    if (ip.type() == IPv6) {
+        struct sockaddr_in6 *tmpaddr = (struct sockaddr_in6 *)&serveraddr;
+        memset((char *) tmpaddr, 0, sizeof(struct sockaddr_in));
```
s-hadinger, line 227: Shouldn't the size be `sizeof(sockaddr_in6)`?

s-hadinger: Actually you can remove both `memset` by simply declaring

`struct sockaddr_storage serveraddr = {};`

which forces the compiler to initialize the structure with zeros

tablatronix:
Is this expected?

```
[  2257][V][WiFiServer.h:41] WiFiServer(): WiFiServer::WiFiServer(port=80, ...)
[  2257][V][WebServer.cpp:87] WebServer(): WebServer::Webserver(port=80)
[  2266][E][WiFiServer.cpp:91] begin(): ---------------- IPv6
[ 67108][V][WebServer.cpp:296] handleClient(): New client: client.localIP()=0.0.0.0
```

sgryphon:
> Is this expected? localip ?
> 
> ```
> [ 67108][V][WebServer.cpp:296] handleClient(): New client: client.localIP()=0.0.0.0
> ```

Could be the bug that @s-hadinger mentions above, where the sprintf uses the wrong bytes?

nuclearcat:
I am going to work on updating PR in 2 days, just was too busy. Will address this too...

s-hadinger:
@nuclearcat I have already fixed and expanded this PR including using `ip_addr_t`, and adding support for UDP. It is now used by default in Tasmota for a few weeks and proved stable and solid.

I don't know how to do a PR on this PR so I will push a new PR, although I don't want to shadow your work.

The branch is here.
https://github.com/tasmota/arduino-esp32/tree/PR_IPv6

I will do a proper PR later today

nuclearcat:
If it includes any of my work please include my Signed-off-by, then i think it will be ok, i will close mine after you send yours.

s-hadinger:
https://github.com/espressif/arduino-esp32/pull/7722 sent as a superset for this PR.

tablatronix:
I am still seeing the issue I raised

- framework-arduinoespressif32 @ 2.0.7

tablatronix:
What is the status of this afaik this is not fixed

[ 37114][V][WebServer.cpp:296] handleClient(): New client: client.localIP()=0.0.0.0

```
*wm:[2] [SYS] WM version:  v2.0.15-rc.1
*wm:[2] [SYS] Arduino version:  2.0.10
*wm:[2] [SYS] ESP SDK version:  4.4.5.230614
*wm:[2] [SYS] Free heap:        180812
*wm:[2] [SYS] Chip ID: 2665141460
*wm:[2] [SYS] Chip Model: ESP32-PICO-D4
```

s-hadinger:
We will shortly propose a new IPv6 PR for Arduino 3 and esp-idf 5.1. It's already in extensive testing in Tasmota.

PR #7430, opened 2022-11-05
Add TWAI receive example

designer2k2:
## Description of Change

Example showing how to receive messages using the TWAI interface and a CAN transceiver.
It also uses the alerts to show how to set them up and react to them.

It is based on the samples shown in the api: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html

## Tests scenarios
I used an esp32-c3-devkit (ESP32-D0WDQ6) with a SN65HVD230 CAN transceiver. 
This was attached between 2 Arduino Nano´s with a MCP2515/MCP2562 CAN interface, whereby they replayed a recording from a quite busy CAN bus.
This replay was then observed by the ESP32 and send with this example to the host.

PR #7550, opened 2022-12-05
TWAI: Add receive example (Copy of #7430)

Dazza0:
Copy of https://github.com/espressif/arduino-esp32/pull/7430 but squashed and rebased.

### DIFF
--- a/libraries/ESP32/examples/TWAI/TWAIreceive/TWAIreceive.ino
+++ b/libraries/ESP32/examples/TWAI/TWAIreceive/TWAIreceive.ino
@@ -1,0 +1,127 @@
+/* ESP32 TWAI receive example.
+  Receive messages and sends them over serial.
+
+  Connect a CAN bus transceiver to the RX/TX pins.
+  For example: SN65HVD230
+
+  TWAI_MODE_LISTEN_ONLY is used so that the TWAI controller will not influence the bus.
+
+  The API gives other possible speeds and alerts:
+  https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html
+
+  Example output from a can bus message:
+  -> Message received
+  -> Message is in Standard Format
+  -> ID: 604
+  -> Byte: 0 = 00, 1 = 0f, 2 = 13, 3 = 02, 4 = 00, 5 = 00, 6 = 08, 7 = 00
+
+  Example output with alerts:
+  -> Alert: A (Bit, Stuff, CRC, Form, ACK) error has occurred on the bus.
+  -> Bus error count: 171
+  -> Alert: The RX queue is full causing a received frame to be lost.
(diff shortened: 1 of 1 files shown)


Dazza0:
@designer2k2 I've rebased and squashed your PR. PTAL.

VojtechBartoska:
@PilnyTomas Can you please do one more "quick" review and after that, we can merge it.

libraries/ESP32/examples/TWAI/TWAIreceive/TWAIreceive.ino
```diff
@@ -0,0 +1,117 @@
...
+  twai_general_config_t g_config = TWAI_GENERAL_CONFIG_DEFAULT((gpio_num_t)TX_PIN, (gpio_num_t)RX_PIN, TWAI_MODE_LISTEN_ONLY);
+  twai_timing_config_t t_config = TWAI_TIMING_CONFIG_500KBITS();  //Look in the api-reference for other speed sets.
+  twai_filter_config_t f_config = TWAI_FILTER_CONFIG_ACCEPT_ALL();
+
+  // Install TWAI driver
+  if (twai_driver_install(&g_config, &t_config, &f_config) == ESP_OK) {
+    Serial.println("Driver installed");
+  } else {
+    Serial.println("Failed to install driver");
+    return;
```
me-no-dev, line 50: if this fails, `setup()` will return and start the `loop()` which will check for alerts on non-working bus. I suggest setting a flag that TWAI started OK, which will allow then the code in `loop()` to execute (else just delay and do nothing)

libraries/ESP32/examples/TWAI/TWAIreceive/TWAIreceive.ino
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  // Start TWAI driver
+  if (twai_start() == ESP_OK) {
+    Serial.println("Driver started");
+  } else {
+    Serial.println("Failed to start driver");
+    return;
```
me-no-dev, line 58: same as above

libraries/ESP32/examples/TWAI/TWAIreceive/TWAIreceive.ino
```diff
@@ -0,0 +1,127 @@
...
+      Serial.printf(" %d = %02x,", i, message.data[i]);
+    }
+    Serial.println("");
+  }
+}
+
+void loop() {
+  if (!driver_installed) {
+    // Driver not installed
+    Serial.println("TWAI driver not installed!");
```
me-no-dev, line 95: you can replace this line with a simple `delay(1000);` which will leave the reason as last printed line and sleep from then on. As is, this line will be printed as fast as the UART can push it out in a loop

me-no-dev (approved):

PR #7172, opened 2022-08-23
DRAFT: Board menu split

PilnyTomas:
## Description of Change
Draft of splitting the board menu.

## Tests scenarios
none

## Related links
Discussion for community meeting #7167

PR #7557, opened 2022-12-06
Sorted boards.txt + added board-sorting script for future use

PilnyTomas:
The new script `.sort_boards.sh` sorts contents in file `boards.txt` alphabetically based on `.name` parameter to be sorted for users in Arduino IDE.
Official Espressif boards are listed on top, sorted separately.
The file `boards.txt` in this PR contains result of this script.

This is a replacement for the previously proposed (impossible solution) splitting of the boards menu #7172

#### Known issues:

- Removes empty lines
- Removes commented lines


VojtechBartoska:
Thanks @PilnyTomas for the PR. Final result are good. 

We just need to make sure it's working correctly for new added boards and is also okay for LibBuilder.

@me-no-dev & @SuGlider please help with reviewing this script and the approach. Thanks.

pedrominatel:
Hi guys,

I would suggest creating some basic rules for adding new boards to prevent a huge number of boards on the list.
Most of the boards doesn't have any special peripheral, pins, or configuration and could be used just as a generic ESP32-x.

SuGlider (commented):
I think it is a nice tool, not not sure it would fit along with Arduino Core. 
Nice work @PilnyTomas!

It could stay within Lib Builder? @me-no-dev 
Or as @pedrominatel suggested, we shall think a better way to solve the growing list of ESP32 3rd party boards.

VojtechBartoska:
@PilnyTomas up to the discussion today, the approach and script looks okay. 

Only thing, please remove the changes in boards.txt file in this PR. (as it was there just to show the result of the script)

PilnyTomas:
> please remove the changes in boards.txt file

The main addition is the sorted list, not the script - the script is something extra that can be used in the future if needed.

Velocet:
> It could stay within Lib Builder? @me-no-dev Or as @pedrominatel suggested, we shall think a better way to solve the growing list of ESP32 3rd party boards.

Maybe use aliases and compile the boards file from it? This way boards with just another name could be used in Arduino IDE without too much hussle. Most beginners will likely look for their specific board instead of the ESP32 counterpart.

Phetdanai:
<details><summary>Details</summary>
<p>



</p>
</details>

PR #7559, opened 2022-12-07
Added onDisconnect with gatt parameters to BLEServerCallbacks, equivalent to the existing onConnect

### LINKED ISSUE #6008: BLEServer.h callback onDisconnect() missing esp_ble_gatts_cb_param_t parameter
The BLEServer.h defines callbacks for connection and disconnection.

For the connect, there are 2 virtual methods, 1 without and 1 with esp_ble_gatts_cb_param_t.

For the disconnect, there is only 1 virtual method, without that parameter.

At least for coherency, there should be an onDisconnect() with that parameter. This parameter it is even in the comments!!! and it is missing in the code.

That parameter is useful for example to know the client that disconnects.

This fix wouldn't even impact the code already written, as it is an extra virtual method.

Look at the comment of the missing parameter esp_ble_gatts_cb_param_t:

class BLEServerCallbacks {
public:
        virtual ~BLEServerCallbacks() {};
        /**
         * @brief Handle a new client connection.
         *
         * When a new client connects, we are invoked.
         *
         * @param [in] pServer A reference to the %BLE server that received the client connection.
         */
        virtual void onConnect(BLEServer* pServer);
        virtual void onConnect(BLEServer* pServer, esp_ble_gatts_cb_param_t *param);
        /**
         * @brief Handle an existing client disconnection.
         *
         * When an existing client disconnects, we are invoked.
         *
         * @param [in] pServer A reference to the %BLE server that received the existing client disconnection.  <--- HERE!!!
         */
        virtual void onDisconnect(BLEServer* pServer);
}; // BLEServerCallbacks

RM-Alltrons:
-----------
## Description of Change
This change adds a onDisconnect with parameters to BLEServerCallbacks, equivalent to onDisconnect, which allows the code to know which (of the multiple connections) is disconnecting. Similar to how onConnect can be used with esp_ble_gatts_cb_param_t now.

## Tests scenarios
Tested on ESP32-S3-DevKitC-1, Platform-IO 5.1.1 (framework=Arduino), while connecting with 2 BLE devices (android and ios) simultaneous, reconnecting them one at a time.

## Related links

Closes #6008

### DIFF
--- a/libraries/BLE/src/BLEServer.cpp
+++ b/libraries/BLE/src/BLEServer.cpp
@@ -374,6 +374,12 @@ void BLEServerCallbacks::onDisconnect(BLEServer* pServer) {
 	log_d("BLEServerCallbacks", "<< onDisconnect()");
 } // onDisconnect
 
+void BLEServerCallbacks::onDisconnect(BLEServer* pServer, esp_ble_gatts_cb_param_t* param) {
+	log_d("BLEServerCallbacks", ">> onDisconnect(): Default");
+	log_d("BLEServerCallbacks", "Device: %s", BLEDevice::toString().c_str());
+	log_d("BLEServerCallbacks", "<< onDisconnect()");
+} // onDisconnect
+
 void BLEServerCallbacks::onMtuChanged(BLEServer* pServer, esp_ble_gatts_cb_param_t* param) {
 	log_d("BLEServerCallbacks", ">> onMtuChanged(): Default");
 	log_d("BLEServerCallbacks", "Device: %s MTU: %d", BLEDevice::toString().c_str(), param->mtu.mtu);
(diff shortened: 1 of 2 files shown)


me-no-dev:
@RM-Alltrons could you please sign the CLA so we can merge your changes?

RM-Alltrons:
> @RM-Alltrons could you please sign the CLA so we can merge your changes?
When I pushed, I tried signing a few times, but that did not seem to  work. 
I now tried a different browser, it says that "You have agreed to the CLA for espressif/arduino-esp32", but  CLA-batch does not confirm that.

me-no-dev:
@RM-Alltrons I guess your local github settings cause you to commit as "Admin" which is not your username and that causes the problem. I will merge it anyway :) Thanks!

PR #7560, opened 2022-12-07
Fix conflict with ESP-IDF's esp_crt_bundle

me-no-dev:
Conflict is reproducible if you have both RainMake and WiFiClientSecure included


SuGlider (approved):
LGTM.
Solves conflict with IDF X.509 Certificate Bundle functions.
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_crt_bundle.html

PR #7563, opened 2022-12-08
apply the same USB descriptor's iSerialNumber logic for ESP32-S2 as t…

lyusupov:
## Description of Change

ESP32-S**3** is currently using certain method to assign a default USB _iSerialNumber_ field for USB descriptor by specifying "_ _MAC_ _" keyword.
This PR applies the same method to the ESP32-S**2** target as well.


cores/esp32/USB.cpp
```diff
@@ -34,7 +34,7 @@
 #define USB_PRODUCT ARDUINO_BOARD
 #endif
 #ifndef USB_SERIAL
-#if CONFIG_IDF_TARGET_ESP32S3
+#if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
```
me-no-dev, line 37: The USB bootloader inside ESP32-S2 uses Serial "0". This is here for that exact reason. Please revert this change and set it manually if you need something specific.

lyusupov: done in https://github.com/espressif/arduino-esp32/pull/7563/commits/537dc7ce99a6db7e2528cd85f9f7951ec22ea995

PR #7583, opened 2022-12-12
Workaround for when USB CDC is unplugged

### LINKED ISSUE #6983: ESP32C3M1: Serial.print() calls when not connected to USB cause delays.
### Board

ESP32C3M1

### Device Description

Custom board:
IMG_7922

### Hardware Configuration

GPIO 18 & 19 are connected to USB. USB is used to program the device and to monitor Serial.print output.

### Version

v2.0.3

### IDE Name

Arduino

### Operating System

MacOS 10.15.7

### Flash frequency

40Mhz

### PSRAM enabled

yes

### Upload speed

115200

### Description

USB is used to program the device and to monitor Serial.print output. However, when the device is unplugged, and the Serial.print() lines are left in the code, they cause delays of a second or so. Plugging back in will make the delays go away. If I remove the Serial.print() lines, it makes the delays disappear too.

### Sketch

```cpp
#include <Wire.h>
#include <Adafruit_NeoPixel.h>

#define LED_PIN 6
#define NEOPIXEL_PIN LED_PIN
Adafruit_NeoPixel pixels(1, NEOPIXEL_PIN, NEO_GRB + NEO_KHZ800);

void setup() {
  Serial.begin(115200);

}

void loop() {

  pixels.begin();  // INITIALIZE NeoPixel strip object (REQUIRED)
  
  Serial.print("Red... ");
  // drawMessage("Test Red LED");
  pixels.setPixelColor(0, pixels.Color(64, 0, 0));
  pixels.show();
  delay(500);

    Serial.print("Green... ");
// drawMessage("Test Green LED");
  pixels.setPixelColor(0, pixels.Color(0, 64, 0));
  pixels.show();
  delay(500);

    Serial.print("Blue... ");
// drawMessage("Test Blue LED");
  pixels.setPixelColor(0, pixels.Color(0, 0, 64));
  pixels.show();
  delay(500);
  
    Serial.print("White... ");
// drawMessage("Test Whitish LED");
  pixels.setPixelColor(0, pixels.Color(64, 64, 64));
  pixels.show();
  delay(500);

  Serial.print("Off... ");
  pixels.clear();
  pixels.show();
  delay(500);
  Serial.println("Done. Looping");
}```
```

### Debug Message

```plain
There is no debug or error message.
```

### Other Steps to Reproduce

Run the sketch. While the device is plugged into a computer with Arduino, the lights flash faster. When you unplug the USB cable the lights slow down.

With more or longer Serial.print output, the delay can be more than a second.

That same code above is running on the two boards in this video. When you plug the USB in, you can see that the flash rate increases on the one plugged in.

https://user-images.githubusercontent.com/1689095/178857842-551736c3-3ac0-4adc-9bf8-758c5a6fa6ad.mov

 

SuGlider:
## Description of Change
By default Tx Timeout is 200 and it causes a delay every time that HW  USB CDC is written when the USB is unplugged.
Setting it to 0 by default solves this issue and it has no side effect.
As stated in the issue #6983, it affects boards that use Batteries and are not connected to a USB port.

## Tests scenarios
This issue affects ESP32-C2 and ESP32-S3 with HW Serial.

## Related links
Fix #6983

### DIFF
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -32,7 +32,7 @@ static uint8_t rx_data_buf[64];
 static intr_handle_t intr_handle = NULL;
 static volatile bool initial_empty = false;
 static xSemaphoreHandle tx_lock = NULL;
-static uint32_t tx_timeout_ms = 200;
+static uint32_t tx_timeout_ms = 0;  // workaround for when USB CDC is not connected
 static esp_event_loop_handle_t arduino_hw_cdc_event_loop_handle = NULL;
 
 static esp_err_t arduino_hw_cdc_event_post(esp_event_base_t event_base, int32_t event_id, void *event_data, size_t event_data_size, BaseType_t *task_unblocked){


cores/esp32/HWCDC.cpp
```diff
@@ -32,7 +32,7 @@ static uint8_t rx_data_buf[64];
 static intr_handle_t intr_handle = NULL;
 static volatile bool initial_empty = false;
 static xSemaphoreHandle tx_lock = NULL;
-static uint32_t tx_timeout_ms = 200;
+static uint32_t tx_timeout_ms = 0;  // workaround for when USB CDC is not connected
```
me-no-dev, line 35: with this being 0, it means that is the ring buffer is full, the code will not give it time to empty some space and will just discard the input. Maybe wait if the is not enough space? Fixing it for when USB is disconnected, but breaking it for when connected is not all that nice :)

SuGlider: Waiting when there is no space (timeout) is exacty the issue... when the board is turned on with the USB unplugged. That is what causes the delay from the point when the RingBuffer gets full and it never is flushed because there is no USB connection.

SuGlider: There is no event that tells the user that USB is or not plugged. Even the ARDUINO_HW_CDC_CONNECTED_EVENT works only the first time, but if we unplug the board, it doesn't tell anything.

SuGlider: @me-no-dev - Added code to test if USB is plugged and the application has never changed Tx Timeout. In that case, it will set Tx Timeout to default 100ms. 
Otherwise (USB has never been plugged or the Application has explicitly changed Tx Timeout Value), it will keep it zero or the new value set by the application.

tshcherban:
Hi
I still have the same issue with board hang after serial monitor closed (ESP32-S3). However im using VS Code with arduino plugin.
With Arduino IDE (1.8) it works fine.

PR #7591, opened 2022-12-15
Rainmaker Example Changes

sanketwadekar:
## Description of Change
1. Changed default OTA type in examples.
2. Added patch that will prevent crashes during provisioning due to low memory.

## Tests scenarios
I have tested all examples on esp32 using master branch.


me-no-dev (approved):

pedrominatel (approved):
LGTM

PR #7593, opened 2022-12-15
Fix Ethernet library link status and missing functionality with recent IDF versions

### LINKED ISSUE #6105: ESP32 ETH.h linkUp never reports true
### Board

ESP32 All

### Device Description

LoLin D32 Pro. ESP32-CAM, TTGO - all

### Hardware Configuration

LAN 8720 Ethernet Adapter

### Version

v2.0.1

### IDE Name

Arduino 1.8.1

### Operating System

ESP32

### Flash frequency

1

### PSRAM enabled

yes

### Upload speed

900k+

### Description

ETH.h has no code that ever sets the link up flag to true. This results the linkUp function to always report false even when the link is up and passing data. The link up flag is used when compiled for any ESP32 platform.

### Sketch

```cpp
Example sketch from the repository modified to call ETH.linkUp and print result.
```

### Debug Message

```plain
No debug message needed. Code review identified this issue
```

### Other Steps to Reproduce

_No response_

ppescher:
## Description of Change
The variable `eth_link` was never updated, so it was always reporting link down.
Proposed fix is to use the status bits managed by WiFi callbacks, that are updated correctly.
Link duplex mode did not have an implementation for IDF versions > 3 and actually the version check was done incorrectly with the preprocessor.
Proposed fix is using the same ioctl mechanism as for the link speed.

## Tests scenarios
Tested with PlatformIO and the latest stable release 2.0.5 on a custom hardware using LAN8710 and LAN8740.

## Related links
Closes #6105

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -220,9 +220,6 @@ ETHClass::ETHClass()
      ,eth_handle(NULL)
 #endif
      ,started(false)
-#if ESP_IDF_VERSION_MAJOR > 3
-     ,eth_link(ETH_LINK_DOWN)
-#endif
 {
 }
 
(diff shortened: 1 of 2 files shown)


me-no-dev (approved):

me-no-dev:
Thanks @ppescher for the fix!

TD-er:
Just tested this and it seems to be working fine.
Ethernet connected state is reliable again and also the reported network speed/mode is working fine.

I did test by plugging and re-plugging a number of times and also by selecting the Ethernet link mode on the switch.
Half-duplex, Full duplex and auto is all reported fine. Also 10 vs. 100 Mbps link speed.

However, 10 Mbps half-duplex is not really usable as it  seems to loose quite a lot of packets.
But I doubt anyone will ever use that again these days.
When the speed gets limited from the ESP side of the connection, you probably have other issues like unstable voltage to the ESP.

TL;DR
A think this should be merged.

P-R-O-C-H-Y (approved):

Issue #6961, opened 2022-07-09
CORRUPT HEAP assert failed: multi_heap_free when connecting to BLE Server


savejeff:
### Board

ESP32-S3

### Device Description

DevKitC-1, soldered and Breadboard using multiple Sensors over I2C and SPI, running on both cors with Multiple Tasks

Core0: running Task0 (higher prio) and Task2 (lower prio)
Core1: running Task1 (higher prio) and Task3 (lower prio)

### Hardware Configuration

SARA R422M8N over Serial Port
SD Card over SPI
IMU over I2C
etc ...

### Version

v2.0.3

### IDE Name

PlatformIO

### Operating System

Windows 10

### Flash frequency

240Mhu

### PSRAM enabled

yes

### Upload speed

115200

### Description

Since using the ESP32 S3 with Both cores with the Core version 2.0.0 i get heap corrupted errors on BLE connect

The Code runs fine until i connect through my smartphone to the BLE Server running on the ESP32-S3 

this is what I'm getting
![image](https://user-images.githubusercontent.com/11009455/178100143-23f5072b-1534-4e1a-aa97-063b6bcaf1e3.png)

![image](https://user-images.githubusercontent.com/11009455/178100160-f5df44a5-f39d-4b6e-a254-20b248f0d70a.png)

![image](https://user-images.githubusercontent.com/11009455/178100164-8ad47f7c-73cd-4527-a472-14a756bfbb5f.png)

### Debug Message

```plain
CORRUPT HEAP: Bad tail at 0x3fcb728c. Expected 0xbaad5678 got 0xbaad5600

assert failed: multi_heap_free multi_heap_poisoning.c:253 (head != NULL)


Decoding stack results
0x403776b2: panic_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp_system/panic.c line 402
0x4037eea1: esp_system_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp_system/esp_system.c line 128
0x40384cbd: __assert_func at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/newlib/assert.c line 85
0x40384917: multi_heap_free at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/heap/multi_heap_poisoning.c line 245
[...]

PR #7597, opened 2022-12-18
Fix potential CORRUPT HEAP problem on libraries/BLE/src/BLEDevice.cpp

CurryEx:
## Description of Change
Let `BLEDevice::removePeerDevice` on `libraries/BLE/src/BLEDevice.cpp` serializable to avoid potential CORRUPT HEAP problem

## Tests scenarios
_excuse for my bad English_
First of all, to recur this problem, you can try these code
```cpp
#include <BLEDevice.h>
#include <Esp.h>

static BLEClient *pClient;
const BLEUUID serviceUUID("ebe0ccb0-7a0a-4b0c-8a1a-6ff2997da3a6");
const BLEUUID charUUID("ebe0ccc1-7a0a-4b0c-8a1a-6ff2997da3a6");
static BLERemoteService *pService;
static BLERemoteCharacteristic *pCharacteristic;

//a simple connect function, after connecting, to operate one servcice
bool connectAndSetNotify(BLEAddress address)
{
	if (pClient->connect(address) == true)
	{
		Serial.printf("signal strength %d \n", pClient->getRssi());

		pService = pClient->getService(serviceUUID);
		if (pService == nullptr)
		{
			Serial.print("connection broken: fail to get service");
			if (pClient->isConnected())
				pClient->disconnect();
			return false;
		}
		pCharacteristic = pService->getCharacteristic(charUUID);
		if (pCharacteristic == nullptr)
		{
			Serial.print("connection broken: fail to get characteristic");
			if (pClient->isConnected())
				pClient->disconnect();
			return false;
		}
		else
		{
			pCharacteristic->registerForNotify([](BLERemoteCharacteristic *pBLERemoteCharacteristic_THB, uint8_t *pData, size_t length, bool isNotify) {});
			return true;
		}
	}
	else
	{
		Serial.println("fail to connect");
		return false;
	}
}

void setup()
{
	Serial.begin(115200);
	BLEDevice::init("ESP32");
	pClient = BLEDevice::createClient();
}

void loop()
{
	BLEAddress addressAvailable("a4:c1:38:00:00:00");
	BLEAddress addressNotAvailable("a4:c1:38:00:00:01");
	//the first device MUST be connectable and the second device MUST be not connectable
	Serial.println("connecting to an available device");
	connectAndSetNotify(addressAvailable);
	pCharacteristic->registerForNotify(NULL);
	pClient->disconnect();

	//wait for disconnect
	delay(2000);
	Serial.println("connecting to a not available device");
	connectAndSetNotify(addressNotAvailable);
	Serial.println("!!can not reach!!");
	delay(10 * 1000);
}
```

In my opinion, there is no problem, but you will get CORRUPT HEAP exception just like issue #6961 and this
image

After debugging, I found there may have some situations that invoking this function simultaneously, such as while connecting to a not available device will invoke this function after `rc != ESP_GATT_OK` on `BLEClient.cpp` and `ESP_GATTC_DISCONNECT_EVT`on `BLEClient::gattClientEventHandler`.

Invoke simultaneously may cause concurrency problem of unsafe map `m_connectedClientsMap`

Therefore, my solution is to add a `mutex lock` to this function, if there is some better way to fix, please revise my PR

I have tested my Pull Request on framework-arduinoespressif32 3.20004.0(the environment I use is PIO I copy the version info from `package.json`) with ESP32 and ESP32-CAM Board

## Related links
maybe related to issue #6961


libraries/BLE/src/BLEDevice.cpp
```diff
@@ -629,10 +629,15 @@ void BLEDevice::addPeerDevice(void* peer, bool _client, uint16_t conn_id) {
 	m_connectedClientsMap.insert(std::pair<uint16_t, conn_status_t>(conn_id, status));
```
me-no-dev, line 629: shouldn't the same mux be used here and any place where `m_connectedClientsMap` is accessed?

CurryEx: As far as I'm noticed, the `insert` and other operations to `m_connectedClientsMap` will not be invoked concurrently. So, it should not be necessary🤔.

me-no-dev: are you positive that insert and find/erase can not happen at the same time?

CurryEx: `erase` operation will happen at the same time above the situation of this pull request. `find` operation will not cause concurrent problem. Regretfully, I haven't found enough evidence so far to promise if the `insert` operation will be invoked simultaneously in this library, however this is unlikely to happen. Therefore, I only made the `BLEDevice::removePeerDevice` serializable.

PR #7613, opened 2022-12-21
Removes a few Warnings and fixes Math rand() to work like Arduino mainstream

SuGlider:
## Description of Change
This is part of a work to clean up Warnings from the core. The flags below are not necessary anymore:
```
-Wno-frame-address 
-Wno-error=unused-function 
-Wno-error=unused-variable 
-Wno-error=deprecated-declarations 
-Wno-sign-compare 
-Wwrite-strings
-Wno-error=unused-but-set-variable 
-Wno-old-style-declaration
```

But we still need `-Wno-unused-parameter` because IDF inline declarations use some parameters that are not used dependening on  `#define`

This PR also fixes Math rand(negative parameter) in order to make it work exactly as it does in Arduino mainstream.
``` cpp
  Serial.println("random(-5) ::");
  for (uint8_t i = 0; i < 10; i++) {
  Serial.print(random(-5));
  Serial.print(" | ");
  }
  Serial.println();

  Serial.println("random(-3, 3) ::");
  for (uint8_t i = 0; i < 10; i++) {
  Serial.print(random(-3, 3));
  Serial.print(" | ");
  }
  Serial.println();
```

## Tests scenarios
Tested with ESP32 WROVER-E, using a modified `platform.txt` with this change:
```
compiler.c.flags.esp32=-Wno-unused-parameter -mlongcalls -ffunction-sections -fdata-sections -ggdb -freorder-blocks -fstack-protector -fstrict-volatile-bitfields -fno-jump-tables -fno-tree-switch-conversion -std=gnu99  -MMD -c

compiler.cpp.flags.esp32=-Wno-unused-parameter -mlongcalls -ffunction-sections -fdata-sections -ggdb -freorder-blocks -fstack-protector -fstrict-volatile-bitfields -fno-jump-tables -fno-tree-switch-conversion -std=gnu++11 -fexceptions -fno-rtti  -MMD -c
```

## Related links
It is a pre requisite for merging #7060


me-no-dev (approved):

cores/esp32/WMath.cpp
```diff
@@ -36,25 +36,17 @@ void randomSeed(unsigned long seed)
...
-            l = uint32_t(m);
-        }
-    }
-    return m >> 32;
+  if ( howbig == 0 )
+  {
+    return 0 ;
+  }
+  if (howbig < 0) {
+    return (random(0, -howbig));
```
P-R-O-C-H-Y, line 47: @SuGlider Lets say I call `random(-10)` but get result of 0 to 10.
Shouldn't this be `return -(random(0, -howbig));` so we invert the value in return back on what it should be?

SuGlider: No... For some reason, this is how it works in Arduino mainstream.
But in Arduino (other boards), when calling random(-5), thre results are positive and between 0 and 4.

cores/esp32/stdlib_noniso.c
```diff
@@ -163,15 +163,15 @@ char * dtostrf(double number, signed int width, unsigned int prec, char *s) {
...
+    for (unsigned int  i = 0; i < prec; ++i)
         rounding *= 10.0;
     rounding = 1.0 / rounding;
 
     number += rounding;
 
     // Figure out how big our number really is
     double tenpow = 1.0;
-    int digitcount = 1;
+    unsigned int  digitcount = 1;
```
P-R-O-C-H-Y, line 174: One more space added after int, but no big deal.

cores/esp32/WMath.cpp
SuGlider: I tested that with other 3 non-ESP32 Arduino frameworks to make sure it was not a "bug".
There are also some discussion in Arduino Forum about it and all point to returning positive values for any rand(long value)
The only way to get negative values is using range with rand(val1, val2)

cores/esp32/stdlib_noniso.c
SuGlider: :-)
I see you don't like spaces....

SuGlider: done.

cores/esp32/WMath.cpp
P-R-O-C-H-Y: Okey thanks for nice explanation 👍

P-R-O-C-H-Y (approved):

Issue #6624, opened 2022-04-25
I want the ESP32 default setting to be HT20.


CheolsoonKwon:
### Board

ESP32-WROOM-32D

### Device Description

Devkit

### Hardware Configuration

GPIO 16, 17pin is UART 

### Version

v2.0.2

### IDE Name

Arduino

### Operating System

Windows10

### Flash frequency

80M

### PSRAM enabled

no

### Upload speed

921600

### Description

Hello Everyone

I want the ESP32 default setting to be HT20.

I changed the settings because I couldn't capture in Wireshark.

Change to HT20 -> HT40+ to capture.

ESP32 Arduino seems to work with HT40+.

Previously, the test condition was Laptop(Station) <-> ESP32(AP).

It worked without any problems, and when I changed the Laptop(station) to an HT20 dedicated device, a data transmission/reception problem occurred.

Is it possible to change ESP32's HT40 setting to HT20 on Arduino?

### Sketch

```cpp
//Server Configuration
    Serial.println("Configuring access point...");
    // You can remove the password parameter if you want the AP to be open.
    WiFi.softAPConfig(local_IP, gateway, subnet);
    WiFi.softAP(Read_SSID, Read_PW);
    IPAddress myIP = WiFi.softAPIP();
    Serial.print("AP IP address: ");
    Serial.println(myIP);
    server.begin();
    server.setNoDelay(true);
    Serial.println("Server started");
    SSID_PW_Change_Flag = 2;
    SSID_PW_Diff_Flag = 2;
    Serial.println("AP START");
```

### Debug Message

```plain
No Debug Message
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

lbernstone:
[esp_wifi docs](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv416wifi_interface_t)
```
#include <WiFi.h>
#include <esp_wifi.h>

void setup() {
  Serial.begin(115200);
  WiFi.softAP("led101010", "test1234");
  esp_wifi_set_bandwidth(WIFI_IF_AP, WIFI_BW_HT20);
}

void loop() {

```
#include <WiFi.h>
#include <esp_wifi.h>
void setup() {
  Serial.begin(115200);
  WiFi.softAP("esp32", "test1234");
[...]

PR #7619, opened 2022-12-23
Added WiFi Bandwidth Setting Methods for AP and STA modes.

gonzabrusco:
-----------
## Description of Change
Now you can control the bandwidth of the WiFi in AP and STA modes. Before this change, the WiFi always defaulted to HT40 (at least on my ESP32). Not you can control if you want to use HT20 or HT40. This is not tested, but I expect the HT20 working mode to have a better range. 

## Tests scenarios
I tested this on a devkit and also on a custom board with an ESP32.

## Related links
https://github.com/espressif/arduino-esp32/issues/6624


gonzabrusco:
Ok. Don't merge just yet. I think that by chance I hit a bug in the IDF. Don't know why, but if I set the STA bandwidth or the AP bandwidth.... Ethernet begins to fail (I receive the ETH_CONNECTED event but never the ETH_GOT_IP event). Instead, I start to receive these error messages:

```
E (120597) esp.emac: emac_esp32_transmit(229): insufficient TX buffer size
E (135597) esp.emac: emac_esp32_transmit(229): insufficient TX buffer size
E (150598) esp.emac: emac_esp32_transmit(229): insufficient TX buffer size
```

Maybe the first, second, third time works ok. But after plugging and unplugging the Ethernet cable. It begins to fail. If I remove the bandwidth() and softAPbandwidth() calls, everything works ok....

How come this is related? 🤪

mrengineer7777:
As a side effect this also starts SoftAP (AP) or WiFi (STA) modes.  I vote for checking WiFi.mode() rather than starting AP or STA.

gonzabrusco:
I did this that way because if SoftAP or STA is not enabled, then the bandwidth IDF call gets ignored. I expected that if somebody is setting the WiFi bandwidth, then I guess it's about to use it... so why not enable it? That's what other methods in this classes do.

PR #7629, opened 2022-12-26
feat: adding possibility to manually set MD5 checksum and authorization for HTTP update

vlastahajek:
I've written [a library](https://github.com/bonitoo-io/ESP-Github-Updater) that allows updating from Github releases. 
To verify firmware integrity HTTPUpdater already supports MD5 checksum but it is only read from the server response header. This is not possible if there is no control of the server.
If the checksum is known from another source, it is not possible to set it currently. 
This PR allows this.
The same addition was already done for [ESP8266](https://github.com/esp8266/Arduino/pull/8204).

Update 23.2.2023:
Adding also setting of authorization to allow connecting to a private server.
This is already part of ESP8266 HTTPUpdate.


libraries/HTTPUpdate/src/HTTPUpdate.h
```diff
@@ -121,6 +126,7 @@ class HTTPUpdate
 private:
     int _httpClientTimeout;
     followRedirects_t _followRedirects;
+    String _md5Sum;
```
me-no-dev, line 129: `_md5Sum = String();` should be added in the class constructor after `_followRedirects` is set.

vlastahajek: But `String _md5Sum;` means it is automatically initialized empty String instance, isn't it?

me-no-dev: the compiler will complain. The String is not necessarily initialized. You can initialize it in two general ways:
```cpp
HTTPUpdate::HTTPUpdate(int httpClientTimeout): _httpClientTimeout(httpClientTimeout), _ledPin(-1), _md5Sum(){
  _followRedirects = HTTPC_DISABLE_FOLLOW_REDIRECTS;
}
```

Or
```cpp
HTTPUpdate::HTTPUpdate(int httpClientTimeout): _httpClientTimeout(httpClientTimeout), _ledPin(-1){
  _followRedirects = HTTPC_DISABLE_FOLLOW_REDIRECTS;
  _md5Sum = String();
}
```

vlastahajek: Ok, understood. Thanks for the correction 👍. Fixed.

vlastahajek:
I will follow this PR by adding also authorization. This will finally make this on par with ESP8266 HTTPUpdate.

vlastahajek:
@me-no-dev, what with this PR? 
I've seen another PR, that added a request update callback, was accepted even to the update release. Its feature allows also adding authorization, however, authorization is a common feature that deserves its own API functions.

me-no-dev:
Hi @vlastahajek ! We will have this merged for v3.0.0

Issue #5949, opened 2021-11-30
ESP32 ETH no ETH.end / disable


Typhoor:
I want do `end` or `stop` the ETH/ethernet component to conserve energy:

In ETH.H/CPP there is a `begin` method but no corressponding `end`method.
In the underlying Espressif esp-idf there is a `esp_eth_start` and a corresponding `esp_eth_stop`.

I'm using WT32_ETH01 and/or Olimex ESP32 POE mostly both are using the LAN 8720 PHY
Some have power pin some have not.

I observe that power raises from 60-80mA to 150mA when starting ETH (ethernet), but there is no way back.

Workaround so far is the set some static (flash/eeprom) markers and then reboot (not activating Ethernet). 
But this disturbes the other working tasks...

VojtechBartoska:
Hello @Typhoor, are you able to propose changes in PR?

playmiel:
Hello, I have the same problem, I would like to disable the Ethernet but I can't find any way to do it, do you have an idea?

aminzakizebarjad:
I think there is a way to do that   
`esp_eth_stop(ETH.eth_handle);`

Tarik2142:
https://github.com/espressif/esp-idf/blob/fcc43ef8ed8a4e03c3a795cb03ba560787254da6/components/esp_eth/test/test_emac.c#L193

VojtechBartoska:
Hello, for version 3.0.0 Networking was refactored. Can you please give it a try? 

I am for now closing this ticker as it is quite old, if needed, you can reopen it.

me-no-dev:
@hitecSmartHome here is the fix: https://github.com/espressif/arduino-esp32/pull/11048

hitecSmartHome:
Thank you! Will check it out!

PR #7632, opened 2022-12-27
Register Ethernet `netif` in WiFiGeneric to

s-hadinger:

### Checklist

-----------
## Description of Change

Register the ethernet `netif` object in `WiFiGeneric` `esp_netifs`. Without this registry, IP6 events are never sent back to event callback, because of the tests here: https://github.com/espressif/arduino-esp32/blob/b63f947037bece7a499faf0043636dd8531b885e/libraries/WiFi/src/WiFiGeneric.cpp#L453-L464

## Tests scenarios

Tested on Olimex POE ESP32 board, now connecting Ethernet to a network supporting IPv6 SLAAC, `ARDUINO_EVENT_ETH_GOT_IP6` is correctly sent to callbacks. Before this fix, no event was called.


mrengineer7777:
Code analysis performed on arduino-esp32 v2.0.5.  Disclaimer: I don't fully understand why WiFiGeneric tracks eth_netif_t entries, or how ETH module works.

Discussion:

add_esp_interface_netif() isn't used anywhere else in code.  If used, it should probably be exported in WiFiGeneric.h.

WiFiGeneric.cpp keeps track of network interfaces in private array `static esp_netif_t* esp_netifs[ESP_IF_MAX] = {NULL, NULL, NULL}`.  Entries for AP and STA interfaces appear to be directly created in `wifiLowLevelInit()`, but the ethernet entry is not setup(?).

Ethernet interface is created in ETH begin:
``
bool ETHClass::begin(uint8_t phy_addr, int power, int mdc, int mdio, eth_phy_type_t type, eth_clock_mode_t clock_mode, bool use_mac_from_efuse)
{
...
    esp_netif_config_t cfg = ESP_NETIF_DEFAULT_ETH();
    esp_netif_t *eth_netif = esp_netif_new(&cfg);
...
}
```

I agree `eth_netif` should be registered with WiFiGeneric.  I also agree `add_esp_interface_netif(ESP_IF_ETH, eth_netif);` is the correct way to register it.  Not sure on placement of add_esp_interface_netif(), but seems reasonable.  Definitely needs to be before esp_eth_start(), and makes sense to not add it until after `esp_netif_attach()`.

Note `ETHClass::begin()` creates a new eth_netif_t entry every time it is called.  If ETHClass::end() [#5949] is added later, end() may need to unregister the ESP_IF_ETH entry.

Would it be better to create the ethernet entry in `wifiLowLevelInit()`?  Side effects unknown.

TD-er:
Yep, WiFi events also need to be unregistered, or else you will get them multiple times.

N.B. For WiFi events you need to register their callback function(s) each time you enable some interface that wasn't enabled before. e.g. AP only will not register STA events.

So it makes sense to register these events when you call begin() and unregister them when you call end().
However, if my memory serves me well, there wasn't an end() function for ETH.

If there still isn't an end() function, then you can register those events in the lowlevel init function (given it will only called once)

mrengineer7777:
> Yep, WiFi events also need to be unregistered, or else you will get them multiple times.
wifiLowLevelInit() checks for event handle before registering again. Looks safe.

> N.B. For WiFi events you need to register their callback function(s) each time you enable some interface that wasn't enabled before. e.g. AP only will not register STA events.
AP and STA registered once in wifiLowLevelInit().
 
> So it makes sense to register these events when you call begin() and unregister them when you call end(). However, if my memory serves me well, there wasn't an end() function for ETH.
Correct.  Currently no ETH.end().

TD-er:
> AP and STA registered once in wifiLowLevelInit().

But when you add your own event callback functions, the ones meant for interfaces not being enabled, will not be registered.

PR #7638, opened 2022-12-29
Update HTTPClient.cpp to send big payload

ekapujiw2002:
## Description of Change
This update will handle sending big payload packet correctly by sending it per determined chunk size, for example per 1000 bytes. Also fix content length header on subsequence redirection request.

## Tests scenarios
Tested on ESP32-CAM AiThinker using Arduino-esp32 core v3.3.1 for sending photo as base64 payload to GDrive via Google App Script. If payload size is more than 16384 bytes then **HTTPClient::sendRequest** will fail.

## Related links
None

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -7,7 +7,7 @@
  * Copyright (c) 2015 Markus Sattler. All rights reserved.
  * This file is part of the HTTPClient for Arduino.
  * Port to ESP32 by Evandro Luis Copercini (2017), 
- * changed fingerprints to CA verification. 												 
+ * changed fingerprints to CA verification. 
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -605,8 +605,8 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
         if(payload && size > 0) {
             addHeader(F("Content-Length"), String(size));
         } else {
-			// force 0 content length for redirect request
-			addHeader(F("Content-Length"), String("0"));
+            // force 0 content length for redirect request
+            addHeader(F("Content-Length"), String("0"));
         }
 
         // add cookies to header, if present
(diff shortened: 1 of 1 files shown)


libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -619,10 +622,21 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
 
         // send Payload if needed
         if(payload && size > 0) {
-            if(_client->write(&payload[0], size) != size) {
-                return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+			
+			// this will fail for big payload
```
atanisoft, line 626: Instead of leaving the old code in comments, it would be better to simply drop it.

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -619,10 +622,21 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
...
+//             if(_client->write(&payload[0], size) != size) {
+//                 return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+//             }
+			
+			// send the payload per 1000 bytes
+			log_d("Sending %u byte ...\n", size);
+            size_t sent_bytes = 0;
+			for (size_t id_data = 0; id_data < size; id_data+=1000)
```
atanisoft, line 634: instead of a for loop using fixed sizes for chunks, please consider a loop similar to below (note this is not tested):
```
while(sent_bytes < size) {
  size_t sent = _client->write(&payload[sent_bytes], std::min(size - sent_bytes, _chunkSize));
  if (sent < 0) {
    log_e("Failed to send chunk!");
    break;
  }
  sent_bytes += sent;
}
```
Note that there is *ZERO* guarantee that the write call will send the full block size you request it to send thus you must check that it actually sent data and only then move the data starting point.  The check for negative sent byte count is to check for sending errors.

P-R-O-C-H-Y (changes requested):
@ekapujiw2002 Did you take a look on comments from @atanisoft? If not, please do the changes you were asked for.

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -619,10 +622,21 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
...
-            if(_client->write(&payload[0], size) != size) {
-                return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+			
+			// this will fail for big payload
+//             if(_client->write(&payload[0], size) != size) {
+//                 return returnError(HTTPC_ERROR_SEND_PAYLOAD_FAILED);
+//             }
+			
+			// send the payload per 1000 bytes
+			log_d("Sending %u byte ...\n", size);
```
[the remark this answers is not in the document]
ekapujiw2002: @atanisoft Fixed. Please kindly check it. Thank you.

libraries/HTTPClient/src/HTTPClient.cpp
ekapujiw2002: @atanisoft Fixed. Please kindly check it. Thank you.

libraries/HTTPClient/src/HTTPClient.cpp
ekapujiw2002: @atanisoft Fixed. Please kindly check it. Thank you.

me-no-dev:
@ekapujiw2002 I invite you to see the changes that I have made on the topic of large uploads [here](https://github.com/espressif/arduino-esp32/pull/8006/files). I have tested it and also added some more safeguards and error propagation.

me-no-dev:
closing as already implemented

Issue #7459, opened 2022-11-10
Let the WiFiClientSecure set an hostname different from the ip address


lu-zero:
### Related area

WiFiClientSecure

### Hardware specification

Any

### Is your feature request related to a problem?

The function

```
int start_ssl_client(sslclient_context *ssl_client, const char *host, uint32_t port, int timeout, const char *rootCABuff, bool useRootCABundle, const char *cli_cert, const char *cli_key, const char *pskIdent, const char *psKey, bool insecure, const char **alpn_protos
```

uses `host` to resolve the ip and use the it as expected CN. In some cases the CN information is provided by other means 

### Describe the solution you'd like

Provide a mean to set the CN when it is different.

### Additional context

https://github.com/espressif/arduino-esp32/issues/7350 is tangentially related.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

cziter15:
There is a room for refactoring in context of SSL and WiFiClientSecure.
When you look at WiFiClientSecure, you'll see that it simply converts IPAddress to string, passing it to start_ssl_client function. 

Why? I suspect that the idea was to have a proper hostname for certificate validation (parameter of mbedtls_ssl_set_hostname). The solution is simple - introduce another optional (NULL by default) sslHostname parameter and replace current hostname parameter with IPAddress.

```
int WiFiClientSecure::connect(IPAddress ip, uint16_t port, const char *CA_cert, const char *cert, const char *private_key)
{
    return connect(ip.toString().c_str(), port, CA_cert, cert, private_key);
}

int WiFiClientSecure::connect(const char *host, uint16_t port, const char *CA_cert, const char *cert, const char *private_key)
{
    int ret = start_ssl_client(sslclient, host, port, _timeout, CA_cert, _use_ca_bundle, cert, private_key, NULL, NULL, _use_insecure, _alpn_protos);
    _lastError = ret;
    if (ret < 0) {
        log_e("start_ssl_client: %d", ret);
        stop();
        return 0;
    }
    _connected = true;
    return 1;
}
```

cziter15:
I've issued PR that aims to resolve this issue.

- Added another connect variant in WiFiClientSecure that can take both IP and hostname from the user.
- All other connect methods ('higher level') will call that underlying function. Hostname is now translated to IP in WiFiClientSecure, not in ssl_client code.

Parsaabasi:
Hello, 

[...]

Issue #7356, opened 2022-10-14
WiFiClientSecure doesn't respect timeout in send_ssl_data


cziter15:
No details about the device required as whole ecosystem is affected. 

**Affected component**: WiFiClientSecure, ssl_client

### Description

Function send_ssl_data, called by WiFiClientSecure doesn't respect timeout set on socket.

After disconnecting the network (not WiFi directly, just turn off LTE or disconnect network cable), having some small piece of code sending periodically few messages to MQTT broker via SSL, I've encountered freeze inside send_ssl_data function.

After some investigation, it appears to hang inside while loop.

Temporary working solution, but 5000 is a magic constant:

```
int send_ssl_data(sslclient_context *ssl_client, const uint8_t *data, size_t len)
{
    log_v("Writing HTTP request with % bytes...", len); //for low level debug
    int ret = -1;

    unsigned long send_start_time=millis();
    while ((ret = mbedtls_ssl_write(&ssl_client->ssl_ctx, data, len)) <= 0) {
        if((millis()-send_start_time)>5000)
            return -1;

        if (ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE && ret < 0) {
            log_v("Handling error %d", ret); //for low level debug
            return handle_error(ret);
        }
        //wait for space to become available
        vTaskDelay(2);
    }

    return ret;
}
```
The best way to fix this will be to introduce send_timeout inside sslclient_context and set it to socket timeout (passed in connect method by WiFiClientSecure).

cziter15:
I've submitted PR which adds socket_timeout to ssl_context and copies timeout value there. Then this socket_timeout variable is used in send_ssl_data as timeout.

VojtechBartoska:
@cziter15 Thank you for contribution, we will review your PR soon.

TD-er:
Could the timeout (also) be related to this one?
https://github.com/espressif/arduino-esp32/pull/6676

cziter15:
This timeout here is something specific, not related to ms vs second interpretation.

dsilletti:
@cziter15 could you please provide your fix (socket_timeout)?  This is probably the reason why my ESP32 remain stuck if I disconnect the WAN interface of my Wi-Fi router, while connected via Wi-Fi to the MQTT broker (using PubSubClient  with WiFiClientSecure and publishing every second).

The PubSubClient setSocketTimeout does not help.

Thanks

[...]

PR #7643, opened 2022-12-30
Allow passing IP as connect method parameter in WiFiClientSecure

cziter15:
## Description of change
This PR aims to improve the logic of host address parameter passing, taking into account the elimination of unnecessary conversions of the IP address into a string. Thanks to this contribution, it becomes possible to manually provide the IP address with any hostname.

- Added generic 'connect' variant that doesn't 'bind' hostname with IP address (user can provide IP manually).
- This generic variant is now called by all other functions. Hostname translation has been moved to WiFiClientSecure.
- This change also removes redundant hostByName calls (when IP address is passed).

## Tests scenarios
Tested on custom board with ESP-WROOM-32 module without PSRAM. Latest master with applied changes.

## Related links
Targeting: #7459
Related: #7350 #7356


VojtechBartoska:
Thanks for the PR @cziter15, we will review it soon.

PR #7646, opened 2022-12-31
setPins fix (CONFIG_IDF_TARGET_ESP32 only):  "specified pins are not supported by this chip."

tmfarrington:
setPins() bug fix. (Impacts ESP32 only)
LogError: SDMMCFS: specified pins are not supported by this chip.

[from comment: // ESP32 doesn't support SDMMC pin configuration via GPIO matrix.
                         // Since SDMMCFS::begin hardcodes the usage of slot 1, only check if
                         // the pins match slot 1 pins.]

setPins() was testing pin 'd1' against values for D2 and D3. (Just a common cut/paste error.) This would work fine in 1 pin mode when all are -1 but incorrectly fail when specifying the pins for 4 pin mode. I now see this function doesn't really do anything on the ESP32...accept now correctly checks if we are trying to use D2 and D3 of the slot 1 pins.

Tested On: TTGO TS V1.4


P-R-O-C-H-Y (approved):
LGTM! Thanks @tmfarrington for fixing that :)

VojtechBartoska:
@me-no-dev should be merged after your review

Issue #7647, opened 2022-12-31
Compiler creating debug.cfg, debug_custom.json, and esp32.svd files in my sketch folder under version 2.0.6


HomeSpan:
### Board

Any board

### Device Description

Occurs on all Espressif development boards

### Hardware Configuration

Nothing

### Version

other

### IDE Name

Arduino 1.8.19

### Operating System

MacOs 13.1

### Flash frequency

80 MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

In the latest release (2.0.6), whenever you compile a sketch, the compiler is dropping three new files into the sketch folder:  *debug.cfg*, *debug_custom.json*, and *esp32.svd*.  These files have never appeared when compiling with version 2.0.5 or earlier.  Is there a way of turning the automatic output of these files?

### Sketch

```cpp
Any sketch.
```

### Debug Message

```plain
None
```

### Other Steps to Reproduce

Compile any sketch with Arduino IDE 1.8.19 using the latest Arduino-ESP32 version 2.0.6 and the compiler drops three new files, presumably used for debugging/testing, in to the sketch folder. 

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

svdrummer:
I have just been asking this question on the Arduino forum. I have started getting these files being added. 
debug.cfg
debug_custom.json
esp32.svd

me-no-dev:
they are required for JTAG debugging with Arduino IDE 2.0.3+

IeiuniumLux:
@me-no-dev, The above 3 files are generated with the classic Arduino 1.8.19 IDE. Why does it now require those files too?

me-no-dev:
@IeiuniumLux it does not, but we can not detect which version is the IDE in order to copy them only on the new one

HomeSpan:
@me-no-dev, Can the output be an option that is selected from the tool menu, similar to setting the "Core Debug Level" or "Erase all Flash Before Sketch Upload"?  Perhaps something like "Output JTAG files" where the default is "No" so that the files would never be output in version 1 of the Arduino IDE and would be optionally output in version 2?

Separately, is there anything in either "platform.txt" or "programmers.txt" that we can edit to disable the JTAG output?

me-no-dev:
[...]

PR #7655, opened 2023-01-03
Copy JTAG files only if it's enabled in the board menu

me-no-dev:
Fixes: https://github.com/espressif/arduino-esp32/issues/7647

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ _build/
 
 # Test log files
 *.log
+debug.cfg
+debug.svd
+debug_custom.json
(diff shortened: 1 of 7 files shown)


P-R-O-C-H-Y (approved):

SuGlider:
@me-no-dev Do you see the Disabled JTAG menu option with IDE 2.0.3 in the Macbook?

P-R-O-C-H-Y:
@SuGlider I tested only in the old IDE 1.8.19, where it works fine. As there are known issues in the new IDE 2.0.X, where changes in the menu are not updated. Had these problems before with the JTAG testing and tried everything with no chance on Mac. I think as it works in the old IDE, it should work in the new IDE as well.

me-no-dev:
@SuGlider as @P-R-O-C-H-Y said, the new IDE has some horrible caching enabled and that prevent it from grabbing changes for the boards menus. Hopefully it will be OK when released and package updated for the users.

PR #7699, opened 2023-01-11
Fixes BLE data printing

### LINKED ISSUE #7637: Payload used as string, thus "prints" adds garbage
### Board

ESP32-S3

### Device Description

homemade

### Hardware Configuration

homemade

### Version

### IDE Name

ARduino IDE

### Operating System

linux

### Flash frequency

40

### PSRAM enabled

no

### Upload speed

115200

### Description

Hello
v2.0.6

this is the output of the demo program using BLE_client

https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/examples/BLE_client/BLE_client.ino

```
We are now connected to the BLE Server.
Notify callback for characteristic 0000ffe1-0000-1000-8000-00805f9b34fb of data length 2
data: AT⸮⸮⸮?⸮$

```
The data length is correct, I recognize just an 'AT' command from the HC06.
Line 32 of the demo program uses the payload directly as a string (serial.print) .

Maybe just add a warning for readers: this is a payload interpreted wrongly as a string for simplicity of the demo.

https://github.com/espressif/arduino-esp32/blob/b63f947037bece7a499faf0043636dd8531b885e/libraries/BLE/examples/BLE_client/BLE_client.ino#L32

### Sketch

```cpp
bool isNotify) {
    Serial.print("Notify callback for characteristic ");
    Serial.print(pBLERemoteCharacteristic->getUUID().toString().c_str());
    Serial.print(" of data length ");
    Serial.println(length);
    Serial.print("data: ");
    Serial.println((char*)pData);
```

### Debug Message

```plain
x
```

### Other Steps to Reproduce

_No response_

SuGlider:
## Description of Change
BLE data has no '\0' terminator, therefore it can't be printed as a regular C string. This fix just prints the BLE data based on its length.

## Tests scenarios
ESP32 using BLE Client and Server

## Related links
Fix #7637

### DIFF
--- a/libraries/BLE/examples/BLE_client/BLE_client.ino
+++ b/libraries/BLE/examples/BLE_client/BLE_client.ino
@@ -29,7 +29,10 @@ static void notifyCallback(
     Serial.print(" of data length ");
     Serial.println(length);
     Serial.print("data: ");
-    Serial.println((char*)pData);
+    for (size_t i = 0; i < length; i++) {
+      Serial.print((char)pData[i]);
+    }
+    Serial.println();
 }
 
 class MyClientCallback : public BLEClientCallbacks {


mrengineer7777 (approved):
Correct variable types and bounds.  Code style is consistent. LGTM.

mrengineer7777 (approved):
Uses `size_t HardwareSerial::write(const uint8_t *buffer, size_t size)`.  Parameters are correct. LGTM.

P-R-O-C-H-Y (approved):

PR #7752, opened 2023-01-24
HTTPClient Library - Fix on response headers handling

marciopamplona:
## Description of Change
The library wasn't dealing correctly with the response headers. The function collectHeaders wasn't even used.
I rewrote with vector and a class to store the values.

## Tests scenarios
Tested on ESP32-PICO-D4, Arduino core v2.0.6

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -109,8 +109,8 @@ HTTPClient::~HTTPClient()
     if(_client) {
         _client->stop();
     }
-    if(_currentHeaders) {
-        delete[] _currentHeaders;
+    if(_responseHeaders.size()) {
+        _responseHeaders.clear();
     }
     if(_tcpDeprecated) {
         _tcpDeprecated.reset(nullptr);
@@ -589,10 +589,8 @@ int HTTPClient::sendRequest(const char * type, uint8_t * payload, size_t size)
     uint16_t redirectCount = 0;
     do {
         // wipe out any existing headers from previous request
-        for(size_t i = 0; i < _headerKeysCount; i++) {
-            if (_currentHeaders[i].value.length() > 0) {
-                _currentHeaders[i].value.clear();
-            }
+        if(_responseHeaders.size()) {
+            _responseHeaders.clear();
         }
 
         log_d("request type: '%s' redirCount: %d\n", type, redirectCount);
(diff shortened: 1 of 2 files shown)


PilnyTomas:
@marciopamplona thank you for your contribution.
Could you please help us? How did you test it? Could you please provide an example that would demonstrate the broken and fixed behavior?

marciopamplona:
Hello @PilnyTomas, sure.

**How to test it:**
```C++
HTTPClient http;
String queryURL = "http://ip-api.com/json";
http.begin(queryURL.c_str());
int httpResponseCode = http.GET();

if (httpResponseCode == 200) {  
  int numberOfHeaders = http.headers();
  String date = http.header("Date");
  Serial.printf("Number of headers: %d\nDate header: %s\n\n", numberOfHeaders, date.c_str());
}
```

**Expected result:**
```
Number of headers: 6
Date header: Wed, 25 Jan 2023 21:11:46 GMT
```

**Actual result:**
```

Number of headers: 0
Date header: 
```

**Code analysis:**

LINE 1093, function `HTTPClient::headers()` returns `_headerKeysCount`, but this variable is not being set nowhere in the code, except for the LINE 1057 on function `HTTPClient::collectHeaders()` which is NOT BEING USED.
The headers are NOT BEING STORED anywhere, so the function http.header(const char* name) can't return anything.

PilnyTomas (approved):
Thanks for the explanation.
LGTM

P-R-O-C-H-Y (changes requested):
@marciopamplona Code looks overall good, thanks for the fix :)
As it is a change in API, I suggest to milestone it for version 3.0.0.

But `HTTPUpdate` library used `HTTPClients collectHeaders` function. Please adjust the `HTTPUpdate` code there to match your changes? Compiling HTTPUpdate example cast this error:
```
Users/prochy/Documents/Arduino/hardware/espressif/esp32/libraries/HTTPUpdate/src/HTTPUpdate.cpp: In member function 'HTTPUpdateResult HTTPUpdate::handleUpdate(HTTPClient&, const String&, bool)':
/Users/prochy/Documents/Arduino/hardware/espressif/esp32/libraries/HTTPUpdate/src/HTTPUpdate.cpp:224:10: error: 'class HTTPClient' has no member named 'collectHeaders'; did you mean 'connected'?
     http.collectHeaders(headerkeys, headerkeyssize);
```

The same collectHeaders function is used in WebServer. As for now, both function were the same (Server / Client). I think it would be great, if you made the changes in the WebServer side too, to it will be the same again for both libs.

@me-no-dev Do you agree?

me-no-dev:
This change is in effect removing functionality and allowing for using more memory than required.

Correct me if I am wrong, but this change collects all received headers and stores them in memory, until the request is destroyed. The point of `collectHeaders` was to limit the number of headers to only the ones that we need in our code, which saves an unknown amount of memory (because some sites return HUGE headers).

If I am correct, then this is actually a regression which also points to missing proper documentation on how to use `collectHeaders`

me-no-dev:
Thanks for confirming @marciopamplona ! @pedrominatel PTAL on adding docs for this

PR #7774, opened 2023-01-30
WString Return bool

### LINKED ISSUE #7766: Use bool in WString Return values
### Related area

WString.h

### Hardware specification

N/A

I would like our WString library to return the same values as ESP8266 and ArduinoCore-API.

I'd like to change function return values from `unsigned char` to `bool`.  This was done in both ESP8266 (https://github.com/esp8266/Arduino/pull/7939) and ArduinoCore-API (https://github.com/arduino/ArduinoCore-API/pull/147)

I could leave the library as-is, but it makes it very hard to compare changes between ESP32 and ESP8266.

### Additional context

I'm willing to submit a PR with the changes.  I believe this will have no impact on users.  Before I do the work I'd like feedback from repository maintainers that this will be helpful/accepted.

mrengineer7777:
-----------
## Description of Change
Changed function return values from unsigned char to bool. This makes our library consistent with ESP8266 (https://github.com/esp8266/Arduino/pull/7939) and ArduinoCore-API (https://github.com/arduino/ArduinoCore-API/pull/147).  It also makes comparisons easier.

Discussion:
To avoid bugs, this PR focuses ONLY on boolean returns.  It does consolidate the return value in `changeBuffer()` to match ESP8266.  I'm confident the functionality is the same.

The function `unsigned char String::equalsConstantTime(const String &s2) const` should likely be bool.  Not touching it because changing the implementation might affect constant timing.  This matches ESP8266.

## Tests scenarios
Tested on arduino-esp32 v2.0.5 and hardware Adafruit Huzzah32 Feather.  This uses an ESP32-WROOM-32E.  Runs our custom code with no observable changes (as expected).

## Related links
Closes #7766

### DIFF
--- a/cores/esp32/WString.cpp
+++ b/cores/esp32/WString.cpp
@@ -176,26 +176,25 @@ void String::invalidate(void) {
     init();
 }
 
-unsigned char String::reserve(unsigned int size) {
+bool String::reserve(unsigned int size) {
     if(buffer() && capacity() >= size)
-        return 1;
+        return true;
     if(changeBuffer(size)) {
         if(len() == 0)
             wbuffer()[0] = 0;
-        return 1;
+        return true;
     }
-    return 0;
+    return false;
 }
 
-unsigned char String::changeBuffer(unsigned int maxStrLen) {
+bool String::changeBuffer(unsigned int maxStrLen) {
     // Can we use SSO here to avoid allocation?
     if (maxStrLen < sizeof(sso.buff) - 1) {
         if (isSSO() || !buffer()) {
             // Already using SSO, nothing to do
-	    uint16_t oldLen = len();
+            uint16_t oldLen = len();
             setSSO(true);
             setLen(oldLen);
-            return 1;
         } else { // if bufptr && !isSSO()
             // Using bufptr, need to shrink into sso.buff
             char temp[sizeof(sso.buff)];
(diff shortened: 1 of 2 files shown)


VojtechBartoska:
Thanks for quickly delivered PR @mrengineer7777! We'll review it soon to have it in upcoming 2.0.7 release.

me-no-dev (approved):

SuGlider:
@mrengineer7777 - I just have a simple question:
It is the same implementation as for ESP8266, but, in line 224 of WString.h
https://github.com/espressif/arduino-esp32/blob/710db38f7321f90387aec38e98af5c9ea6163e29/cores/esp32/WString.h#L224

https://github.com/esp8266/Arduino/blob/master/cores/esp8266/WString.h#L242

Why it is not also `bool` instead of `unsigned char`?
Should it be changed in both ESP32 and ESP8266 to `bool`?

mrengineer7777:
> @mrengineer7777 - I just have a simple question: It is the same implementation as for ESP8266, but, in line 224 of WString.h
> 
> https://github.com/espressif/arduino-esp32/blob/710db38f7321f90387aec38e98af5c9ea6163e29/cores/esp32/WString.h#L224
> 
> https://github.com/esp8266/Arduino/blob/master/cores/esp8266/WString.h#L242
> 
> Why it is not also `bool` instead of `unsigned char`? Should it be changed in both ESP32 and ESP8266 to `bool`, along with respective code?

Hi @SuGlider .  As I noted in my description above, `equalsConstantTime` should probably be `bool`.  However to apply that you'll have to modify the implementation of the function, which could break the constant time functionality.  I didn't want to risk introducing a bug and figured the ESP8266 guys knew what they were doing, so I left it alone.  I feel any such change should be done in a separate PR and coordinated with both platforms.  For now I recommend merging this PR as is.

SuGlider (approved):

SuGlider:
@me-no-dev - Ready for merging.
Thanks @mrengineer7777!

PR #7778, opened 2023-01-31
Added methods + example to retrive local MAC for BT

PilnyTomas:
It was not easy or obvious how to retrieve a MAC address of a local BT device.
There is an IDF function `esp_bt_dev_get_address` which accomplishes that and is used in this solution.
However, there was no `BluetoothSerial`method that would facilitate this.
This PR adds three new methods, slightly extends the `BTAddress::toString`, and adds a new example `GetLocalMAC` which demonstrates the usage of the new additions.

### DIFF
--- a/libraries/BluetoothSerial/src/BTAddress.cpp
+++ b/libraries/BluetoothSerial/src/BTAddress.cpp
@@ -44,7 +44,7 @@ BTAddress::BTAddress() {
  *
  * @param [in] stringAddress The hex representation of the address.
  */
-BTAddress::BTAddress(std::string stringAddress) {
+BTAddress::BTAddress(String stringAddress) {
 	if (stringAddress.length() != 17) return;
 
 	int data[6];
@@ -97,7 +97,7 @@ esp_bd_addr_t *BTAddress::getNative() const {
  * ```
  * @return The string representation of the address.
  */
-std::string BTAddress::toString(bool capital) const {
+String BTAddress::toString(bool capital) const {
 	auto size = 18;
 	char *res = (char*)malloc(size);
 	if(capital){
--- a/libraries/BluetoothSerial/src/BTAddress.cpp
+++ b/libraries/BluetoothSerial/src/BTAddress.cpp
@@ -91,7 +91,7 @@ esp_bd_addr_t *BTAddress::getNative() const {
  * ```
  * xx:xx:xx:xx:xx:xx
  * ```
- * When the parameter `caputal` == true the format uses capital letters:
+ * When the parameter `capital` == true the format uses capital letters:
  * ```
  * XX:XX:XX:XX:XX:XX
  * ```
@@ -100,11 +100,11 @@ esp_bd_addr_t *BTAddress::getNative() const {
 std::string BTAddress::toString(bool capital) const {
 	auto size = 18;
 	char *res = (char*)malloc(size);
-  if(capital){
-    snprintf(res, size, "%02X:%02X:%02X:%02X:%02X:%02X", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
-  }else{
-    snprintf(res, size, "%02x:%02x:%02x:%02x:%02x:%02x", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
-  }
+	if(capital){
+		snprintf(res, size, "%02X:%02X:%02X:%02X:%02X:%02X", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
+	}else{
+		snprintf(res, size, "%02x:%02x:%02x:%02x:%02x:%02x", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
+	}
 	std::string ret(res);
 	free(res);
 	return ret;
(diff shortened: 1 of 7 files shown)


libraries/BluetoothSerial/src/BTAddress.cpp
```diff
@@ -86,19 +86,25 @@ esp_bd_addr_t *BTAddress::getNative() const {
...
- *
- * A string representation of an address is in the format:
- *
+ * @param [in] capital changes the letter size
+ * By default the parameter `capital` == false and the string representation of an address is in the format:
  * ```
  * xx:xx:xx:xx:xx:xx
  * ```
- *
+ * When the parameter `caputal` == true the format uses capital letters:
```
P-R-O-C-H-Y, line 94: Fix typo ... parameter **`capital`** == true ...

libraries/BluetoothSerial/src/BTAddress.cpp
```diff
@@ -86,19 +86,25 @@ esp_bd_addr_t *BTAddress::getNative() const {
...
-std::string BTAddress::toString() const {
+std::string BTAddress::toString(bool capital) const {
 	auto size = 18;
 	char *res = (char*)malloc(size);
-	snprintf(res, size, "%02x:%02x:%02x:%02x:%02x:%02x", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
+  if(capital){
+    snprintf(res, size, "%02X:%02X:%02X:%02X:%02X:%02X", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
+  }else{
+    snprintf(res, size, "%02x:%02x:%02x:%02x:%02x:%02x", m_address[0], m_address[1], m_address[2], m_address[3], m_address[4], m_address[5]);
+  }
```
P-R-O-C-H-Y, line 107: Align the if/else with spaces to match the format in the function.

libraries/BluetoothSerial/src/BluetoothSerial.cpp
```diff
@@ -1183,4 +1181,18 @@ std::map<int, std::string> BluetoothSerial::getChannels(const BTAddress &remoteA
...
+
+BTAddress BluetoothSerial::getBtAddressObject(){
+  uint8_t mac_arr[ESP_BD_ADDR_LEN];
+  getBtAddress(mac_arr);
+  return BTAddress(mac_arr);
+}
+
+std::string BluetoothSerial::getBtAddressString(){
+  return getBtAddressObject().toString(true);
+}
```
P-R-O-C-H-Y, line 1197: I have checked that the formatting style isn't the same in whole file, but can you align the spaces as is in the function above?

P-R-O-C-H-Y (changes requested):
@PilnyTomas please check my comments (typo and format issues), example is working great! Thank for the PR. After fixes it will be ready to merge.

P-R-O-C-H-Y (approved):
LGTM!

libraries/BluetoothSerial/src/BTAddress.cpp
```diff
@@ -86,19 +86,25 @@ esp_bd_addr_t *BTAddress::getNative() const {
...
  * ```
- *
+ * When the parameter `capital` == true the format uses capital letters:
+ * ```
+ * XX:XX:XX:XX:XX:XX
+ * ```
  * @return The string representation of the address.
  */
-std::string BTAddress::toString() const {
+std::string BTAddress::toString(bool capital) const {
```
me-no-dev, line 100: I see that originally the method returns `std::string`, but please change it to Arduino `String` instead.

libraries/BluetoothSerial/src/BluetoothSerial.cpp
```diff
@@ -1183,4 +1181,31 @@ std::map<int, std::string> BluetoothSerial::getChannels(const BTAddress &remoteA
...
+    uint8_t mac_arr[ESP_BD_ADDR_LEN];
+    getBtAddress(mac_arr);
+    return BTAddress(mac_arr);
+}
+/**
+ * @brief      Gets the MAC address of local BT device as string.
+ *
+ * @return     The BT MAC address string.
+ */
+std::string BluetoothSerial::getBtAddressString() {
```
me-no-dev, line 1208: Same here `std::string` =>> `String`

libraries/BluetoothSerial/src/BluetoothSerial.h
```diff
@@ -85,6 +85,9 @@ class BluetoothSerial: public Stream
         const int MAX_INQ_TIME = (ESP_BT_GAP_MAX_INQ_LEN * INQ_TIME);
         
         operator bool() const;
+        void getBtAddress(uint8_t *mac);
+        BTAddress getBtAddressObject();
+        std::string getBtAddressString();
```
me-no-dev, line 90: you forgot to change the type here

libraries/BluetoothSerial/src/BTAddress.h
```diff
@@ -24,12 +24,12 @@ class BTAddress {
 public:
```
me-no-dev, line 24: up in the includes, you need to change `#include <string>` to `#include "Arduino.h"`

me-no-dev (changes requested):
Change include and left over `std::string` in BluetoothSerial.h

me-no-dev (approved):

PR #7791, opened 2023-02-03
Fix for negative temp in Eddystone TLM; solving #7618

### LINKED ISSUE #7618: BLE Beacon Scanner example might not handle negative temperatures?
### Board

Olimex ESP32 PoE ISO

### Device Description

PoE PCB using the ESP32 WROOM 32UE

### Hardware Configuration

Nothing special

### Version

### IDE Name

Arduino IDE

### Operating System

All

### Flash frequency

80Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

https://github.com/espressif/arduino-esp32/blob/master/libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino#L118

When low temperatures occur, and go below 0 C, the calculations do not seem to be correct.  It appears that using these 16-bit values in a 32-bit INT doesn't detect/use twos complement to handle the negative numbers properly.

In my final code I moved to a "signed short" which seems to fix the issue.

### Sketch

```cpp
signed short value;

          // extract the temperature
          value = (cServiceData[TEMP_MSB] << 8) + cServiceData[TEMP_LSB];
```

### Debug Message

```plain
No debug ... just wrong calculated values.
```

### Other Steps to Reproduce

_No response_

### LINKED ISSUE #7858: BLE on ESP32: BLEBeacon::getProximityUUID() returns reversed UUID string compared to expected value
### Board

ESP32 Dev Module

### Device Description

ESP-WROOM-32 board.

### Hardware Configuration

No external connections except USB connection to the Host PC.

### Version

v2.0.6

### IDE Name

Arduino IDE 2.0

### Operating System

Windows 10

### Flash frequency

40 MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Calling **BLEBeacon::getProximityUUID()** after constructing BLEBeacon object **returns reversed UUID string** compared to what is expected.

**The code:** 
```
  BLEBeacon oBeacon = BLEBeacon();
  oBeacon.setData(strManufacturerData);
  // strManufacturerData is obtained via calling advertisedDevice.getManufacturerData() during BLE Scan
  
  System.println(oBeacon.getProximityUUID().toString().c_str());

```
outputs:                c1655d97-b7ac-d1b2-e940-23f200eb28d4
expected value is: d428eb00-f223-40e9-b2d1-acb7975d65c1

After analyzing the related code, the following was found in BLEBeacon.cpp:

```
BLEUUID BLEBeacon::getProximityUUID() {
	return BLEUUID(m_beaconData.proximityUUID, 16, **false**);
}

```
The last parameter decides whether bytes should be reversed in memory in the implementation of BLEUUID.

After modyfying getProximityUUID method to:
```
BLEUUID BLEBeacon::getProximityUUID() {
	return BLEUUID(m_beaconData.proximityUUID, 16, **true**);
}

```
**The above code returns UUID value as expected:**

outputs:                d428eb00-f223-40e9-b2d1-acb7975d65c1
expected value is: d428eb00-f223-40e9-b2d1-acb7975d65c1

**Conclusion:**
the last parameter in BLEUUID constructor call in BLEBeacon::getProximityUUID should be changed to true.

### Sketch

```cpp
/*
   Based on Neil Kolban example for IDF: https://github.com/nkolban/esp32-snippets/blob/master/cpp_utils/tests/BLE%20Tests/SampleScan.cpp
   Ported to Arduino ESP32 by Evandro Copercini
   Changed to a beacon scanner to report iBeacon, EddystoneURL and EddystoneTLM beacons by beegee-tokyo
*/

#include <Arduino.h>

#include <BLEDevice.h>
#include <BLEUtils.h>
#include <BLEScan.h>
#include <BLEAdvertisedDevice.h>
#include <BLEBeacon.h>

#define ENDIAN_CHANGE_U16(x) ((((x)&0xFF00) >> 8) + (((x)&0xFF) << 8))

int scanTime = 5; //In seconds

BLEScan *pBLEScan;

class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
{
    void onResult(BLEAdvertisedDevice advertisedDevice)
    {
        if (true == advertisedDevice.haveManufacturerData())
        {
          std::string strManufacturerData = advertisedDevice.getManufacturerData();

          uint8_t cManufacturerData[100];
          strManufacturerData.copy((char *)cManufacturerData, strManufacturerData.length(), 0);

          int iRSSI = -9999;

          if (25 == strManufacturerData.length() && 0x4c == cManufacturerData[0] && 0x00 == cManufacturerData[1])
          {
            if (advertisedDevice.haveRSSI())
            {
              iRSSI = advertisedDevice.getRSSI();
            }

            Serial.println("Found an iBeacon!");
			      Serial.println("Raw data:");
			      Serial.print("    ");
			      for (int iCnt = 0; 25 > iCnt; iCnt++) 
			      {
              Serial.printf("[%02X]", cManufacturerData[iCnt]);
			      }
			      Serial.println();

            BLEBeacon oBeacon = BLEBeacon();

            oBeacon.setData(strManufacturerData);

			      Serial.println("\nDecoded data:");
            Serial.printf("    ID: %04X Major: %d Minor: %d UUID: %s Power: %d\n", oBeacon.getManufacturerId(), ENDIAN_CHANGE_U16(oBeacon.getMajor()), ENDIAN_CHANGE_U16(oBeacon.getMinor()), oBeacon.getProximityUUID().toString().c_str(), oBeacon.getSignalPower());
            Serial.printf("    RSSI: %04d\n", iRSSI);

         }
        }

        return;
      }
};

void setup()
{
  Serial.begin(115200);

  BLEDevice::init("");

  pBLEScan = BLEDevice::getScan(); //create new scan
  pBLEScan->setAdvertisedDeviceCallbacks(new MyAdvertisedDeviceCallbacks());
  pBLEScan->setActiveScan(true); //active scan uses more power, but get results faster
  pBLEScan->setInterval(100);
  pBLEScan->setWindow(99); // less or equal setInterval value

  Serial.println("Configuration finished...");
}

void loop()
{
  Serial.println("******************************************************************************");

  // Start scanning
  BLEScanResults foundDevices = pBLEScan->start(scanTime, false);

  // Print scan results summary
  Serial.printf("Devices found: %d total\n", foundDevices.getCount());

  // Clear scanning results to release memory
  pBLEScan->clearResults();

  Serial.println("Scan done!");
  Serial.println("==============================================================================");

  delay(10000);
}
```

### Debug Message

```plain
No error messages. This is output of the snip. You can clearly see how UUID is reversed compared to raw data (UUID start on index 4 and ends on index 19 of Manufacturer data for iBeacon). ID, Major, Minor and Power values are as expected:

Found an iBeacon!
Raw data:
    [4C][00][02][15][D4][28][EB][00][F2][23][40][E9][B2][D1][AC][B7][97][5D][65][C1][00][01][00][10][BF]

Decoded data:
    ID: 004C Major: 1 Minor: 16 UUID: c1655d97-b7ac-d1b2-e940-23f200eb28d4 Power: -65
```

### Other Steps to Reproduce

Reproduced also on Linux.

PilnyTomas:
This PR changes the conversion of 8.8 to float and vise versa to accommodate negative values.

Fix #7618 
Fix #7858


SuGlider (changes requested):

me-no-dev:
@SuGlider PTAL

SuGlider:
@PilnyTomas 

I have tested the original version of `BLE_EddystoneTLM_Beacon.ino` example using the [nRF Connect Android APP](https://play.google.com/store/apps/details?id=no.nordicsemi.android.mcp&gl=US) nad it displays the ESP32 TLM information when I click on the name of the device.

It works perfectly well. No need to change it.

image

SuGlider:
For the screenshot above this is the Serial output:

```
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x5 (DSLEEP),boot:0x8 (SPI_FAST_FLASH_BOOT)
pro cpu reset by JTAG
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3808,len:0x44c
load:0x403c9700,len:0xbec
load:0x403cc700,len:0x2920
entry 0x403c98d8
start ESP32 14
deep sleep (284s since last reset, 21s since last boot)
Random temperature is 26.37C
Converted to 8.8 format 1A5E
Advertizing started for 10s ...
enter deep sleep for 10s
```

When I click on "RAW", I can see the Advertsing block of data which has such information, including the Temperature as `0x1A5E`. See the picture below:

image

SuGlider:
I'll do a full review of your PR by Wednesday.

libraries/BLE/src/BLEEddystoneTLM.h
```diff
@@ -18,33 +22,34 @@
...
-	void		setTemp(float temp);
-	void		setCount(uint32_t advCount);
-	void		setTime(uint32_t tmil);
+  BLEEddystoneTLM();
+  std::string getData();
+  BLEUUID   getUUID();
+  uint8_t   getVersion();
+  uint16_t  getVolt();
+  uint16_t  getTemp();
+  float     getFloatTemp();
```
SuGlider, line 31: In the Original code there was just
`float	   getTemp();`

Now this has changed to `float getFloatTemp();`
and `float getFloatTemp();`

This is a breaking change... 
I don't think that it is a good idea.

libraries/BLE/src/BLEEddystoneTLM.cpp
```diff
@@ -46,10 +44,14 @@ uint16_t BLEEddystoneTLM::getVolt() {
 	return ENDIAN_CHANGE_U16(m_eddystoneData.volt);
 } // getVolt
 
-float BLEEddystoneTLM::getTemp() {
-	return ENDIAN_CHANGE_U16(m_eddystoneData.temp) / 256.0f;
+uint16_t BLEEddystoneTLM::getTemp() {
+	return ENDIAN_CHANGE_U16(m_eddystoneData.temp);
 } // getTemp
```
SuGlider, line 49: As said, this is a breaking change. I think it should stay the same:
`float BLEEddystoneTLM::getTemp()`

The original code works correctly.

libraries/BLE/src/BLEEddystoneTLM.cpp
```diff
@@ -46,10 +44,14 @@ uint16_t BLEEddystoneTLM::getVolt() {
...
 
+float BLEEddystoneTLM::getFloatTemp() {
+  return EDDYSTONE_TEMP_U16_TO_FLOAT(ENDIAN_CHANGE_U16(m_eddystoneData.temp));
+} // getFloatTemp
+
```
SuGlider, line 54: As said, this is a breaking change. I think it should stay the same:
`float BLEEddystoneTLM::getTemp()`

The original code works correctly.

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
           Serial.printf("TX power %d\n", foundEddyURL.getPower());
           Serial.println("\n");
         }
-        else if (payLoad[11] == 0x20)
+        if (advertisedDevice.getPayloadLength() >= 22 && payLoad[22] == 0x20)
         {
```
SuGlider, line 103: Are you sure that this is right? [11] --> [22]?

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
...
-          Serial.println("\n");
-          Serial.print(foundEddyURL.toString().c_str());
-          Serial.println("\n");
+          Serial.printf("Found an EddystoneTLM beacon! payload length = %d B; minus 22 = %d\n", advertisedDevice.getPayloadLength(), advertisedDevice.getPayloadLength() - 22);
+          BLEEddystoneTLM eddystoneTLM;
+          //std::string TLMframe(payLoad[22], advertisedDevice.getPayloadLength() - 22);
+          //eddystoneTLM.setData(TLMFrame);
+          eddystoneTLM.setData(std::string((char*)payLoad+22, advertisedDevice.getPayloadLength() - 22));
+          Serial.printf("Reported battery voltage: %dmV\n", eddystoneTLM.getVolt());
+          Serial.printf("Reported temperature: %.2f°C (raw data=0x%04X)\n", eddystoneTLM.getFloatTemp(), eddystoneTLM.getTemp());
```
SuGlider, line 110: Here it may need a change back using the MACROs...
Please see the comentaries I made in `BLEEddystoneTLM.h/.cpp`

SuGlider (changes requested):
In general, I'd say that the less we change the API and the original code, the better.
Changing the API will cause a breaking change, which is not desirable.

libraries/BLE/src/BLEEddystoneTLM.h
SuGlider: The suggestion here is to keep the original API `float getTemp();` and use the MACROs for converting temperature types whenever necessary.

libraries/BLE/src/BLEEddystoneTLM.cpp
SuGlider: Suggestion is to eliminate `uint16_t BLEEddystoneTLM::getTemp()` and just keep original `float BLEEddystoneTLM::getTemp()` that converts U16 BigEndian to Float (return value) using the MACRO.

libraries/BLE/src/BLEEddystoneTLM.cpp
SuGlider: `m_eddystoneData.temp` is in Bigendian 8.8 fixed point (16 bits) and `EDDYSTONE_TEMP_U16_TO_FLOAT()` already does all the job.

No need to use `ENDIAN_CHANGE_U16()` here.

Please eliminate this method because it causes a breaking change to the API.

libraries/BLE/src/BLEEddystoneTLM.cpp
SuGlider: This should be the final code:
``` cpp
float BLEEddystoneTLM::getTemp() {
	return EDDYSTONE_TEMP_U16_TO_FLOAT(m_eddystoneData.temp);
} // getTemp
```

libraries/BLE/src/BLEEddystoneTLM.h
SuGlider: This should be the final code:
``` cpp
  BLEEddystoneTLM();
  std::string getData();
  BLEUUID   getUUID();
  uint8_t   getVersion();
  uint16_t  getVolt();
  float  getTemp();
```

libraries/BLE/src/BLEAdvertising.cpp
```diff
@@ -443,6 +443,11 @@ void BLEAdvertisementData::setPartialServices(BLEUUID uuid) {
  * @brief Set the service data (UUID + data)
  * @param [in] uuid The UUID to set with the service data.  Size of UUID will be used.
  * @param [in] data The data to be associated with the service data advert.
+ * Data frame:
+ * | Field  || Len | Type | UUID        | EddyStone Frame |
+ * | Offset || 0   | 1    | 2           | 4               |
+ * | Len    || 1   | 1    | 2           | up to 20        |
+ * | Data   || ??  | ??   | 0xAA | 0xFE | ???             |
```
SuGlider, line 450: This is a generic BLE advertising API.
Does this commentary relate to this?

libraries/BLE/examples/BLE_EddystoneTLM_Beacon/BLE_EddystoneTLM_Beacon.ino
```diff
@@ -48,10 +49,10 @@ void setBeacon()
...
   uint16_t beconUUID = 0xFEAA;
   uint16_t volt = random(2800, 3700); // 3300mV = 3.3V
-  float tempFloat = random(2000, 3100) / 100.0f;
-  Serial.printf("Random temperature is %.2fC\n", tempFloat);
-  int temp = (int)(tempFloat * 256); //(uint16_t)((float)23.00);
-  Serial.printf("Converted to 8.8 format %0X%0X\n", (temp >> 8), (temp & 0xFF));
+  float tempFloat = random(-3000, 3000) / 100.0f;
+  Serial.printf("Random temperature is %.2f°C\n", tempFloat);
+  int temp = EDDYSTONE_TEMP_FLOAT_TO_U16(tempFloat);
+  Serial.printf("Converted to 8.8 format 0x%04X\n", temp);
```
SuGlider, line 55: The original code was working correclty. There is no need to change it.
Please revert the change in lines 51 to 55.

I tested the proposed new code and it doesn't work.

The problem with the new code is that `EDDYSTONE_TEMP_FLOAT_TO_U16()` returns a `big endian uint16_t` and the original code was using `temp` as `little endian` which was turned into `big endian` with the lines: 
``` cpp
  beacon_data[4] = (temp >> 8);           // Beacon temperature
  beacon_data[5] = (temp & 0xFF);           //

```

SuGlider: Please revert to:
``` cpp
float tempFloat = random(2000, 3100) / 100.0f;
  Serial.printf("Random temperature is %.2fC\n", tempFloat);
  int temp = (int)(tempFloat * 256); //(uint16_t)((float)23.00);
  Serial.printf("Converted to 8.8 format %0X%0X\n", (temp >> 8), (temp & 0xFF));
```

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
SuGlider: OK. I checked it with nRF APP and it is correct. The positionn is 22.

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
           Serial.printf("TX power %d\n", foundEddyURL.getPower());
           Serial.println("\n");
         }
-        else if (payLoad[11] == 0x20)
+        if (advertisedDevice.getPayloadLength() >= 22 && payLoad[22] == 0x20)
         {
-          Serial.println("Found an EddystoneTLM beacon!");
-          BLEEddystoneTLM foundEddyURL = BLEEddystoneTLM();
-          std::string eddyContent((char *)&payLoad[11]); // incomplete EddystoneURL struct!
```
SuGlider, line 106: I just changed it there to:
`std::string eddyContent((char *)&payLoad[22]); // incomplete EddystoneURL struct!`

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
...
-
-          eddyContent = "01234567890123";
-
-          for (int idx = 0; idx < 14; idx++)
-          {
-            eddyContent[idx] = payLoad[idx + 11];
```
SuGlider, line 112: I just changed Line 112 to:
`eddyContent[idx] = payLoad[idx + 22];`

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
...
-          }
-
-          foundEddyURL.setData(eddyContent);
-          Serial.printf("Reported battery voltage: %dmV\n", foundEddyURL.getVolt());
-          Serial.printf("Reported temperature from TLM class: %.2fC\n", (double)foundEddyURL.getTemp());
-          int temp = (int)payLoad[16] + (int)(payLoad[15] << 8);
```
SuGlider, line 118: I just changed Line 118 to:
`int16_t temp = payLoad[22+5] + (payLoad[22+4] << 8);`

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
           Serial.printf("TX power %d\n", foundEddyURL.getPower());
           Serial.println("\n");
         }
-        else if (payLoad[11] == 0x20)
```
SuGlider, line 102: I just changed Line 102  to:
`else if (payLoad[22] == 0x20)`

in order to make it work.

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
```diff
@@ -99,30 +99,17 @@ class MyAdvertisedDeviceCallbacks : public BLEAdvertisedDeviceCallbacks
...
-          Serial.printf("Reported temperature from TLM class: %.2fC\n", (double)foundEddyURL.getTemp());
-          int temp = (int)payLoad[16] + (int)(payLoad[15] << 8);
-          float calcTemp = temp / 256.0f;
-          Serial.printf("Reported temperature from data: %.2fC\n", calcTemp);
-          Serial.printf("Reported advertise count: %d\n", foundEddyURL.getCount());
-          Serial.printf("Reported time since last reboot: %ds\n", foundEddyURL.getTime());
-          Serial.println("\n");
-          Serial.print(foundEddyURL.toString().c_str());
-          Serial.println("\n");
+          Serial.printf("Found an EddystoneTLM beacon! payload length = %d B; minus 22 = %d\n", advertisedDevice.getPayloadLength(), advertisedDevice.getPayloadLength() - 22);
```
SuGlider, line 104: I guess that printing the payload length is a left over of debugging, right?

libraries/BLE/examples/BLE_Beacon_Scanner/BLE_Beacon_Scanner.ino
SuGlider: As suggested in `BLEEddystoneTLM.h/.cpp`, `eddystoneTLM.getTemp()` should be kept as in the original API, returning a `float`.
IF necessary, you can use `EDDYSTONE_TEMP_FLOAT_TO_U16()` to get the raw data.

SuGlider:
@PilnyTomas 

A last **VERY** important thing... why the original code looks at TLM data in the position [11] and not in the position [22]?
Well... the truth is that this position may change from Beacon to Beacon and it shall be interpreted based on the BLE Advertising Service Data Block information. 

So the position 22 works for the `BLE_EddystoneTLM_Beacon.ino` current example, but if we changed the Beacon name `oAdvertisementData.setName("TLMBeacon");` to something else, the postion of TLM data will change in the advertising service data block. You can try it and verify it.

Therefore, this postion shall be calculated based on the interpretation the this frame, looking for the filed TYPE 0x16 with LEN = 17 and 0x20 in the 3rd byte of the VALUE.

image

PilnyTomas:
> postion shall be calculate

In that case i can fix this in the other PR. But what to do now when we want it ASAP? Just roll back to original?

SuGlider:
> > postion shall be calculate
> 
> In that case i can fix this in the other PR. But what to do now when we want it ASAP? Just roll back to original?

I've added the necessary commit to make it find the Eddystone data dynamicaly.
I shall work fine now, including for the URL beacon.

SuGlider (approved):
@PilnyTomas @me-no-dev @VojtechBartoska 
I have tested Eddystone TLM and URL.
Everything is working fine.
Ready to merge.

SuGlider:
Added a FIX for issue #7858

SuGlider:
> Added a FIX (commit [a64613d](https://github.com/espressif/arduino-esp32/commit/a64613d3a32bfb9b32811eee09ab688064dadc5a)) for issue #7858

Tested. Working correctly now. Issue #7858 solved.

me-no-dev (approved):

PR #7801, opened 2023-02-05
Adds softAp(String) to make it compatible with ESP8266

### LINKED ISSUE #7780: API compatibility of WiFi.softAP() with ESP8266
### Related area

API compatibility of WiFi.softAP() with ESP8266

### Hardware specification

ESP32 vs ESP8266

I was porting some code from ESP8266 to ESP32.
Encounter the following at compile:

With ESP8266 chip, the following

String myString="name of my AP net";
WiFi.softAP(myString, "my password);

is valid

Under ESP32
it does not compile, as API expect const char * and not String as valid type.

Therefore if you could add this behaviour, for the sake of simplicity and not breaking API behaviour.
WiFi.softAP(myString, "my password);

Thanks.

Add API:
WiFi.softAP(String acceptThis, String andAlsoThis)

Modify source code.

### Additional context

_No response_

SuGlider:
## Description of Change
Adds API compatibility of WiFi.softAP() with ESP8266, by adding `String` as possible overload parameters.

## Tests scenarios
Using ESP32 works correctly. 

## Related links
Fix #7780

### DIFF
--- a/libraries/WiFi/src/WiFiAP.h
+++ b/libraries/WiFi/src/WiFiAP.h
@@ -38,6 +38,10 @@ class WiFiAPClass
 public:
 
     bool softAP(const char* ssid, const char* passphrase = NULL, int channel = 1, int ssid_hidden = 0, int max_connection = 4, bool ftm_responder = false);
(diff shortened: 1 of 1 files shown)


artofit:
Happy to see swift fix

PR #7807, opened 2023-02-06
Add missing function getSocketOption

rtpmsys:
There is a function for setting socket options (setSocketOption) but the corresponding function to read the socket options is missing.
This PR provides the missing function getSocketOption.

## Tests scenarios
Tested this on ESP32 with the following code:

```
WiFiClient client;
int rcvbuf_size = 0;
client.getSocketOption(SOL_SOCKET, SO_RCVBUF, &rcvbuf_size, sizeof(int));
```

The int rcvbuf_size will contain the receive buffer size of the socket.

This is the second part of #7361 (now closed) which was split into PR #7686 (merged) and this one


mrengineer7777:
Oops.  Function name collision with LWIP getsockopt().

```
In file included from /home/runner/Arduino/hardware/espressif/esp32/libraries/WiFi/src/WiFiClient.cpp:22:
/home/runner/Arduino/hardware/espressif/esp32/tools/sdk/esp32c3/include/lwip/lwip/src/include/lwip/sockets.h:635:81: note:   initializing argument 5 of 'int getsockopt(int, int, int, void*, socklen_t*)'
 static inline int getsockopt(int s,int level,int optname,void *opval,socklen_t *optlen)
```

PR #7809, opened 2023-02-07
Added function timerAttachInterruptFlag to timer api

davidk88:
## Description of Change
Added function timerAttachInterruptFlag to timer api, which allows passing interrupt allocation flags. For example ESP_INTR_FLAG_IRAM for low latency application.

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.6 with ESP32

### DIFF
--- a/cores/esp32/esp32-hal-timer.c
+++ b/cores/esp32/esp32-hal-timer.c
@@ -221,11 +221,15 @@ bool IRAM_ATTR timerFnWrapper(void *arg){
     return false;
 }
 
-void timerAttachInterrupt(hw_timer_t *timer, void (*fn)(void), bool edge){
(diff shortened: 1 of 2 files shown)


me-no-dev:
I'm not sure how to act here, because this API will not exist in the next major release (v3.0.0). ESP-IDF Driver API has changed much. You will be able to provide interrupt flags in `timerBegin` though.

davidk88:
If 2.0.6 is the last release before version 3 then ignore this PR.

VojtechBartoska:
Hello @davidk88, thanks for the PR and new function. We will have more releases based on IDF 4.4. Pretty soon, 2.0.7 will be out and possibly we'll add one more bug fix release a bit later.

@P-R-O-C-H-Y Can you please help with review? From my point of view, we can add this PR if it works as expected.

PR #7812, opened 2023-02-08
Reword "ESP-IDF as Component" in readme link

remenyo:
## Description of Change
I think "Arduino as an ESP-IDF component" or just "As ESP-IDF component" instead of  "ESP-IDF as Component" are a more correct ways to name [the link in the readme about Arduino as an IDF component.](https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html)

1. "ESP-IDF as Component" would imply that ESP-IDF is some sort of library for Arduino, which is IMO misleading, because it's true the other way around.
2. It's written as "Arduino as an ESP-IDF component" on the webpage it points to as well.

Also I removed the capitalization from "Component" as I have not found a reason why is it capitalized.

## Related links
[https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html](https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html)


me-no-dev:
Thanks @remenyo :) makes sense

PR #7817, opened 2023-02-08
LILYGO T-Display S3 has 512kb SRAM

Eistee82:


Jason2866:
The SRAM is memory mapped and usable is the actual value.

me-no-dev:
I will close this for now and investigate later if numbers needs adjusting

Randomblock1:
Yep, look at the datasheet: ["512 KB on-chip SRAM"](https://www.espressif.com/sites/default/files/documentation/esp32-s3_datasheet_en.pdf#subsubsection.3.1.2) 
Meanwhile the ESP32-S2 [has 320Kb SRAM](https://www.espressif.com/sites/default/files/documentation/esp32-s2_datasheet_en.pdf#e)
This should definitely be reopened, it seems like a pretty big issue to me.

me-no-dev:
@Randomblock1 yet, when you flash the chip, you do not have access to the full amount of SRAM. 512KB is not the amount that should be put in the board config.

PR #7832, opened 2023-02-10
Avoid starting AP Mode even when the password is too short

### LINKED ISSUE #7831: SoftAP fails to assign SSID/PWD with pasword size 1 to 7 characters and still turns on AP
### Board

ESP32 Devkit

### Device Description

Just the Devkit

### Hardware Configuration

None

### Version

v2.0.6

### IDE Name

Arduino IDE 1.8.15

### Operating System

Windows 11

### Flash frequency

QIO 80MHz

### PSRAM enabled

no

### Upload speed

Tried 115200 and 921600

### Description

When running the WiFiAccessPoint example and changing the SSID with a short password (less than 8 characters) it assigns an unexpected SSID and leaves it as open nerwork (no passord).
The example works fine with passwords with more than 7 characters and when it is NULL (open network).
Otherwise it fails, returning `false` but it still turns on the AP with an ESP32_XXXX SSID.

It seems that the issue is in the WiFiAP.cpp code that should DISABLE AP before returning.
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiAP.cpp#L151-L155

``` cpp
bool WiFiAPClass::softAP(const char* ssid, const char* passphrase, int channel, int ssid_hidden, int max_connection, bool ftm_responder)
{

    if(!WiFi.enableAP(true)) {
        // enable AP failed
        log_e("enable AP first!");
        return false;
    }

    if(!ssid || *ssid == 0) {
        // fail SSID missing
        log_e("SSID missing!");
        return false;
    }

    if(passphrase && (strlen(passphrase) > 0 && strlen(passphrase) < 8)) {
        // fail passphrase too short
        log_e("passphrase too short!");
        return false;
    }
  ```

### Sketch

```cpp
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/examples/WiFiAccessPoint/WiFiAccessPoint.ino
```

### Debug Message

```plain
ets Jul 29 2019 12:21:46

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1344
load:0x40078000,len:13516
load:0x40080400,len:3604
entry 0x400805f0

Configuring access point...
[   121][E][WiFiAP.cpp:153] softAP(): passphrase too short!
AP IP address: 192.168.4.1
Server started
```

### Other Steps to Reproduce

NONE

SuGlider:
## Description of Change
Using Arduino 2.0.6 with the WiFiAccessPoint example, if the password is changed to have between 1 and 7 characters, an error message is sent, but even though the AP starts to work.

It will start with a different SSID and no password, which should not happen given that the `WiFiAPClass::softAP()` returned `false` and failed with an error message.

This PR fixes it by changing the order of the parameter verification to happen before setting the AP mode.

## Tests scenarios
Tested with ESP32

## Related links
Fix #7831

### DIFF
--- a/libraries/WiFi/examples/WiFiAccessPoint/WiFiAccessPoint.ino
+++ b/libraries/WiFi/examples/WiFiAccessPoint/WiFiAccessPoint.ino
@@ -32,7 +32,11 @@ void setup() {
   Serial.println("Configuring access point...");
 
   // You can remove the password parameter if you want the AP to be open.
-  WiFi.softAP(ssid, password);
+  // a valid password must have more than 7 characters
+  if (!WiFi.softAP(ssid, password)) {
+    log_e("Soft AP creation failed.");
+    while(1);
+  }
   IPAddress myIP = WiFi.softAPIP();
   Serial.print("AP IP address: ");
   Serial.println(myIP);
--- a/libraries/WiFi/src/WiFiAP.cpp
+++ b/libraries/WiFi/src/WiFiAP.cpp
@@ -136,12 +136,6 @@ void wifi_softap_config(wifi_config_t *wifi_config, const char * ssid=NULL, cons
 bool WiFiAPClass::softAP(const char* ssid, const char* passphrase, int channel, int ssid_hidden, int max_connection, bool ftm_responder)
 {
 
-    if(!WiFi.enableAP(true)) {
-        // enable AP failed
-        log_e("enable AP first!");
-        return false;
-    }
-
     if(!ssid || *ssid == 0) {
         // fail SSID missing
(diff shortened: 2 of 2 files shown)


mrengineer7777 (approved):
This is exactly how I would fix WiFiAPClass::softAP(). LGTM.

Issue #6172, opened 2022-01-21
Fix inconsistent behavior in serveStatic() of WebServer on ESP32 using LittleFS


technik-gegg:
### Board

ESP32DEV

### Device Description

NodeMCU

### Hardware Configuration

Not relevant

### Version

v2.0.2

### IDE Name

PlatformIO

### Operating System

Windows 10

### Flash frequency

default

### PSRAM enabled

no

### Upload speed

115200

### Description

I'm working on a firmware which is offering a web browser interface for controlling the behavior for both the ESP8266 and ESP32. I've figured inconsistencies in those frameworks when using the *serveStatic()* method of the *WebServer* class. 
The *serveStatic()* method mainly serves two purposes: 
One is to build a facade to abstract complex page names into a simple URI, i.e.:

`serveStatic("/dump", FS, "/www/debug/dump_runtime_vars.html")`

The other is to tell the WebServer in which folder(s) to look for files being requested by the browser unconditionally, i.e:

`serveStatic("/css", FS, "/www/css/")`

In the latter case, any file requested that starts with the URI /css/ is being concatenated to the physical path (/www/css/) and returned to the caller (browser).  

This is the way it works in the ESP8266 framework, regardless whether SPIFFS or LittleFS is being used. In the ESP32 framework this doesn't work the same way if **LittleFS** is being used. One has to explicitly define every possible file which could be requested by the browser, which is not only inconvenient but also requires to modify, recompile and flash the firmware every time new files are being added to the file system.

I've traced down this issue and found out, that the culprit is the constructor of the *[StaticRequestHandler](https://github.com/espressif/arduino-esp32/blob/master/libraries/WebServer/src/detail/RequestHandlersImpl.h)* class (lines 65 ff) utilized by the WebServers *serveStatic()* method. 
The main issue occurs because of the first line this code:

 `    _isFile = fs.exists(path);` <--
`     log_v("StaticRequestHandler: path=%s uri=%s isFile=%d ...`
`      _baseUriLength = _uri.length();`

Though this works well on SPIFFS it doesn't on LittleFS, since LittleFS returns **true** from the *.exists()* method if the given path is a valid directory.
To overcome this issue one has to change the **_isFile =** line to:
`_isFile = fs.exists(path) && !fs.isDirectory(path);`
[...]

PR #7861, opened 2023-02-17
Custom wrapper stream for HTTPUpdate

Samureimer:
The user might want to manipulate or add headers to the OTA binaries. Allowing the user to parse the TCPStream through a custom stream allows just that. In my case I needed to be able to encrypt the binary for security reasons.

To do this I add a header with the IV of the AES encryption to the binary. I then return that overhead of that header through the callback with the new stream wrapper that handles the header parsing and decryption.

The overhead is then subtracted from the total binary length as it should not be written to flash.

I plan on backporting this to idf-release/v4.4 as this is where I actually need it.

### Checklist

-----------
## Description of Change
Added a callback the the HTTPUpdate class to get a custom stream for parsing binaries.

## Tests scenarios
Tested on ESP32-C3.

## Related links
This feature is similar to the one requested here, but is more flexible and simpler in my opinion #6172
Also similar to the feature in the ESP_HTTPS_OTA [Link](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/kconfig.html#config-esp-https-ota-decrypt-cb)

### DIFF
--- a/docs/source/troubleshooting.rst
+++ b/docs/source/troubleshooting.rst
@@ -123,6 +123,18 @@ Sample code to check SDK WPA3 support at compile time:
     #warning "No WPA3 support."
     #endif
 
+Serial not printing
+*******************
+
+I have uploaded firmware to the ESP32 device, but I don't see any response from a Serial.print (HardwareSerial).
+
+Solution
+^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


me-no-dev:
Please provide example :) We will review then

Issue #7558, opened 2022-12-06
WiFi UDP misuses new/delete and crashes


davepl:
### Board

All (eg: HeltecWifiKit32)

### Device Description

All

### Hardware Configuration

All with WiFi

### Version

latest development Release Candidate (RC-X)

### IDE Name

VSCode

### Operating System

FreeRTOS

### Flash frequency

40

### PSRAM enabled

no

### Upload speed

115200

### Description

WiFiUdp::parsePacket makes extensive use of new/delete while checking the return value.  It crashes in low memory conditions because new doesn't return null, it throws an exception, which is not caught.   They could just specify __nothrow but they don't.  Or it could be replaced with malloc/free.

I've fixed it privately, which has corrected this crash for me at least, and would like to fix it in the original.

``` C++
int WiFiUDP::parsePacket(){
  if(rx_buffer)
    return 0;
  struct sockaddr_in si_other;
  int slen = sizeof(si_other) , len;
  char * buf = new char[1460];
  if(!buf){
    return 0;
  }
  if ((len = recvfrom(udp_server, buf, 1460, MSG_DONTWAIT, (struct sockaddr *) &si_other, (socklen_t *)&slen)) == -1){
    delete[] buf;
    if(errno == EWOULDBLOCK){
      return 0;
    }
    log_e("could not receive data: %d", errno);
    return 0;
  }
  remote_ip = IPAddress(si_other.sin_addr.s_addr);
  remote_port = ntohs(si_other.sin_port);
  if (len > 0) {
    rx_buffer = new cbuf(len);
    rx_buffer->write(buf, len);
  }
  delete[] buf;
  return len;
}
```

Here’s an example from the log:

```
  #0  0x40084e09:0x3ffea2e0 in panic_abort at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/panic.c:402
  #1  0x4008f1f1:0x3ffea300 in esp_system_abort at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/esp_system/esp_system.c:128
  #2  0x40094c71:0x3ffea320 in abort at /Users/ficeto/Desktop/ESP32/ESP32S2/esp-idf-public/components/newlib/abort.c:46
  #3  0x400d89af:0x3ffea3a0 in TerminateHandler() at src/main.cpp:471
[...]

PR #7868, opened 2023-02-20
Replace new with malloc for non-class calls

### LINKED ISSUE #7849: Use new(std::nothrow) to prevent crashes
### Board

NA

### Device Description

NA

### Hardware Configuration

Arduino ESP32 Feather

### Version

### IDE Name

PlatformIO

### Operating System

Win11

### Flash frequency

80Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

I noticed in ESP8266 that they have changed their new calls to new(std::nothrow).  This could prevent some crashes in our libraries.

See https://github.com/esp8266/Arduino/commit/692598228412bfbb4561805b88b4370f19f2ef91

cbuf is one of the libraries that was fixed.

### Sketch

```cpp
#include "Arduino.h"
#include "cbuf.h"

void setup() {
    cbuf *p = new cbuf(128);
    p->resize(100);
}

void loop() {
}
```

### Debug Message

```plain
Crash on low memory
```

### Other Steps to Reproduce

_No response_

mrengineer7777:
## Description of Change
This PR resolves potential crashing issues when allocating memory using `new` keyword.  Will be replacing with `malloc` for non-class calls. Targeting 3.0.x branch.

## Tests scenarios
arduino-esp32 core v2.0.7 with Adafruit Feather ESP32.

## Related links

Started by issue #7558.  PR #8065 will fix the issue on 2.x branch.

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -225,30 +225,31 @@ String EspClass::getSketchMD5()
     const esp_partition_t *running = esp_ota_get_running_partition();
     if (!running) {
         log_e("Partition could not be found");
-
         return String();
     }
+
     const size_t bufSize = SPI_FLASH_SEC_SIZE;
-    std::unique_ptr<uint8_t[]> buf(new uint8_t[bufSize]);
-    uint32_t offset = 0;
-    if(!buf.get()) {
+    uint8_t *pb = (uint8_t *)malloc(bufSize);
+    if(!pb) {
         log_e("Not enough memory to allocate buffer");
-
         return String();
     }
+    uint32_t offset = 0;
+
     MD5Builder md5;
     md5.begin();
-    while( lengthLeft > 0) {
+    while(lengthLeft > 0) {
         size_t readBytes = (lengthLeft < bufSize) ? lengthLeft : bufSize;
-        if (!ESP.flashRead(running->address + offset, reinterpret_cast<uint32_t*>(buf.get()), (readBytes + 3) & ~3)) {
+        if (!ESP.flashRead(running->address + offset, (uint32_t *)pb, (readBytes + 3) & ~3)) {
+            free(pb);
             log_e("Could not read buffer from flash");
-
             return String();
         }
-        md5.add(buf.get(), readBytes);
+        md5.add(pb, readBytes);
         lengthLeft -= readBytes;
         offset += readBytes;
     }
+    free(pb);
     md5.calculate();
     result = md5.toString();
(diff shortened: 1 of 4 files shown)


me-no-dev:
Given that all places that need changing are byte buffers, why not switch to malloc and free instead?

mrengineer7777:
> Given that all places that need changing are byte buffers, why not switch to malloc and free instead?

I had the same thought.  This PR will likely switch those lines to malloc but also resolve other potential `new` issues.  This PR will take a lot of research and also a test project to demonstrate that it DOES resolve potential crashes.

mrengineer7777:
I've decided to limit the scope of this PR to just replacing `new` with malloc in cases where a class constructor isn't called.  This will limit conflicts with ongoing refactoring work.  I'm not aware of any crashes related to missing `nothrow`, and from what I understand a class should have a `nothrow` variant to guarantee it won't crash.

mrengineer7777:
Help needed: `std::unique_ptr<uint8_t[]> buf(new uint8_t[bufSize]);` 
https://github.com/espressif/arduino-esp32/blob/0d84018d969309addacbcc3e3782c1fadc95fbc8/cores/esp32/Esp.cpp#L224
Why did this use `unique_ptr`?

Reworked here: https://github.com/espressif/arduino-esp32/pull/7868/commits/e20bcfacee0dc881026ffdbe24a734fd48c2d7ba
Needs review!

me-no-dev:
> Help needed: `std::unique_ptr<uint8_t[]> buf(new uint8_t[bufSize]);`
> 
> https://github.com/espressif/arduino-esp32/blob/0d84018d969309addacbcc3e3782c1fadc95fbc8/cores/esp32/Esp.cpp#L224
> 
> 
> Why did this use `unique_ptr`?

`std::unique_ptr` is a smart pointer that owns and manages another object through a pointer and disposes of that object when the `unique_ptr` goes out of scope. More or less it will automatically free the buffer on function exit

VojtechBartoska:
@me-no-dev What about this PR? Is it still valid?

mrengineer7777:
Rebased PR on latest master.  No conflicts.  I'm open to suggestions on what parts of this PR you guys want to keep.

We should at least include the crash fix from #8065 to WiFiUDP.cpp.  In fact, why didn't the fix on the 2.x branch get added to 3.x?  Is it because we were going to merge this PR instead?

me-no-dev:
> Is it because we were going to merge this PR instead?

Yes :)

PR #7883, opened 2023-02-23
ADD: heltec wsl v3 pinout information

gguerreromx:
Added the required pin out file for Heltec Wireless Stick Lite v3 (Board from November 2022).

Information gathered from the site:
https://heltec.org/project/wireless-stick-lite-v2/ (it shares url with v2)

And Esp32 s3 technical document when required (SPI/I2C pins).

### DIFF
--- a/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
+++ b/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
@@ -3,7 +3,7 @@
 
 #include <stdint.h>
 
-#define Wireless_Stick_Lite true
+#define Wireless_Stick_Lite_V3 true
 #define DISPLAY_HEIGHT 0
 #define DISPLAY_WIDTH  0
 
@@ -19,8 +19,8 @@ static const uint8_t LED_BUILTIN = 35;
 #define BUILTIN_LED  LED_BUILTIN // backward compatibility
 #define LED_BUILTIN LED_BUILTIN
 
-static const uint8_t TX = 44;
-static const uint8_t RX = 43;
+static const uint8_t TX = 43;
+static const uint8_t RX = 44;
 
 static const uint8_t SDA = 2;
 static const uint8_t SCL = 3;
@@ -37,14 +37,14 @@ static const uint8_t A3 = 4;
 static const uint8_t A4 = 5;
 static const uint8_t A5 = 6;
 static const uint8_t A6 = 7;
-static const uint8_t A10 = 12;
-static const uint8_t A12 = 14;
-static const uint8_t A13 = 15;
-static const uint8_t A14 = 16;
-static const uint8_t A15 = 17;
-static const uint8_t A16 = 18;
-static const uint8_t A17 = 19;
-static const uint8_t A18 = 20;
+static const uint8_t A7 = 12;
+static const uint8_t A8 = 14;
+static const uint8_t A9 = 15;
+static const uint8_t A10 = 16;
+static const uint8_t A11 = 17;
+static const uint8_t A12 = 18;
+static const uint8_t A13 = 19;
+static const uint8_t A14 = 20;
 
 static const uint8_t T0 = 1;
 static const uint8_t T1 = 2;
--- a/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
+++ b/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
@@ -8,7 +8,7 @@
 #define DISPLAY_WIDTH  0
 
 #define EXTERNAL_NUM_INTERRUPTS 16
-#define NUM_DIGITAL_PINS        40
+#define NUM_DIGITAL_PINS        20
 #define NUM_ANALOG_INPUTS       15
 
 #define analogInputToDigitalPin(p)  (((p)<20)?(analogChannelToDigitalPin(p)):-1)
--- a/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
+++ b/variants/heltec_wireless_stick_lite_v3/pins_arduino.h
@@ -1,0 +1,60 @@
+#ifndef Pins_Arduino_h
+#define Pins_Arduino_h
+
+#include <stdint.h>
+
+#define Wireless_Stick_Lite true
+#define DISPLAY_HEIGHT 0
+#define DISPLAY_WIDTH  0
+
+#define EXTERNAL_NUM_INTERRUPTS 16
+#define NUM_DIGITAL_PINS        40
+#define NUM_ANALOG_INPUTS       15
+
+#define analogInputToDigitalPin(p)  (((p)<20)?(analogChannelToDigitalPin(p)):-1)
+#define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
+#define digitalPinHasPWM(p)         (p < 34)
+
+static const uint8_t LED_BUILTIN = 35;
+#define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN LED_BUILTIN
+
+static const uint8_t TX = 44;
+static const uint8_t RX = 43;
+
+static const uint8_t SDA = 2;
+static const uint8_t SCL = 3;
+
+static const uint8_t SS    = 34;
+static const uint8_t MOSI  = 35;
+static const uint8_t SCK   = 36;
+static const uint8_t MISO  = 37;
+
+static const uint8_t A0 = 1;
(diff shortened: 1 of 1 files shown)


variants/heltec_wireless_stick_lite_v3/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
+#ifndef Pins_Arduino_h
+#define Pins_Arduino_h
+
+#include <stdint.h>
+
+#define Wireless_Stick_Lite true
```
platypii, line 6: Please use a different macro name so we can distinguish this from V1. I would suggest
`#define Wireless_Stick_Lite_v3 true`

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
...
+static const uint8_t A17 = 19;
+static const uint8_t A18 = 20;
+
+static const uint8_t T0 = 1;
+static const uint8_t T1 = 2;
+static const uint8_t T2 = 3;
+static const uint8_t T3 = 4;
+static const uint8_t T4 = 5;
+static const uint8_t T5 = 6;
+static const uint8_t T6 = 7;
```
platypii, line 55: Should be touch `T1` to `T14` for the ESP32-S3. See page 19:
https://www.espressif.com/sites/default/files/documentation/esp32-s3_datasheet_en.pdf

See also the Heltec Wifi Lora V3 pins:
https://github.com/espressif/arduino-esp32/blob/master/variants/heltec_wifi_lora_32_V3/pins_arduino.h

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
...
+static const uint8_t A5 = 6;
+static const uint8_t A6 = 7;
+static const uint8_t A10 = 12;
+static const uint8_t A12 = 14;
+static const uint8_t A13 = 15;
+static const uint8_t A14 = 16;
+static const uint8_t A15 = 17;
+static const uint8_t A16 = 18;
+static const uint8_t A17 = 19;
+static const uint8_t A18 = 20;
```
platypii, line 47: Should be analog ADC `A1` to `A19` for the ESP32-S3. See page 19:
https://www.espressif.com/sites/default/files/documentation/esp32-s3_datasheet_en.pdf

See also the Heltec Wifi Lora V3 pins:
https://github.com/espressif/arduino-esp32/blob/master/variants/heltec_wifi_lora_32_V3/pins_arduino.h

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
...
+#define analogInputToDigitalPin(p)  (((p)<20)?(analogChannelToDigitalPin(p)):-1)
+#define digitalPinToInterrupt(p)    (((p)<40)?(p):-1)
+#define digitalPinHasPWM(p)         (p < 34)
+
+static const uint8_t LED_BUILTIN = 35;
+#define BUILTIN_LED  LED_BUILTIN // backward compatibility
+#define LED_BUILTIN LED_BUILTIN
+
+static const uint8_t TX = 44;
+static const uint8_t RX = 43;
```
platypii, line 23: These appear to be switched. See pinout diagram:
https://resource.heltec.cn/download/Wireless_Stick_Lite_V3/HTIT-WSL_V3.png

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <stdint.h>
+
+#define Wireless_Stick_Lite true
+#define DISPLAY_HEIGHT 0
+#define DISPLAY_WIDTH  0
+
+#define EXTERNAL_NUM_INTERRUPTS 16
+#define NUM_DIGITAL_PINS        40
+#define NUM_ANALOG_INPUTS       15
```
platypii, line 12: `NUM_ANALOG_INPUTS` should be 20 for ESP32-S3

platypii (changes requested):

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: To my understanding, four main variants of the esp32/lora heltec boards exist, with I believe, three revisions each. While I understand the reasoning behind placing the version in the definition, this attribute is mainly used with their library (https://github.com/HelTecAutomation/Heltec_ESP32).

This definition allows the library to know what features are available in the board (big screen, small screen, no screen), allow external voltage, etc. (See code: https://github.com/HelTecAutomation/Heltec_ESP32/blob/master/src/heltec.cpp).

Since there is no "board hardware difference" between versions 1,2,3, placing this definition allows the library to work as expected and without depending on that library to update (since it hasn't updated for some revisions).

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: I didn't placed them since they do not exist as an available pin in the board itself (https://resource.heltec.cn/download/Wireless_Stick_Lite_V3/HTIT-WSL_V3.png).

They can be placed if required?

Should I place them?

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
platypii: The `Wireless_Stick_Lite` and the `Wireless_Stick_Lite_V3` are extremely different board hardware. The only thing similar is the form factor. The V3 uses a new ESP32-S3 cpu, different lora module, totally different pin numbers, and even different numbers of pins.

I understand you're trying to make this work with the heltec library, but please don't use a duplicate board macro here just to make an external library work. I have a [PR](https://github.com/HelTecAutomation/Heltec_ESP32/pull/100) open in that repo to support newer boards. You can also fork that repo and use your own version of the heltec library. Until they merge my PR there, I've been pointing PlatformIO at my own fork:
```
lib_deps = https://github.com/platypii/Heltec_ESP32.git
```

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
platypii: That's a totally fair point. I added all of them because that's how it was [here](https://github.com/Heltec-Aaron-Lee/WiFi_Kit_series/blob/master/esp32/variants/WIFI_LoRa_32_V3/pins_arduino.h). I kind of agree with your point that we should only include pins which are actually available. But all the other heltec v3 pin files include all of them. Either way seems fine.

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
platypii: Same comment as above about which pins to include. Either way works.

However, why do your GPIO pins change from being A+1 to A+2 starting at A10? That doesn't seem right?

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: I do agree in your point since editing the repo is the actual solution and not doing it half way as I was attempting.

I just saw you PR and I believe it is a great idea to actually separate the features from the boards. Solves current issues and make it easier to add new boards in the future. 

I hope it is accepted promptly.

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: I am hardly a tinkerer when it comes to electronics and for that reason I find it easier to think about what i can create/modify if there is a correlation between what I see in hardware and in code. 

That is why I believe showing only available pins is the right way to go.

If there is no issue I would like to be kept as is.

In the other it didn't make much sense my logic of separating them by ADC#_ and Ax and A1x so I did fix the numeration to be continuous.

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: Fixed, it didn't make sense (see above)

variants/heltec_wireless_stick_lite_v3/pins_arduino.h
gguerreromx: Fixed

P-R-O-C-H-Y (approved):

PR #7911, opened 2023-03-01
CI HW Testing Update

P-R-O-C-H-Y:
## Description of Change
Updated upload-artifact action to v3.
Edited build_dir path where the sketches are built in to + paths for upload-artifact actions
Added `if-no-files-found: error` to upload-artifact action


.github/scripts/sketch_utils.sh
```diff
@@ -121,16 +121,16 @@ function build_sketch(){ # build_sketch <ide_path> <user_path> <path-to-ino> [ex
     if [ -n "$ARDUINO_BUILD_DIR" ]; then
         build_dir="$ARDUINO_BUILD_DIR"
     elif [ $len -eq 1 ]; then
-        # build_dir="$sketchdir/build"
-        build_dir="$HOME/.arduino/build.tmp"
+        build_dir="$sketchdir/build"
+        # build_dir="$HOME/.arduino/build.tmp"
```
me-no-dev, line 125: different solution is needed here. It really is not nice to run CI locally and generate all those files in your  local fork. Maybe generate the artifacts elsewhere?

P-R-O-C-H-Y: OK will try to make it different way :)

P-R-O-C-H-Y: Is it ok now? Should be done the way you proposed to.

me-no-dev: yup. that is better :)

me-no-dev (approved):

mrengineer7777:
Does this also apply to branch 3.0.0?

P-R-O-C-H-Y:
> Does this also apply to branch 3.0.0?

@mrengineer7777 yes, as this PR is fixing the CI HW testing. But for 3.0 there might be a some more changes for the CI.

PR #7913, opened 2023-03-01
Fix the F_CPU frequency definition for the ESP32-S3 in esp32-hal.h

### LINKED ISSUE #8005: esp32-s3 cannot find F_CPU for FastLed
### Board

ESP32-S3

### Device Description

 esp32-s3 cannot find F_CPU for FastLed

### Hardware Configuration

NO

### Version

v2.0.7

### IDE Name

ESP-IDF

### Operating System

Windows10

### Flash frequency

40Mhz

### PSRAM enabled

yes

### Upload speed

921600

### Description

 esp32-s3 cannot find F_CPU for FastLed

### Sketch

```cpp
#ifndef HAL_ESP32_HAL_H_
#define HAL_ESP32_HAL_H_

#include <stdint.h>
#include <stdbool.h>
#include <stdio.h>
#include <stdlib.h>
#include <stdarg.h>
#include <inttypes.h>
#include <string.h>
#include <math.h>
#include "sdkconfig.h"
#include "esp_system.h"
#include "esp_sleep.h"
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "freertos/queue.h"
#include "freertos/semphr.h"
#include "freertos/event_groups.h"

#ifdef __cplusplus
extern "C" {
#endif

#ifndef F_CPU
#if CONFIG_IDF_TARGET_ESP32 // ESP32/PICO-D4
#define F_CPU (CONFIG_ESP32_DEFAULT_CPU_FREQ_MHZ * 1000000U)
#elif CONFIG_IDF_TARGET_ESP32S2
#define F_CPU (CONFIG_ESP32S2_DEFAULT_CPU_FREQ_MHZ * 1000000U)

//not esp32-s3
```

### Debug Message

```plain
esp32-s3 cannot find F_CPU for FastLed
```

### Other Steps to Reproduce

_No response_

devrim-oguz:
## Description of Change
Hello, I was using the FastLED library and it was complaining about F_CPU not being defined. So, I just noticed that it is not defined for the ESP32-S3 module. So I made this change in the header file and it compiled. Therefore I wanted to propose this change to the HAL library to improve compatibility. Thank you for your time.

## Tests scenarios
I have tested this with the ESP32-S3-WROOM-1-N16R2 module for the FastLED library. Haven't done any extensive testing since it is a trivial change (I hope).

#Fix #8005


mrengineer7777:
Interesting.  LGTM.  I wonder if C3 and other variants also need an entry?

devrim-oguz:
> Interesting. LGTM. I wonder if C3 and other variants also need an entry?

They probably would need to. And every single board that comes after. This is not a good way of doing it tbh, but that's how its defined in the sdkconfig. The interesting part is that this error does not occur in PlatformIO, probably because it gets defined in somewhere else and this part is skipped. But it does so in esp-idf when arduino used as a component. So this is how I solved it for my own application. It would be better if the other boards are added to this, but I didn't look much about how many other variants there are.

SuGlider:
> > Interesting. LGTM. I wonder if C3 and other variants also need an entry?
> 
> They probably would need to. And every single board that comes after. This is not a good way of doing it tbh, but that's how its defined in the sdkconfig. The interesting thing is that this error does not occur in PlatformIO, probably because it gets defined in somewhere else and this part is skipped. But it does so in esp-idf when arduino used as a component. So this is how I solved it for my own application. It would be better if the other boards are added to this, but I didn't look much about how many other variants there are.

@mrengineer7777 @devrim-oguz 

I just modified the PR to be generic, based on the sdkconfig definition of each SoC.
It shall now work for the ESP32, ESP32S2, ESP32C3 and ESP32S3.

@devrim-oguz - please test and confirm that it works for the FastLED library with the S3.

SuGlider:
> > > Interesting. LGTM. I wonder if C3 and other variants also need an entry?
> > 
> > 
> > They probably would need to. And every single board that comes after. This is not a good way of doing it tbh, but that's how its defined in the sdkconfig. The interesting thing is that this error does not occur in PlatformIO, probably because it gets defined in somewhere else and this part is skipped. But it does so in esp-idf when arduino used as a component. So this is how I solved it for my own application. It would be better if the other boards are added to this, but I didn't look much about how many other variants there are.
> 
> @mrengineer7777 @devrim-oguz
> 
> I just modified the PR to be generic, based on the sdkconfig definition of each SoC. It shall now work for the ESP32, ESP32S2, ESP32C3 and ESP32S3.
> 
> @devrim-oguz - please test and confirm that it works for the FastLED library with the S3.

OK... Actually the fix I did works for the IDF 5.1 and Arduino 3.0.0.
For Arduino 2.0.8, it shall follow the previous model.
Let me create 2 PRs, one for each Arduino Core version.

mrengineer7777 (approved):
LGTM

PR #7926, opened 2023-03-04
Use enum for C++ to avoid macro pitfalls in esp32_hal_uart.h

dok-net:
C prepropressor macros are generally considered deprecated in C++, in this particular case, their use for the serial configuration breaks the enum from the EspSoftwareSerial library, release 8.0.0 onward.

This merge request provides the previous macro defines for C language builds, but replaces them functionally identical with a C++ enumeration for C++ builds, which is the default Arduino language anyway.

The ESP8266 Arduino Core uses a C++ enum already.


SuGlider (approved):
Thanks @dok-net for the PR!

PR #7932, opened 2023-03-05
RainMaker - Improve sysProvEvent function

FernandoGarcia:
Improve sysProvEvent function to prevent [this](https://github.com/espressif/esp-rainmaker/issues/247#issuecomment-1450208851) type of error.


libraries/RainMaker/examples/RMakerCustom/RMakerCustom.ino
```diff
@@ -26,24 +26,67 @@ static Device my_device("Dimmer", "custom.device.dimmer", &gpio_dimmer);
...
-#if CONFIG_IDF_TARGET_ESP32S2
-            Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on SoftAP\n", service_name, pop);
-            printQR(service_name, pop, "softap");
-#else
-            Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on BLE\n", service_name, pop);
-            printQR(service_name, pop, "ble");
-#endif
-            break;
-        case ARDUINO_EVENT_PROV_INIT:
-            wifi_prov_mgr_disable_auto_stop(10000);
```
sanketwadekar, line 40: This code was added to handle provisioning failures when the log level is set to "Debug". This shouldn't be removed.

libraries/RainMaker/examples/RMakerCustom/RMakerCustom.ino
```diff
@@ -26,24 +26,67 @@ static Device my_device("Dimmer", "custom.device.dimmer", &gpio_dimmer);
...
-            break;
-        case ARDUINO_EVENT_PROV_CRED_SUCCESS:
-            wifi_prov_mgr_stop_provisioning();
-            break;
-        default:;
-    }
+  switch (sys_event->event_id)
+  {
+    case ARDUINO_EVENT_WIFI_STA_GOT_IP:
+      {
```
sanketwadekar, line 32: Please look at the code format in the example and format it accordingly.

sanketwadekar: Update other examples as well.

sanketwadekar (changes requested):
Uploading actual logs while provisioning will help.

libraries/RainMaker/examples/RMakerCustom/RMakerCustom.ino
```diff
@@ -26,24 +26,67 @@ static Device my_device("Dimmer", "custom.device.dimmer", &gpio_dimmer);
...
-#else
-            Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on BLE\n", service_name, pop);
-            printQR(service_name, pop, "ble");
-#endif
-            break;
-        case ARDUINO_EVENT_PROV_INIT:
-            wifi_prov_mgr_disable_auto_stop(10000);
-            break;
-        case ARDUINO_EVENT_PROV_CRED_SUCCESS:
-            wifi_prov_mgr_stop_provisioning();
```
sanketwadekar, line 43: This shouldn't be removed

libraries/RainMaker/examples/RMakerCustom/RMakerCustom.ino
FernandoGarcia: > Please look at the code format in the example and format it accordingly.

Hi!

Could you add a lint with clang-format or similar to Github action to auto format according to Espressif style on PR?

Best regards.

FernandoGarcia:
Hi @VojtechBartoska !

Which answer are you waiting for?

Best regards.

sanketwadekar:
This PR does provide more useful information to the user but does not help in solving the WiFi connection failure issue addressed [here](https://github.com/espressif/esp-rainmaker/issues/247#issuecomment-1454853581). Solving that issue would require changes in the Arduino WiFi library.

The WiFi disconnection error code 202 (AUTH_FAIL) is considered a non-reconnectable reason and only one reconnection attempt is made. Adding the reconnection code in the rainmaker sketch won't solve the issue.
One solution to this would be to try reconnection for all disconnection reasons.
Other would be to add this error code 202 to the list of reconnectable reasons. This still leaves us with reconnection issues with error codes not on the list.
@me-no-dev @VojtechBartoska @SuGlider @shahpiyushv what do you think about this?

me-no-dev:
@sanketwadekar AUTH_FAIL means that PASS was wrong. No? If so, then how is that reconnectable?

sanketwadekar:
> @sanketwadekar AUTH_FAIL means that PASS was wrong. No? If so, then how is that reconnectable?



> @sanketwadekar AUTH_FAIL means that PASS was wrong. No? If so, then how is that reconnectable?

You're right, it should not be reconnectable. However, @FernandoGarcia reported that the device did not connect to WiFi after a reboot. See comment [here](https://github.com/espressif/esp-rainmaker/issues/247#issuecomment-1450208851). I am not sure what could be the reason behind it.

shahpiyushv:
> @sanketwadekar if the ESP returns AUTH_FAIL and then later on connect with the same credentials, then the issue is somewhere else. If we patch this, it would fix the issue for @FernandoGarcia but it will break the general functionality for everyone else. I would rather have this resolved upstream in ESP-IDF and have it working correctly.

@me-no-dev , AUTH_FAIL is sometimes returned even when the signal strength is poor. The esp-idf examples' common component does not have any check for disconnect reason as you can see [here](https://github.com/espressif/esp-idf/blob/master/examples/common_components/protocol_examples_common/wifi_connect.c#L63)

me-no-dev:
> > It isn't the case because I was pretty close from router and my board has external antenna.
> 
> @FernandoGarcia What I mean here is that wrong credentials may not necessarily be the reason for AUTH_FAIL. It could be received in other cases too.

But you will get it when credentials are wrong, so that would cause pointless reconnect on loop.

me-no-dev:
how about we address AUTH_FAIL issue separately. Seems unconnected to this PR. BTW @FernandoGarcia please fix the conflicts.

AUTH_FAIL: I propose a method to be added to WiFiSTA to control wether reconnects will be attempted on AUTH_FAIL. This way it can help the folks with that issue, while not causing pointless reconnects to the rest of the users.

me-no-dev:
@FernandoGarcia please fix the build errors so we can merge it :)

me-no-dev:
@shahpiyushv @sanketwadekar I am looking at our code and do not see a reason why WiFi would not reconnect on AUTH_FAIL

PR #7941, opened 2023-03-08
Completely inline the helper pure abstract __FlashStringHelper class

### LINKED ISSUE #7940: FPSTR not portable between ESP8266 and ESP32
### Board

ESP32 Dev Module

### Device Description

DevKitC

### Hardware Configuration

Irrelevant.

### Version

### IDE Name

Arduino IDE

### Operating System

Windows 10

### Flash frequency

irrelevant

### PSRAM enabled

yes

### Upload speed

115200

### Description

After all these years, the following MCVE still does not compile without changes on both ESP8266 Core master and ESP32 Core master.
Shouldn't `FPSTR` correctly be defined as a noop on ESP32 instead of "StringHelper"?

### Sketch

```cpp
#if defined(ESP8266)
#include <ESP8266mDNS.h>
#elif defined(ESP32)
#include <ESPmDNS.h>
#endif

static const char MDNSNAME[] PROGMEM = "esp"; // your device's MDNS host name

void setup()
{
    MDNS.begin(FPSTR(MDNSNAME));
}

void loop()
{
}
```

### Debug Message

```plain
Compiling on ESP8266 works, on ESP32 fails:


fpstr_issue.ino: In function void setup()
 
fpstr_issue.ino: 11:31: error: no matching function for call to 'MDNSResponder::begin(const __FlashStringHelper*)
   MDNS.begin(FPSTR(MDNSNAME))
 
fpstr_issue.ino:4: In file included from
ESPmDNS.h:57: note  candidate  bool MDNSResponder  begin(const char*)
   bool begin(const char* hostName)
   ^~~~~
ESPmDNS.h:57: note    no known conversion for argument 1 from const __FlashStringHelper* to const char*
Error compiling project sources
Build failed for project 'fpstr_issue'
```
```


### Other Steps to Reproduce

_No response_

### PR DESCRIPTION
**Pull request by dok-net**
On ESP32, it's not needed - all the files using it are different from their ESP8266 counterparts anyway.

Fixes #7940

### CODE REVIEW

**Comment by mrengineer7777** [comment]
Breaking change! Affects ESPAsyncWebServer and possibly other libraries.

**Comment by dok-net** [comment]
@mrengineer7777 Thanks for pointing this out. Fixed.

**Comment by me-no-dev** [comment]
Why remove code instead of adding what is incompatible with 8266? Now you are risking causing harm to other libraries that might be using those functions.

**Comment by dok-net** [comment]
> Why remove code instead of adding what is incompatible with 8266? Now you are risking causing harm to other libraries that might be using those functions.

Remove? What exactly are you seeing that I have missed? There should be no removals, everything is just inlined now because most of it can be / has to be no-ops.

**Comment by me-no-dev** [comment]
> > Why remove code instead of adding what is incompatible with 8266? Now you are risking causing harm to other libraries that might be using those functions.
> 
> Remove? What exactly are you seeing that I have missed? There should be no removals, everything is just inlined now because most of it can be / has to be no-ops.

Ah yes... my bad :)

**Comment by brentru** [comment]
@P-R-O-C-H-Y @me-no-dev Thank you very much for adding a check for the external libraries. This will be incredibly useful down the line!

However, FYI (unless I am reading into this incorrectly), there are 10 errors and 1 warning within the external library check on this **breaking change** (mentioned above): https://github.com/espressif/arduino-esp32/actions/runs/4627758830 yet the CI shows it passes. Is this an issue with the CI tooling?

**Comment by P-R-O-C-H-Y** [comment]
> @P-R-O-C-H-Y @me-no-dev Thank you very much for adding a check for the external libraries. This will be incredibly useful down the line!
> 
> However, FYI (unless I am reading into this incorrectly), there are 10 errors and 1 warning within the external library check on this **breaking change** (mentioned above): https://github.com/espressif/arduino-esp32/actions/runs/4627758830 yet the CI shows it passes. Is this an issue with the CI tooling?

Hi @brentru,

As the CI job is universal for PRs and Schedule runs, I did not think about this point of view that it can be confusing, that the CI passed successfully, when there were errors. When the External Library Test is successful, it means that the test run finishes without any complications, like wrong inputs (json with libs and sketches) or issues with CI actions. It needs to be successful as the compilation error is a result too. But I can think of it a bit and maybe do some changes.

As you can see there were errors in the run you mentioned, doesn't have to mean, that the PR breaks anything, it may even fix some libraries or at least fix some warnings. Where to look first, is the [result table printed to comment by the GH action](https://github.com/espressif/arduino-esp32/pull/7941#issuecomment-1498848139). You have to do manual check for now, to see if the PR breaks some library. As you can see, in this PR the compilation BEFORE is the same as AFTER (BEFORE->AFTER PR). You can check [docs](https://espressif-docs.readthedocs-hosted.com/projects/arduino-esp32/en/latest/external_libraries_test.html) to read more information about the External Library Test.

Hope this help, but feel free to ask there or maybe in the discussions [CI Testing for External Libraries](https://github.com/espressif/arduino-esp32/discussions/8041). Even if you know some libraries that are highly used, feel free to add a comment there and we can add them later on :)

**Comment by bblanchon** [comment]
This change breaks the following use case:

```c++
const __FlashStringHelper* str = F("lorem ipsum");
```

mrengineer7777:
@bblanchon We are aware this PR has issues.  Does [this fix](https://github.com/espressif/arduino-esp32/pull/8147) help?

bblanchon:
Thank you for the quick reply, @mrengineer7777.
Yes, #8147 should fix this issue.

PR #4577, opened 2020-11-25
[1.0.5] Update Heltec ESP32 series boards definition

Heltec-Aaron-Lee:
- Added LoRaWAN features option for Heltec ESP32 + LoRa series boards;
- Added a new board Heltec Wireless Stick Lite.

PR #7942, opened 2023-03-08
Update boards.txt file to fix some of the broken board.build entries that should be upper case

FrightRisk:
Fix some of the board.build entries that are lower case to make them upper case. In particular, the Heltec boards were broken in the Arduino IDE with version 2.0.7  or the ESP32 library when they were changed from HELTEC_WIFI_32 to heltec_wifi_32, etc.

-----------
## Description of Change
Changes many, but not all of the incorrectly lowecase board.build names to be uppercase. When they were changed to lowercase, they broke conditional compiles based on board type. The standard since these are constants, is to be upper case.

## Tests scenarios
Arduino IDE 1.8.09 and 2.0.4. Tested with the Heltec wifi board and wifi board v3

## Partially fixes issue

Fixes https://github.com/espressif/arduino-esp32/issues/7943

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -3831,7 +3831,7 @@ ttgo-t-oi-plus.build.target=esp
 ttgo-t-oi-plus.build.mcu=esp32c3
 ttgo-t-oi-plus.build.core=esp32
 ttgo-t-oi-plus.build.variant=ttgo-t-oi-plus
-ttgo-t-oi-plus.build.board=TTGO-T-OI-PLUS_DEV
+ttgo-t-oi-plus.build.board=TTGO_T_OI_PLUS_DEV
 ttgo-t-oi-plus.build.bootloader_addr=0x0
 
 ttgo-t-oi-plus.build.cdc_on_boot=0
@@ -19251,7 +19251,7 @@ lionbit.build.target=esp32
 lionbit.build.mcu=esp32
 lionbit.build.core=esp32
 lionbit.build.variant=lionbit
-lionbit.build.board=Lion:BIT_DEV_BOARD
+lionbit.build.board=LIONBIT_DEV_BOARD
 
 lionbit.build.f_cpu=240000000L
 lionbit.build.flash_size=4MB


platypii:
If you're going to fix some of them, seems like you might just fix them all? Here are the remaining ones:
```
grep 'build.board=[a-z]' boards.txt 
```

```
esp32vn-iot-uno.build.board=esp32vn_iot_uno
heltec_wireless_stick.build.board=heltec_wireless_stick
sensesiot_weizen.build.board=sensesiot_weizen
OpenKB.build.board=openkb
uPesy_wrover.build.board=uPesy_WROVER
uPesy_wroom.build.board=uPesy_WROOM
atmegazero_esp32s2.build.board=atmegazero_esp32s2
connaxio_espoir.build.board=connaxio_espoir
unphone8.build.board=unphone8
unphone9.build.board=unphone9
roboheart_hercules.build.board=roboheart_hercules
```

mrengineer7777:
@FrightRisk I couldn't find a commit that specifically made those entries lower case.  It looks like they were originally lower case when added.  I agree constants should be fully uppercase.  Most of the `.build.board =` entries are uppercase.

boards.txt
```diff
@@ -13126,7 +13126,7 @@ CoreESP32.build.target=esp32
 CoreESP32.build.mcu=esp32
 CoreESP32.build.core=esp32
 CoreESP32.build.variant=Microduino-esp32
-CoreESP32.build.board=CoreESP32
+CoreESP32.build.board=CORESP32
```
atanisoft, line 13129: an E got dropped on this line it seems.

mrengineer7777:
> #4577 made the heltec ones lowercase

Now that's digging deep.  Nice catch!

FrightRisk:
> If you're going to fix some of them, seems like you might just fix them all? Here are the remaining ones:
> 
> ```
> grep 'build.board=[a-z]' boards.txt 
> ```
> 
> ```
> esp32vn-iot-uno.build.board=esp32vn_iot_uno
> heltec_wireless_stick.build.board=heltec_wireless_stick
> sensesiot_weizen.build.board=sensesiot_weizen
> OpenKB.build.board=openkb
> uPesy_wrover.build.board=uPesy_WROVER
> uPesy_wroom.build.board=uPesy_WROOM
> atmegazero_esp32s2.build.board=atmegazero_esp32s2
> connaxio_espoir.build.board=connaxio_espoir
> unphone8.build.board=unphone8
> unphone9.build.board=unphone9
> roboheart_hercules.build.board=roboheart_hercules
> ```

Happy to do the rest. I wasn't sure about a couple of them, though I thought they should all be in caps, you guys confirmed it. I can update the PR Thursday sometime. It is Wednesday night for me now.

boards.txt
FrightRisk: Got it

FrightRisk:
Still unclear about several other lower case boards like a LoRa and "Pocket32", the grep doesn't seem to show them, but they are there. Should I change every build.board entry to upper case?

atanisoft:
> Should I change every build.board entry to upper case?

I would if for no other reason than to be consistent across the board on what the `#define` will end up as.

platypii:
> the grep doesn't seem to show them

Good point, here's a better regex. It will find boards with any lowercase or `-` characters. Dash should probably be replaced by underscore too since it's not a legal envvar.

```
grep 'build.board=\w*[a-z-]' boards.txt
```

atanisoft:
> I see "variants" and left those alone though some have dashes and some have underscores

The `board.variant` entry needs to be an exact match for the board specific directory listed in https://github.com/espressif/arduino-esp32/tree/master/variants.

FrightRisk:
Ok, I think I am good then. All the build boards are uppercase and all use underscores. I also changed the Lion:Bit board. It is referenced elsewhere as "lionbit" and I don't think there should be colons in constants, so made it LIONBIT_DEV_BOARD. I believe it is read for review.

VojtechBartoska:
@P-R-O-C-H-Y Please review this PR.

me-no-dev:
Wouldn't changing ALL boards now cause other conditional statements to fail? I know for a fact that M5 have some conditions that depends on the correct name as-is

FrightRisk:
I guess that decision rests with someone above me, but the standard is that these should be all caps and that is what I would vote for. You really can't have a hodge podge mess. I agree that it might break something else if people were used to a constant listed incorrectly, but is a very simple fix to just change to all caps in any existing code. M5 can issue an update.

P-R-O-C-H-Y:
@Tinyu-Zhao Can you please take a look if this PR will cause issues for your boards? Thanks

Tinyu-Zhao:
> @Tinyu-Zhao Can you please take a look if this PR will cause issues for your boards? Thanks

I will check it，Thanks.

tobozo:
hi,

please don't do this to all boards.

as @me-no-dev mentioned, this will break all third party libraries that rely on board names for device detection at compile time, and there are many, I'm only listing here a few ones from memory but I know there are more:

- LovyanGFX
- M5Unified
- M5Stack
- M5Core2
- ESP32-Chimera-Core
- ESP32-targz
- ESP32-USB-Soft-Host

And "the fix is to add an `if defined`" looks as ugly as this:

image

Please consider the tradeoff of being semantically correct vs the workload and the bugs this will generate downstream.

atanisoft:
@tobozo There will be breakages regardless of what is done or not done unfortunately. The mixed case names should never have been included in the first place.

I agree the workaround is a bit ugly but perhaps it could be mitigated by adding a board pins variant header that defines the old mixed case version of the define?

tobozo:
@atanisoft unless I missed a document saying these are the standards, `The standard since these are constants, is to be upper case.` sounds more like a convention to me (e.g. linux kernel doesn't respect that).

If it is only a convention, then is it required?

I'm just weighing the tech debt between following that convention and only fixing the heltect board, and I don't understand why an upper or mixed-case macro is important at this stage or even in the future, please enlighten me.

>  mitigated by adding a board pins variant header that defines the old mixed case version of the define

Although easier to manage that's still a tech debt: trading a global compile flag with a scoped will only affect multidevice libraries/projects who do not explicitely include that header.

atanisoft:
> trading a global compile flag with a scoped will only affect multidevice libraries/projects who do not explicitely include that header.

If they include Arduino.h it will be automatically included (I haven't seen any library that doesn't include this header).

> If it is only a convention, then is it required?

Definitely more of a convention than standard, but having a mixed set of pre-compiler constants lacks consistency and style. Is it required, that is up for debate.

tobozo:
Can we (library maintainers, board vendors, end users) get a vote before such a tech debt is deferred on us or will it just happen whatever we discuss or debate?

Diako22:
Я вас очень прошу делайте так чтобы я ушол на покой😔

VojtechBartoska:
Hi all,

we have discussed this internally and also on [last Community Meeting](https://github.com/espressif/arduino-esp32/discussions/8062) and I'm sharing here the approach we come up with.

1. In this PR fix just HELTEC_WIFI_32 board definition which is causing the issues
2. Leave other boards as they are now to not brake them on other places (libraries and so on, mentioned above by @tobozo)
3. Set up a rule to use only Upper case naming from now on for new additions

Please let us know if this make sense for you, we're looking forward for your feedback.

tobozo:
that's good news, thank you for taking the time to discuss this :+1:

mrengineer7777:
>     1. In this PR fix just HELTEC_WIFI_32 board definition which is causing the issues
> 
>     2. Leave other boards as they are now to not brake them on other places (libraries and so on, mentioned above by @tobozo)
> 
>     3. Set up a rule to use only Upper case naming from now on for new additions
> 
> 
> Please let us know if this make sense for you, we're looking forward for your feedback. cc @atanisoft, @mrengineer7777, @FrightRisk, @brentru

@VojtechBartoska I agree 100% with this approach.  Who is setting up the rule to enforce Upper case naming?

VojtechBartoska:
> > ```
> > 1. In this PR fix just HELTEC_WIFI_32 board definition which is causing the issues
> > 
> > 2. Leave other boards as they are now to not brake them on other places (libraries and so on, mentioned above by @tobozo)
> > 
> > 3. Set up a rule to use only Upper case naming from now on for new additions
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Please let us know if this make sense for you, we're looking forward for your feedback. cc @atanisoft, @mrengineer7777, @FrightRisk, @brentru
> 
> @VojtechBartoska I agree 100% with this approach. Who is setting up the rule to enforce Upper case naming?

We can include it into Boards test script to check if the PR meet all the requirements.
https://github.com/espressif/arduino-esp32/blob/master/.github/workflows/boards.yml

VojtechBartoska:
@P-R-O-C-H-Y up to my last comment, will you be able to add one more check to boards.yml to validate if the new board addition use Capital letters?

After this modification, we can close this PR.

me-no-dev:
We can close this now as the relevant features have already been merged :)

PR #7978, opened 2023-03-20
Avoid global defines for default ETH configs

antlanc7:
I just moved the default values directly into the prototype, in order to delete the defines that are visible to the whole user project.

### DIFF
--- a/libraries/Ethernet/src/ETH.h
+++ b/libraries/Ethernet/src/ETH.h
@@ -25,30 +25,6 @@
 #include "esp_system.h"
 #include "esp_eth.h"
 
-#ifndef ETH_PHY_ADDR
-#define ETH_PHY_ADDR 0
-#endif
-
-#ifndef ETH_PHY_TYPE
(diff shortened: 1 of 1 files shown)


mrengineer7777:
This appears to reduce flexibility.  What issue does this solve?

mrengineer7777:
The whole point of those defines is so you can define them previously (e.g. in sdkconfig.h) and then they will be automatically included.  Put your defines in a .h file and include it before including ETH.h.

mrengineer7777:
Or use your custom defines (use different names) directly in the function call:
```
if(!ETH.begin(EC_PHY_ADDR, EC_PHY_RST, EC_PHY_MDC, EC_PHY_MDIO, ETH_PHY_KSZ8081, ETH_CLOCK_GPIO0_IN, false)) {
        log_e("ETH failed to start");
    }
```

antlanc7:
Ok so I'm supposed to use the same names in a previously included header file in order to call ETH.begin with redefined default values. Thank you for the clarification

PR #8006, opened 2023-03-29
Add support for large uploads to HTTPClient

me-no-dev:
Partial duplicate of https://github.com/espressif/arduino-esp32/pull/7638


ekapujiw2002 (commented):
Is without yield() that while loop will not make esp soft reset?

me-no-dev:
> Is without yield() that while loop will not make esp soft reset?

Will not, because write itself will yield internally.

PR #8010, opened 2023-03-30
Kernel panic when delete BLEClient object with descriptor characteristic

chesterkwan:
Fix kernel panic issue when trying to release BLEClient when the target device have a ble descriptor.

## Description of Change
When I try to release BLEClient to free memory, sometimes it causes a kernel panic. After tracing down the code, I discover that when it try to delete the characteristic map of the BLEClient, it will also have to remove the descriptors under it if it exists.

When removing the descriptor from the map in BLERemoteCharacteristic::removeDescriptors() function, the code will first remove the map entry and then will try to access that deleted entry to remove the descriptor object. Which caused the kernel panic as entry are not exist.

The fix is simply removing the line which remove the map entry, and let the following code to remove then descriptor object. The final clear after the for-loop will ensure the map being cleared. 

## Tests scenarios
Tested on ESP32C3 and ESP32


SuGlider:
@chesterkwan - Thanks for the PR!
Could you please provide a testing case code that causes the issue?

VojtechBartoska:
Any updates on this @chesterkwan ?

chesterkwan:
Thanks for reminding me of the pull request, too busy on other projects and nearly forgot about it.

The following section is callback class to handle the connect and disconnect of the BLE device.
What I have done different from normal approach is deleting the BLEClient to prevent memory leak for multiple connect and disconnection.

The faulty condition is that when you try to disconnect a device with write or notify characteristic, it will cause a kernel panic. But if you try to disconnect a device only with read characteristic, everything goes fine.

After drilling into the library, and find out the write and notify both characteristic include a descriptor. And the kernel panic were caused by the code trying to remove an item which have already been removed.

Thanks


`class HandCallback : public BLEClientCallbacks {
  void onConnect(BLEClient* pclient) {
    connected_hand = pclient;
    Serial.println("Connected");
  }

  void onDisconnect(BLEClient* pclient) {
    Serial.println("Disconnect");
    delete(pclient);
    connected_hand = NULL;
    if (connection_state == Connected){
      connection_state = Disconnect;
    }
  }
};`

chesterkwan:
> Any updates on this @chesterkwan ?

Sorry is there any question regarding this pull request?

VojtechBartoska:
Hi @chesterkwan, not right now, it's awaiting the review from our side.

lucasssvaz (approved):
LGTM

libraries/BLE/src/BLERemoteCharacteristic.cpp
```diff
@@ -520,7 +520,6 @@ void BLERemoteCharacteristic::registerForNotify(notify_callback notifyCallback,
 void BLERemoteCharacteristic::removeDescriptors() {
 	// Iterate through all the descriptors releasing their storage and erasing them from the map.
 	for (auto &myPair : m_descriptorMap) {
-	   m_descriptorMap.erase(myPair.first);
 	   delete myPair.second;
 	}
 	m_descriptorMap.clear();   // Technically not neeeded, but just to be sure.
```
lucasssvaz, line 525: Just remove this comment to avoid the clear being deleted in the future.

```suggestion
	m_descriptorMap.clear();
```

lucasssvaz (changes requested):
Looks good. Just a small suggestion.

P-R-O-C-H-Y (approved):

libraries/BLE/src/BLERemoteCharacteristic.cpp
P-R-O-C-H-Y: @chesterkwan Can you please commit this suggested change, so we can merge it?

chesterkwan:
Committed. Thanks

lucasssvaz (approved):

PR #8034, opened 2023-04-05
 recompile the .a files to enable CONFIG_FREERTOS_USE_TRACE_FACILITY

playmiel:


me-no-dev:
Such pull requests need to be addressed in the lib-builder repository and not here. They will get overwritten again by the lib-builder. https://github.com/espressif/esp32-arduino-lib-builder

Issue #6451, opened 2022-03-20
BluetoothSerial / esp_bt_mem_release / memory consumption in arduino api


ferbar:
### Related area

BT, BluetoothSerial

### Hardware specification

any

### Is your feature request related to a problem?

Using BluetoothSerial is taking about 140kB ram of the ESP32. That could be reduced by recompiling the arduino esp idf with a reduced set of options or by using  esp_bt_mem_release(BTLE) for example if bluetooth LE isn't required. Since recompiling takes a lot of time I would suggest some options in the ESP32 arduino sdk.

### Describe the solution you'd like

When `btInUse()` is returning true about 70kB ram are in use since boot. When `btStart()` is called additional 70kB are required. If BluetoothSerial is used, I don't think that BTLE will be used in general, so I suggest an option to BluetoothSerial or BluetoothSerial::begin to disable BTLE and BluetoothSerial::end to disable BT in general (and call `esp_bt_mem_release()`)

a) `btStart()` has to be changed because of
esp_bt_controller_config_t cfg = BT_CONTROLLER_INIT_CONFIG_DEFAULT();
   .mode = ESP_BT_MODE_CLASSIC_BT,

b) `esp_bt_mem_release()` has to be called before `esp_bt_controller_init()` and after `BluetoothSerial::end()`

I already have those modifications and would file a PR when https://github.com/espressif/arduino-esp32/pull/6380 is merged.
@me-no-dev: what do you think about
* add a btStartMode(BT_classic / BTLE / DM) -> releases BTLE or BT_classic ram if not used
* add a btStop_memrelease() -> which frees every BT ram
* Is there any reason why esp32-hal-bt.c is a .c, not .cpp file (I'm asking because overloading those functions would be more convinient)
* everything could be done in BluetoothSerial.cpp, however I think esp32-hal-bt.h: btStart() / btStop is be the right place.

Thanks!

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

antoinnneee:
I have the same issue. I need ~50Ko more memory for my project, and I'm lock due to bluetooth ram usage. I'm not using BLE in project. 

For you're question : _"Is there any reason why esp32-hal-bt.c is a .c, not .cpp file (I'm asking because overloading those functions would be more convinient)"_
All the esp32 frameworks is built with C. And it's a better to keep C for performance and ram usage. 
[...]

PR #8051, opened 2023-04-09
Bluetooth-classic: release BLE memory when BT classic only is requested

ferbar:
### Checklist

-----------
## Description of Change
If CONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY is defined BLE is disabled, therefore the allocated memory can get freed to save about 14kB ram.

## Tests scenarios

Arduino-esp32:2,0.6 (current master) with ESP32 under platformio with and without -DCONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY=y

## Related links
https://github.com/espressif/arduino-esp32/issues/6451


cores/esp32/esp32-hal-bt.c
```diff
@@ -20,10 +20,10 @@ bool btInUse(){ return true; }
 
 #include "esp_bt.h"
 
-#ifdef CONFIG_BTDM_CONTROLLER_MODE_BTDM
-#define BT_MODE ESP_BT_MODE_BTDM
-#elif defined(CONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY)
+#ifdef CONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY
 #define BT_MODE ESP_BT_MODE_CLASSIC_BT
+#elif defined(CONFIG_BTDM_CONTROLLER_MODE_BTDM)
+#define BT_MODE ESP_BT_MODE_BTDM
```
me-no-dev, line 26: does switching defines do anything here?

ferbar: My intention is to tell btStart() that I want to use bluetooth-classic only. CONFIG_BTDM_CONTROLLER_MODE_BTDM is defined in a default aduino-esp32 configuration, so it is the first #if that matches and can not be overridden. When we switch the defines BR_EDR_ONLY can be forced with a -DCONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY.

If you prefer an other #define like ARDUINO_BLUETOOTH_EDR_ONLY or a new function for BT-classic initialization please let me know.

me-no-dev: how about you instead `-DBT_MODE=ESP_BT_MODE_CLASSIC_BT` yourself and only `#ifndef BT_MODE` do the config check.

cores/esp32/esp32-hal-bt.c
```diff
@@ -34,16 +34,28 @@ bool btStarted(){
 
 bool btStart(){
     esp_bt_controller_config_t cfg = BT_CONTROLLER_INIT_CONFIG_DEFAULT();
+#if CONFIG_IDF_TARGET_ESP32
+#ifdef CONFIG_BTDM_CONTROLLER_MODE_BR_EDR_ONLY
```
me-no-dev, line 38: This would need to check `#if BT_MODE == ESP_BT_MODE_CLASSIC_BT` instead

cores/esp32/esp32-hal-bt.c
```diff
@@ -20,30 +20,44 @@ bool btInUse(){ return true; }
...
 
 bool btStart(){
     esp_bt_controller_config_t cfg = BT_CONTROLLER_INIT_CONFIG_DEFAULT();
+#if BT_MODE == ESP_BT_MODE_CLASSIC_BT
+    // esp_bt_controller_enable(MODE) This mode must be equal as the mode in “cfg” of esp_bt_controller_init().
+    esp_bt_controller_mem_release(ESP_BT_MODE_BLE);
+#elif BT_MODE == ESP_BT_MODE_BLE
+    esp_bt_controller_mem_release(ESP_BT_MODE_CLASSIC_BT);
+#endif
+    cfg.mode=BT_MODE;
```
me-no-dev, line 45: `cfg.mode` is only available on ESP32. You should have kept the whole block filtered for ESP32 target. Also as far as I remember, doing mem release of the classic BT will also break BLE. If BT/BLE is not used, that memory is already cleared when Arduino is starting up. So keep just `esp_bt_controller_mem_release(ESP_BT_MODE_BLE);`

ferbar: sorry, you were too fast.
1. esp_bt_controller_mem_release(ESP_BT_MODE_CLASSIC_BT) crashes the esp32 without an error message when using BLE
2. unexpectedly #if BT_MODE == ESP_BT_MODE_CLASSIC_BT doesn't work. The preprocessor only likes integer values, I will find a solution. Thinking of:
 ```
    cfg.mode=BT_MODE;
    if(cfg.mode == ESP_BT_MODE_CLASSIC_BT) {
      esp_bt_controller_mem_release(ESP_BT_MODE_BLE);
    }
```
tomorrow....

Comment (me-no-dev) [reply cores/esp32/esp32-hal-bt.c:45]:
```cpp
#if CONFIG_IDF_TARGET_ESP32
cfg.mode=BT_MODE;
if(cfg.mode == ESP_BT_MODE_CLASSIC_BT) {
  esp_bt_controller_mem_release(ESP_BT_MODE_BLE);
}
#endif
```

Looks like a good plan 👍

VojtechBartoska:
@me-no-dev Can you please check? Updates are there, thanks @ferbar.

ferbar:
If you are fine with my changes I would squash them before the PR merge.

cores/esp32/esp32-hal-bt.c
sanketwadekar: @me-no-dev Is it not possible to clear memory used by BT_CLASSIC if I'm using only BLE? In the case of Rainmaker, we often run into low memory since both Classic BT and BLE are on during WiFi provisioning. Tagging @shahpiyushv.

me-no-dev: @sanketwadekar as far as I remember and it seems that @ferbar confirms it above, it is not possible to release classic memory even if you are using just BLE

ferbar:
@me-no-dev do you need any additional information?

Issue #6922, opened 2022-06-28
ESP32S3 - HID  USBHID  SendReport error and USB Replug


RichardsATcn:
### Board

ESP32-S3-WROOM-1  Moudle

### Device Description

ESP32-S3-WROOM-1

### Hardware Configuration

![image](https://user-images.githubusercontent.com/8119891/176147483-4e5f1d81-746c-451c-b316-3e27de209e14.png)
 only use USB ；

### Version

v2.0.3

### IDE Name

Arduino  1.8.19

### Operating System

Windows 10

### Flash frequency

QIO 80

### PSRAM enabled

no

### Upload speed

115200

### Description

[ESP32-S3-USB.zip](https://github.com/espressif/arduino-esp32/files/8999685/ESP32-S3-USB.zip)

[src.zip](https://github.com/espressif/arduino-esp32/files/8999715/src.zip)

USB device will appear error after using for a while.

If the CDC serial port is enabled, it will be easier to send at fixed time.

Devices will be re enumerated repeatedly

The error code segment is as follows

`  bool res = ready();
    if (!res)
    {
        log_e("not ready");
    }
    else
    {
        res = tud_hid_n_report(0, id, data, len);
        if (!res)
        {
            log_e("report %u failed", id);
        }
        else
        {
            xSemaphoreTake(tinyusb_hid_device_input_sem, 0);
            if (xSemaphoreTake(tinyusb_hid_device_input_sem, timeout_ms / portTICK_PERIOD_MS) != pdTRUE)
            {
                log_e("report %u wait failed", id);
                res = false;
            }
        }
    }
    xSemaphoreGive(tinyusb_hid_device_input_mutex);

//Added code snippet
    if (tinyusb_hid_device_input_sem)
    {
        xSemaphoreGive(tinyusb_hid_device_input_sem);
    }
    log_d("report %u wait over %s", id, res == true ? "true" : "false");
//Added code snippet

    return res;`

usb设备使用一会儿会出现错误。
如果cdc串口开启定时发送时候会更容易出现。
设备会反复的 重新枚举

### Sketch

```cpp
//Added code snippet
    if (tinyusb_hid_device_input_sem)
    {
        xSemaphoreGive(tinyusb_hid_device_input_sem);
    }
    log_d("report %u wait over %s", id, res == true ? "true" : "false");
//Added code snippet
```

### Debug Message

```plain
[...]

PR #8058, opened 2023-04-10
Makes Gamepad example able to be tested with Windows 10/11

SuGlider:
## Description of Change
Initial example code had no effect with Win10/11 because BUTTON_START was not recognized. 
This PR makes it visible in the Windows Game Controller Testing TAB.

## Tests scenarios
ESP32-S2 and ESP32-S3 (OTG mode)

## Related links
https://github.com/espressif/arduino-esp32/issues/6922#issuecomment-1500791745
https://github.com/espressif/arduino-esp32/issues/6922#issuecomment-1501858344

### DIFF
--- a/libraries/USB/examples/Gamepad/Gamepad.ino
+++ b/libraries/USB/examples/Gamepad/Gamepad.ino
@@ -7,7 +7,9 @@ void loop(){}
 #include "USBHIDGamepad.h"
 USBHIDGamepad Gamepad;
 
-const int buttonPin = 0;
+// This sketch works correctly for the ESP32-S2 and ESP32-S3 in OTG mode (TinyUSB)
+
+const int buttonPin = 0;  // GPIO 0 is the BOOT button of the board
 int previousButtonState = HIGH;
 
 void setup() {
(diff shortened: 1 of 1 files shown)


libraries/USB/examples/Gamepad/Gamepad.ino
```diff
@@ -18,9 +20,14 @@ void setup() {
 
 void loop() {
   int buttonState = digitalRead(buttonPin);
-  if ((buttonState != previousButtonState) && (buttonState == LOW)) {
-    Gamepad.pressButton(BUTTON_START);
-    Gamepad.releaseButton(BUTTON_START);
+  if (buttonState != previousButtonState) {
+    if (buttonState == LOW) { // BOOT Button pressed
+      // changes many states
+      Gamepad.send(-100, 100, 50, 100, -100, -50, HAT_DOWN_RIGHT, BUTTON_TL);
```
me-no-dev, line 26: last argument here is mask, so correct value would be `(1 << BUTTON_TL)`.

SuGlider:
After testing all combinations, it seems that no button is recognized by Win11.
Maybe the report is different.

me-no-dev:
Does it all work (buttons and axis) when you have that window Active, or only some things are working? Generally that is very strange... usb report is really standard.

SuGlider:
> Does it all work (buttons and axis) when you have that window Active, or only some things are working? Generally that is very strange... usb report is really standard.

All working fine. I'll submit the changes I made to the example that are able to test everything.

Issue #6947, opened 2022-07-06
WiFi (specifically WifiGeneric) handling of events is subject to race conditions


egnor:
### Board

any board (with networking)

### Device Description

any device (with networking)

### Hardware Configuration

no special configuration

### Version

latest development Release Candidate (RC-X)

### IDE Name

any IDE

### Operating System

any OS

### Flash frequency

any frequency

### PSRAM enabled

yes

### Upload speed

any speed

### Description

It's hard to write a specific repro case for this since it depends on race conditions. (Also, my understanding may be off, please correct!)

The `WiFi` instance allows user code to register for events with `WiFi.onEvent()`. This isn't documented, but it is present in most example code, described in various tutorials and copied into many/most network-using sketches. This is implemented the `WiFiGeneric` class (one of the base classes for `WiFiClass`, of which `WiFi` is an instance) using an independently running FreeRTOS task (pinned to core 0(?) by default):
https://github.com/espressif/arduino-esp32/blob/fcd4799c6de6eb5a5a8eba94818adf770238ecc0/libraries/WiFi/src/WiFiGeneric.cpp#L299

That free-running task reads from a queue and invokes `WiFiGenericClass::_eventCallback()`:
https://github.com/espressif/arduino-esp32/blob/fcd4799c6de6eb5a5a8eba94818adf770238ecc0/libraries/WiFi/src/WiFiGeneric.cpp#L924

After translating the event, `_eventCallback()` loops through all registered callbacks and invokes them serially:
https://github.com/espressif/arduino-esp32/blob/fcd4799c6de6eb5a5a8eba94818adf770238ecc0/libraries/WiFi/src/WiFiGeneric.cpp#L1049

Nowhere in this process is locking or other synchronization done, that I can figure out. I see these implications?

1. If code adds/removes callbacks at the same time as callbacks are being invoked, arbitrarily bad undefined behavior can result due to conflict over `cbEventList`. Most code will add callbacks at the start and leave them there forever, but if that's the expectation it seems worth documenting?
[...]

Issue #7227, opened 2022-09-06
WiFi Hostname is only set when changing mode to WIFI_MODE_STA


tbnobody:
### Board

generic

### Device Description

-

### Hardware Configuration

No

### Version

v2.0.4

### IDE Name

VSCode with PlatformIO

### Operating System

Windows 11

### Flash frequency

-

### PSRAM enabled

no

### Upload speed

115200

### Description

There are several issues in issue trackers out there in the internet which are stating that the hostname is not set correctly. When diving deeper into the library, it seems that the hostname is only set on one location: https://github.com/espressif/arduino-esp32/blob/77065bffb2c57704ee46ad0a43270e50860814f5/libraries/WiFi/src/WiFiGeneric.cpp#L1122

That means set a call of `WiFi.setHostname()` does nothing in the first step. It's not even enough to call `WiFi.mode(WIFI_MODE_STA);` if the mode was already WIFI_MODE_STA. Because if current mode equals new mode: https://github.com/espressif/arduino-esp32/blob/77065bffb2c57704ee46ad0a43270e50860814f5/libraries/WiFi/src/WiFiGeneric.cpp#L1109
no changes are applied.

To set the hostname correctly the following snippet has to be used:
```
WiFi.mode(WIFI_MODE_APSTA); // or any other mode
WiFi.setHostname("asdf");
WiFi.mode(WIFI_MODE_STA);
```

If the behavior is known to the user of the class it's  simple to use but it's a little bit unintuitive. Does it make sense to at least document this behavior?

### Sketch

```cpp
-
```

### Debug Message

```plain
-
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

domdfcoding:
I've run into the same issue of the hostname being set, with internet unhelpfully guides suggesting to do it they way I was already. 
I've tried the approach listed here (with the exception of `WIFI_STA` instead of `WIFI_MODE_STA`, not sure what the difference is) and it works as I would expect. 

Could this be fixed, or at least the limitation documented, please?

I'm running on an ESP32-C3, if it's relevant.

pedrominatel:
Hi, @domdfcoding,

This will be added to the documentation. On the code side, I'll discuss with the team if there is something to do.

Parsaabasi:
Hello, 

[...]

Issue #6700, opened 2022-05-05
WiFi.setHostname() doesn't always work


mrengineer7777:
### Board

ESP32-WROOM-32UE

### Device Description

Custom hardware, but this is not a hardware issue.
FYI we need an option for "Version" below: arduino-esp32 v2.0.3.

### Hardware Configuration

Interfaces with 2 UARTs.

### Version

latest master (checkout manually)

### IDE Name

VSCode with PIO

### Operating System

Win10

### Flash frequency

?

### PSRAM enabled

no

### Upload speed

460800

### Description

WiFi.setHostname() doesn't always set the hostname.  After looking at #6086, #6278, and studying WiFiSTA, WiFiGeneric, I can see the issue.

setHostname() just sets a string (cstr) in memory.  It is not applied until WiFiGenericClass::mode() is called, which happens when WiFi.config(), WiFi.begin(), WiFi.enableSTA(), etc are called.

The issue is that if the WiFi.mode() is already STA, then the hostname will not be set even if WiFi is not connected.  I would expect hostname to be set in WiFi.begin(), or WiFi.mode() to handle setting it when STA is enabled but not connected.

In our case we scan WiFi networks on startup, which apparently enables STA mode, so setHostname() is never applied.

Impact: It would be nice if it worked.

Workaround: If I stop wifi [`WiFi.mode(WIFI_MODE_NULL);`] before setting the hostname, then it works.

### Sketch

```cpp
Extracted pieces of my code:

void setup(void) {
    WDT_Start();    //Reboot if our main loop stops responding
    Serial.begin(115200);
    Serial.println("Booting...");  // USB won't be connected yet, so we will only see this on a reboot
    delay(2000);  // Delay for debugging. Allows time for USB to connect to computer.

    WiFiCtrl.StartEventHandler();
    WiFiCtrl.StartPingSvc();  // Make sure SoftAP or WiFi are up
    StartScanAsync();         // Start scan before starting WiFi as scan will fail if WiFi isn't connected. Completion of WiFi scan starts WiFi.
}

void loop {
    WDT_KeepAlive();
    delay(100);             // Allow other tasks to run
    DispCtrl_ReadEvents();  // Process any events from display (currently only factory reset msg)
}

//NOTE: WiFiCtrl class is a wrapper I wrote to extend the WiFi class. It basically handles events.
[...]

Issue #6278, opened 2022-02-13
WiFi.setHostname() not working


dsilletti:
### Board

ESP32

### Device Description

plain module

### Hardware Configuration

N/A

### Version

latest master

### IDE Name

Arduino IDE

### Operating System

macOS 12.1

### Flash frequency

40Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

WiFi.setHostname() does not work, hostname remain always esp32-XXXXXX

### Sketch

```cpp
WiFi.disconnect(true,true);
delay(1000);
WiFi.mode(WIFI_STA);
esp_wifi_set_country_code("IT", true);
WiFi.setHostname("hostname");
```

### Debug Message

```plain
N/A
```

### Other Steps to Reproduce

WiFi.setHostname("hostname");

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

me-no-dev:
how did you test this? it's often that the DHCP server might cache the hostname

dsilletti:
@me-no-dev I power cycled the Wi-Fi router that provides the DHCP server several times, power cycled the esp32 several times, but the hostname is always esp32-XXXXXX

JonGordon123:
On most routers if you want to clear the cache, you have to reset it using the reset button... and not just power cycle it. You could... connect to a different router, or reset the router, or change the esp32 MAC address to confirm the hostname shows up in the router.

me-no-dev:
I have mikrotiks here for routers and they have "clear leases" option that allowed me to confirm the hostname change, else I also thought it did not work.

dsilletti:
Interesting that even if I change the MAC address, the IP address I've got from the DHCP is different, but the hostname I see on the router is still the same esp32-XXXXXX. @me-no-dev could you please share your relevant code lines, this is in my running sketch (including the MAC address change I applied now for testing):
```
  WiFi.disconnect(true,true);
  delay(1000);
  WiFi.mode(WIFI_STA);
  uint8_t newMACAddress[] = {0x32, 0xAE, 0xA4, 0x07, 0x0D, 0x66};
  esp_wifi_set_mac(WIFI_IF_STA, &newMACAddress[0]);
  esp_wifi_set_country_code("IT", true);
  WiFi.setHostname(hostname);
```

gonzabrusco:
[...]

PR #8081, opened 2023-04-16
Add documentation on threading issues with WiFi.onEvent() to examples

egnor:
## Description of Change
Adds comments explaining that the handler registered with `WiFi.onEvent()` will be called from another thread, and thus must be written carefully to be thread safe. Add `// FIXME` comments to specific handlers which are apparently written unsafely (or can't be easily validated as safe).

## Tests scenarios
Untested, but only changes comments.

## Related links
Addresses https://github.com/espressif/arduino-esp32/issues/6947#issuecomment-1506398379


me-no-dev:
this is nothing strange. I am not sure it warrants the comments. Why `setHostname` is called where it is has it's reason and is not unknown. The method can not be called before the interface is started and you would also like it to be set before start talking to DHCP servers.  Many things in our Arduino core run in threads and most of our APIs are thread-safe, or at least the ones that make sense to be, like SPI, Wire, Serial and so on. Some caution always needs to exercised.

egnor:
> this is nothing strange. I am not sure it warrants the comments. Why `setHostname` is called where it is has it's reason and is not unknown. The method can not be called before the interface is started and you would also like it to be set before start talking to DHCP servers.

IMHO, those sequencing requirements are not clear for the end user, hence the various bugs in the topic of "I called `setHostname()` and it didn't work". The various sequencing conflicts with regard to the the TCP/IP subsystem are definitely an internal thing from the perspective of an Arduino user. And notably they differ between `WiFi` and `ETH`...

> Many things in our Arduino core run in threads
Threads are often used internally but it's relatively uncommon that end-user code is run in a thread without that being mentioned. I have seen a large number of cases of code patterned after these examples which did non threadsafe things with the app's own data structures. A lot of programmers using the Arduino API have never even heard of threads.

> and most of our APIs are thread-safe, or at least the ones that make sense to be, like SPI, Wire, Serial and so on.
It's far from clear to the observer which APIs might "make sense to be" thread safe. Specifically many of the methods on WiFi itself are not in fact thread safe!

> Some caution always needs to exercised.
It's hard to exercise caution when you don't know the pitfalls-- the intent here was to provide a small warning sign next to one such pitfall.

In my understanding, generally speaking, Arduino APIs (as opposed to ESP-IDF APIs) are intended to be usable by less-technically-experienced users, and anything as gnarly as "this callback runs in a separate thread" really needs to be called out.

Normally I'd say that the fact that `.onEvent` is undocumented would be enough of a red flag, but the code I'm suggesting comments for is _example code_ users are specifically encouraged to copy and adapt...

I'm certainly happy to revise these comments to give more explicit reasoning for why things are done the way they are, if that seems helpful to you?

me-no-dev:
> I'm certainly happy to revise these comments to give more explicit reasoning for why things are done the way they are, if that seems helpful to you?

I think it makes much more sense some info to be added to our documentation, rather than in examples. It could promote the functionality and warn the users that would not look at those examples as well. What do you think?

c.c. @pedrominatel

pedrominatel:
@me-no-dev and @egnor,

We can add this information to the WiFi API and create a new document on Tutorials to show how to use it instead of adding comments on each example (hard to maintain in future examples or any changes to the WiFi API).

Do you want to propose this change on the docs, @egnor?

mrengineer7777:
Do we have a WiFi .rst file were we can put comments rather than on the examples?

me-no-dev:
@mrengineer7777 like [this one](https://github.com/espressif/arduino-esp32/blob/master/docs/source/api/wifi.rst)?

egnor:
Note that the [current `WiFi` documentation](https://github.com/espressif/arduino-esp32/blob/master/docs/source/api/wifi.rst) does not mention `WiFi.onEvent()` at all; I have mixed feelings about documenting it given that it invokes threads, but if it's used in the example code it should be, and I can attempt to add a description. We can also document the limitations on `WiFi.setHostname()` there -- I'm not sure I 100% understand those limitations, but I can make a stab.

Note also that the [current `ETH` documentation](https://github.com/espressif/arduino-esp32/blob/master/docs/source/api/ethernet.rst) is barely more than a stub, and simply points to examples. We can try to expand it, or reference the `WiFi` document.

I would also put in a final plug for ALSO adding SOME comments to the example code. It's true that we can't ensure that every example gets documented, but many, many, maybe even most people use these APIs by starting out looking at examples, and it's quite handy if any significant pitfalls are called out directly therein. I think the surprise introduction of threads counts as a significant pitfall.

mrengineer7777:
@egnor Adding comments to each example clutters them up unnecessarily.  I have had no issues calling other functions from WiFi.onEvent() function handlers.  That said, I believe documenting WiFi.onEvent() in WiFi.rst is very valuable.  It's important to know about that separate thread in case the user notices odd behavior or crashes.

Documenting WiFi.setHostname() would also be interesting but may be a bigger task than you want to tackle.  The short story on setHostname() is this: Arduino saves a copy of the hostname when you call the function, but it is not applied except when WiFi.mode goes from NONE -> STA.  Not sure about Ethernet.  It has something to do with the underlying IDF libraries.  To make sure the hostname is set, you have to call it before WiFi.begin().  See the following related issues:

https://github.com/espressif/arduino-esp32/issues/7227
https://github.com/espressif/arduino-esp32/issues/6700
https://github.com/espressif/arduino-esp32/issues/6278
https://github.com/espressif/arduino-esp32/issues/6086
https://github.com/espressif/arduino-esp32/issues/2537

I agree more documentation on Ethernet would be good.  Should probably do that on a separate PR.

egnor:
I think my philosophy is different than yours about threads and debugging, but, I shall endeavor to add something to WiFi.rst and thin out the comments. (I'd like to leave a small warning; I submit that its value is extremely high, as threads are quite finicky and dangerous for the unfamiliar. Occasional weird hard-to-debug glitches are very. annoying. especially in embedded systems...)

egnor:
OK I've added some content to `wifi.rst`, and greatly thinned out the comments in example code. Take another look, @me-no-dev and @mrengineer7777 ?

for onEvent:

image

for setHostname:

image

pedrominatel (approved):
Sorry for the long delay, @egnor. LGTM.
cc @me-no-dev and @VojtechBartoska

VojtechBartoska (approved):
LGTM

pedrominatel:
Can we merge, @VojtechBartoska and @me-no-dev? I need to resolve some conflicts in #8650 to finish the deployment process.

PR #8091, opened 2023-04-17
(Update camera_pins.h) Flash does not work on the esp32cam

tnagorran:
it is planned to use the flash led but a normal led is configured instead

### DIFF
--- a/libraries/ESP32/examples/Camera/CameraWebServer/camera_pins.h
+++ b/libraries/ESP32/examples/Camera/CameraWebServer/camera_pins.h
@@ -155,7 +155,8 @@
 #define HREF_GPIO_NUM     23
 #define PCLK_GPIO_NUM     22
 
-#define LED_GPIO_NUM      33
+// 4 for flash led or 33 for normal led
+#define LED_GPIO_NUM       4
 
 #elif defined(CAMERA_MODEL_TTGO_T_JOURNAL)
 #define PWDN_GPIO_NUM      0


Rob58329:
Is changing “#define LED_GPIO_NUM 33“ to “#define LED_GPIO_NUM 4” the best idea, as most ESP32-CAM boards do have a normal LED on GPIO33, and the flash LED on GPIO4 is not present on all ESP32-CAM boards, and is a bit special as well as GPIO4 is also connected to HS2_Data on the SD-Socket so you have to disable HS2_Data on the SD-socket if you want to use the flash at the same time as the SD-socket.

Might it not be a better idea to leave “#define LED_GPIO_NUM 33“ and create a new variable “#define FLASH_GPIO_NUM 4“?

dc740:
I'm trying to get the flash to work without success. I tried two boards already. I have 3. The third one has the Prusa ESP32 CAM Firmware, and the flash works just fine. The flash is in pin 4 (for the prusa firmware too). I wonder if I have to disable something to get it to work. The SD is not inserted in any of the three modules.

Rob58329:
@dc740:  You do not say which board you are having issues with, or detail the code you are using.  However it you are using one of the boards which has the flash-led connected to the same pin as the SD-card HS2_Data pin, and you .begin() the SD card, you will not be able to use the flash-led (irrespective of if you have an SD-card in the slot or not).

In this case, you instead have to either (a) not .begin() the SD card, or (b)  .begin() the SD-card in 1bit-mode, so it doesn't use the HS2_Data pin.

See "https://randomnerdtutorials.com/esp32-cam-ai-thinker-pinout/" other similar websites
(do a search on "ESP32cam sd 1bit mode").

Typically, these say that the relevant command is:
`bool begin(const char * mountpoint="/sdcard", bool mode1bit=false); // cf. SD_MMC.h`

And typical usage is:
`SD_MMC.begin("/sdcard", true); // Start the SD card in 1bit mode.`

dc740:
I apologize for the lack of details. The board is an ESP32 CAM version 1 detailed in this article
https://blog.prusa3d.com/development-of-the-open-source-camera-firmware-solution-from-the-community-for-prusa-printers_94421/

I had three at home and was experimenting with them, and it was weird that the prusa firmware would work. And you explained it perfectly. They are initializing the SD in 1bit-mode.
https://github.com/prusa3d/Prusa-Firmware-ESP32-Cam/blob/master/ESP32_PrusaConnectCam/micro_sd.cpp#L118

In my case I'm using the example code as-is. So I was expecting there would be no SD initialization, but I must be wrong, since I can't get it to work.
https://github.com/espressif/arduino-esp32/tree/master/libraries/ESP32/examples/Camera/CameraWebServer

I'd guess there is code to initialize the SD, but it's not inside the example itself and I'm just missing it.

PR #8096, opened 2023-04-18
fix path for bootloaders, changed from `sdk` to `esp32-arduino-libs`

Jason2866:
without the needed bootloader files are not found.

@me-no-dev 
Supporting development with Platformio is currently nearly impossible since the toolchains 12.2.0 are not available in Platformio registry. https://github.com/espressif/crosstool-NG/issues/41


me-no-dev:
ahh... must have missed that part

PR #8126, opened 2023-04-26
LEDC Refactoring - Peripheral manager implemented

P-R-O-C-H-Y:
## Description of Change
Refactored LEDC peripheral to use peripheral manager, with that goes API changes.
Updated documentation.
Only 1 tone pin can be used at the same time now, to match Arduino functionality.

## Tests scenarios
Tested on all supported chips.


pedrominatel:
@P-R-O-C-H-Y, it will be nice if you add the readme files to the examples.

docs/source/api/ledc.rst
pedrominatel, line 1: LGTM

cores/esp32/esp32-hal-ledc.c
```diff
@@ -157,110 +216,80 @@ uint32_t ledcWriteNote(uint8_t chan, note_t note, uint8_t octave){
...
+    ledc_channel_handle_t bus = (ledc_channel_handle_t)perimanGetPinBus(pin, ESP32_BUS_TYPE_LEDC);
+    if(bus == NULL && perimanSetPinBus(pin, ESP32_BUS_TYPE_INIT, NULL)){
+
+        if(ledcAttach(pin, analog_frequency, analog_resolution) == 0){
+            log_e("analogWrite setup failed (freq = %u, resolution = %u). Try setting different resolution or frequency");
+            return;
+        }
     }
-    ledcWrite(pin_to_channel[pin] - 1, value);
+    ledcWrite(pin, value);
```
SuGlider, line 277: One issue I think that there is with analogWrite() would be that it allocates ("eats") ledc channels and there is no way to release them.
A suggestion is to allow it to release/detach the channel when <value> is Zero.
It may have to make sure that the GPIO goes to low level...

cores/esp32/Tone.cpp
```diff
@@ -6,20 +6,18 @@
 
 static TaskHandle_t _tone_task = NULL;
 static QueueHandle_t _tone_queue = NULL;
```
SuGlider, line 8: Not sure, but I think that @me-no-dev want us to try to do not use Tasks within the peripherals.
I don't see a clear motivation for using a task in the `Tone` implementation...

cores/esp32/esp32-hal-ledc.c
P-R-O-C-H-Y: > I think that analogWrite() has a potential issue:
This function allocates ("eats") LEDC channels and there is no way to release them.

You may be wrong with this, as the peripheral manager will be implemented in 3.0.0, if you setup any other peripheral on that pin, the peripheral manager will call ledcDetachBus and free that LEDC channel.

cores/esp32/Tone.cpp
P-R-O-C-H-Y: Maybe @PilnyTomas can take a look on this in separate PR, as he wrote this implementation :)

me-no-dev: Motivation here is to be able to "setup" a melody and let it play, without having to wait for it to be over. Same goes for when you want to just set it to beep for 100ms and not having to wait. We could look into doing it blocking or maybe coming up with another idea.

cores/esp32/esp32-hal-ledc.c
me-no-dev: I agree with @P-R-O-C-H-Y here. It's not necessary to switch to GPIO if the result is what is expected to be. Maybe there is a hardware reason to be done in that way on AVR

me-no-dev:
@P-R-O-C-H-Y PTAL on the conflicts

P-R-O-C-H-Y:
> @P-R-O-C-H-Y PTAL on the conflicts

conflicts resolved

PR #8129, opened 2023-04-27
TOUCH - Peripheral manager implementation

P-R-O-C-H-Y:
## Description of Change
Implemented peripheral manager into touch driver.
As touch is already using pins as input, there was no need to change any API, just added `touchDetachBus` function for the Peripheral manager, which now call `touch_pad_deinit()`, if no more touchpads are used.

## Tests scenarios
Tested on ESP32-S3.


P-R-O-C-H-Y:
HW test fails on timer, not related to this PR. HW timer test fixed by PR #8135.

PR #8179, opened 2023-05-10
GPIO - Peripheral manager implementation

P-R-O-C-H-Y:
## Description of Change
Implemented peripheral manager into gpio driver.
As gpio functions are already using pins as input, there was no need to change any API, just added gpioDetachBus function that is required for the Peripheral manager, which just returns true.

## Related links

### DIFF
--- a/cores/esp32/esp32-hal-gpio.c
+++ b/cores/esp32/esp32-hal-gpio.c
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "esp32-hal-gpio.h"
+#include "esp32-hal-periman.h"
 #include "hal/gpio_hal.h"
 #include "soc/soc_caps.h"
 
@@ -89,6 +90,10 @@ static InterruptHandle_t __pinInterruptHandlers[SOC_GPIO_PIN_COUNT] = {0,};
 
 #include "driver/rtc_io.h"
 
+static bool gpioDetachBus(void * bus){
+    return true;
+}
+
 extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
 {
 #ifdef RGB_BUILTIN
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-gpio.c
```diff
@@ -130,24 +143,41 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
...
+                const uint8_t comm_val = val != 0 ? RGB_BRIGHTNESS : 0;
+                neopixelWrite(RGB_BUILTIN, comm_val, comm_val, comm_val);
+                return;
+            }
+        #endif
+        gpio_set_level((gpio_num_t)pin, val);
+    }
+    else {
+        log_e("Pin is not set as GPIO.");
+    }
```
me-no-dev, line 169:
```cpp
  #ifdef RGB_BUILTIN
      if(pin == RGB_BUILTIN){
          //use RMT to set all channels on/off
          const uint8_t comm_val = val != 0 ? RGB_BRIGHTNESS : 0;
          neopixelWrite(RGB_BUILTIN, comm_val, comm_val, comm_val);
          return;
      }
  #endif
     if(perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) != NULL){
         gpio_set_level((gpio_num_t)pin, val);
     } else {
         log_e("Pin is not set as GPIO.");
     }

```

cores/esp32/esp32-hal-gpio.c
```diff
@@ -130,24 +143,41 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
...
 }
 
 extern int ARDUINO_ISR_ATTR __digitalRead(uint8_t pin)
 {
-	return gpio_get_level((gpio_num_t)pin);
+    if(perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) != NULL){
+	    return gpio_get_level((gpio_num_t)pin);
```
me-no-dev, line 175: some extra spaces here

P-R-O-C-H-Y: removed

cores/esp32/esp32-hal-gpio.c
P-R-O-C-H-Y: fixed, Thanks :)

P-R-O-C-H-Y:
@me-no-dev Fixed everything and tested `digitalRead()` on pin used for LEDC and result was always 0, but led channel was setup with resolution 8 and duty 255 so full ON. So I think we can leave the check for the BUS there, as the `digitalRead()` won't work anyways.

PR #8182, opened 2023-05-10
HardwareSerial: fix begin() lock issue on error path

pillo79:
If the user supplied a wrong UART number, the `begin()` method would return without releasing the lock. Add missing `UNLOCK()` call.

No other unbalanced locks exist in this file.

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -367,6 +367,7 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
 #endif
             default:
                 log_e("Bad UART Number");
+                HSERIAL_MUTEX_UNLOCK();
                 return;
         }
     }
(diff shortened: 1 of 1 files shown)


cores/esp32/HardwareSerial.cpp
```diff
@@ -367,6 +367,7 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
 #endif
             default:
                 log_e("Bad UART Number");
+                HSERIAL_MUTEX_UNLOCK();
```
me-no-dev, line 370: This default case is here just to make the compiler happy. In fact, you will never get here with wrong port number, because this check has already been made on line 330

SuGlider: I just tested the compilation and it doesn't complain about the lack of `default:` case.
I'll just remove this case from the code to make it clearer.

SuGlider:
@pillo79 - Thanks for noticing it and for the PR.
As wrote in the review, such default case will never happen because the `_uart_nr` is tested at the beginning of the function code.
I have removed such `case` from the `switch` in order to make the code clearer.

@me-no-dev - It can be merged to the 2.x and 3.0.0 branches.

pillo79:
> @pillo79 - Thanks for noticing it and for the PR. As wrote in the review, such default case will never happen because the `_uart_nr` is tested at the beginning of the function code. 
Sorry for missing that! :unamused: Removing the default case is even better, thanks :+1:

PR #8183, opened 2023-05-10
Update PlatformIO scripts for the upcoming 3.0 core

valeros:
This PR enables PlatformIO checks in CI and updates related scripts according to the changes planned for core v3.0:

- Added new package `framework-arduinoespressif32-libs` with prebuilt IDF libraries. In the current implementation the libraries are cloned directly from https://github.com/espressif/esp32-arduino-libs.git with a specific commit from that repository specified in the `FRAMEWORK_LIBS_REVISION` variable. 
- Updated toolchain and esptoolpy packages
- Minor changes according to recent updates in PlatformIO's `espressif32` dev-platform.

### DIFF
--- a/.github/scripts/install-platformio-esp32.sh
+++ b/.github/scripts/install-platformio-esp32.sh
@@ -3,8 +3,10 @@
 export PLATFORMIO_ESP32_PATH="$HOME/.platformio/packages/framework-arduinoespressif32"
 PLATFORMIO_ESP32_URL="https://github.com/platformio/platform-espressif32.git"
 
-TOOLCHAIN_VERSION="8.4.0+2021r2-patch5"
-ESPTOOLPY_VERSION="~1.40400.0"
+TOOLCHAIN_VERSION="12.2.0+20230208"
+ESPTOOLPY_VERSION="~1.40501.0"
+FRAMEWORK_LIBS_REPOSITORY="https://github.com/espressif/esp32-arduino-libs.git"
+FRAMEWORK_LIBS_REVISION="1d21c0e389" # points to the "2004bf4e11" revision of IDF v5.1.0
 ESPRESSIF_ORGANIZATION_NAME="espressif"
 
 echo "Installing Python Wheel ..."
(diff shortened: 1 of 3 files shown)


me-no-dev:
Very nice, but I am worried that the versions are fixed, rather than "use latest" or "use the same one as in the env". Any chance for that to happen?

valeros:
> Very nice, but I am worried that the versions are fixed, rather than "use latest" or "use the same one as in the env". Any chance for that to happen?

Sorry, what versions exactly? I just used the same packages as in [ackage_esp32_index.template.json](https://github.com/espressif/arduino-esp32/blob/idf-release/v5.1/package/package_esp32_index.template.json#L38-L92).

me-no-dev:
Yes, you use the ones currently in the json, but what happens when those change? libs and tools get updated from the lib-builder and will often change. Understandably, toolchain version might differ for a few days, so we could think of handling that version in a different way, but at least for libs, version should be read from the package template json in order to test with current libs.

valeros:
Thanks for the clarification, I'll try to make it a bit more dynamically configured.

me-no-dev:
@valeros maybe it makes sense for libs version to be updated automatically by the lib-builder, when it's pushing an update to this repo? https://github.com/espressif/esp32-arduino-lib-builder/blob/release/v5.1/tools/push-to-arduino.sh#L168

valeros:
> maybe it makes sense for libs version to be updated automatically by the lib-builder, when it's pushing an update to this repo?

@me-no-dev Sounds good, but it seems there might be some obstacles. Please correct me if I'm wrong.

- I believe I cannot simply replace the version of `esp32-arduino-libs` package in the `install-platformio-esp32.sh` file while running CI in the `esp32-arduino-lib-builder` repo. Mainly because it clones the `arduino-esp32` repo from the default branch (master) which won't contain the latest changes from this PR until `idf-release/v5.1` branch is merged to master. 
- I probably can create a new Python script to autogenerate the `install-platformio-esp32.sh` file from scratch in the `esp32-arduino-lib-builder` CI and then push it as you described via https://github.com/espressif/esp32-arduino-lib-builder/blob/release/v5.1/tools/push-to-arduino.sh#L168
- Third approach was already mentioned, I just read already available `package_esp32_index.template.json` in this repo and extract version for `esp32-arduino-libs`.

Any thoughts?

me-no-dev:
I would say, go for third option. We will ensure to keep the version string with the same structure

valeros:
@me-no-dev Please take a look at the latest changes. Now the package with prebuilt libraries is added dynamically.

me-no-dev:
Thanks @valeros ! Let's give it a shot :)

Issue #7748, opened 2023-01-23
Provisioning example fails to scan networks


edg2411:
### Board

esp32 wroom

### Device Description

it is a custom board but uses the esp32 wroom 32e

### Hardware Configuration

nothing strange

### Version

latest master (checkout manually)

### IDE Name

platformio

### Operating System

windows 10

### Flash frequency

40mhz

### PSRAM enabled

yes

### Upload speed

115200

### Description

the provisioning example provided in the arduino libraries dont find networks by scanning or not scanning at all i dont know. The esp-idf version of the example works just fine im using the android provisioning app.
I see other issues but no one seems to help me

### Sketch

```cpp
it is the example without modifications, im using softap but same problem happens with bt
```

### Debug Message

```plain
cant see any message, just not finding it in the app
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

PilnyTomas:
Hi @edg2411 , could you please send a link to an exact example?
Please try to change the debug level to verbose, compile & flash again and post the serial output.
Can you please tell me something about the WiFis you expect to be found?

edg2411:
Hi thank you for replying

https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFiProv/examples/WiFiProv/WiFiProv.ino
it is the example that offers arduino ide
i did some debug and it is scanning the networks that are available just wifis from routers and common APs like my home network..
As I said when using the same example but from the idf sdk which use the same libraries i think it works just fine.. the networks appear in the app.
https://github.com/espressif/esp-idf/tree/master/examples/provisioning this is the idf example that its working for me
https://github.com/espressif/esp-idf-provisioning-android and this is the app that im using for both examples..
I will do some debug with verbose level tomorrow if i can

edg2411:
I changed the compiler option to verbose and now its scanning and showing the networks

[    91][D][WiFiGeneric.cpp:931] _eventCallback(): Arduino Event: 0 - WIFI_READY
[    93][I][WiFiProv.cpp:137] beginProvision(): Starting provisioning AP using SOFTAP. service_name : Prov_123, pop : abcd1234
[    93][V][WiFiGeneric.cpp:515] _arduino_event_cb(): Provisioning Initialized!
[   190][D][WiFiGeneric.cpp:931] _eventCallback(): Arduino Event: 33 - PROV_INIT
[...]

PR #8209, opened 2023-05-16
Provisioning fix and upgrades

PilnyTomas:
Based on issue #7748
The main problem is simply case size for the device name:

The IDF example is using upper case: `const char *ssid_prefix = "PROV_";` [line 210](https://github.com/espressif/esp-idf/blob/54576b7528b182256c027de86eb605a172bc2821/examples/provisioning/wifi_prov_mgr/main/app_main.c#L210)

While the Arduino is using lower case (only the first letter is upper case):
https://github.com/espressif/arduino-esp32/blob/3ec5f4efa1de4342aaab742008dc630091e5e035/libraries/WiFiProv/examples/WiFiProv/WiFiProv.ino#L50

This PR changes the device name to the same as the IDF name and also to what is expected by the app on default settings.
This should remove the confusion for first-time users.

This PR also added function for printing QR code (simply copied from the IDF example)

This PR also updates the README file.

### DIFF
--- a/libraries/WiFiProv/src/WiFiProv.cpp
+++ b/libraries/WiFiProv/src/WiFiProv.cpp
@@ -166,7 +166,6 @@ void WiFiProvClass :: beginProvision(prov_scheme_t prov_scheme, scheme_handler_t
 
 // Copied from IDF example
 void  WiFiProvClass :: printQR(const char *name, const char *pop, const char *transport){
-#if __has_include("qrcode.h")
     if (!name || !transport) {
         log_w("Cannot generate QR code payload. Data missing.");
         return;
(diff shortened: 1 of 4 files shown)


libraries/WiFiProv/src/WiFiProv.cpp
```diff
@@ -26,6 +26,7 @@
 #include <esp_wifi.h>
 #include <esp_event.h>
 #include <esp32-hal.h>
+#include "qrcode.h"
```
me-no-dev, line 29: you should check if that include exists and only then include and provide the new function. When building as component and qrcode component is not additionally installed, compilation will fail: https://github.com/espressif/arduino-esp32/actions/runs/4989420314/jobs/8933324447?pr=8209#step:4:1408

PilnyTomas: I was expecting this to be available for all SoCs, I can see that is not, but do not understand why?
How can I test it?

me-no-dev: it exists from the lib-builder, because it's installed with RainMaker, but as clean IDF+Arduino-as-component it does not exist for ANY chip.

PilnyTomas: Should be resolved - CI passed

PR #8213, opened 2023-05-16
Fixed compiler warning

pilinux:
fixed warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]

-----------
## Description of Change

When `dl_math.hpp` is used [`#include "dl_math.hpp"`], compiler generates warnings because the implemented functions `sqrt_quick()` and `sqrt_reciprocal_quick()` violate the strict aliasing rules of C/C++ language standard. To avoid the warning and adhere to strict aliasing rules, a type-safe approach is used in this commit.

## Tests scenarios

The pull request is built for ESP32, ESP32-C3, ESP32-S2 and ESP32-S3 chips and tested on ESP32-S3 dev module.

Furthermore, as a reference, the following C++ code can be used to verify the underlying math.

`g++ main.cpp -o main`

`./main`

```
#include <cmath>
#include <iostream>

inline float sqrt_quick(float x)
{
    //    const int result = 0x1fbb4000 + (*(int *)&x >> 1);
    //    return *(float *)&result;
    union
    {
        float f;
        int i;
    } pun;

    pun.f = x;
    pun.i = 0x1fbb4000 + (pun.i >> 1);
    return pun.f;
}

inline float sqrt_reciprocal_quick(float x)
{
    float xhalf = 0.5f * x;
    // int i = *(int *)&x;             // get bits for floating value
    int i;
    memcpy(&i, &x, sizeof(float));
    i = 0x5f375a86 - (i >> 1); // gives initial guess y0
    // x = *(float *)&i;               // convert bits back to float
    memcpy(&x, &i, sizeof(float));
    x = x * (1.5f - xhalf * x * x); // Newton step, repeating increases accuracy
    return x;
}

int main()
{
    // test cases for sqrt_quick()
    float inputs0[] = {4.0f, 9.0f, 16.0f, 25.0f, 100.0f};
    float expectedResults0[] = {2.0f, 3.0f, 4.0f, 5.0f, 10.0f};

    int numTests0 = sizeof(inputs0) / sizeof(inputs0[0]);

    for (int i = 0; i < numTests0; i++)
    {
        float input = inputs0[i];
        float expectedResult = expectedResults0[i];

        // call the function under test
        float result = sqrt_quick(input);

        // check if the result matches the expected result
        if (std::abs(result - expectedResult) < 0.1f)
        {
            std::cout << "Test case " << i + 1 << " passed!" << std::endl;
        }
        else
        {
            std::cout << "Test case " << i + 1 << " failed!" << std::endl;
            std::cout << "Expected: " << expectedResult << ", Got: " << result << std::endl;
        }
    }

    // test cases for sqrt_reciprocal_quick()
    float inputs1[] = {4.0f, 9.0f, 16.0f, 25.0f, 100.0f};
    float expectedResults1[] = {0.5f, .33f, .25f, .2f, 0.1f};

    int numTests1 = sizeof(inputs1) / sizeof(inputs1[0]);

    for (int i = 0; i < numTests1; i++)
    {
        float input = inputs1[i];
        float expectedResult = expectedResults1[i];

        // call the function under test
        float result = sqrt_reciprocal_quick(input);

        // check if the result matches the expected result
        if (std::abs(result - expectedResult) < 0.01f)
        {
            std::cout << "Test case " << i + 1 << " passed!" << std::endl;
        }
        else
        {
            std::cout << "Test case " << i + 1 << " failed!" << std::endl;
            std::cout << "Expected: " << expectedResult << ", Got: " << result << std::endl;
        }
    }

    return 0;
}
```

### DIFF
--- a/tools/sdk/esp32/include/esp-dl/include/math/dl_math.hpp
+++ b/tools/sdk/esp32/include/esp-dl/include/math/dl_math.hpp
@@ -38,8 +38,15 @@ namespace dl
          */
         inline float sqrt_quick(float x)
         {
-            const int result = 0x1fbb4000 + (*(int *)&x >> 1);
-            return *(float *)&result;
+            union
+            {
+                float f;
+                int i;
+            } pun;
+
+            pun.f = x;
+            pun.i = 0x1fbb4000 + (pun.i >> 1);
+            return pun.f;
         }
 
         /**
@@ -51,9 +58,10 @@ namespace dl
         inline float sqrt_reciprocal_quick(float x)
         {
             float xhalf = 0.5f * x;
-            int i = *(int *)&x;             // get bits for floating value
+            int i;
+            memcpy(&i, &x, sizeof(float));  // copy the bits of x into an integer
             i = 0x5f375a86 - (i >> 1);      // gives initial guess y0
-            x = *(float *)&i;               // convert bits back to float
+            memcpy(&x, &i, sizeof(float));  // copy the integer bits back into x
             x = x * (1.5f - xhalf * x * x); // Newton step, repeating increases accuracy
             return x;
         }
--- a/tools/sdk/esp32c3/include/esp-dl/include/math/dl_math.hpp
+++ b/tools/sdk/esp32c3/include/esp-dl/include/math/dl_math.hpp
@@ -38,8 +38,15 @@ namespace dl
          */
         inline float sqrt_quick(float x)
         {
-            const int result = 0x1fbb4000 + (*(int *)&x >> 1);
-            return *(float *)&result;
+            union
+            {
+                float f;
+                int i;
+            } pun;
+
+            pun.f = x;
+            pun.i = 0x1fbb4000 + (pun.i >> 1);
+            return pun.f;
         }
 
         /**
@@ -51,9 +58,10 @@ namespace dl
         inline float sqrt_reciprocal_quick(float x)
         {
             float xhalf = 0.5f * x;
-            int i = *(int *)&x;             // get bits for floating value
+            int i;
+            memcpy(&i, &x, sizeof(float));  // copy the bits of x into an integer
             i = 0x5f375a86 - (i >> 1);      // gives initial guess y0
-            x = *(float *)&i;               // convert bits back to float
+            memcpy(&x, &i, sizeof(float));  // copy the integer bits back into x
             x = x * (1.5f - xhalf * x * x); // Newton step, repeating increases accuracy
             return x;
         }
--- a/tools/sdk/esp32s2/include/esp-dl/include/math/dl_math.hpp
+++ b/tools/sdk/esp32s2/include/esp-dl/include/math/dl_math.hpp
@@ -38,8 +38,15 @@ namespace dl
          */
         inline float sqrt_quick(float x)
         {
-            const int result = 0x1fbb4000 + (*(int *)&x >> 1);
-            return *(float *)&result;
+            union
+            {
+                float f;
+                int i;
+            } pun;
+
+            pun.f = x;
+            pun.i = 0x1fbb4000 + (pun.i >> 1);
+            return pun.f;
         }
 
         /**
@@ -51,9 +58,10 @@ namespace dl
         inline float sqrt_reciprocal_quick(float x)
         {
             float xhalf = 0.5f * x;
-            int i = *(int *)&x;             // get bits for floating value
+            int i;
+            memcpy(&i, &x, sizeof(float));  // copy the bits of x into an integer
             i = 0x5f375a86 - (i >> 1);      // gives initial guess y0
-            x = *(float *)&i;               // convert bits back to float
+            memcpy(&x, &i, sizeof(float));  // copy the integer bits back into x
             x = x * (1.5f - xhalf * x * x); // Newton step, repeating increases accuracy
             return x;
         }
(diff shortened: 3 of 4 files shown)


mrengineer7777:
@pilinux This is one of the warnings thrown during CI.  I've been trying to figure out how to fix it.  I still don't understand what `const int result = 0x1fbb4000 + (*(int *)&x >> 1);` does, but your code fix looks promising!

atanisoft:
Because this modifies files under `tools/sdk` it can not be merged here but instead needs to be merged upstream in [ESP-DL](https://github.com/espressif/esp-dl) targeting [this](https://github.com/espressif/esp-dl/blob/master/include/math/dl_math.hpp) file.

> I still don't understand what `const int result = 0x1fbb4000 + (*(int *)&x >> 1);` does, but your code fix looks promising!

That's a rather complex way to type cast the float to an int and drop the first bit. At least that is how I read it. I think the proposed solution in this PR is a much cleaner way to do this. The only concern that may be there is if float and int end up diverging in byte size.

mrengineer7777:
@atanisoft Good catch on needing to modify the upstream file.

>>That's a rather complex way to type cast the float to an int and drop the first bit. At least that is how I read it. 

`>> 1` is a bit shift.  It effectively divides by 2.
I'm guessing the hard-coded numbers are the address of a table.  Thus this looks up pre-computed values in a flash table.

>>The only concern that may be there is if float and int end up diverging in byte size.

That's true.  We can add this bit of code to test for that:
_Static_assert(sizeof(int) == sizeof(float), L"This function will only work if flash and int are the same size");

pilinux:
@mrengineer7777 Ok, I created a new [pull request](https://github.com/espressif/esp-dl/pull/124).

`_Static_assert(sizeof(int) == sizeof(float), L"This function will only work if float and int are the same size");`

This one works at compiler-level which you can add in my previous test code sample. I got no error.

Today, I tested on esp32-c3 using arduino code:

```
#include <cmath>

inline float sqrt_quick(float x)
{
    //    const int result = 0x1fbb4000 + (*(int *)&x >> 1);
    //    return *(float *)&result;
    union
    {
        float f;
        int i;
    } pun;

    pun.f = x;
    pun.i = 0x1fbb4000 + (pun.i >> 1);
    return pun.f;
}

inline float sqrt_reciprocal_quick(float x)
{
    float xhalf = 0.5f * x;
    // int i = *(int *)&x;             // get bits for floating value
    int i;
    memcpy(&i, &x, sizeof(float));
    i = 0x5f375a86 - (i >> 1); // gives initial guess y0
    // x = *(float *)&i;               // convert bits back to float
    memcpy(&x, &i, sizeof(float));
    x = x * (1.5f - xhalf * x * x); // Newton step, repeating increases accuracy
    return x;
}

void setup()
{
    Serial.begin(115200);
    Serial.println("Test started...");
    delay(1000);

    Serial.print("Size of int: ");
    Serial.println(sizeof(int));
    Serial.print("Size of float: ");
    Serial.println(sizeof(float));

    // test cases for sqrt_quick()
    float inputs0[] = {4.0f, 9.0f, 16.0f, 25.0f, 100.0f};
    float expectedResults0[] = {2.0f, 3.0f, 4.0f, 5.0f, 10.0f};

    int numTests0 = sizeof(inputs0) / sizeof(inputs0[0]);

    for (int i = 0; i < numTests0; i++)
    {
        float input = inputs0[i];
        float expectedResult = expectedResults0[i];

        // call the function under test
        float result = sqrt_quick(input);

        // check if the result matches the expected result
        if (std::abs(result - expectedResult) < 0.1f)
        {
            Serial.print("Test case ");
            Serial.print(i + 1);
            Serial.println(" passed!");
        }
        else
        {
            Serial.print("Test case ");
            Serial.print(i + 1);
            Serial.println(" failed!");
            Serial.print("Expected: ");
            Serial.print(expectedResult);
            Serial.print(", Got: ");
            Serial.println(result);
        }
    }

    // test cases for sqrt_reciprocal_quick()
    float inputs1[] = {4.0f, 9.0f, 16.0f, 25.0f, 100.0f};
    float expectedResults1[] = {0.5f, .33f, .25f, .2f, 0.1f};

    int numTests1 = sizeof(inputs1) / sizeof(inputs1[0]);

    for (int i = 0; i < numTests1; i++)
    {
        float input = inputs1[i];
        float expectedResult = expectedResults1[i];

        // call the function under test
        float result = sqrt_reciprocal_quick(input);

        // check if the result matches the expected result
        if (std::abs(result - expectedResult) < 0.01f)
        {
            Serial.print("Test case ");
            Serial.print(i + 1);
            Serial.println(" passed!");
        }
        else
        {
            Serial.print("Test case ");
            Serial.print(i + 1);
            Serial.println(" failed!");
            Serial.print("Expected: ");
            Serial.print(expectedResult);
            Serial.print(", Got: ");
            Serial.println(result);
        }
    }
}

void loop()
{
}
```

And here's the output:

```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x1 (POWERON),boot:0xc (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5810,len:0x438
load:0x403cc710,len:0x91c
load:0x403ce710,len:0x25b0
entry 0x403cc710
Test started...
Size of int: 4
Size of float: 4
Test case 1 passed!
Test case 2 passed!
Test case 3 passed!
Test case 4 passed!
Test case 5 passed!
Test case 1 passed!
Test case 2 passed!
Test case 3 passed!
Test case 4 passed!
Test case 5 passed!
```

Thank you and @atanisoft for your time to check this pull request. If there is no further input from your side, I will close this pull request after 1-2 days 👍🏻

Issue #7903, opened 2023-02-28
XIAO ESP32-C3 Rainmaker example crash


MingyaoLiu:
### Board

Seeed XIAO ESP32-C3

### Device Description

XIAO ESP32

### Hardware Configuration

Nothing connected.

### Version

v2.0.7

### IDE Name

Arduino IDE

### Operating System

windows 11

### Flash frequency

80 or 40 tried both

### PSRAM enabled

yes

### Upload speed

921600

### Description

Loaded a rainmaker example from arduino IDE on XIAO esp32c3, build successfully, crashes on XIAO. I tried 2 examples one is RMakerCustom, one is RMakerSwitch, both crashes.

Tried wifi provision and some BLE examples, all fine.

### Sketch

```cpp
//This example demonstrates the ESP RainMaker with a standard Switch device.
#include "RMaker.h"
#include "WiFi.h"
#include "WiFiProv.h"

#define DEFAULT_POWER_MODE true
const char *service_name = "PROV_1234";
const char *pop = "abcd1234";

//GPIO for push button
static int gpio_0 = 9;
static int gpio_switch = 7;

/* Variable for reading pin status*/
bool switch_state = true;

//The framework provides some standard device types like switch, lightbulb, fan, temperaturesensor.
static Switch my_switch;

void sysProvEvent(arduino_event_t *sys_event) {
  switch (sys_event->event_id) {
    case ARDUINO_EVENT_PROV_START:
#if CONFIG_IDF_TARGET_ESP32S2
      Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on SoftAP\n", service_name, pop);
      printQR(service_name, pop, "softap");
#else
      Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on BLE\n", service_name, pop);
      printQR(service_name, pop, "ble");
#endif
      break;
    case ARDUINO_EVENT_PROV_INIT:
      wifi_prov_mgr_disable_auto_stop(10000);
      break;
    case ARDUINO_EVENT_PROV_CRED_SUCCESS:
      wifi_prov_mgr_stop_provisioning();
      break;
    default:;
  }
}

void write_callback(Device *device, Param *param, const param_val_t val, void *priv_data, write_ctx_t *ctx) {
  const char *device_name = device->getDeviceName();
  const char *param_name = param->getParamName();

  if (strcmp(param_name, "Power") == 0) {
[...]

PR #8243, opened 2023-05-23
fixes ESP32-C3 WiFiProv and btInUse()

SuGlider:
## Description of Change

This change is necessary within Arduino Core 2.0.x and 3.0.x.

When testing ESP32-C3 with WiFiProv.ino example to use Bluetooth provisioning, it fails because `btInUse()` returns false due to a weak/strong symbol resolving within the linker.

This PR sets the origin of the `btInUse()` function to be with `esp32-hal-bt.c` and its symbol is an external one for `esp32-hal-misc.c` making possible for the user to redefine it, is necessary. This solves the Core library C3 linker issue.

## Tests scenarios
The PR is tested by using `WiFiProv.ino` example by removing the `#if CONFIG_IDF_TARGET_ESP32` in order to use it with the C3.
This shall use this call: `    WiFiProv.beginProvision(WIFI_PROV_SCHEME_BLE, WIFI_PROV_SCHEME_HANDLER_FREE_BTDM, WIFI_PROV_SECURITY_1, "abcd1234", "Prov_123");`

Tested with ESP32, ESP32-C3, ESP32-S3.

## Related links
Fix https://github.com/espressif/arduino-esp32/issues/7903

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -209,9 +209,8 @@ bool verifyRollbackLater() { return false; }
 #endif
 
 #ifdef CONFIG_BT_ENABLED
-//overwritten in esp32-hal-bt.c
-bool btInUse() __attribute__((weak));
-bool btInUse(){ return false; }
+//from esp32-hal-bt.c
+extern bool btInUse();
 #endif
 
 void initArduino()
--- a/cores/esp32/esp32-hal-bt.c
+++ b/cores/esp32/esp32-hal-bt.c
@@ -16,7 +16,8 @@
 
 #ifdef CONFIG_BT_ENABLED
 
-bool btInUse(){ return true; }
+// user may want to change it to free resources
+__attribute__((weak)) bool btInUse(){ return true; }
 
 #include "esp_bt.h"
 


SuGlider:
@sanketwadekar - please test it and let me know. Thanks!

sanketwadekar:
@SuGlider, I have tested the Rainmaker Switch example which uses the WiFiProv library on esp32c3 and it works fine. Have you checked other weak symbols in esp32-hal-misc.c? Also, Arduino 2.0.9 is already released right? You might need to add this PR to the next release milestone.

Tagging @vikramdattu, @dhavalgujar

sanketwadekar:
Closes https://github.com/espressif/arduino-esp32/issues/7903

cores/esp32/esp32-hal-misc.c
```diff
@@ -209,9 +209,8 @@ bool verifyRollbackLater() { return false; }
 #endif
 
 #ifdef CONFIG_BT_ENABLED
-//overwritten in esp32-hal-bt.c
-bool btInUse() __attribute__((weak));
```
vikramdattu, line 213: @SuGlider I am wondering if this change is also applicable for other weak symbols in this file?

cc @sanketwadekar

SuGlider: I have verified it. 
The issue occurs when a weak and strong symbols are part of the same `core.a` library file. 
In the building process with Arduino, it first compiles all the Core Files and build a reusable Library.
The other weak functions are changed/redefined in the Sketch, therefore, its strong symbol are correctly linked.

The issue only happens with this very specific `btInUse()` function because its strong symbol is in the same `core.a` file and it is ignored by the RISC-V Linker.

SuGlider:
> @SuGlider, I have tested the Rainmaker Switch example which uses the WiFiProv library on esp32c3 and it works fine. 

Thanks for testing and confirming.

>Have you checked other weak symbols in esp32-hal-misc.c? 

Yes, all other symbols are OK. They can be defined in the Arduino Sketch and the strong symbols are linked correctly (xtensa and RV). The issue is very specific to symbols within the same Core Library (`core.a`) 

>Also, Arduino 2.0.9 is already released right? You might need to add this PR to the next release milestone.

Yes. We don't have yet a 2.0.10 version. 
I'll talk to @VojtechBartoska and @me-no-dev to check how to include this fix in the Arduino Core.

SuGlider:
@VojtechBartoska  - I just created the 2.0.10 TAG. I also added another PR to 2.0.10
https://github.com/orgs/espressif/projects/3/views/18?layout=board&filterQuery=milestone%3A%222.0.10%22+

PR #8247, opened 2023-05-24
Fixes HWCDC::end()

### LINKED ISSUE #8224: HWCDC::end() does not set tx_lock to NULL, causing crashes
### Board

ESP32-C3 / ESP32-S3

### Device Description

Any ESP32-C3 or -S3 using HWCDC

### Hardware Configuration

Any ESP32-C3 or -S3 board using HWCDC

### Version

v2.0.9

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

40MHz

### PSRAM enabled

yes

### Upload speed

115200

### Description

HWCDC::end() does not reset the semaphore.
So after a call to `end()` and a new call to `begin()` the `tx_lock` semaphore is invalid.
Thus trying to write something to that port again will cause a crash.

https://github.com/espressif/arduino-esp32/blob/3ec5f4efa1de4342aaab742008dc630091e5e035/cores/esp32/HWCDC.cpp#L190-L205

### Sketch

```cpp
-
```

### Debug Message

```plain
-
```

### Other Steps to Reproduce

_No response_

SuGlider:
## Description of Change

If we call `serial.end()` followed by `serial.begin` using JTAG/CDC interface, it will never initialize the semaphore `tx_lock` and it will reset the board when data is written.
There is a missing `tx_lock = NULL;` in the `HWCDC::end()` function.

## Tests scenarios
Tested with C3 and S3.

## Related links
Fixes #8224

### DIFF
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -195,6 +195,7 @@ void HWCDC::end()
     intr_handle = NULL;
     if(tx_lock != NULL) {
         vSemaphoreDelete(tx_lock);
+        tx_lock = NULL;
     }
     setRxBufferSize(0);
     setTxBufferSize(0);
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -28,7 +28,7 @@ ESP_EVENT_DEFINE_BASE(ARDUINO_HW_CDC_EVENTS);
 
 static RingbufHandle_t tx_ring_buf = NULL;
 static xQueueHandle rx_queue = NULL;
-static uint8_t rx_data_buf[64];
+static uint8_t rx_data_buf[64] = {0};
 static intr_handle_t intr_handle = NULL;
 static volatile bool initial_empty = false;
 static xSemaphoreHandle tx_lock = NULL;


mrengineer7777 (approved):
LGTM

TD-er:
Yep, very complex fix :)
LGTM.

I will also have a quick look at the other static values to see if we might have missed others too.

TD-er:
While we're at it, maybe also initialize the rx_data_buf to 0?
```c++
static uint8_t rx_data_buf[64]{};
```

I did a quick search for other calls of `vSemaphoreDelete` and there are more places where the pointer isn't set to NULL afterwards. This isn't an issue In destructors where the pointer used is a member.
But in `esp32-hal-tinyusb.c` it is a good candidate for future bugs when that function is later extended. (now it is a local variable inside that function and the function is left after this delete call)

SuGlider:
> While we're at it, maybe also initialize the rx_data_buf to 0?

Ok, done.

> I did a quick search for other calls of `vSemaphoreDelete` and there are more places where the pointer isn't set to NULL afterwards. This isn't an issue In destructors where the pointer used is a member. But in `esp32-hal-tinyusb.c` it is a good candidate for future bugs when that function is later extended. (now it is a local variable inside that function and the function is left after this delete call)

`esp32-hal-tinyusb.c` uses the semaphore internally only. 
It will never be extended because it is just used to check the installation of the JTAG driver.

I found  `vSemaphoreDelete` with "no NULL" in `HardwareSerial::~HardwareSerial()` and `TwoWire::~TwoWire()`, but those are the destructors and all object memory will be gone anyway. The construtor will set it to NULL.

 All other places are fine.

Issue #8231, opened 2023-05-19
RainMaker device callback issue


lishnih:
### Board

LilyGo T-Display-S3

### Device Description

LilyGo T-Display-S3

### Hardware Configuration

Amended esp-rainmaker\examples\switch example
gpio_switch = PIN_21

### Version

v2.0.9

### IDE Name

Arduino IDE

### Operating System

Windows 11

### Flash frequency

No idea

### PSRAM enabled

yes

### Upload speed

115200

### Description

I elaborate the esp-rainmaker\examples\switch example. Originally it works fine, but after adding a new device with his callback the switch callback become the same as new device callback.

### Sketch

```cpp
// This example demonstrates the ESP RainMaker with a standard Switch device.
#include "RMaker.h"
#include "WiFi.h"
#include "WiFiProv.h"
#include "AppInsights.h"

#define DEFAULT_POWER_MODE true
const char *service_name = "PROV_1234";
const char *pop = "abcd1234";

// GPIO for push button
static int gpio_0 = 0;
static int gpio_switch = PIN_21;

/* Variable for reading pin status*/
bool switch_state = true;

// The framework provides some standard device types like switch, lightbulb,
// fan, temperaturesensor.
static Switch *my_switch = NULL;


static Device *my_menu = NULL;


void sysProvEvent(arduino_event_t *sys_event)
{
    switch (sys_event->event_id) {
    case ARDUINO_EVENT_PROV_START:
        Serial.printf("\nProvisioning Started with name \"%s\" and PoP \"%s\" on SoftAP\n",
                      service_name, pop);
        printQR(service_name, pop, "softap");
        break;
    case ARDUINO_EVENT_PROV_INIT:
        wifi_prov_mgr_disable_auto_stop(10000);
        break;
    case ARDUINO_EVENT_PROV_CRED_SUCCESS:
        wifi_prov_mgr_stop_provisioning();
        break;
    default:;
    }
}

void write_callback(Device *device, Param *param, const param_val_t val,
                    void *priv_data, write_ctx_t *ctx)
{
    const char *device_name = device->getDeviceName();
    const char *param_name = param->getParamName();

    if (strcmp(param_name, "Power") == 0) {
        Serial.printf("Received value = %s for %s - %s\n",
                      val.val.b ? "true" : "false", device_name, param_name);
        switch_state = val.val.b;
        (switch_state == false) ? digitalWrite(gpio_switch, LOW)
[...]

PR #8249, opened 2023-05-25
rainmaker: multiple device callbacks not getting registered

sanketwadekar:
## Description of Change
This PR fixes an issue of multiple device callbacks (one callback per device) were not getting registered and invoked.
Consider an example where I create two devices, a switch and a fan; each of them having their own write callbacks. On controlling either switch/fan through Rainmaker app, the callback that got registered at last gets invoked. This is also seen in the issue reported [here](https://github.com/espressif/arduino-esp32/issues/8231)
## Test Scenarios
I have tested this PR by creating four devices; Switch, Lightbulb, Temp. sensor and Fan; each of them having their own callback. On controlling the device through app, respective callbacks were getting invoked.

## Related links
Closes https://github.com/espressif/arduino-esp32/issues/8231

### DIFF
--- a/libraries/RainMaker/src/RMakerDevice.cpp
+++ b/libraries/RainMaker/src/RMakerDevice.cpp
@@ -5,27 +5,31 @@
 static esp_err_t err;
 typedef void (*deviceWriteCb)(Device*, Param*, const param_val_t val, void *priv_data, write_ctx_t *ctx);
 typedef void (*deviceReadCb)(Device*, Param*, void *priv_data, read_ctx_t *ctx);
-
-void (*write_cb)(Device*, Param*, param_val_t, void*, write_ctx_t*);
-void (*read_cb)(Device*, Param*, void*, read_ctx_t*);
-Device device;
-Param param;
+typedef struct {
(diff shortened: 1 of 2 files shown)


sanketwadekar:
@shahpiyushv @vikramdattu PTAL

lishnih:
Good day,
I tried this commit but it's not successful for me: Core 1 panic'ed (InstructionFetchError). Exception was unhandled.

sanketwadekar:
> Good day,
> I tried this commit but it's not successful for me: Core 1 panic'ed (InstructionFetchError). Exception was unhandled.
> Maybe I can assist somehow?

@lishnih I tried this PR with the example you shared in your issue description and it worked fine. Please share your sketch that gives `InstructionFetchError` exception.

lishnih:
> @lishnih I tried this PR with the example you shared in your issue description and it worked fine. Please share your sketch that gives `InstructionFetchError` exception.

I withdraw my comment, I tried it again and found out that I defined Switch variable inside setup() therefore I got the exception. With global variable I have no issue.

PR #8274, opened 2023-05-31
M5CoreS3 board and partitions update

tobozo:
followup to #8161 

- adjusted default build values
    1) cdc default enabled
    2) 16MB flash + partition scheme copied from Core2 default selected
- added missing sections 
- added two custom partitions schemes to the variant folder

the new custom partitions layout is open for discussions.

[link to clean diff](https://github.com/espressif/arduino-esp32/pull/8274/files?w=1)

PR #8161, opened 2023-05-04
Create M5CoreS3 pins_arduino.h and board

Tinyu-Zhao:
M5CoreS3 related [documents](https://docs.m5stack.com/en/core/CoreS3)

PR #8276, opened 2023-05-31
M5CoreS3 board and partitions update

tobozo:
followup to #8161 #8274 

- adjusted default build values
    1) cdc default enabled
    2) 16MB flash + partition scheme copied from Core2 default selected
- added missing sections 
- added two custom partitions schemes to the variant folder ([see them in use](https://www.youtube.com/watch?v=zdIoBEfVvf4))

the new custom partitions layout is open for discussions.


Tinyu-Zhao:
CoreS3 PSRAM is best set to on by default, otherwise it will affect the normal work of the camera.
image

Tinyu-Zhao:
Hi @tobozo ,CoreS3 PSRAM is best set to on by default, otherwise it will affect the normal work of the camera.
image

me-no-dev:
> Hi @tobozo ,CoreS3 PSRAM is best set to on by default, otherwise it will affect the normal work of the camera.
> image

AMF you do not need that menu at all. Board comes with 32MB Flash and 8MB QSPI PSRAM. Clocks should be set to 80MHz and configuration options for those items should not exist.

tobozo:
@Tinyu-Zhao thanks for your input I'll remove the PSRAM block as suggested.

> locks should be set to 80MHz and configuration options for those items should not exist.

@me-no-dev Should I also remove the FlashMode block, or maybe just the `Flashmode.*.flash_freq` lines?

Tinyu-Zhao:
But in the absence of these configurations, Arduino will automatically select Flash Size as 4Mb and PSRAM is set to off, I think this default setting is a bit troublesome for the user to configure manually.

Tinyu-Zhao:
@tobozo I think the default setting for theUSB CDC On Boot: Disabled is a better choice, otherwise many examples and the kernel is used are USBSerial, there will be convenient not through the case.
image

tobozo:
@Tinyu-Zhao I understand it affects the example sketch, however changing the default option value in the boards.txt will not prevent users from changing it in the menu, and the error can happen anyway. 

Moreover, using `Serial` everywhere is much more intuitive, e.g when re-using sketches from other ESP32 products that do not handle USBCDC, or when the sketch is using a different library (M5Unified, LovyanGFX, TFT_eSPI, Arduino_GFX).

The official USBSerial example suggests using a [safeguard](https://github.com/espressif/arduino-esp32/blob/master/libraries/USB/examples/USBSerial/USBSerial.ino#L8-L14) to properly assign objects in a transparent fashion.

image

Some context: many ESP32-S3 board profiles enable CDC at boot: enabling this option solves many problems that occur in default situation e.g. debug logs not visible, late init, and no tty visible from the PC when in a crash loop.

In practice, [enabling USB-CDC at boot aliases Serial instead of USBSerial](https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/HWCDC.h#LL103C8-L103C13).
Since Serial and USBSerial are both hardware CDC, using `Serial.begin()` with USB-CDC is the same as using `USBSerial.begin()` without USB-CDC.

Please let me know what the M5 team decides and I'll comply.

P-R-O-C-H-Y:
@tobozo @Tinyu-Zhao Is this PR ready for reviewing / merging? :) 2.0.10 release is close. Thanks

Tinyu-Zhao:
@tobozo It is recommended to keep USBSerial as the serial port, as many users will not be able to compile after the update if the new release is not properly adapted.

tobozo:
@Tinyu-Zhao I've updated the branch accordingly, however I have some afterthoughts:

whatever choice is made, users will have troubles:

- without CDC on boot users may experience problems to flash CoreS3 when *not* using <M5CoreS3.h> library
- with CDC on boot user have to manually alias the USBSerial when using M5CoreS3 library

I think it is safe to assume the problem is isolated to M5CoreS3 library and could be adressed by patching M5CoreS3.h with the conditional aliasing, so that USBSerial is available in both situations.

me-no-dev:
I have modified the M5CoreS3 library locally to not use `USBSerial`, but a combination of `Serial` + `log_x`, in order to get best compatibility. `CDC_ON_BOOT` is ON. IMHO M5CoreS3 lib should be adjusted. @Tinyu-Zhao some of the drivers are missing there as well. Could those be added? Publishing the source of the `FactoryTest` default firmware could also be helpful!

VojtechBartoska:
any updates on this?

me-no-dev:
@tobozo I agree with the changes. Let's hope that @Tinyu-Zhao will also agree

tobozo:
@me-no-dev the CoreS3 User Demo project (the FactoryTest default firmware) can be found [here](https://github.com/imliubo/CoreS3-UserDemo)

tobozo:
all checks have passed but git complains, I'm not sure what to do with this, should I merge or rebase? 

image

Tinyu-Zhao:
I'll try it later.

me-no-dev:
What do we do with this PR now? We are about to release the last 2.x, so if we miss this, it will go for 3.0.0, where many things are different.

Tinyu-Zhao:
> What do we do with this PR now? We are about to release the last 2.x, so if we miss this, it will go for 3.0.0, where many things are different.

Is it too late to merge? It's been checked and it's fine.

me-no-dev:
Not too late!

PR #8282, opened 2023-06-01
Print full chip report when log level is sufficient

me-no-dev:
When the user selects log level INFO or above, full chip data will be printed. Example data below:

#### Info printed before setup() is called
```
=========== Before Setup Start ===========
Chip Info:
------------------------------------------
  Model             : ESP32-S3
  Package           : 0
  Revision          : 0
  Cores             : 2
  Frequency         : 240 MHz
  Embedded Flash    : No
  Embedded PSRAM    : No
  2.4GHz WiFi       : Yes
  Classic BT        : No
  BT Low Energy     : Yes
  IEEE 802.15.4     : No
------------------------------------------
INTERNAL Memory Info:
------------------------------------------
  Total Size        :   397676 B ( 388.4 KB)
  Free Bytes        :   365332 B ( 356.8 KB)
  Allocated Bytes   :    27328 B (  26.7 KB)
  Minimum Free Bytes:   360352 B ( 351.9 KB)
  Largest Free Block:   294900 B ( 288.0 KB)
------------------------------------------
SPIRAM Memory Info:
------------------------------------------
  Total Size        :  2097152 B (2048.0 KB)
  Free Bytes        :  2095104 B (2046.0 KB)
  Allocated Bytes   :        0 B (   0.0 KB)
  Minimum Free Bytes:  2095104 B (2046.0 KB)
  Largest Free Block:  2064372 B (2016.0 KB)
  Bus Mode          : QSPI
------------------------------------------
Flash Info:
------------------------------------------
  Chip Size         :  8388608 B (8 MB)
  Block Size        :    65536 B (  64.0 KB)
  Sector Size       :     4096 B (   4.0 KB)
  Page Size         :      256 B (   0.2 KB)
  Bus Speed         : 80 MHz
  Bus Mode          : QIO
------------------------------------------
Partitions Info:
------------------------------------------
                nvs : addr: 0x00009000, size:    20.0 KB, type: DATA, subtype: NVS
            otadata : addr: 0x0000E000, size:     8.0 KB, type: DATA, subtype: OTA
               app0 : addr: 0x00010000, size:  1280.0 KB, type:  APP, subtype: OTA_0
               app1 : addr: 0x00150000, size:  1280.0 KB, type:  APP, subtype: OTA_1
             spiffs : addr: 0x00290000, size:  1408.0 KB, type: DATA, subtype: SPIFFS
           coredump : addr: 0x003F0000, size:    64.0 KB, type: DATA, subtype: COREDUMP
------------------------------------------
Software Info:
------------------------------------------
  Compile Date/Time : Jun  2 2023 00:15:19
  Compile Host OS   : macosx
  ESP-IDF Version   : v5.1-dev-4734-g8a08cfe7d1
  Arduino Version   : 3.0.0
------------------------------------------
Board Info:
------------------------------------------
  Arduino Board     : ESP32S3_DEV
  Arduino Variant   : esp32s3
  Arduino FQBN      : espressif:esp32:esp32s3:JTAGAdapter=default,PSRAM=enabled,FlashMode=qio,FlashSize=4M,LoopCore=1,EventsCore=1,USBMode=hwcdc,CDCOnBoot=default,MSCOnBoot=default,DFUOnBoot=default,UploadMode=default,PartitionScheme=default,CPUFreq=240,UploadSpeed=921600,DebugLevel=verbose,EraseFlash=none
============ Before Setup End ============
```

#### Info printed after setup() is called
```
=========== After Setup Start ============
INTERNAL Memory Info:
------------------------------------------
  Total Size        :   397676 B ( 388.4 KB)
  Free Bytes        :   362340 B ( 353.8 KB)
  Allocated Bytes   :    29840 B (  29.1 KB)
  Minimum Free Bytes:   360352 B ( 351.9 KB)
  Largest Free Block:   294900 B ( 288.0 KB)
------------------------------------------
SPIRAM Memory Info:
------------------------------------------
  Total Size        :  2097152 B (2048.0 KB)
  Free Bytes        :  2094828 B (2045.7 KB)
  Allocated Bytes   :      260 B (   0.3 KB)
  Minimum Free Bytes:  2094828 B (2045.7 KB)
  Largest Free Block:  2064372 B (2016.0 KB)
------------------------------------------
GPIO Info:
------------------------------------------
                  0 : GPIO
                  8 : I2C_MASTER
                  9 : I2C_MASTER
                 11 : SPI_MASTER
                 12 : SPI_MASTER
                 13 : SPI_MASTER
============ After Setup End =============
```

### DIFF
--- a/cores/esp32/chip-debug-report.cpp
+++ b/cores/esp32/chip-debug-report.cpp
@@ -1,0 +1,346 @@
+#include "esp_heap_caps.h"
+#include "esp_chip_info.h"
+#include "esp_idf_version.h"
+#include "esp_arduino_version.h"
+#include "esp_rom_spiflash.h"
+#include "esp_flash.h"
+#include "esp_partition.h"
+#include "esp_app_format.h"
+#include "soc/efuse_reg.h"
+#include "soc/rtc.h"
+#include "soc/spi_reg.h"
+#if CONFIG_IDF_TARGET_ESP32S2
+#include "esp32s2/rom/spi_flash.h"
+#endif
+#include "esp_bit_defs.h"
+
+#include "Arduino.h"
+#include "esp32-hal-periman.h"
+
+#define chip_report_printf log_printf
+
+#define printMemCapsInfo(caps) _printMemCapsInfo(MALLOC_CAP_##caps, #caps)
+#define b2kb(b) ((float)b/1024.0)
+#define b2mb(b) ((float)b/(1024.0*1024.0))
+static void _printMemCapsInfo(uint32_t caps, const char * caps_str){
+  multi_heap_info_t info;
+  size_t total = heap_caps_get_total_size(caps);
+  heap_caps_get_info(&info, caps);
+  chip_report_printf("%s Memory Info:\n", caps_str);
+  chip_report_printf("------------------------------------------\n");
+  chip_report_printf("  Total Size        : %8u B (%6.1f KB)\n", total, b2kb(total));
+  chip_report_printf("  Free Bytes        : %8u B (%6.1f KB)\n", info.total_free_bytes, b2kb(info.total_free_bytes));
+  chip_report_printf("  Allocated Bytes   : %8u B (%6.1f KB)\n", info.total_allocated_bytes, b2kb(info.total_allocated_bytes));
+  chip_report_printf("  Minimum Free Bytes: %8u B (%6.1f KB)\n", info.minimum_free_bytes, b2kb(info.minimum_free_bytes));
+  chip_report_printf("  Largest Free Block: %8u B (%6.1f KB)\n", info.largest_free_block, b2kb(info.largest_free_block));
+}
+
+static void printPkgVersion(void){
+  chip_report_printf("  Package           : ");
+#if CONFIG_IDF_TARGET_ESP32
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_BLK0_RDATA3_REG, EFUSE_RD_CHIP_PACKAGE);
+  switch(pkg_ver){
+    case EFUSE_RD_CHIP_VER_PKG_ESP32D0WDR2V3: chip_report_printf("D0WD-R2-V3"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32D0WDQ6  : chip_report_printf("D0WD-Q6"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32D0WDQ5  : chip_report_printf("D0WD-Q5"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32D2WDQ5  : chip_report_printf("D2WD-Q5"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32U4WDH   : chip_report_printf("U4WD-H"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32PICOD4  : chip_report_printf("PICO-D4"); break;
+    case EFUSE_RD_CHIP_VER_PKG_ESP32PICOV302: chip_report_printf("PICO-V3-02"); break;
+  }
+#elif CONFIG_IDF_TARGET_ESP32S2
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SPI_SYS_3_REG, EFUSE_PKG_VERSION);
+  switch (pkg_ver) {
+    case 1: chip_report_printf("FH16"); break;
+    case 2: chip_report_printf("FH32"); break;
+    default: chip_report_printf("%lu", pkg_ver); break;
+  }
+#elif CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SPI_SYS_3_REG, EFUSE_PKG_VERSION);
+  chip_report_printf("%lu", pkg_ver);
+#elif CONFIG_IDF_TARGET_ESP32C2
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_BLK2_DATA1_REG, EFUSE_PKG_VERSION);
+  chip_report_printf("%lu", pkg_ver);
+#elif CONFIG_IDF_TARGET_ESP32H2
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SYS_4_REG, EFUSE_PKG_VERSION);
+  chip_report_printf("%lu", pkg_ver);
+#else
+  chip_report_printf("Unknown");
+#endif
+  chip_report_printf("\n");
+}
+
+static void printChipInfo(void){
+  esp_chip_info_t info;
+  esp_chip_info(&info);
+  chip_report_printf("Chip Info:\n");
+  chip_report_printf("------------------------------------------\n");
+  chip_report_printf("  Model             : ");
+  switch(info.model){
+    case CHIP_ESP32: chip_report_printf("ESP32\n"); break;
+    case CHIP_ESP32S2: chip_report_printf("ESP32-S2\n"); break;
+    case CHIP_ESP32S3: chip_report_printf("ESP32-S3\n"); break;
+    case CHIP_ESP32C2: chip_report_printf("ESP32-C2\n"); break;
+    case CHIP_ESP32C3: chip_report_printf("ESP32-C3\n"); break;
+    case CHIP_ESP32C6: chip_report_printf("ESP32-C6\n"); break;
+    case CHIP_ESP32H2: chip_report_printf("ESP32-H2\n"); break;
+    default: chip_report_printf("Unknown %d\n", info.model); break;
+  }
+  printPkgVersion();
+  chip_report_printf("  Revision          : ");
+  if(info.revision > 0xFF){
+    chip_report_printf("%d.%d\n", info.revision >> 8, info.revision & 0xFF);
+  } else {
+    chip_report_printf("%d\n", info.revision);
+  }
+  chip_report_printf("  Cores             : %d\n", info.cores);
+  rtc_cpu_freq_config_t conf;
+  rtc_clk_cpu_freq_get_config(&conf);
+  chip_report_printf("  Frequency         : %lu MHz\n", conf.freq_mhz);
+  chip_report_printf("  Embedded Flash    : %s\n", (info.features & CHIP_FEATURE_EMB_FLASH)?"Yes":"No");
+  chip_report_printf("  Embedded PSRAM    : %s\n", (info.features & CHIP_FEATURE_EMB_PSRAM)?"Yes":"No");
+  chip_report_printf("  2.4GHz WiFi       : %s\n", (info.features & CHIP_FEATURE_WIFI_BGN)?"Yes":"No");
+  chip_report_printf("  Classic BT        : %s\n", (info.features & CHIP_FEATURE_BT)?"Yes":"No");
+  chip_report_printf("  BT Low Energy     : %s\n", (info.features & CHIP_FEATURE_BLE)?"Yes":"No");
+  chip_report_printf("  IEEE 802.15.4     : %s\n", (info.features & CHIP_FEATURE_IEEE802154)?"Yes":"No");
+}
+
+static void printFlashInfo(void){
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
+  #define ESP_FLASH_IMAGE_BASE 0x1000
+#else
+  #define ESP_FLASH_IMAGE_BASE 0x0000
+#endif
+// REG_SPI_BASE is not defined for S3/C3 ??
+#if CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C3
+  #ifndef REG_SPI_BASE
+  #define REG_SPI_BASE(i)     (DR_REG_SPI1_BASE + (((i)>1) ? (((i)* 0x1000) + 0x20000) : (((~(i)) & 1)* 0x1000 )))
+  #endif // REG_SPI_BASE
+#endif // TARGET
+
+  chip_report_printf("Flash Info:\n");
+  chip_report_printf("------------------------------------------\n");
+  uint32_t hw_size = 1 << (g_rom_flashchip.device_id & 0xFF);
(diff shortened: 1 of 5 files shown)


cores/esp32/chip-debug-report.cpp
```diff
@@ -0,0 +1,338 @@
...
+  log_printf("  Cores             : %d\n", info.cores);
+  rtc_cpu_freq_config_t conf;
+  rtc_clk_cpu_freq_get_config(&conf);
+  log_printf("  Frequency         : %lu MHz\n", conf.freq_mhz);
+  log_printf("  Embedded Flash    : %s\n", (info.features & CHIP_FEATURE_EMB_FLASH)?"Yes":"No");
+  log_printf("  Embedded PSRAM    : %s\n", (info.features & CHIP_FEATURE_EMB_PSRAM)?"Yes":"No");
+  log_printf("  2.4GHz WiFi       : %s\n", (info.features & CHIP_FEATURE_WIFI_BGN)?"Yes":"No");
+  log_printf("  Classic BT        : %s\n", (info.features & CHIP_FEATURE_BT)?"Yes":"No");
+  log_printf("  BT Low Energy     : %s\n", (info.features & CHIP_FEATURE_BLE)?"Yes":"No");
+  log_printf("  IEEE 802.15.4     : %s\n", (info.features & CHIP_FEATURE_IEEE802154)?"Yes":"No");
```
pedrominatel, line 103: I think including the info "Ethernet: Yes/No" will also be good.

cores/esp32/chip-debug-report.cpp
```diff
@@ -0,0 +1,338 @@
...
+        }
+        log_printf("\n");
+      }
+      it = esp_partition_next(it);
+    }
+    //esp_partition_iterator_release(iterator);
+  }
+}
+
+static void printSoftwareInfo(void){
```
pedrominatel, line 199: I think it is worth printing MAC addresses for BLE, WiFi, and Ethernet when it is available.

pedrominatel (commented):
I think this will help a lot! Thanks @me-no-dev

cores/esp32/chip-debug-report.cpp
me-no-dev: why? how will that help us?

cores/esp32/chip-debug-report.cpp
me-no-dev: this info comes from the efuses, while other "features" like available peripherals are something very different. We know that only ESP32 has ethernet. This info is more for us than to tell the user what they have on-board.

cores/esp32/chip-debug-report.cpp
pedrominatel: It's just because sometimes people don't know how to print the MAC or get confused with the 3 MACs.
If you think this info is useless for us then it's ok.

me-no-dev: I would rather not collect such unique information and I do not think that it will help us troubleshooting anything. If users can't get MAC addresses themselves, we can help separately.

pedrominatel (approved):
LGTM

cores/esp32/chip-debug-report.cpp
```diff
@@ -0,0 +1,338 @@
...
+  }
+#elif CONFIG_IDF_TARGET_ESP32S2
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SPI_SYS_3_REG, EFUSE_PKG_VERSION);
+  switch (pkg_ver) {
+    case 1: log_printf("FH16"); break;
+    case 2: log_printf("FH32"); break;
+    default: log_printf("%lu", pkg_ver); break;
+  }
+#elif CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6
+  uint32_t pkg_ver = REG_GET_FIELD(EFUSE_RD_MAC_SPI_SYS_3_REG, EFUSE_PKG_VERSION);
```
TD-er, line 57: Is there any documentation on what this pkg_ver is telling us for the S3/C3/etc?
I couldn't find anything yesterday.

me-no-dev: no and they all return 0 so far. It will probably make more sense in the future. Chips are too new

TD-er: Well the S3 already has seen [quite a few iterations and versions](https://espeasy.readthedocs.io/en/latest/ESPEasy/ESPchips.html#quad-octal-spi-mode).
So I find it very strange (frustrating) that there seems to be no way to read from some efuses whether the chip has embedded flash/PSRAM, its size and how it is wired, without actually access either flash or PSRAM and hopefully using the correct nr. of lanes.

me-no-dev: you must have skipped the top of the log?

TD-er: It does indeed show whether flash or PSRAM is embedded, but you can't simply see whether octal or quad mode is needed for PSRAM.
Thus it seems impossible to detect which build might be needed when initially flashing the board.

I have 2 boards as an example:
- (A) LilyGo T-dongle, which has 16M quad flash, no PSRAM
- (B) S3 "Wemos dev pro" with 16M quad flash, 8MB octal PSRAM

Flashing a build for quad flash, quad PSRAM will work fine on board A, which doesn't have PSRAM , but on  board B with 8M PSRAM you can't detect it. (and it isn't embedded in the chip, thus the "embedded PSRAM" efuse isn't set)

And when flashing a build with quad flash, octal PSRAM to board A which hasn't (such) PSRAM, you end up with a bootloop.

N.B. this board I have with 16M quad flash and 8M PSRAM (not embedded) does have an official Espressif module, so it would have been nice if Espressif did mark with some efuses how to recognize those features before flashing.

This makes it quite a trial-and-error run to properly detect what features are present on a board.

And you might say, well I don't need the PSRAM, so it doesn't matter, but using GPIO33-37 will still be wired to the PSRAM, thus using them for something else might have side-effects.

You can still read the efuses to try figuring out whether these specific pins are 'powered' via either the internal LDO or by the ESP's Vdd. But this is quite complex as you can also have override efuses active to ignore the bootstrap pin GPIO 45(??) which controls this internal voltage and I'm not 100% sure whether reading those states is actually reading an efuse, or the actual state of the power supply for GPIO 33-37 and thus maybe related to having PSRAM support enabled, etc.

TL;DR
I still think it could have been done a lot simpler by Espressif...

me-no-dev: There is an efuse set for octal PSRAM and you will get an error when booting in wrong mode. This dump is printed AFTER everything is setup and after any error from PSRAM.

TD-er: > There is an efuse set for octal PSRAM

Which one?
That would be of great help to let a user know there might be a better suited build available if it can be detected what mode to use.
Use case: I'm thinking of something very similar to what Tasmota is doing with their "Safeboot" builds.
So if that build could simply detect the presence of PSRAM (or at least the required SPI mode) without crashing, then it would be a lot more user friendly compared to flashing and hoping it won't end up in a bootloop.

me-no-dev: hmm I am not finding it in the lists... maybe it was for flash only, though I do not find that one either.

ESP32-S3 so far comes with two PSRAM options:
- 2MB QSPI PSRAM
- 8MB OPI PSRAM

So if you check the markings on the module and you see NxxR8, that means you have OPI PSRAM, R2 will be QSPI

TD-er: > hmm I am not finding it in the lists... maybe it was for flash only, though I do not find that one either.

This is what I'm using so far for the OPI flash wired efuse.
Not sure how useful it is, as it will probably not boot when the wrong one was set.
But esptool.py could perhaps at some point do a check for it...

Label in the datasheet is "flash type" if I remember correctly.

```c++
bool getFlashChipOPI_wired() {
  #ifdef ESP32_CLASSIC
    return false;

  #else
  // Source: https://github.com/espressif/esptool/commit/b25606b95920bd06df87aff9202c7a15377d4a30
  const uint32_t data = REG_GET_BIT(EFUSE_RD_REPEAT_DATA3_REG, BIT(9));
  return data != 0;
  #endif
}
```

me-no-dev: 5.1 supports flash auto-detection and that is turned on for ESP32S3. Should boot either way. The same technique is not possible for PSRAM

TD-er: > 5.1 supports flash auto-detection and that is turned on for ESP32S3. Should boot either way.

That's good to know, will at least save a lot of hassle.

I guess there is some other checking being done when accessing the PSRAM as I've also seen some logs showing a bit more info on the detected PSRAM like brand (??).
So would be really nice if there could still be some runtime check to be found for trying PSRAM SPI mode. It would already be useful if it were only for embedded PSRAM as those are more likely to already have an efuse set indicating it has PSRAM.

mrengineer7777:
Intriguing. Looks helpful for issue reports. Would it make more sense to print at DEBUG level rather than INFO?

cores/esp32/chip-debug-report.cpp
TD-er: I have some ESP32-S2 unit here, which does have 2M of embedded PSRAM and 4MB embedded flash.
However the chip features flags do not mention embedded flash nor embedded PSRAM.

<html>
<body>


ESP Chip ID: | 5609182 (0x5596DE)
-- | --
ESP Chip Frequency: | 240 MHz
ESP Crystal Frequency: | 40 MHz
ESP APB Frequency: | 80 MHz
ESP Chip Model: | ESP32-S2FNR2
ESP Chip Features: | Wi-Fi bgn
ESP Chip Revision: | 0.0
ESP Chip Cores: | 1


</body>
</html>

The `ESP Chip Features` string is generated with this code:

```c++
    String features;

    if (getChipFeatures().wifi_bgn) { features += F("Wi-Fi bgn / "); }
    if (getChipFeatures().bluetooth_ble) { features += F("BLE / "); }
    if (getChipFeatures().ieee_802_15_4) { features += F("IEEE 802.15.4 / "); }
    if (getChipFeatures().embeddedFlash) { features += F("Emb. Flash / "); }
    if (getChipFeatures().embeddedPSRAM) { features += F("Emb. PSRAM"); }
    features.trim();

    if (features.endsWith(F("/"))) { features = features.substring(0, features.length() - 1); }
    addHtml(features);
```

```c++
struct esp32_chip_features {
  bool embeddedFlash{};
  bool wifi_bgn{};
  bool bluetooth_ble{};
  bool bluetooth_classic{};
  bool ieee_802_15_4{};
  bool embeddedPSRAM{};
};

esp32_chip_features getChipFeatures() {
  static esp32_chip_features res;
  static bool loaded = false;
  if (!loaded) {
    esp_chip_info_t chip_info;
    esp_chip_info(&chip_info);

    res.embeddedFlash = chip_info.features & CHIP_FEATURE_EMB_FLASH;
    res.wifi_bgn      = chip_info.features & CHIP_FEATURE_WIFI_BGN;
    res.bluetooth_ble = chip_info.features & CHIP_FEATURE_BLE;
    res.bluetooth_classic = chip_info.features & CHIP_FEATURE_BT;
    res.ieee_802_15_4 = chip_info.features & CHIP_FEATURE_IEEE802154;
    res.embeddedPSRAM = chip_info.features & CHIP_FEATURE_EMB_PSRAM;

    loaded = true;
  }
  return res;
}
```

This is one of those purple "Wemos" modules with the bare chip on the PCB. So both flash and PSRAM are for sure embedded in the chip and thus Espressif should have burnt those efuse bits, right?

In other words, those `chip_info.features` may not represent the correct state.

me-no-dev:
> Intriguing. Looks helpful for issue reports. Would it make more sense to print at DEBUG level rather than INFO?

that is one conversation to be had. Not sure what is the best option here.

cores/esp32/chip-debug-report.cpp
me-no-dev: > In other words, those chip_info.features may not represent the correct state.

I can not comment on that. I'm just dumping what ESP-IDF provides.

atanisoft: > In other words, those `chip_info.features` may not represent the correct state.

In most cases that I've seen where they don't match up they are using a SoC that doesn't have anything embedded and instead uses external flash / psram. I have yet to find any Espressif branded modules that exhibit different behaviors.

SuGlider:
> > Intriguing. Looks helpful for issue reports. Would it make more sense to print at DEBUG level rather than INFO?
> 
> that is one conversation to be had. Not sure what is the best option here.

I think that the PR is very complete, considering what is possible to be reported.
My suggestion goes about making it available, by default, within VERBOSE or DEBUG level only (not INFO).

-----------

Another suggestion would be about allowing any beginner users to activate it by just adding a `REPORT_ENABLED(true)` (or something similar) into the Sketch. It would force issuing the report, independently of the `Core debug Level` selected.
This may be useful for us, when investigating an open issue, to just ask this user to add this declaration in the sketch and then ask reporting back the Serrial output to us.

To achieve it, we could create some weak function that, by default, doesn't issue the report. 
When the user includes the declaration, it changes the behaviour to always issue the report.

cores/esp32/main.cpp
```diff
@@ -9,6 +9,10 @@
 #endif
 #endif
```
SuGlider, line 11: This is my suggestion about having a weak function:

``` cpp
__attribute__((weak)) bool chipReportEnabled(void) {
    return false;
}
#define REPORT_ENABLED(cr_enabled) bool chipReportEnabled(void) { return cr_enabled; }

```

cores/esp32/main.cpp
```diff
@@ -39,7 +43,13 @@ __attribute__((weak)) size_t getArduinoLoopTaskStackSize(void) {
 
 void loopTask(void *pvParameters)
 {
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO
+    printBeforeSetupInfo();
+#endif
     setup();
```
SuGlider, line 49: Adding the effect for `REPORT_ENABLED(true)` option here:

``` cpp
if (chipReportEnabled()) printBeforeSetupInfo();
```

cores/esp32/main.cpp
```diff
@@ -39,7 +43,13 @@ __attribute__((weak)) size_t getArduinoLoopTaskStackSize(void) {
...
+#if ARDUHAL_LOG_LEVEL >= ARDUHAL_LOG_LEVEL_INFO
+    printAfterSetupInfo();
+#endif
     for(;;) {
```
SuGlider, line 53: and here:

``` cpp
if (chipReportEnabled()) printAfterSetupInfo();
```

SuGlider (commented):
@me-no-dev - I have suggested a possible improvement.

PR #8301, opened 2023-06-09
Added support for M5Stamp Pico from M5Stack

MWolfahrt:
## Description of Change
Added support for the M5Stamp Pico from M5Stack.
See here for description: https://docs.m5stack.com/en/core/stamp_pico

## Tests scenarios
Tested the implementation on my PlatformIO locally.
Blink program compiled uploaded and work correctly.

## Related links
[https://docs.m5stack.com/en/core/stamp_pico](https://docs.m5stack.com/en/core/stamp_pico)


P-R-O-C-H-Y (approved):
LGTM

Issue #8318, opened 2023-06-15
Possible bug WebServer::handleClient method


supersjellie:
### Board

ESP32C3

### Device Description

Different custom boards communication by digital pins and/or I2C

### Hardware Configuration

Switches/sensors

### Version

latest master (checkout manually)

### IDE Name

Arduino IDE

### Operating System

Windows`10

### Flash frequency

80

### PSRAM enabled

yes

### Upload speed

115200

### Description

About 5 to 10% of the calls to the (default) webserver fails. The issue also occurs on the ESP8266 Arduino (same code for both devices). So this is a copy of https://github.com/esp8266/Arduino/issues/8941 . Making sure the same object is used solves it (not even a single timeout anymore in the (calling) openHAB log. Details check above link.

PR : https://github.com/espressif/arduino-esp32/pull/8319

@earlephilhower : as promised, here's an issue and PR for the ESP32.

### Sketch

```cpp
Details check above link.
```

### Debug Message

```plain
Details check above link.
```

### Other Steps to Reproduce

Details check above link.

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

Parsaabasi:
Hello, 

Due to the overwhelming volume of issues currently being addressed, we have decided to close the previously received tickets. If you still require assistance or if the issue persists, please don't hesitate to reopen the ticket.

Thanks.

Jason2866:
Closing since the related PR https://github.com/espressif/arduino-esp32/pull/8319 is merged

PR #8319, opened 2023-06-15
Fix for timeout issues on WebServer.cpp

supersjellie:
## Description of Change
Removing client temporary variable. Arduino doesn't like copying objects. And for the logisch de class variable _currentClient can be used from the start of the method

## Tests scenarios
My openHAB server calls 3 ESP32 devices every minute. In the openHAB log about 5-10% fails on a timeout (3000 ms is set timeout on openHAB). With above code modification none (running for 3 days now) of the calls failed.

## Related links
https://github.com/espressif/arduino-esp32/issues/8318

(same issue and resolution for ESP8266, check https://github.com/esp8266/Arduino/issues/8941 )

### DIFF
--- a/libraries/WebServer/src/WebServer.cpp
+++ b/libraries/WebServer/src/WebServer.cpp
@@ -285,17 +285,16 @@ void WebServer::serveStatic(const char* uri, FS& fs, const char* path, const cha
 
 void WebServer::handleClient() {
   if (_currentStatus == HC_NONE) {
-    WiFiClient client = _server.available();
-    if (!client) {
+    _currentClient = _server.available();
+    if (!_currentClient) {
       if (_nullDelay) {
         delay(1);
       }
       return;
     }
 
-    log_v("New client: client.localIP()=%s", client.localIP().toString().c_str());
+    log_v("New client: client.localIP()=%s", _currentClient.localIP().toString().c_str());
 
-    _currentClient = client;
     _currentStatus = HC_WAIT_READ;
     _statusChange = millis();
   }


SuGlider:
@supersjellie @me-no-dev 

I have tested this PR and it works with no problems.
I used the `WebServer` `HelloServer.ino` example to test `server.handleClient();`
I couldn't notice any difference from the original code to the PR code.
I get a few errors in the WiFi Client in both cases:

```
[ 16123][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 26460][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 26463][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 36087][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 46242][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 56364][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
[ 66167][D][WiFiClient.cpp:546] connected(): Disconnected: RES: 0, ERR: 128
```

supersjellie:
> I couldn't notice any difference from the original code to the PR code.

No differences in response time when your refresh 'like crazy'? In my setup (of course I use a bigger/more complicated sketch than the helloserver.ino) the difference is large. Most requests are the same but with the old code about 5-10% had large response times. And with the new code this disappeared.

SuGlider:
>What would mean that for the ESP32 this code modification is more cosmetic, saving an unneeded variable and keeping it in synch with the ESP8266 code.

Yes, this PR may save a few cycles and some stack allocation in execution time.
@me-no-dev  - It is an improvement, anyway, therefore, I think that we should merge it.

SuGlider (approved):

PR #8327, opened 2023-06-18
Adds HardwareSerial to the Peripheral Manager

SuGlider:
## Description of Change

Adds UART to the Peripheral Manager. 
Moved all UART GPIO information to ESP32-HAL layer.

`Serial.end()` will detach all UART GPIOs.
When another Peripheral uses any UART GPIO (Rx/Tx/Rts/Cts), the UART driver is terminated and all the GPIOs are detached.

## Tests scenarios
Tested with ESP32 and ESP32S3.

## Related links
None.


SuGlider:
Closed due to using a wrong branch -- too many commits.

PR #8351, opened 2023-06-26
Add checks for SOC defines

P-R-O-C-H-Y:
## Description of Change
Added checks to drivers/libraries for SoC defines, if peripheral or feature is supported by chip.


me-no-dev:
Missing SPI, Wire and the rest of the USB classes in libraries

PR #8359, opened 2023-06-27
Fixes RMT filter & idle timing and setup

SuGlider:
## Description of Change
Initial values of Filter and IDLE time for RMT RX functions were not working correctly.
This PR fixes it and sets it to minimum and maximum limits by default.
It also verifiy the values before setting them to make warn possible problems in the console.

## Tests scenarios
Tested with ESP32S3 using a HC-SR04 sketch

``` cpp
const uint8_t TRIG_PIN = 11;
const uint8_t ECHO_PIN = 12;

uint32_t getEcho(){
  rmt_data_t trig_pulse = {
    .duration0 = 10,
    .level0 = 1,
    .duration1 = 10,
    .level1 = 0
  };
  if(!rmtWriteAsync(TRIG_PIN, &trig_pulse, 1)){
    Serial.println("Send Trig Failed!");
    return 0;
  }
  
  rmt_data_t echo_pulse;
  size_t received = 1;
  if(!rmtRead(ECHO_PIN, &echo_pulse, &received, 1000) || !received){
    Serial.println("Read Echo Failed!");
    return 0;
  }
  Serial.printf("  P[0]: ticks: %u, level: %u\n", echo_pulse.duration0, echo_pulse.level0);
  Serial.printf("  P[1]: ticks: %u, level: %u\n", echo_pulse.duration1, echo_pulse.level1);

  if(echo_pulse.level0){
    return echo_pulse.duration0;
  }
  return echo_pulse.duration1;
}

void setup() {
  Serial.begin(115200);
  Serial.setDebugOutput(true);
  rmtInit(TRIG_PIN, RMT_TX_MODE, RMT_MEM_NUM_BLOCKS_1, 500000);
  rmtInit(ECHO_PIN, RMT_RX_MODE, RMT_MEM_NUM_BLOCKS_1, 500000);
}

void loop() {
  float cm = (0.034 * getEcho());
  Serial.print("Distance: ");
  Serial.print(cm, 3);
  Serial.println(" cm");
  delay(1000);
}
```

## Related links
None

### DIFF
--- a/cores/esp32/esp32-hal-rmt.c
+++ b/cores/esp32/esp32-hal-rmt.c
@@ -16,6 +16,7 @@
 #include "driver/gpio.h"
 #include "driver/rmt_tx.h"
 #include "driver/rmt_rx.h"
+#include "hal/rmt_ll.h"
 
 #include "esp32-hal-rmt.h"
 #include "esp32-hal-periman.h"
@@ -55,6 +56,7 @@ struct rmt_obj_s {
   EventGroupHandle_t rmt_events;             // read/write done event RMT callback handle
   bool rmt_ch_is_looping;                    // Is this RMT TX Channel in LOOPING MODE?
   size_t *num_symbols_read;                  // Pointer to the number of RMT symbol read by IDF RMT RX Done
+  uint32_t frequency_Hz;                     // RMT Frequency
 
 #if !CONFIG_DISABLE_HAL_LOCKS
   xSemaphoreHandle g_rmt_objlocks;           // Channel Semaphore Lock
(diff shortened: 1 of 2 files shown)


me-no-dev:
I think that combining both filter and rx threshold functions into one will make most sense. Maybe something like `setRxFilter(int pin, uint16_t min_ticks, uint16_t max_ticks)` would make most sense and also argument in nanoseconds is not the best option for Arduino users IMHO. Using ticks is better.

SuGlider:
@me-no-dev - Done.

PR #8360, opened 2023-06-27
Add TWAI transmit example

designer2k2:
## Description of Change
Example showing how to transmit messages using the TWAI interface and a CAN transceiver.
It also uses the alerts to show how to set them up and react to them.

This is based on the receive example, adopted with the example from the api-reference for transmission: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html#message-transmission

## Tests scenarios
I used an esp32-c3-devkit (ESP32-D0WDQ6) with a SN65HVD230 CAN transceiver.
This was attached between 2 Arduino Nano´s with a MCP2515/MCP2562 CAN interface, whereby they replayed a recording from a quite busy CAN bus.
The ESP32 is used to additionally transmit messages with this example into the bus and the output was observed.

Caution, when only another ESP32 is used to receive by the TWAI receive example the TWAI_MODE_LISTEN_ONLY  has to be replaced with TWAI_MODE_NORMAL, otherwise there is no confirmation and the transmit example will fail.

## Related links

The TWAI receive example: https://github.com/espressif/arduino-esp32/pull/7430

### DIFF
--- a/libraries/ESP32/examples/TWAI/TWAItransmit/TWAItransmit.ino
+++ b/libraries/ESP32/examples/TWAI/TWAItransmit/TWAItransmit.ino
@@ -1,0 +1,123 @@
+/* ESP32 TWAI transmit example.
+  This transmits a message every second.
+
+  Connect a CAN bus transceiver to the RX/TX pins.
+  For example: SN65HVD230
+
+  The API gives other possible speeds and alerts:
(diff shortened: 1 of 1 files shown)


SuGlider:
@designer2k2 - Thank you for the example!

PR #8474, opened 2023-08-01
Create pins_arduino.h for Lionbit S3 STEM Dev Board

PradeepKotu:
lionbit s3 pins pins_arduino.h file

I am writing to request the inclusion of the remarkable new development board, the "Lionbit S3 STEM Dev Board," into the official board files. this is the pins_arduino.h file. 

## Description of Change
lionbit s3 pins pins_arduino.h file for variant 

## Tests scenarios
we have tested all pins works fine

## Related links
lionbit.cc


PradeepKotu:
@CLAassistant  is there any issue with this request?

Issue #8502, opened 2023-08-08
Partition tables, CSV in Variants folder, file not found


WanaGo:
### Board

ESP32-S3 DEV

### Device Description

Using a custom module, however its the same for the ESP32-S3 DEV

### Hardware Configuration

N/A

### Version

latest master (checkout manually)

### IDE Name

Arduino IDE 2.1.1

### Operating System

Windows 11

### Flash frequency

80Mhz

### PSRAM enabled

yes

### Upload speed

921600

### Description

https://github.com/espressif/arduino-esp32/blob/master/docs/source/tutorials/partition_table.rst

This tutorial describes how to set up a Partition scheme CSV file, however it does not touch on at all how to put the CSV's into the Variants folder inside the core, rather than in the tools/partitions/ folder.

Wanting to submit a pull request soon to add our custom board into the boards.txt etc, however we just cannot get the partitions working out of the variants folder, it fails to compile as it cannot fine 'partitions.csv'.
Doing the exact same thing with an Adafruit board, which has its CSV file in its Variants folder, the compile looks for its CSV file in the variants folder, and copies it to the temp folder partitions.csv just fine - but why ours doesnt, I have no idea.

I dont know what I am missing.

I have edited the boards.txt file, and added in the partition listings.
I have created my board variant folder, which is referenced in the boards.txt file.

When it goes to compile, it says it cant find the partitions.csv file though.

When I select say the Adafruit MagTag board, which is set up similar for partitions, its CSV file is in its variants folder, nothing too different in its boards.txt file, but when that compiles, the compile text is quite different and it specifically looks at the Variants folder and the specific CSV selected in the Partition scheme menu, and copies that as is required to make the partitions.csv file.

Mine does not do that, it never looks in the variants folder for the CSV file, to copy over, and I have no idea why. I must be missing something, but I just cant figure out what it is.

If I have the partition csv files in the tools/partitions/ folder, it works fine - however that is not where it 'should' be I think, as I want to eventually submit this to merge into the ESP32 core, and it needs to then be in the Variants folder for my custom board.

Does anyone have any idea what is missing?

Error is:

```
FQBN: esp32:esp32:xxx-ESP32-S3R8n16
Using board 'xxx-ESP32-S3R8n16' from platform in folder: C:\Users\xxx\AppData\Local\Arduino15\packages\esp32\hardware\esp32\2.0.9
[...]

PR #8507, opened 2023-08-09
Update partition_table.rst with Variants folder detail, and new 3rd Party Board

James-4DSystems:
## Description of Change
1st - Adding missing detail regarding custom Partitions being added under the variants folder, as the tutorial only covers having the csv in your sketch folder, or in the tools/partitions folder.
This information is critical for anyone designing commercial systems who wants to integrate different partitions to those which are standard, as the Sketch folder or tools/partitions are not really valid options in this scenario, as they are specific to a companies hardware / variant.

2nd - Adding in new 3rd Party Boards from 4D Systems.

## Tests scenarios
General testing on this Tutorial information with own boards. 
More detail likely could be added for the Tutorial, however this is the core information that was missing and could not be found elsewhere.
Testing on the boards addition using our new hardware, has been sucessful.

## Related links
https://github.com/espressif/arduino-esp32/issues/8502

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -24112,3 +24112,160 @@ nebulas3.menu.EraseFlash.all=Enabled
 nebulas3.menu.EraseFlash.all.upload.erase_cmd=-e
 
 ##############################################################
+
+gen4-ESP32-S3R8n16.name=4D Systems gen4-ESP32 16MB Modules (ESP32-S3R8n16)
+gen4-ESP32-S3R8n16.vid.0=0x303a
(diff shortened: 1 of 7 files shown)


James-4DSystems:
@VojtechBartoska - Hello, noticed you were the one progressing other 3rd party additions, are you able to assist us with ours, so we can make it for the next release? If this Pull needs to be split up so the Tutorial is separate, please can you do that? Thank you in advance.

James-4DSystems:
Deleted and raised 2 separate PR's based on 2 different branches

PR #8509, opened 2023-08-10
ledc.c: Fix analogWrite() last channel available verification

rlipas:
## Description of Change
The last channel allocated is number 0, which conflicted with the value given to an uninitialized pin, giving the "No more analogWrite channels available!" error when trying to use it.

Pins are now given the value -1 to indicate that they are not used so channel 0 can be used without errors.

## Tests scenarios
Tested on a custom board with ESP32-S3-WROOM-1 module (8 ledc channels available.)

### DIFF
--- a/cores/esp32/esp32-hal-ledc.c
+++ b/cores/esp32/esp32-hal-ledc.c
@@ -210,7 +210,7 @@ uint32_t ledcChangeFrequency(uint8_t chan, uint32_t freq, uint8_t bit_num)
     return ledc_get_freq(group,timer);
 }
 
-static int8_t pin_to_channel[SOC_GPIO_PIN_COUNT] = { 0 };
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y (approved):
Nice catch, TY @rlipas

me-no-dev:
This would not work though, because only the first element of the `pin_to_channel` array will be `-1`. The rest will be `0`. I suggest to keep zero, but use the channel number incremented by 1, so that if `pin_to_channel[pin]` equals `1`, then channel is `0`, and if it equals `5`, then channel would be `4`, so on. This would ensure that all channels are used and that `0` is handled as unused channel

rlipas:
Oof... I totally forgot how array initializations worked :sweat_smile: 
I fixed this by incrementing the channel as suggested. Sorry for the mix up.
Tomorrow I'll have access to an oscilloscope and I'll be able to properly test this.

PR #8517, opened 2023-08-12
fix flash mode read out for C6

Jason2866:
Currently the flash mode for the C6 always is `QOUT`. This is fixed by the PR. Now the correct flash mode is returned
@me-no-dev


me-no-dev:
Thanks @Jason2866

Issue #3546, opened 2019-12-08
When FFat is mounted before SD, then filesystem reports wrong system sizes


ullix:
I use an ESP32 in Arduino environment (more precise: UECIDE). I use both the internal Flash (about 1.4MB) with FFat and an external SD card (32GB) with SD. 

When I first mount the Flash with FFat, and then the SD with SD, all calls to totalBytes and freeBytes, usedBytes, resp., show only the values valid for FFat. 

When I reverse the order of mounting, so doing SD first, then all comes out correct.

(Note: FFat can't do usedBytes,  SD can't do freeBytes.)

```
#include "FS.h"
#include "FFat.h"
#include "SD.h"

void setup(){
    Serial.begin(115200);
    uint64_t totalB;    // total Bytes
    uint64_t freeB;     // free Bytes
    uint64_t usedB;     // used Bytes = total Bytes - free Bytes    
    uint64_t cardSize;  // only for SD: size of card

    //************ First ******************************************
    Serial.println("\nMounting FFat Flash");
    if (!FFat.begin(false)) {     
        Serial.println("An Error has occurred while mounting FFAT");
        return;
    }
    // end first **************************************************

    //************ Second *****************************************
    Serial.println("\nMounting SD card");
    if(!SD.begin()){
        Serial.println("Card Mount Failed");
        return;
    }
    //end second***************************************************
    
    Serial.println();
    
    totalB = SD .totalBytes();
    usedB  = SD .usedBytes();
    freeB  = totalB - usedB;   // no freeBytes in SD
    cardSize = SD.cardSize() / (1024 * 1024);    
    Serial.printf("SD: total: %llu,  free: %llu,  used: %llu, card size: %llu \n", totalB, freeB, usedB, cardSize);  
    
    totalB = FFat.totalBytes();
    freeB  = FFat.freeBytes();
    usedB  = totalB - freeB;    // no usedBytes in FFat
    Serial.printf("FFat: total: %llu,  free: %llu,  used: %llu \n", totalB, freeB, usedB);

    totalB = SD .totalBytes();
    usedB  = SD .usedBytes();
    freeB  = totalB - usedB;   // no freeBytes in SD
    cardSize = SD.cardSize() / (1024 * 1024);
    Serial.printf("SD: total: %llu,  free: %llu,  used: %llu, card size: %llu \n", totalB, freeB, usedB, cardSize);  
    
    totalB = FFat.totalBytes();
    freeB  = FFat.freeBytes();
[...]

PR #8520, opened 2023-08-13
When FFat is mounted before SD, then filesystem reports wrong system sizes

holgerlembke:
Depending on filesystem init order, SD reports wrong usedBytes()/totalBytes() results.

The reason was a fixed drive id.

See:

https://github.com/espressif/arduino-esp32/issues/3546#issuecomment-1676429313


VojtechBartoska:
@P-R-O-C-H-Y Is it a good candidate for 2.0.12 release?

PR #8530, opened 2023-08-15
New 3rd Party Boards - 4D Systems

James-4DSystems:
## Description of Change
Adding new board range from 4D Systems
gen4-ESP32 Range, featuring ESP32-S3R8 SoC, 16MB Flash
Custom Partitions to suit our modules and 16MB Flash

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -24112,3 +24112,160 @@ nebulas3.menu.EraseFlash.all=Enabled
 nebulas3.menu.EraseFlash.all.upload.erase_cmd=-e
 
 ##############################################################
+
+gen4-ESP32-S3R8n16.name=4D Systems gen4-ESP32 16MB Modules (ESP32-S3R8n16)
+gen4-ESP32-S3R8n16.vid.0=0x303a
+gen4-ESP32-S3R8n16.pid.0=0x1001
+
+gen4-ESP32-S3R8n16.bootloader.tool=esptool_py
+gen4-ESP32-S3R8n16.bootloader.tool.default=esptool_py
+
+gen4-ESP32-S3R8n16.upload.tool=esptool_py
+gen4-ESP32-S3R8n16.upload.tool.default=esptool_py
+gen4-ESP32-S3R8n16.upload.tool.network=esp_ota
+
+gen4-ESP32-S3R8n16.upload.maximum_size=1310720
+gen4-ESP32-S3R8n16.upload.maximum_data_size=327680
+gen4-ESP32-S3R8n16.upload.flags=
+gen4-ESP32-S3R8n16.upload.extra_flags=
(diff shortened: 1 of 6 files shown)


P-R-O-C-H-Y:
Hi @James-4DSystems, can you please provide a link for your new board?
Btw. is it intended to have all SPI pins declared as -1 in pins_arduino.h? Thanks

James-4DSystems:
> Hi @James-4DSystems, can you please provide a link for your new board? Btw. is it intended to have all SPI pins declared as -1 in pins_arduino.h? Thanks

Hi @P-R-O-C-H-Y 
The website link is not yet finished, should be ready in about a week I believe.
Can I email you the pre-release datasheets for now?

SPI pins we decided to set to -1 as the pins_arduino.h file is shared with all our ESP32 products, and we have different SPI requirements for each. The SDFAT library seems to be the only thing that uses these 'defaults' when first included (when using our library for our products), but when we call the library for actual use, the pins are set to the required pins for what the library is actually being used for, so this seems to have no negative effect. So yes, it is intended. 
Our library which is used with our products will be releasing soon, deals with setting these up correctly. 
We did have them all defined for one of the products, but it was 'wrong' for the others, but it had no negative effects as when we call our library it sets everything correctly anyway. 
Setting them as -1 seemed to be the best compromise, so it was not incorrect for anything. -1 is there to define cases like this we thought. 
We had thought about splitting them up, but its actually for no gain, as everything else in the pins_arduino.h and variant folder, other than those pins, are common.

If its a strict requirement we have to have these set, then we will have to split the products up, just to have these pins different, and more work will flow on for us for our library and IDE.

Please advise if this is your requirement.

If you want a copy of the pre-release datasheets, please flick me an email and I will happily provide them. Just dont want to make them public until the website is up and the range is announced. email me at james [at] 4dsystems.com.au

P-R-O-C-H-Y (approved):

James-4DSystems:
@P-R-O-C-H-Y 
> Thank you @James-4DSystems for all the informations. I don't need the datasheet as you don't declare many pins so there is no need to check them. But it's better that in the PR is a datasheet linked. So maybe you can edit the description later when the website is ready, even if the PR will be already merged :)

Product line has been released:
https://4dsystems.com.au/product-category/intelligent-display-modules/gen4-esp32-display-modules/

2.4", 2.8", 3.2", 3.5", 4.3", 5.0" all released, and 7.0" and 9.0" coming soon. Non-Touch, Capacitive Touch, and Resistive Touch available for various models.

Datasheets available from the individual product pages, linked from the above link, along with the Github link for the Arduino library which goes along with these modules (Also available from the Library Manager in the Arduino IDE).

Thanks

Issue #6129, opened 2022-01-12
ESP32 BLE/WiFi throws an error [E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes" - still not resolved


sudheeshsmadhav:
### Board

ESP32 devkit v4

### Device Description

I am using ESP32 devkit v4

### Hardware Configuration

no GPIO connections, ESP32 scans the BLE tags and sends their id and rssi values over mqtt

### Version

latest master

### IDE Name

Platform IO also uses arduino IDE 1.8.57.0

### Operating System

Windows10

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

ESP32 scans the available BLE tags and publishing the collected information over mqtt (Raspberry Pi 4 B is the broker). It was working fine in the last month. 
now it started to print error in the console something like [E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes" and after some time it restarts. BLE scan runs in a freeRTOS task. MQTT publishing is in loop().

it works fine for around 5min(time interval is random) then fails and reboot as the log mentioned. 

[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 113, "Software caused connection abort"

### Sketch

```cpp
will provide if requested.
```

### Debug Message

```plain
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 11, "No more processes"
[E][WiFiClient.cpp:395] write(): fail on fd 54, errno: 113, "Software caused connection abort"
PUB Result: 0
[...]

PR #8541, opened 2023-08-18
Reimplemented flush in WiFiClient

PilnyTomas:
The WiFiClient doesn't seem responsible for the actual reading, yet the function `WiFiClient::flush` was calling `recv()`.
This resulted in errors.
I have reimplemented the `WiFiClient::flush` functionality into the `WiFiClientRxBuffer` and the `WiFiClient::flush` simply calls that: `_rxBuffer->flush();`
Solves #6129


sudheeshsmadhav:
the issue is not solved yet. Even after accepting these changes, Still the esp log prints [1522598][E][WiFiClient.cpp:430] write(): fail on fd 48, errno: 11, "No more processes"

https://github.com/espressif/arduino-esp32/issues/6129

PR #8564, opened 2023-08-25
Adding Lion:Bit S3 STEM Dev Board

PradeepKotu:
## Description of Change
LionBit.cc has developed a new board call LionBits3 need to add  Lion:Bit S3 STEM Dev Board to official ESP list 

## Tests scenarios
Tested and already in the market

## Related links
LionBit.cc

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -24112,3 +24112,225 @@ nebulas3.menu.EraseFlash.all=Enabled
 nebulas3.menu.EraseFlash.all.upload.erase_cmd=-e
 
 ##############################################################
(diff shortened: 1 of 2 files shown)


atanisoft (changes requested):
The change to boards.txt appears to be reformatting the entire file making it impossible to verify what was changed. Please update the PR with only the additional entries being created and not a complete reformat of the file.

PradeepKotu:
@atanisoft I am going to close this and thank you for your advice. i am resubmitting it. if you can help me it is a grate help.

PR #8574, opened 2023-08-28
Add support for custom panic handler

me-no-dev:


lucasssvaz (approved):
LGTM. Do you think it would be nice to provide an usage example or is it too niche for that ?

me-no-dev:
> LGTM. Do you think it would be nice to provide an usage example or is it too niche for that ?

it is indeed quite niche and we need to first test it in some dangerous ways, before we can provide a good info and limitations. Ex. Write info to FS, Send over network, etc.

PR #8562, opened 2023-08-25
Adds S3 GPIO48 in  digitalPinToInterrupt() + GPIO INTR Review

SuGlider:
## Description of Change
A user has open an issue about not being able to add an Interrupt to ESP32-S3 GPIO 48 because the GPIO num shall be lower than 48 in `digitalPinToInterrupt()` macro.

Based on it, this PR has a initial review of the macros and the values used in `pins_arduio.h` for the main SoCs (ESP32, S2, S3 and C3).

## Tests scenarios
Simples test done in ESP32/S2/S3/C3 with `attachInterrupt()`.

## Related links
Closes #8560

PR #8533, opened 2023-08-15
added lilygo t-display (esp32-based) board definition

PBrunot:
## Description of Change

This change adds a new board lilygo-t-display based on esp32. Board variants have 4 MB or 16 MB flash memory. All boards come with a TFT display. This board is the predecessor to the already included lilygo-t-display-s3.

[Manufacturer Board webpage](https://www.lilygo.cc/products/lilygo%C2%AE-ttgo-t-display-1-14-inch-lcd-esp32-control-board)

[Pinmap](https://github.com/Xinyuan-LilyGO/TTGO-T-Display/blob/master/image/pinmap.jpg)

## Tests scenarios

I have tested my pull request with a Lilygo TDisplay HW rev 1.1 by configuring PlatformIO to use the platform in https://github.com/PBrunot/arduino-esp32#feature-lilygo-tdisplay as framework for arduino-esp32, and uploading a sketch to a destination board successfully. Testing sketch is using serial, TFT display over SPI.

platformio.used for testing
```
[env:T-Display]
platform = https://github.com/platformio/platform-espressif32.git
board = lilygo_t-display
framework = arduino
monitor_speed = 115200
build_flags = -D ARDUINO_LILYGO_T_DISPLAY_S3
platform_packages =
    framework-arduinoespressif32 @ https://github.com/PBrunot/arduino-esp32#feature-lilygo-tdisplay
lib_deps = 
	moononournation/GFX Library for Arduino@^1.3.7
	Wire
	SPI
```

## Related links

This is required to add support in platformio for the same board, as per platformio esp32 policies : https://github.com/platformio/platform-espressif32/issues/209

PR #8608, opened 2023-09-06
lilygo-t-display : rename pins_arduino.h.txt to pins_arduino.h

PBrunot:
## Description of Change
This PR fixes a wrongly named file in the merged PR #8533 
User @SuGlider has proposed the fix in his PR #8562 which is however delayed for 3.0. 
Therefore this PR is created to fix the problem early.

## Related links
See PR #8533


mrengineer7777 (commented):
LGTM.  Maintainers: is this targeting the desired branch?

PR #8620, opened 2023-09-08
Default Console UART Rx Tx defined in HardwareSerial Constructor

### LINKED ISSUE #8324: Use pin I2C pin 43 as I/O for DS18B20 temp sensor
### Board

ESP32 S3R8

### Device Description

LilyGO T-Display S3 

### Hardware Configuration

GPIO43 is suggested by LilyGO as the I2C pins, they are brought out to the JST connector.
I am trying to connect a DS18B20 "oneWire" temp sensor that will work on other GPIO but mot pin 43 or 44.
I2C devices work on these pins as expected.

### Version

v2.0.6 at first and then deleted it and installed v2.0.9, no change!

### IDE Name

Arduino IDE v2.1.0

### Operating System

MacOS Ventura 13.4

### Flash frequency

QIO 80

### PSRAM enabled

yes

### Upload speed

460800

### Description

I am trying to connect a Dallas Semiconductor DS18B20 temperature sensor to either the GPIO43 or the JST pin 43, which are probably the same circuit.
There is no response from the sensor at all, the oscilloscope shows 3.3v steady.
The sensor works on GPIO 1 (or GPIO 18, e.g.).
The pin 43 can be programmed as an input or an output successfully, but not the serial serial I/O of the DS18B20.
Further another user says that the JST pin 43 works in ESPHome, but we cannot get it to work on Arduino.  I cannot understand why it works in only one software.
My question therefore is:
"Is there a configuration setup for the LilyGO T-Display S3 that configures GPIO43 for I2C only? How can we free it up?"

### Sketch

```cpp
#include <OneWire.h>
#include <DallasTemperature.h>
const int oneWireBus = 43;     
OneWire oneWire(oneWireBus);
DallasTemperature sensors(&oneWire);

void setup() {
  // Start the Serial Monitor
  Serial.begin(115200);
  delay(200);
  Serial.println(" setup  ");
  delay(200);
   sensors.begin();
}

void loop() {
  sensors.requestTemperatures(); 
  float temperatureC = sensors.getTempCByIndex(0);
  Serial.print(temperatureC);
  Serial.println("ºC");
  Serial.println("   ");
  delay(2000);
```
```


### Debug Message

```plain
Do not think this is the issue!
```

### Other Steps to Reproduce

It works on ESPHome, but not Arduino IDE.

### LINKED ISSUE #8573: HardwareSerial::end() won't detach RX/TX from IO MUX when it is executed isolated

``` cpp
void setup() {
  
  // code as usual...
  Serial.begin();
  delay(200);
  Serial.println(" setup ");
  delay(200);
  sensors.begin();
}
```

SuGlider:
## Description of Change
ESP32 SoCs attach RX and TX pins in boot time, conneting those GPIOs to IOMUX or GPIO Matrix.
Therefore, if the user calls `HardwareSerial::setPins()` or `HardwareSerial::end()` with no previous `HardwareSerial::begin()`,
those default already attached pins won't be detached, preventing them to be used for other purpose.

This PR shall be merged along with #8619 in order to make full effect with `HardwareSerial::setPins()`

## Tests scenarios
``` cpp
void setup() {
  ets_printf("\nThis Line is printed fine because it uses TX Pin attached in Boot Time.\n\n");
  ets_printf("1:: This Log line shall not be displayed!\n"); // using IDF console ets_printf() -- TX unattached!
  
  Serial.flush();

  Serial.setPins(-1, 2);    // shall set UART0 to GPIO2 and detach default TX -- DEPENDS on MERGING #8619 
  Serial.flush();
  
  Serial.end();
  Serial.setPins(-1, 2); // it has not effect because Serial is ended... 
  Serial.flush();
}

void loop() {}
```

## Related links
Closes #8324
Closes #8573

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -348,28 +348,36 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
 #endif
 
     HSERIAL_MUTEX_LOCK();
+    // make sure that if rxPin|txPin has changed, the previous GPIOs (_rxPin|_txPin) are detached
+    // Example: Serial.begin(115200); Serial.begin(115200, 2, 5); ==> second begin() shall detach RX/TX from first one
+    if (_uart != NULL && rxPin >= 0 && rxPin != _rxPin) {
+        uartDetachPins(_uart, _rxPin, -1, -1, -1);
+    }
+    if (_uart != NULL && txPin >= 0 && txPin != _txPin) {
+        uartDetachPins(_uart, -1, _txPin, -1, -1);
+    }
     // First Time or after end() --> set default Pins
     if (!uartIsDriverInstalled(_uart)) {
         switch (_uart_nr) {
             case UART_NUM_0:
                 if (rxPin < 0 && txPin < 0) {
-                    rxPin = SOC_RX0;
-                    txPin = SOC_TX0;
+                    rxPin = _rxPin < 0 ? SOC_RX0 : _rxPin;
+                    txPin = _txPin < 0 ? SOC_TX0 : _txPin;
                 }
             break;
 #if SOC_UART_NUM > 1                   // may save some flash bytes...
             case UART_NUM_1:
                if (rxPin < 0 && txPin < 0) {
-                    rxPin = RX1;
-                    txPin = TX1;
+                    rxPin = _rxPin < 0 ? RX1 : _rxPin;
+                    txPin = _txPin < 0 ? TX1 : _txPin;
                 }
             break;
 #endif
 #if SOC_UART_NUM > 2                   // may save some flash bytes...
             case UART_NUM_2:
                if (rxPin < 0 && txPin < 0) {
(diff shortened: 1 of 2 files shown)


cores/esp32/HardwareSerial.cpp
```diff
@@ -97,14 +97,18 @@ void serialEvent2(void) {}
 
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
 #if ARDUINO_USB_CDC_ON_BOOT //Serial used for USB CDC
-HardwareSerial Serial0(0);
+// sets UART0 (default console) RX/TX default pins already configured in boot
+HardwareSerial Serial0(0, SOC_RX0, SOC_TX0);
```
me-no-dev, line 101: if this is for UART0 only, would it not be safer to keep the old signature and handle the default pins in the constructor?

SuGlider: Yes, this is a better way, less intrusive.

PR #8622, opened 2023-09-11
Support relative paths in UNIX search PATH

me-no-dev:
Fixes: https://github.com/espressif/arduino-esp32/issues/8610

### DIFF
--- a/.github/scripts/on-push.sh
+++ b/.github/scripts/on-push.sh
@@ -4,6 +4,9 @@ set -e
 
 export ARDUINO_BUILD_DIR="$HOME/.arduino/build.tmp"
 
+# Relative path test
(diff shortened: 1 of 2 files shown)


SuGlider (approved):
It sounds good. Using absolute path for `bash` increases security.

Anyway, it doesn't affect Windows. 
We may think about something similar for `cmd.exe`

I'll test it here as well using Win ENV VAR `ComSpec`

P-R-O-C-H-Y (approved):

VojtechBartoska:
@me-no-dev this will be part of 2.0.15 release, right?

me-no-dev:
if released :) i'll port it over to 3.0.0 too

PR #8629, opened 2023-09-13
Uart detach 2.0.13

### LINKED ISSUE #8324: Use pin I2C pin 43 as I/O for DS18B20 temp sensor

### LINKED ISSUE #8573: HardwareSerial::end() won't detach RX/TX from IO MUX when it is executed isolated

### LINKED ISSUE #8607: HardwareSerial::setPins does not unset previous pin

SuGlider:
## Description of Change
There is a bug in `HardwareSerial::setPins()` that sets the UART pins internally to the wrong value at the end of the operation, which may lead to errors in `HardwareSerial::end()`, not dettaching the right GPIO from UART.

This PR also makes sure that `HardwareSerial::setPins()` will always detach previous attached GPIOs.

ESP32 SoCs attach RX0 and TX0 console pins in boot time, connecting those GPIOs to IOMUX or GPIO Matrix.
The PR makes sure that if the application calls `Serial.begin(baud, rxPin, txPin)` at first to setup UART, it will detach Console RX0/TX0.
The same, if the user calls `HardwareSerial::setPins()` or `HardwareSerial::end()` with no previous `HardwareSerial::begin()`,

Other improvements/fixes:
1- `setPins()` can be called before `begin()`, or in any order.
2- `end()` can be called at any time, even before `begin()` detaching UART Pins.
3- Once RX/TX are set with `begin(baud, rx, tx)` or with `setPins()`, calling `begin(baud)` will not force setting default board's RX/TX.

## Tests scenarios
Use Cases:
0) First time calling end() or setPins() todetach default RX0/TX0 
1) First time calling begin(baud) --> it will use SoC default RX/TX pins
2) First calls begin(baud, rx, tx) then calls begin(baud) --> it will keep rx, tx from previous begin() +  detaches default RX/TX
3) First calls begin(baud), then setPin(rx, tx), then calls begin(baud) --> NEEDS #8620 :: detaches default RX/TX + attaches rx/tx in second begin()
4) First calls begin(baud), then end(), then calls begin(baud, rx, tx) --> it will use rx/tx pins + detaches default RX/TX 
5) First calls begin(baud) then begin(baud, rx, tx) --> it will use rx/tx from second begin() + detaches default RX/TX
6) First calls begin(baud, rx, tx) --> detaches default RX/TX (UART0 only) + attachs RX/TX from begin()
7) First calls begin(baud, rx1, tx1) then calls begin(baud, rx2, tx2) --> detaches rx1/tx1 + attachs rx2/tx2
8) First calls begin(baud) then calls begin(baud, -1, tx) --> detaches default TX, attaches tx and default RX 
9) First calls begin(baud, -1, tx) --> attaches tx, leaves RX unchenged, but in UART0.end(), it will force dettach default RX
10) First calls begin(baud, rx) --> attaches rx, leaves TX unchanged, but in UART0.end(), it will force dettahc default TX
11) First calls begin(baud, rx) then calls begin(baud, rx2, tx2) --> detaches default RX/TX (UART0 only) + detaches rx + attaches rx2/tx2  

``` cpp
void setup() {
  ets_printf("\nThis Line is printed fine because it uses TX Pin attached in Boot Time.\n\n");
  ets_printf("1:: This Log line shall not be displayed!\n"); // using IDF console ets_printf() -- TX unattached!
  
  Serial.flush();

  Serial.setPins(-1, 2);    // shall set UART0 to GPIO2 and detach default TX -- DEPENDS on MERGING #8619 
  Serial.flush();
  
  Serial.end();
  Serial.setPins(-1, 2); // it has not effect because Serial is ended... 
  Serial.flush();
}

void loop() {}
```

``` cpp

// Check previous GPIOs with a Logic Analyzer
void setup() {
  Serial.begin(115200);  // sets default RX and TX
  Serial.println("This will go out in the default TX pin");
  Serial.flush();
  
  Serial.setPins(-1, 2);  // Now TX shall be GPIO2 and default TX GPIO can't send any data
  Serial.println("This line can't be seen in the default TX GPIO!");
  Serial.flush();
}

void loop(){
}
```

## Related links
Closes #8324
Closes #8573 
Closes #8607

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -133,7 +133,7 @@ void serialEventRun(void)
 #define HSERIAL_MUTEX_UNLOCK()  
 #endif
 
-HardwareSerial::HardwareSerial(int uart_nr) : 
+HardwareSerial::HardwareSerial(uint8_t uart_nr) : 
 _uart_nr(uart_nr), 
 _uart(NULL),
 _rxBufferSize(256),
@@ -336,7 +336,7 @@ void HardwareSerial::_uartEventTask(void *args)
 
 void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, int8_t txPin, bool invert, unsigned long timeout_ms, uint8_t rxfifo_full_thrhd)
 {
-    if(0 > _uart_nr || _uart_nr >= SOC_UART_NUM) {
+    if(_uart_nr >= SOC_UART_NUM) {
         log_e("Serial number is invalid, please use numers from 0 to %u", SOC_UART_NUM - 1);
         return;
     }
@@ -436,11 +436,11 @@ void HardwareSerial::begin(unsigned long baud, uint32_t config, int8_t rxPin, in
     // detach previous attached RX/TX pins when it has changed
     if (_uart != NULL) {
         if (rxPin >= 0 && rxPin != _rxPin) {
-            uartDetachPins(_uart, _rxPin, -1, -1, -1);
+            uartDetachPins(_uart_nr, _rxPin, -1, -1, -1);
             _rxPin = rxPin;
         }
         if (txPin >= 0 && txPin != _txPin) {
-            uartDetachPins(_uart, -1, _txPin, -1, -1);
+            uartDetachPins(_uart_nr, -1, _txPin, -1, -1);
             _txPin = txPin;
         }        
     }    
@@ -466,7 +466,7 @@ void HardwareSerial::end(bool fullyTerminate)
 
         _rxFIFOFull = 0; 
 
-        uartDetachPins(_uart, _rxPin, _txPin, _ctsPin, _rtsPin);
+        uartDetachPins(_uart_nr, _rxPin, _txPin, _ctsPin, _rtsPin);
         _rxPin = _txPin = _ctsPin = _rtsPin = -1;
 
     }
@@ -571,11 +571,6 @@ void HardwareSerial::setRxInvert(bool invert)
 // negative Pin value will keep it unmodified
 bool HardwareSerial::setPins(int8_t rxPin, int8_t txPin, int8_t ctsPin, int8_t rtsPin)
 {
-    if(_uart == NULL) {
-        log_e("setPins() shall be called after begin() - nothing done\n");
-        return false;
-    }
-
     // map logical pins to GPIO numbers
     rxPin = digitalPinToGPIONumber(rxPin);
     txPin = digitalPinToGPIONumber(txPin);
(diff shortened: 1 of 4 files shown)


SuGlider:
@me-no-dev  - Please review and merge this PR to 2.0.13

PR #8654, opened 2023-09-20
new industrial board namino bianco

a-sapuppo:


P-R-O-C-H-Y (approved):

me-no-dev:
Also it seems that you have added two extra bootloaders for variants without `pins_arduino.h` and `boards.txt` definition

PR #8671, opened 2023-09-25
Fix memory leak and undefined behavour in Updater.cpp (UpdaterClass)

### LINKED ISSUE #7984: Arduino OTA Updater has memory leak in certain conditions
### Board

any board using updater.cpp

### Device Description

tested on many ESP32 dev boards

### Hardware Configuration

mainly WLED strings, not much else.

### Version

### IDE Name

PlatformIO

### Operating System

Windows 10/11

### Flash frequency

40Mhz/80Mhz

### PSRAM enabled

no

### Upload speed

921600, 115200

### Description

On an OTA firmware attempt, if the transfer fails, and updateClass is defined statically, the buffers used for the update are not freed.

There is no destructor.

### Sketch

```cpp
any sketch using update.cpp
```

### Debug Message

```plain
ESP.getFreeHeap() shows heap memory from buffers being allocated, and not used.
```

### Other Steps to Reproduce

_No response_

MathewHDYT:
## Description of Change
The pull request aims to fix undefined behaviour and a memory leak of 16 bytes every time the `UpdaterClass` is used, this occurs because the `_skipBuffer` is not freed correctly and the `_buffer` is deleted but using the C++ style `delete[]`, even tough the C style memory allocation `malloc` was used. Which is not compatible and is undefined behavour. Meaning it might work on some devices on others it might crash and on some it might do other things.

The changes are compatible with the current API and do not change anything about the `UpdaterClass` public or private method. It doesn't even change the private members. It only adjusts the memory allocation to always use C++ style `new` and to actually free the created memory with `delete[]`.

----------------------------

A more detailed description can be found in the [issue](https://github.com/espressif/arduino-esp32/issues/7984#issuecomment-1717151822) mentioned below as well.

## Tests scenarios
The changed code has been tested with an ESP32 and updating the device with the `UpdaterClass` over OTA.

## Related links
Closes #7984


mrengineer7777 (approved):
Codefix looks good to me.  Is this targeting the correct branch?

MathewHDYT:
> Codefix looks good to me. Is this targeting the correct branch?

@mrengineer7777 Not sure perhaps targeting `release/v2.x` makes more sense. If I need to pull to another branch instead I can simply change the branch of this pull request if so required.

me-no-dev:
> perhaps targeting release/v2.x makes more sense

no. leave it as is. We will not update 2.x anymore and will merge 5.1 into master to focus on 3.0.0

P-R-O-C-H-Y (commented):
@MathewHDYT When updating this part of code, can you change the error messages as suggested? :) Thanks, otherwise LGTM.

P-R-O-C-H-Y (approved):

PR #8672, opened 2023-09-26
Fix race condition in WiFiGenericClass::hostByName

MattiasTF:
## Description of Change
`dns_gethostbyname`, as used in `hostByName`, is required to run in lwIP’s TCP/IP context. This can be verified by enabling `LWIP_CHECK_THREAD_SAFETY` in the sdkconfig.

Calling `dns_gethostbyname` from the Arduino task can trigger race conditions in lwIP or lower layers. One possibility is a corruption of IDF's Ethernet buffers, causing an unstoppable flood of "insufficient TX buffer size" errors, effectively severing all Ethernet connectivity.

`hostByName` cannot be executed in the TCP/IP context because it blocks until completion, preventing the TCP/IP task from receiving the DNS response to unblock itself, causing a deadlock. Therefore, this PR patches `hostByName` to execute `dns_gethostbyname` in the TCP/IP context while still blocking only the Arduino task.

## Tests scenarios
I have tested my PR on Arduino-esp32 master as of 02e31b4001892535602bb0c0cc9b42e14d0c4901 on a custom ESP32 board.

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -37,6 +37,7 @@ extern "C" {
 #include <esp_wifi.h>
 #include <esp_event.h>
 #include <esp_mac.h>
+#include <esp_netif.h>
 #include "lwip/ip_addr.h"
 #include "lwip/opt.h"
 #include "lwip/err.h"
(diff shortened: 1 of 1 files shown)


me-no-dev:
This is good to have! @MattiasTF could you please fix the merge conflicts, so that we can merge this?

MattiasTF:
What would be your preferred fix? I can rebase the branch to give you the simplest fix.

me-no-dev:
@MattiasTF that would be great. We just can't merge it right now and can not rebase it for you, because you did not select to allow us to do that when you created the PR.

MattiasTF:
There we go.

> We just can't merge it right now and can not rebase it for you, because you did not select to allow us to do that when you created the PR.

Thanks for the info. I will look for that option next time.

me-no-dev:
The checkbox next to the green "Create Pull Request" button

MicrosoftTeams-image (1)

MattiasTF:
Thanks, but I just found out that that option only exists when creating PRs from user-owned forks. It’s not possible to use that option with forks living in company repos, which is why I don’t have it.

PR #8697, opened 2023-10-04
3.0.0 - UART Peripheral Manager + Detach UART pins on begin()/setPins()

### LINKED ISSUE #8573: HardwareSerial::end() won't detach RX/TX from IO MUX when it is executed isolated

### LINKED ISSUE #8607: HardwareSerial::setPins does not unset previous pin

SuGlider:
## Description of Change
This PR implements the same changes as described in #8629 
It also adds to the UART Peripheral Manager the control over UART pin and sets default console UART0 RX/TX.

The `BareMinimum.ino` example (empty sketch) will report RX/TX correctly when Verbose/Debug mode are set by using the Chip Debug Report feature.

Summary of changes:

- It is now possible to dettach UART0 pins by calling Serial.end() with no previous begin()
- It is now possible to call setPins() before begin() or in any order
- setPins() will detach any previous pins that have been changed
- begin(baud, rx, tx) will detach any previous attached pins
- setPins() or begin(baud, rx, tx), when called at first, will detach console RX0/TX0, attached in boot
- any pin set as -1 in begin() or setPins() won't be changed nor detached
- begin(baud) will not changed any pins that have been set before this call through a previous begin(baud, rx, tx) or setPin()
- If the application only uses RX or TX, begin(baud, -1, tx) or begin(baud, rx) will change only the assigned pin and keep the other unchanged.

## Tests scenarios
Using ESP32

``` cpp
void setup() {
  printf("\nThis Line is printed fine because it uses TX Pin attached in Boot Time.\n\n");
  delay(100);
  printf("1:: This Log line shall not be displayed!\n"); // using IDF console ets_printf() -- TX unattached!
  delay(100);
  
  Serial.flush();

  Serial.setPins(-1, 2);    // shall set UART0 to GPIO2 and detach default TX 
  Serial.flush();
  
  Serial.end();
  Serial.setPins(-1, 2); // it sets TX as GPIO2
  Serial.flush();
}

void loop() {}
```

## Related links

Fix #8573 
Fix #8607 

Close Draft PRs: #8619 #8620

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -148,7 +148,7 @@ void serialEventRun(void)
 #define HSERIAL_MUTEX_UNLOCK()
 #endif
 
-HardwareSerial::HardwareSerial(int uart_nr) :
+HardwareSerial::HardwareSerial(uint8_t uart_nr) :
 _uart_nr(uart_nr),
 _uart(NULL),
 _rxBufferSize(256),
@@ -172,6 +172,12 @@ _eventTask(NULL)
         }
     }
 #endif
+    // sets UART0 (default console) RX/TX pins as already configured in boot
+    if (uart_nr == 0) {
+        setPins(SOC_RX0, SOC_TX0);
+    }
+    // set deinit function in the Peripheral Manager
+    uart_init_PeriMan();
 }
 
 HardwareSerial::~HardwareSerial()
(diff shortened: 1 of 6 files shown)


P-R-O-C-H-Y (approved):
LGTM! Tested and worked as expected.

lucasssvaz (approved):
Works as expected. LGTM.

SuGlider:
Force Rebase to master in #8719

PR #8708, opened 2023-10-05
new industrial board namino bianco

a-sapuppo:


a-sapuppo:
prepare new code

PR #8710, opened 2023-10-05
added namino bianco board + other fix

a-sapuppo:


variants/namino_bianco/pins_arduino.h
```diff
@@ -0,0 +1,205 @@
...
+#define USB_PID 0x1001
+
+#define NAMINO_BIANCO_BOARD
+
+#define NUM_DIGITAL_PINS            SOC_GPIO_PIN_COUNT    // GPIO 0..48
+#define NUM_ANALOG_INPUTS           20                    // GPIO 1..20
+#define EXTERNAL_NUM_INTERRUPTS     NUM_DIGITAL_PINS      // All GPIOs
+#define analogInputToDigitalPin(p)  (((p)<NUM_ANALOG_INPUTS)?(analogChannelToDigitalPin(p)):-1)
+#define digitalPinToInterrupt(p)    (((p)<NUM_DIGITAL_PINS)?(p):NOT_AN_INTERRUPT)
+#define digitalPinHasPWM(p)         (p < NUM_DIGITAL_PINS)
```
P-R-O-C-H-Y, line 21: Can you please remove this part? For 3.0.0 this will be automatically set in` Arduino.h` file for all SOCs and it's not needed to be defined there anymore.
For reference PR #8706.

P-R-O-C-H-Y (changes requested):
@a-sapuppo PTAL on my comment.
One more thing is, as you don't have an option se select more than 4MB of flash memory, I assume the board is sold with this setup. Why do you need all those partitions which may not be suitable for 4MB flash? Also PSRAM, you have there both options QSPI PSRAM and OPI PSRAM. This depends too on the chip you have on your board. If you have only one configuration of the board, one of this option will cause panic errors if selected and flashed.

PR #8750, opened 2023-10-10
Rename the RV32 toolchain on release to fit paths in GCC limit

me-no-dev:
Some include paths in the RV32 toolchain on Windows end up too long for GCCs 200 char limit and compilation fails. While a fix is provided upstream, we workaround this issue by shortening the tool name and version, thus saving 26 chars of the path.


me-no-dev:
@igrr PTAL

lucasssvaz (approved):
LGTM

Issue #8735, opened 2023-10-06
Refactor the network stack to allow chips without WiFi and better Ethernet and ppp compatibility


me-no-dev:
### Related area

Networking

### Hardware specification

Proper H2 and P4 support

### Is your feature request related to a problem?

Currently almost nothing about networking can operate without WiFi, so for chips like the ESP32-H2 and ESP32-P4 that do not have WiFi radios, almost nothing will compile, even if Ethernet is added.

### Describe the solution you'd like

- Create a central `Network` class, that will keep track of interfaces, handle events and resolve domains (replacement for WiFiGeneric)
- WiFiClient, WiFiServer and WiFiUdp to be moved to the above networking library, renamed and added defines to the old names for compatibility
- Update all other libraries to not reference `WiFi` anymore
- Rewrite AP and STA interfaces by the Ethernet model (Create abstract class, based on `esp_netif` for common functionality)
- Provide interface to the old WiFi API for compatibility
- Discuss possible deprecation of the old APIs/names

### Describe alternatives you've considered

We could probably get around without rewriting, with just a bunch `#ifdef`, but that will be ugly and still require `WiFi` code on a chip without WiFi.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

lucasssvaz:
IMO we should to this as cleanly as possible even if it means rewriting some code. This will be much better in the long run for the code maintainability.

TD-er:
I think it would also make sense to have each "network" device being split up into some kind of "driver" structure like how it is done in the ESP-IDF 5.1 and also have some "state" structure/class which has some generic interface to query simple states like "connected" and "gotIP" etc.

Right now this is a complete mess as you need to query WiFi related code to use Ethernet.
Also disabling WiFi after you processed an Ethernet DHCP will clear out DNS server records etc.

Don't get me started on all possible state changes when connecting to WiFi. (or while being connected)
These are way too many and should be dealt with in a specific piece of code only dealing with those for WiFi.
For example the WiFi disconnect reason can be very important on what to do next.
The most intuitive order of events is:
- Connected
- GotIP

But that doesn't always happen in this order and it should be dealt with in such a state machine struct/class specific to WiFi.

So these state classes should also keep track of their own "addresses". (not sure how it will be called for 802.15.4, have not yet looked into it)
[...]

PR #8760, opened 2023-10-13
3.0.0 Network Refactoring

me-no-dev:
Better explanation will come later


libraries/ESPmDNS/src/ESPmDNS.cpp
```diff
@@ -115,7 +113,7 @@ void MDNSResponder::disableArduino(){
 void MDNSResponder::enableWorkstation(esp_interface_t interface){
     char winstance[21+_hostname.length()];
     uint8_t mac[6];
-    esp_wifi_get_mac((wifi_interface_t)interface, mac);
+    esp_base_mac_addr_get(mac);
```
TD-er, line 116: How does the ESP_Network_Manager know which MAC address to use here?
The function does get an interface as argument, which is probably for compatibilty reasons. But I truly don't see how the Network Manager or this MDNSResponder knows about which interface is queried here as it doesn't seem to have any members which look like an interface, nor does `ESP_Network_Events`.

So it seems you're only dealing with the base MAC address here and not adding any offset for the WiFi SoftAP, Bluetooth or Ethernet interface.

me-no-dev: This is subject to change once all interfaces are implemented/reworked. The point of the change currently is to disconnect WiFi from the rest, so things compile.

VojtechBartoska:
Closes https://github.com/espressif/arduino-esp32/issues/8735

Jason2866:
@me-no-dev Tried with the commits from today. Tasmota compiles and works. Well done!

TD-er:
Tried running the code from this PR too on ESPEasy and I no longer need the delay to get an IPv6 address on ESP32(classic).

However I seem to get the `ARDUINO_EVENT_ETH_GOT_IP6` event twice for both local and global IPv6 address.
My current implementation uses the onEvent from the WiFi class for Ethernet too. I noticed there is a new `ESP_Network_Events` class which has its own `onEvent`. Is it possible these events get duplicated because of legacy support for Eth events via the `WiFi.onEvent()` handler?
Or is there something I'm doing wrong?

me-no-dev:
@TD-er hold your horses. This PR is very far from being done and a lot more is going to be changed. Wait until it's not marked as WIP to comment. And yes, you do get separate events for link-local and global IPv6. Both events come from IDF and you were getting them before any of these changes too. And IPv6 is not fully working on this PR yet.

TD-er:
Nope I meant I get both link local and global IPv6 events for Eth twice. (thus 4 events)

And the reason I'm looking into this is that I also really would love to throw out all my network related kludged code in ESPEasy as soon as possible to use a proper setup structure as soon as possible.
So I thought getting to know what's coming may help me make proper decissions for ESPEasy too as those design issues need to be processed for quite a while in my head to crystalize.

This PR code is by no means going to be used in release code to users until this PR is finished.
It is merely for me to know what will/might be coming and to help you identify some bugs in an early stage.

Just see it as me being very enthusiastic about the upcoming network design changes and I can't wait to actually use it :)

me-no-dev:
v6 works only on ETH at the moment, but I did not get extra events? Will look deeper into it later.

TD-er:
> v6 works only on ETH at the moment

Nope, it also seems to work on WiFi :)
I can access the web interface of my test node via IPv6 while connected via WiFi.

libraries/Networking/src/ESP_Network_Interface.cpp
```diff
@@ -0,0 +1,526 @@
...
+    return String(macStr);
+}
+
+IPAddress ESP_Network_Interface::localIP()
+{
+    if(_esp_netif == NULL){
+        return IPAddress();
+    }
+    esp_netif_ip_info_t ip;
+    if(esp_netif_get_ip_info(_esp_netif, &ip)){
```
sgryphon, line 379: This is the local (as opposed to remote) IP address ; it will usually be a global scope address (but sometimes not). Please align the names of IPv4 and IPv6.

libraries/Networking/src/ESP_Network_Interface.cpp
```diff
@@ -0,0 +1,526 @@
...
+        return (uint8_t)0;
+    }
+    esp_netif_ip_info_t ip;
+    if(esp_netif_get_ip_info(_esp_netif, &ip)){
+        return (uint8_t)0;
+    }
+    return calculateSubnetCIDR(IPAddress(ip.netmask.addr));
+}
+
+IPAddress ESP_Network_Interface::localIPv6()
```
sgryphon, line 457: This is not consistent with usage for IPv4. This is not the local (as opposed to remote) address, with global scope, but the link local scope address.

It should be called `linkScopeIPv6()` or `linkLocalIPv6()` or something similar.

libraries/Networking/src/ESP_Network_Interface.cpp
```diff
@@ -0,0 +1,526 @@
...
+        return IPAddress(IPv6);
+    }
+    static esp_ip6_addr_t addr;
+    if(esp_netif_get_ip6_linklocal(_esp_netif, &addr)){
+        return IPAddress(IPv6);
+    }
+    return IPAddress(IPv6, (const uint8_t *)addr.addr, addr.zone);
+}
+
+IPAddress ESP_Network_Interface::globalIPv6()
```
sgryphon, line 469: Be consistent with IPv4.

The `localIPv4()` function returns a local (as opposed to remote) IPv4 address with global scope.

This function returns a local (as opposed to remote) IPv6 address with global scope -- the same thing, but IPv6 instead of IPv4, so it should have a similar name.

libraries/Networking/src/ESP_Network_Interface.h
```diff
@@ -0,0 +1,73 @@
...
+    ESP_NETIF_ID_PPP,
+    ESP_NETIF_ID_ETH0,
+    ESP_NETIF_ID_ETH1,
+    ESP_NETIF_ID_ETH2,
+    ESP_NETIF_ID_MAX
+} ESP_Network_Interface_ID;
+
+#define ESP_NETIF_ID_ETH ESP_NETIF_ID_ETH0
+
+class ESP_Network_Interface {
```
sgryphon, line 20: Are these underscores standard in class names ... most other classes do not use underscores?

The Arduino Style Guide (https://docs.arduino.cc/learn/contributions/arduino-library-style-guide/) does not specifically mention this or have examples, but it does refer to https://processing.org/reference in the context of camel case function names.  Processing.org has title case class names, e.g. ArrayList.

libraries/Networking/src/ESP_Network_Interface.h
```diff
@@ -0,0 +1,73 @@
...
+        uint8_t * macAddress(uint8_t* mac);
+        String macAddress();
+        IPAddress localIP();
+        IPAddress subnetMask();
+        IPAddress gatewayIP();
+        IPAddress dnsIP(uint8_t dns_no = 0);
+        IPAddress broadcastIP();
+        IPAddress networkID();
+        uint8_t subnetCIDR();
+        IPAddress localIPv6();
```
sgryphon, line 47: As mentioned elsewhere, this should be `linkScopeIPv6()` or `linkLocalIPv6()`, because it has a different result than the IPv4 `localIP()`.  

Usually `localIP()` returns a local (as opposed to remote) IPv4 address with global scope, wheres this function returns link-local scope instead.

libraries/Networking/src/ESP_Network_Interface.cpp
```diff
@@ -0,0 +1,526 @@
...
+        //     arduino_event.event_id = ARDUINO_EVENT_PPP_GOT_IP6;
+        //     Network.setStatusBits(PPP_HAS_IP6_BIT);
+        // } else 
+        if(_interface_id >= ESP_NETIF_ID_ETH && _interface_id < ESP_NETIF_ID_MAX){
+            arduino_event.event_id = ARDUINO_EVENT_ETH_GOT_IP6;
+            if(addr_type == ESP_IP6_ADDR_IS_GLOBAL){
+                Network.setStatusBits(ETH_HAS_IP6_GLOBAL_BIT(_interface_id - ESP_NETIF_ID_ETH));
+            } else if(addr_type == ESP_IP6_ADDR_IS_LINK_LOCAL){
+                Network.setStatusBits(ETH_HAS_IP6_BIT(_interface_id - ESP_NETIF_ID_ETH));
+            }
```
sgryphon, line 129: This list is incomplete; addr_type has some other values, including ULA (unique local address), which are also global scope. The ESP definition of each category is here: https://github.com/espressif/esp-lwip/blob/2.1.2-esp/src/include/lwip/ip6_addr.h#L203

Note that their definition of GLOBAL is not exactly the same as the current definition global scope in RFC 4291 (ULAs have global scope), but is probably the best name for the range which is the only current range used for global assignment. See: https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml

While other types (e.g. site local) are unlikely, ULA addresses could be found in private networks (they are similar to IPv4 private address ranges), including as DNS destinations. They could also be used for NAT66 (like private address range NAT44), so when in use may also expect to reach public DNS locations (via NAT).

libraries/Networking/src/ESP_Network_Manager.cpp
```diff
@@ -0,0 +1,136 @@
...
+int ESP_Network_Manager::hostByName(const char* aHostname, IPAddress& aResult, bool preferV6)
+{
+    err_t err = ERR_OK;
+    gethostbynameParameters_t params;
+
+    // This should generally check if we have a global address assigned to one of the interfaces.
+    // If such address is not assigned, there is no point in trying to get V6 from DNS as we will not be able to reach it.
+    // That is of course, if 'preferV6' is not set to true
+    static bool hasGlobalV6 = false;
+    bool hasGlobalV6Now = (Network.getStatusBits() & NET_HAS_IP6_GLOBAL_BIT) != 0;
```
sgryphon, line 86: ULA addresses should probably also support IPv6 lookup in DNS.

They may be used in private deployments, with either direct DNS support, or NAT66 (like NAT44) where they can connect to public addresses

sgryphon: For example my Thread network at home (Google) uses random ULA ranges to addresses; an Eve switch I own has the addresses:
FE 80 00 00 00 00 00 00 C0 C8 06 A1 78 37 2F 37
FD 62 CB DC AC 0A 00 00 D5 0C F4 56 0A 5B 02 27
FD 1B 49 A9 44 D5 00 01 FA 37 8B B6 11 70 35 6F

I haven't tested a device on it yet, but for this device to reach a public address would use NAT66.  I have seen mention of OTBR (Open Thread Border Router, the underlying open source) mentions NAT64.

In either case DNS (or in the case of Thread mDNS) would need to return IPv6 addresses to be of use.

So, we should do an IPv6 lookup for either ESP GLOBAL   or ULA.

me-no-dev:
@sgryphon I suggest you wait until I finish everything and then comment. Will be helpful if you provide solutions or be able to test what you propose.

libraries/Networking/src/ESP_Network_Interface.cpp
sgryphon: `linkLocalIPv6()` is a win for me. I like it!

Almost all cases will be either:
* 1x Public IPv6, 1x IPv4 (probably a private range), 1x Link-Local IPv6; or
* 1x IPv4, 1x Link-Local IPv6 (if no IPv6 available)

It would be rare to have extra IPv6 as ESP / LWIP only supports RA prefixes; it does not support DHCPv6 stateful, or privacy extensions, so won't have additional addresses that a laptop / tablet / phone might.

Less common scenarios:
* 1x ULA IPv6, either with or without a public IPv6; this could be from someone wanting to test out local IPv6 but not supported by their ISP (no public), or want a private network (similar to IPv4 private address). Not a lot of commercial implementation, although a Thread gateway may use this for the Thread network. May also be present in NAT66 scenarios
* 1x static ULA IPv6 may be used in AP scenarios (possibly 2x with a simple address fd00::1, plus a EUI-64 address)
* Multiple public IPv6: unlikely on one interface. Could be static configuration if you want a simple fixed address plus RA address.
* Multiple link-local IPv6: could be static configuration if you want a simple fixed address (fe80::1), plus an RA address (although ULA are more accessible, e.g. via browser)

sgryphon: `linkLocalIPv6()` aligns with what is available in IDF / LWIP, i.e. `esp_netif_get_ip6_linklocal()`

The initial `globalIPv6()` aligns with `esp_netif_get_ip6_global()`, which I think only returns a 2xxx / 3xxx address.

It would be rare to have multiple public (what ESP calls GLOBAL, i.e. 2xxx or 3xxx) addresses, so I don't think `globalIPv6(int index)` is that useful as there is unlikely more than one. I think it would be more common to have a ULA instead, although if you mean global scope then technically that includes ULAs (but not in ESP). 

Was the index supposed to align with the ` esp_netif_get_all_ip6()` function?  This covers the ULA scenario (single and multiple) as well as the multiple link-local (not very useful for a device I think), as well as where you actually have multiple 2xxx / 3xxx addresses (e.g. one static plus one from RA).

If it did align with the underlying function then maybe `allIPv6(index=0)` would be a better name, as it could return global scope 2xxx/3xxx, ULA, and/or link-local.

me-no-dev: I have two global addresses on my ESP's ETH port, so the is why I proposed the index. And yes, it will come from `esp_netif_get_all_ip6()`

libraries/Ethernet/examples/ETH_W5500_IDF_SPI/ETH_W5500_IDF_SPI.ino
```diff
@@ -94,7 +94,7 @@ void setup()
 {
   Serial.begin(115200);
   Network.onEvent(onEvent);
-  ETH.begin(ETH_TYPE, ETH_ADDR, ETH_CS, ETH_IRQ, ETH_RST, ETH_SPI_HOST, ETH_SPI_SCK, ETH_SPI_MISO, ETH_SPI_MOSI);
+  ETH.begin(ETH_PHY_TYPE, ETH_PHY_ADDR, ETH_PHY_CS, ETH_PHY_IRQ, ETH_PHY_RST, ETH_PHY_SPI_HOST, ETH_PHY_SPI_SCK, ETH_PHY_SPI_MISO, ETH_PHY_SPI_MOSI);
```
TD-er, line 97: From experience, I can tell you that using the other `begin()` function which accepts a SPI host instance as argument is way easier to use.
Especially when you also intend to use the same SPI bus for something else like a display.
So can you at least add some comment to this example advising to use the one with a SPI host argument if they intend to use the bus for something else?

That will for sure save quite a lot of hours for those running into this.

me-no-dev: there are two examples. One with IDF SPI and one with Arduino SPI

sgryphon:
> @sgryphon Will be helpful if you provide solutions or be able to test what you propose.

Happy to test out any code or see what I can do.

I just finished putting up a pull request to ESP-IDF to update their basic example to support all network configurations, so it works no matter if you are IPv4-only, IPv6-only, or dual-stack https://github.com/espressif/esp-idf/pull/13250

I had been trying to get it working, to at least know what options needed to be turned on in ESP-IDF to enable IPv6 properly, so that I know what is needed for ESP32 Arduino.

Their previous example:
* Only worked properly out of the box on IPv4 only networks, but only because IPv6 link-local is always there.
* On dual-stack networks it kind of worked by simply using IPv4 for everything and just ignoring IPv6. This meant IPv6 only destinations didn't work, and it didn't use IPv6 even though it was available.
* Didn't work at all on IPv6-only networks.

I updated the example to fully turn on IPv6, so it gets both an address (SLAAC) and IPv6 DNS (either RDNSS or DHCPv6, whichever is available).  I then changed it instead of waiting for both addresses to wait for any address (the first one), configuring IPv6 preference to a global address.

This means the example runs as soon as it gets either IPv4 or global IPv6 without waiting for the other, allowing the single example configuration to run on any type of network -- IPv6-only, IPv4-only, or dual stack.  I have tested it on all three types of networks, connecting to IPv4-only, IPv6-only, and dual-stack servers.

Happy to test with my device (M5Stack with ESP32 with WiFi) across my networks Astral (dual-stack; has DNS64, but I can turn that off for testing), Wildspace (IPv6-only, with DNS64), and Shadow (IPv4-only; I only ever use this for backwards compatibility testing).

me-no-dev:
@lucasssvaz @SuGlider @P-R-O-C-H-Y this is ready for testing and review in general.

libraries/Networking/src/ESP_Network_Events.h
```diff
@@ -0,0 +1,169 @@
...
+	arduino_event_info_t event_info;
+} arduino_event_t;
+
+typedef void (*NetworkEventCb)(arduino_event_id_t event);
+typedef std::function<void(arduino_event_id_t event, arduino_event_info_t info)> NetworkEventFuncCb;
+typedef void (*NetworkEventSysCb)(arduino_event_t *event);
+
+typedef size_t network_event_handle_t;
+
+class ESP_Network_Events {
```
sgryphon, line 127: Inconsistent class naming using underscores. Other classes use Pascal case.

libraries/Networking/src/ESP_Network_Interface.h
```diff
@@ -0,0 +1,99 @@
...
+static const int ESP_NETIF_STARTED_BIT          = BIT0;
+static const int ESP_NETIF_CONNECTED_BIT        = BIT1;
+static const int ESP_NETIF_HAS_IP_BIT           = BIT2;
+static const int ESP_NETIF_HAS_LOCAL_IP6_BIT    = BIT3;
+static const int ESP_NETIF_HAS_GLOBAL_IP6_BIT   = BIT4;
+static const int ESP_NETIF_WANT_IP6_BIT         = BIT5;
+
+#define ESP_NETIF_ID_ETH ESP_NETIF_ID_ETH0
+
+class ESP_Network_Interface: public Printable {
```
sgryphon, line 33: Inconsistent class naming using underscores. Other classes use Pascal case.

libraries/Networking/src/ESP_Network_Manager.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include "ESP_Network_Events.h"
+#include "IPAddress.h"
+#include "WString.h"
+
+class ESP_Network_Manager : public ESP_Network_Events {
```
sgryphon, line 12: Inconsistent class naming using underscores. Other classes use Pascal case.

libraries/Networking/src/ESP_Network_Interface.h
```diff
@@ -0,0 +1,99 @@
...
+        bool config(IPAddress local_ip = (uint32_t)0x00000000, IPAddress gateway = (uint32_t)0x00000000, IPAddress subnet = (uint32_t)0x00000000, IPAddress dns1 = (uint32_t)0x00000000, IPAddress dns2 = (uint32_t)0x00000000, IPAddress dns3 = (uint32_t)0x00000000);
+        bool dnsIP(uint8_t dns_no, IPAddress ip);
+
+        const char * getHostname() const;
+        bool setHostname(const char * hostname) const;
+
+        bool started() const;
+        bool connected() const;
+        bool hasIP() const;
+        bool hasLocalIPv6() const;
```
sgryphon, line 48: For link-local scope IPv6 use the name `hasLinkScopeIPv6()` (or `hasLinkLocalScopeIPv6()`). Do not use just the name local, as it is incorrect (it is link-local), and also confusing with term local vs remote, used for connections.

libraries/Networking/src/ESP_Network_Interface.h
```diff
@@ -0,0 +1,99 @@
...
+        uint8_t * macAddress(uint8_t* mac) const;
+        String macAddress() const;
+        IPAddress localIP() const;
+        IPAddress subnetMask() const;
+        IPAddress gatewayIP() const;
+        IPAddress dnsIP(uint8_t dns_no = 0) const;
+        IPAddress broadcastIP() const;
+        IPAddress networkID() const;
+        uint8_t subnetCIDR() const;
+        IPAddress localIPv6() const;
```
sgryphon, line 66: For link-local scope IPv6 use the name `hasLinkScopeIPv6()` (or `hasLinkLocalScopeIPv6()`). Do not use just the name local, as it is incorrect (it is link-local), and also confusing with term local vs remote, used for connections, and the term localIP() used for IPv4 (which actually usually returns a global scope address).

Local and remote are the different ends of a connection.

Link-local and global are different scopes.

The well established function `localIP()` for IPv4 returns a local (vs remote) address of global scope.

Do not use the same name for IPv6 to mean a different thing.

libraries/WiFi/src/WiFiAP.cpp
```diff
@@ -432,14 +204,7 @@ bool WiFiAPClass::softAPenableIPv6(bool enable)
...
-    static esp_ip6_addr_t addr;
-    if(WiFiGenericClass::getMode() == WIFI_MODE_NULL){
-        return IPAddress(IPv6);
-    }
-
-    if(esp_netif_get_ip6_linklocal(get_esp_interface_netif(ESP_IF_WIFI_STA), &addr)){
-        return IPAddress(IPv6);
-    }
-    return IPAddress(IPv6, (const uint8_t *)addr.addr, addr.zone);
+    return AP.localIPv6();
```
sgryphon, line 207: If this is a link-local address, then it is probably not the best solution for an access point.  Link-local addresses require a scope suffix and have known issues with access via a browser (Windows happens to work without a suffix if it is the default link; other OS do not work).

For IPv6, an AP should assign a Unique Local Address range (ULA), which is the equivalent of a private address range, and then use Router Advertisement (RA) to assign addresses to clients (similar to DHCP).

You can use either random ULAs and access point addresses, or set known static short values, e.g. set fc00::1 as the static address for the AP (similar to setting 192.168.0.1 as a static IPv4 AP).

sgryphon: i.e. Rather than AP.linkLocalIPv6() this should return `AP.globalIPv6()`, which is similar to IPv4 (IPv4 localIP() returns a global IPv4 address, for the local (vs remote) node).

me-no-dev: I have no issue with link local addresses here. Running on Mac and Linux things (also ESPs). The AP provides LinkLocal IPv6 and this is what is returned here. I'm renaming it to show that. As for Global, I was not able to get that going on my ESP

libraries/Networking/src/NetworkInterface.h
```diff
@@ -0,0 +1,99 @@
...
+        uint8_t * macAddress(uint8_t* mac) const;
+        String macAddress() const;
+        IPAddress localIP() const;
+        IPAddress subnetMask() const;
+        IPAddress gatewayIP() const;
+        IPAddress dnsIP(uint8_t dns_no = 0) const;
+        IPAddress broadcastIP() const;
+        IPAddress networkID() const;
+        uint8_t subnetCIDR() const;
+        IPAddress linkLocalIPv6() const;
```
sgryphon, line 66: :+1: :+1: :+1:

libraries/Networking/src/NetworkManager.cpp
```diff
@@ -0,0 +1,177 @@
...
+ *          else error code
+ */
+int NetworkManager::hostByName(const char* aHostname, IPAddress& aResult, bool preferV6)
+{
+    err_t err = ERR_OK;
+    gethostbynameParameters_t params;
+
+    // This should generally check if we have a global address assigned to one of the interfaces.
+    // If such address is not assigned, there is no point in trying to get V6 from DNS as we will not be able to reach it.
+    // That is of course, if 'preferV6' is not set to true
```
sgryphon, line 90: NOTE: This should work in most cases; I will try and test the different scenario combinations (e.g. v4-only vs v6-only vs dual stack networks and destinations), for host lookup.

However when I was exploring ESP-IDF while I got this approach to work for HTTP calls, it did not work for HTTPS -- for HTTPS I had to pass in the hostname (needed to validate certificate) and the TLS component did the lookup, which I had no control over.

Internally it used lwip_getaddrinfo(), which at best does V4_V6 preference order. 

I will also try and test HTTPS scenarios.

I have a PR up for ESP-LWIP to implement RFC 6724 destination address ordering, based on available source addresses. If Arduino-ESP32 has a similar problem them it will need to wait for the upstream fix. (But I haven't checked yet).

libraries/WiFi/src/AP.cpp
```diff
@@ -0,0 +1,343 @@
...
+    esp_err_t err = esp_wifi_set_bandwidth(WIFI_IF_AP, bandwidth);
+    if(err){
+        log_e("Could not set AP bandwidth! 0x%x: %s", err, esp_err_to_name(err));
+        return false;
+    }
+
+    return true;
+}
+
+String APClass::SSID(void) const{
```
sgryphon, line 284: Is APClass meant to be generic for all networks? i.e. how does it relate to WiFiAP? 

SSID is a WiFi thing, i.e. won't apply for Ethernet.

I guess the name "Access Point" is specific to WiFi, i.e. it doesn't mean a general TCP/IP router/DHCP host (such handing out IP addresses over Ethernet), but only doing so over WiFi.

So probably okay.

me-no-dev: APClass is for WiFiAP, but since WiFiAPClass is taken for now, I used that. You should note that AP, STA and ETH all extend NetworkInterface, so in the implementations you see mostly things specific to that type of interface. SSID for WiFi, linkSpeed for ETH and so on.

libraries/Networking/src/NetworkInterface.cpp
```diff
@@ -0,0 +1,719 @@
...
+    } else {
+        d.ip.u_addr.ip4.addr = 0;
+    }
+    if(esp_netif_set_dns_info(_esp_netif, (esp_netif_dns_type_t)dns_no, &d) != ESP_OK){
+        return false;
+    }
+    return true;
+}
+
+bool NetworkInterface::config(IPAddress local_ip, IPAddress gateway, IPAddress subnet, IPAddress dns1, IPAddress dns2, IPAddress dns3)
```
sgryphon, line 328: The notes in the header say `// For server interfaces (WiFi AP), dns1 is the DHCP lease range start and dns2 is the DNS. dns3 is not used`.

That is a bit weird, having the fourth parameter be the DHCP range, and only one DNS.

Maybe two separate functions for `config()` [normal] and `configServer(local_ip, gateway, subnet, dhcp_range, dns)`.  The functions could check `_is_server_if` and return an error / failure if it is not the right type (e.g. call server config if not a server).

libraries/Networking/src/NetworkInterface.cpp
```diff
@@ -0,0 +1,719 @@
...
+    } else {
+        info.ip.addr = 0;
+        info.gw.addr = 0;
+        info.netmask.addr = 0;
+        d1.ip.u_addr.ip4.addr = 0;
+        d2.ip.u_addr.ip4.addr = 0;
+        d3.ip.u_addr.ip4.addr = 0;
+	}
+
+    if(_is_server_if){
```
sgryphon, line 358: Things missing from the server config is IPv6, which uses a different approach (RA instead of DHCPS).

I don't think adding IPv6 AP should be part of this change, but a separate thing.  Actually maybe the first thing to add would be IPv6 STA config with a static address, DNS, etc. (also a separate thing).

A long journey, step by step.

libraries/Networking/src/NetworkInterface.cpp
me-no-dev: it's more than we have now. We can think of something less weird later. I don't even guarantee that the DNS server is properly advertised by the DHCPS

libraries/Networking/src/NetworkInterface.cpp
```diff
@@ -0,0 +1,719 @@
...
+    bytes += out.print("      ");
+    bytes += out.print("gateway ");
+    bytes += out.print(gatewayIP());
+    bytes += out.print(" dns ");
+    bytes += out.print(dnsIP());
+    bytes += out.println();
+
+    static const char * types[] = { "UNKNOWN", "GLOBAL", "LINK_LOCAL", "SITE_LOCAL", "UNIQUE_LOCAL", "IPV4_MAPPED_IPV6" };
+    esp_ip6_addr_t if_ip6[CONFIG_LWIP_IPV6_NUM_ADDRESSES];
+    int v6addrs = esp_netif_get_all_ip6(_esp_netif, if_ip6);
```
sgryphon, line 707: It would be nice to have a way to access these if needed, e.g. `allIPv6(uint8_t ipv6_no)`, very similar to `dnsIP()`.

The function just calls esp_netif_get_all_ipv6() and if ipv6_no > v6addrs (or zero) then returns the empty default address. Not very efficient if you loop through them all, but better than nothing.

(Not sure if ipv6_no should be 0-based (normal) or 1-based (like `dnsIP()`).  Maybe call it `ipv6_index` for a 0-based array, i.e. 0 returns `IPAddress(IPv6, (const uint8_t *)if_ip6[0].addr, if_ip6[0].zone)` and so on.

me-no-dev: you do have access. All you need to get them is the esp_netif, which you get through `ETH.netif()`, `WiFi.STA.netif()` or `WiFi.AP.netif()`. Just not "Arduino" level API, as such is too early to be decided.

libraries/WiFi/src/WiFiAP.cpp
sgryphon: Happy with the appropriate name; globalIPv6() can come later.

For an IPv6 AP there is quite a bit of set up, similar to setting up DHCPS for IPv4, static address, etc. Except IPv6 uses RA as the primary mechanism.  If all you want is to connect to the device over WiFi, then plain RA is all you need; for an actual router (or host name usage) you would need a way to also provide DNS.  

I don't even know what is possible here in ESP-IDF, but definitely a separate thing.

libraries/Networking/src/NetworkInterface.cpp
sgryphon: That's fine then.

sgryphon: I like the work so far; it is making good progress.

libraries/ESPmDNS/src/ESPmDNS.cpp
lucasssvaz: @me-no-dev Is this still valid ?

libraries/WiFiProv/src/WiFiProv.cpp
```diff
@@ -17,6 +17,10 @@
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
     
 */
+#pragma once
```
SuGlider, line 20: #pragma once is necessary in a cpp file?
```suggestion
```

SuGlider: it seems to be a copy&paste result.

me-no-dev: funny... I do not remember touching this file at all 😆 maybe it felt left out and decided to make a statement

libraries/ESPmDNS/src/ESPmDNS.cpp
me-no-dev: not anymore :)

libraries/Network/src/NetworkClient.h
```diff
@@ -0,0 +1,115 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#ifndef _WIFICLIENT_H_
+#define _WIFICLIENT_H_
```
lucasssvaz, line 21:
```suggestion
#pragma once
```

libraries/Network/src/NetworkClient.h
```diff
@@ -0,0 +1,115 @@
...
+    IPAddress localIP() const;
+    IPAddress localIP(int fd) const;
+    uint16_t localPort() const;
+    uint16_t localPort(int fd) const;
+
+    //friend class NetworkServer;
+    using Print::write;
+};
+
+#endif /* _WIFICLIENT_H_ */
```
lucasssvaz, line 115:
```suggestion
```

libraries/Network/src/NetworkInterface.cpp
```diff
@@ -0,0 +1,719 @@
...
+int NetworkInterface::waitStatusBits(int bits, uint32_t timeout_ms) const {
+    if(!_interface_event_group){
+        return 0;
+    }
+    bits = xEventGroupWaitBits(
+        _interface_event_group,    // The event group being tested.
+        bits,  // The bits within the event group to wait for.
+        pdFALSE,         // bits should be cleared before returning.
+        pdTRUE,        // Don't wait for all bits, any bit will do.
+        timeout_ms / portTICK_PERIOD_MS ) & bits; // Wait a maximum of timeout_ms for any bit to be set.
```
lucasssvaz, line 229:
```suggestion
        _interface_event_group, // The event group being tested.
        bits, // The bits within the event group to wait for.
        pdFALSE, // bits should be cleared before returning.
        pdTRUE, // Don't wait for all bits, any bit will do.
        timeout_ms / portTICK_PERIOD_MS ) & bits; // Wait a maximum of timeout_ms for any bit to be set.
```

libraries/Network/src/NetworkInterface.cpp
```diff
@@ -0,0 +1,719 @@
...
+    for (int i = 0; i < v6addrs; ++i){
+        bytes += out.print("      ");
+        bytes += out.print("inet6 ");
+        bytes += IPAddress(IPv6, (const uint8_t *)if_ip6[i].addr, if_ip6[i].zone).printTo(out, true);
+        bytes += out.print(" type ");
+        bytes += out.print(types[esp_netif_ip6_get_addr_type(&if_ip6[i])]);
+        bytes += out.println();
+    }
+
+    // bytes += out.println();
```
lucasssvaz, line 717:
```suggestion
```

libraries/Network/src/NetworkManager.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "esp_mac.h"
+
+NetworkManager::NetworkManager(){
+
+}
+
+bool NetworkManager::begin(){
+    static bool initialized = false;
+    if(!initialized){
+        initialized = true;
```
lucasssvaz, line 21: I don't think this has any effect
```suggestion
```

libraries/Network/src/NetworkServer.h
```diff
@@ -0,0 +1,62 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+#ifndef _WIFISERVER_H_
+#define _WIFISERVER_H_
```
lucasssvaz, line 20:
```suggestion
#pragma once
```

libraries/Network/src/NetworkServer.h
```diff
@@ -0,0 +1,62 @@
...
+    bool hasClient();
+
+    void end();
+    void close();
+    void stop();
+    operator bool(){return _listening;}
+    int setTimeout(uint32_t seconds);
+};
+
+#endif /* _WIFISERVER_H_ */
```
lucasssvaz, line 62:
```suggestion
```

libraries/Network/src/NetworkUdp.h
```diff
@@ -0,0 +1,77 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * bjoern@cs.stanford.edu 12/30/2008
+ */
+
+#ifndef _WIFIUDP_H_
+#define _WIFIUDP_H_
```
lucasssvaz, line 36:
```suggestion
#ifndef _NETWORKUDP_H_
#define _NETWORKUDP_H_
```

libraries/Network/src/NetworkUdp.h
```diff
@@ -0,0 +1,77 @@
...
+  int read();
+  int read(unsigned char* buffer, size_t len);
+  int read(char* buffer, size_t len);
+  int peek();
+  void flush();
+  IPAddress remoteIP();
+  uint16_t remotePort();
+};
+
+#endif /* _WIFIUDP_H_ */
```
lucasssvaz, line 77:
```suggestion
#endif /* _NETWORKUDP_H_ */
```

libraries/WiFi/src/AP.cpp
```diff
@@ -0,0 +1,343 @@
...
+        bytes += out.print(",STA:");
+        bytes += out.print(clients.num);
+    }
+    return bytes;
+}
+
+
+
+
+#endif /* SOC_WIFI_SUPPORTED */
```
lucasssvaz, line 343:
```suggestion
}

#endif /* SOC_WIFI_SUPPORTED */
```

libraries/WiFi/src/STA.cpp
```diff
@@ -0,0 +1,795 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include "WiFi.h"
+#include "WiFiGeneric.h"
+#include "WiFiSTA.h"
+#if SOC_WIFI_SUPPORTED
+
```
lucasssvaz, line 10:
```suggestion
```

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -784,416 +344,35 @@ WiFiGenericClass::WiFiGenericClass()
...
 
 /**
  * callback for WiFi events
  * @param arg
  */
-esp_err_t WiFiGenericClass::_eventCallback(arduino_event_t *event)
+void WiFiGenericClass::_eventCallback(arduino_event_t *event)
 {
-    static bool first_connect = true;
+    if(!event) return;                                                       //Null would crash this function
```
lucasssvaz, line 371:
```suggestion
    if(!event) return; //Null would crash this function
```

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1589,103 +728,21 @@ bool WiFiGenericClass::setDualAntennaConfig(uint8_t gpio_ant1, uint8_t gpio_ant2
...
-
-/**
- * Resolve the given hostname to an IP address.
- * @param aHostname     Name to be resolved
- * @param aResult       IPAddress structure to store the returned IP address
- * @return 1 if aIPAddrString was successfully converted to an IP address,
- *          else error code
- */
+/*
+ * Deprecated Methods
```
lucasssvaz, line 732: Can we mark all the newly deprecated functions with the `deprecated` attribute ?
https://en.cppreference.com/w/cpp/language/attributes/deprecated

package/package_esp32_index.template.json
lucasssvaz, line 1: Just confirming. Is this supposed updated in this PR or is it a leftover from tests ?

libraries/Network/src/NetworkServer.h
```diff
@@ -0,0 +1,62 @@
+/*
+  Server.h - Server class for Raspberry Pi
```
lucasssvaz, line 2: Maybe we should update this line to reflect the actual description of this file. I think as long as the copyright and the license stays untouched it should be fine. (might be wrong though)

libraries/Network/src/NetworkClient.h
```diff
@@ -0,0 +1,115 @@
+/*
+  Client.h - Base class that provides Client
```
lucasssvaz, line 2: Maybe we should update this line to reflect the actual description of this file. I think as long as the copyright and the license stays untouched it should be fine. (might be wrong though)

lucasssvaz (commented):
Looks very good. Just some suggestions. Everything I've tested so far works flawlessly. Will do more testing and report if anything is out of the ordinary.

libraries/Network/src/NetworkClient.h
me-no-dev: yeah, I am also a bit scared of touching other's headers, so I would rather not

libraries/Network/src/NetworkManager.cpp
me-no-dev: it has in the case where two tasks from separate cores decide to call `Network.begin()`. Not the best solution, but does well so far :)

libraries/WiFi/src/WiFiGeneric.cpp
me-no-dev: No. There is too much code out there that uses those functions. Let's roll out as-is and re-think deprecation with 3.1.

package/package_esp32_index.template.json
me-no-dev: I had to update the libs for some things to work

P-R-O-C-H-Y (approved):

lucasssvaz (approved):

sgryphon:
I know this has already been merged. It is still not working that well for IPv6. I tested across multiple network types:

Network | Dual-Stack | IPv6 | IPv4 | TLS Dual-Stack
IPv4 (Shadow) | IPv4 | Not Possible | Yes | Yes
Dual-stack+NAT64 (Astral) | IPv4* | DNS fail | Yes* | DNS fail
IPv6+NAT64 (Wildspace) | DNS fail | DNS fail | DNS fail | DNS fail

* Dual-stack initially worked for some, but then started failing. I suspect it is lack of support for IPv6, and because I have DNS64 everything has an IPv6 address. Could be a race condition that worked before DNS64 was resolved.

Some of the errors:

Dual-stack+NAT64 network, IPv6 destination:

```
[ 69767][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: Button 2, scenario 1, v0.1.0-156-gaf345f5
[ 69779][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: Global IPv6 2407:8800:bc61:1340:a3a:f2ff:fe65:db28
[ 69794][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: IPv4 192.168.1.154
[ 69802][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: Link-Local IPv6 fe80::a3a:f2ff:fe65:db28%st1
sta: <UP>
      ether 08:3A:F2:65:DB:28
      inet 192.168.1.154 netmask 255.255.255.0 broadcast 192.168.1.255
      gateway 192.168.1.1 dns 192.168.1.1
      inet6 fe80::a3a:f2ff:fe65:db28%st1 type LINK_LOCAL
      inet6 2407:8800:bc61:1340:a3a:f2ff:fe65:db28 type GLOBAL
      inet6 fd7c:e25e:67e8:40:a3a:f2ff:fe65:db28 type UNIQUE_LOCAL
[ 69840][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: DNS0 192.168.1.1
[ 69854][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: DNS1 0.0.0.0
[ 69862][I][Core2Logger.cpp:198] log(): [Core2Logger] CORE2: URL: http://v6.ipv6-test.com/api/myip.php
[ 69875][D][HTTPClient.cpp:303] beginInternal(): protocol: http, host: v6.ipv6-test.com port: 80 url: /api/myip.php
[ 69885][D][HTTPClient.cpp:598] sendRequest(): request type: 'GET' redirCount: 0

[ 76427][E][NetworkManager.cpp:125] hostByName(): DNS Failed for 'v6.ipv6-test.com' with error '-5' and result '-1'
[ 81441][I][NetworkClient.cpp:269] connect(): select returned due to timeout 5000 ms for fd 48
[ 81450][D][HTTPClient.cpp:1163] connect(): failed connect to v6.ipv6-test.com:80
[ 81458][W][HTTPClient.cpp:1486] returnError(): error(-1): connection refused
[ 81465][E][Core2Logger.cpp:192] log(): [Core2Logger] CORE2: HTTP GET error -1: connection refused
```

me-no-dev:
@sgryphon given that you have a full environment and test case, maybe you can figure out what is missing? Try maybe some menuconfig options and rebuild the libs?

IgorTitov:
Will you fix the breaking changes that are mentioned here?: [https://forum.arduino.cc/t/espasync-wifimanager-library-has-no-member-named-setautoconnect/1299741](https://forum.arduino.cc/t/espasync-wifimanager-library-has-no-member-named-setautoconnect/1299741)

Jason2866:
@IgorTitov There are valid reasons for this change. So no.

PR #8762, opened 2023-10-14
HardwareSerial::end(bool) review + Baud Rate detection review and example

SuGlider:
## Description of Change
This PR removes UART pins detaching whenever it is used to detect baud rate.
Such feature only works with ESP32 and ESP32-S2. 
It also fixes error message when trying to use ESP32-S3, ESP32-C3, ESP32-C6 and ESP32-H2 with baud rate detection.

In addition the PR also add an example about how to use baud rate detection feature.

## Tests scenarios
Tested with ESP32 and ESP32-S2 UART0 using the Example provided in this PR.

## Related links
None

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -183,7 +183,7 @@ _eventTask(NULL)
 
 HardwareSerial::~HardwareSerial()
 {
-    end();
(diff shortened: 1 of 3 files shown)


lucasssvaz (approved):
LGTM, thanks for the changes!

Issue #8772, opened 2023-10-16
I2C - Wire Library crashing when setting clock frequency > 142857 Hz


blackhack:
### Board

ESP32 Dev Module

### Device Description

ESP32 Dev Module, nothing connected.

### Hardware Configuration

Just the bare board, it is not necessary to have an I2C device connected to reproduce the crash.

### Version

latest master (checkout manually)

### IDE Name

VSCode PlatformIO, Arduino IDE

### Operating System

Windows 11

### Flash frequency

80 MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

In the latest version with IDF 5.1, something has changed, and whenever we attempt to set the clock speed for the I2C bus to a value greater than 142857 Hz (e.g., 400 KHz), this assertion `assert(scl_high > 13)` fails.

### Sketch

```cpp
#include "Wire.h"

void setup() {
  Serial.begin(115200);
  Wire.begin();
  Wire.setClock(400000);
}

void loop() {
}
```

### Debug Message

```plain
rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:1416
load:0x40078000,len:14804
load:0x40080400,len:4
load:0x40080404,len:3356
entry 0x4008059c

assert failed: i2c_ll_set_bus_timing i2c_ll.h:109 (scl_high > 13)


Backtrace: 0x40082529:0x3ffb2080 0x40088335:0x3ffb20a0 0x4008d362:0x3ffb20c0 0x400da30e:0x3ffb21f0 0x400d229d:0x3ffb2210 0x400d163f:0x3ffb2240 0x400d145d:0x3ffb2260 0x400d333d:0x3ffb2290




ELF file SHA256: 067952623cffaa5f

Rebooting...
ets Jun  8 2016 00:22:57  

DECODED LOG:

#0  0x40082812:0x3ffb4f50 in panic_abort at C:\Users\david\.platformio\packages\framework-espidf\components\esp_system/panic.c:452
  #1  0x40087321:0x3ffb4f70 in esp_system_abort at C:\Users\david\.platformio\packages\framework-espidf\components\esp_system\port/esp_system_chip.c:84
[...]

PR #8777, opened 2023-10-16
Fix Wire.setClock() causing assert

me-no-dev:
When Wire.setClock() is called after Wire.begin() an assert is triggered due to changes in clock calculations

fixes: https://github.com/espressif/arduino-esp32/issues/8772


rottaran:
Hello me-no-dev, great fix! 
I got the same issue with ESP32S3 in the idf-release/v5.1 branch. I copied your changes and it fixed the problem too.

PR #8787, opened 2023-10-21
Fix "fatal error: esp_partition.h: No such file or directory"

Msq001:
according to https://docs.espressif.com/projects/esp-idf/zh_CN/v5.0/esp32/migration-guides/release-5.x/storage.html


me-no-dev:
Thanks!

PR #8798, opened 2023-10-23
Serial as a #define

SuGlider:
## Description of Change
The PR changes `Serial` to be a pre-processor symbol instead of a HardwareSerial/HWCDC/USBCDC object.
Therefore, UART0 is always the `Serial0` object.

If the sketch doesn't use `CDC on Boot`, `Serial` will defined to be `Serial0` by default.
Whenever the sketch is set to use `CDC on Boot`, `Serial` will defined to be `USBSerial` object, which in turn will point to a HWCDC or USBCDC object depending on how it is configured / which SoC is used.

The PR will make `Serial0`, `Serial1` and `Serial2` (if it exists) always available as UART0/1/2.
`USBSerial` will also always be available whenever the SoC has a CDC interface.

If necessary the sketch can change/force the redefinition of `Serial` symbol to whatever necessary.

The PR also adds a `HWCDCSerialEvent()` callback that is executed after `loop()` whenever there are bytes available in the HW Serial Interface.
The same for `USBSerialEvent()` in Native USB CDC. Those two functions can be defined in the sketch by the user and work in the same way as `SerialEvent()` works.

## Tests scenarios
Used "OnReceive_Demo.ino" to test it using ESP32/ESP32S2/ESP32C3/ESP32S3 with different configurations for USB, whenever possible.

## Related links
None

### DIFF
--- a/cores/esp32/HWCDC.h
+++ b/cores/esp32/HWCDC.h
@@ -103,12 +103,13 @@ public:
 
 };
 
-#if ARDUINO_USB_MODE
-#if ARDUINO_USB_CDC_ON_BOOT//Serial used for USB CDC
-extern HWCDC Serial;
-#else
+#if ARDUINO_USB_MODE         // Hardware JTAG CDC selected
+#if ARDUINO_USB_CDC_ON_BOOT  //Serial used for USB CDC
+// When using CDC on Boot, Arduino Serial is the USB device
+#define Serial USBSerial
+#endif
+// USBSerial is always available to used
 extern HWCDC USBSerial;
 #endif
-#endif
 
-#endif /* CONFIG_IDF_TARGET_ESP32C3 */
+#endif /* SOC_USB_SERIAL_JTAG_SUPPORTED */
--- a/cores/esp32/HWCDC.h
+++ b/cores/esp32/HWCDC.h
@@ -102,14 +102,12 @@ public:
     uint32_t baudRate(){return 115200;}
 
 };
-
-#if ARDUINO_USB_MODE         // Hardware JTAG CDC selected
-#if ARDUINO_USB_CDC_ON_BOOT  //Serial used for USB CDC
-// When using CDC on Boot, Arduino Serial is the USB device
-#define Serial USBSerial
+#if ARDUINO_USB_MODE  // Hardware JTAG CDC selected
+#ifndef HWCDC_SERIAL_IS_DEFINED
+#define HWCDC_SERIAL_IS_DEFINED 1
 #endif
-// USBSerial is always available to be used
-extern HWCDC USBSerial;
+// HWCDCSerial is always available to be used
+extern HWCDC HWCDCSerial;
 #endif
 
 #endif /* SOC_USB_SERIAL_JTAG_SUPPORTED */
--- a/cores/esp32/HWCDC.h
+++ b/cores/esp32/HWCDC.h
@@ -108,7 +108,7 @@ public:
 // When using CDC on Boot, Arduino Serial is the USB device
 #define Serial USBSerial
 #endif
-// USBSerial is always available to used
+// USBSerial is always available to be used
 extern HWCDC USBSerial;
 #endif
 
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -122,10 +122,27 @@ HardwareSerial Serial1(1);
 HardwareSerial Serial2(2);
 #endif
 
+#if HWCDC_SERIAL_IS_DEFINED == 1        // Hardware JTAG CDC Event
+extern void HWCDCSerialEvent (void)__attribute__((weak));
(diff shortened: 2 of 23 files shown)


cores/esp32/HardwareSerial.cpp
```diff
@@ -127,11 +124,8 @@ HardwareSerial Serial2(2);
 
 void serialEventRun(void)
 {
-#if ARDUINO_USB_CDC_ON_BOOT //Serial used for USB CDC
-    if(Serial0.available()) serialEvent();
-#else
+    // Serial is always defined as something (UART0 or USBSerial)
     if(Serial.available()) serialEvent();
```
me-no-dev, line 128: Should be `Serial0.available()`

libraries/USB/examples/FirmwareMSC/FirmwareMSC.ino
```diff
@@ -9,13 +9,7 @@ void loop(){}
 #if !ARDUINO_USB_MSC_ON_BOOT
 FirmwareMSC MSC_Update;
 #endif
-#if ARDUINO_USB_CDC_ON_BOOT
 #define HWSerial Serial0
```
me-no-dev, line 12: you no longer need this define. Change the rest of the example code to call `Serial0` for UART and `USBSerial` for CDC. Same for the other applicable examples.

cores/esp32/HWCDC.h
```diff
@@ -103,12 +103,13 @@ class HWCDC: public Stream
 
 };
 
-#if ARDUINO_USB_MODE
-#if ARDUINO_USB_CDC_ON_BOOT//Serial used for USB CDC
-extern HWCDC Serial;
-#else
-extern HWCDC USBSerial;
+#if ARDUINO_USB_MODE         // Hardware JTAG CDC selected
+#if ARDUINO_USB_CDC_ON_BOOT  //Serial used for USB CDC
```
me-no-dev, line 107: See my comment about `USBCDC.h`

cores/esp32/HardwareSerial.h
```diff
@@ -240,15 +241,14 @@ extern void serialEventRun(void) __attribute__((weak));
 #ifndef ARDUINO_USB_CDC_ON_BOOT
 #define ARDUINO_USB_CDC_ON_BOOT 0
 #endif
-#if ARDUINO_USB_CDC_ON_BOOT //Serial used for USB CDC
-#if !ARDUINO_USB_MODE
-#include "USB.h"
-#include "USBCDC.h"
+#if !ARDUINO_USB_CDC_ON_BOOT //Serial used for USB CDC
```
me-no-dev, line 244: here you should have the checks from both CDCs as well. One place for all :)

cores/esp32/HardwareSerial.cpp
SuGlider: I was thinking about it yesterday. If I use `Serial.available()`, it works with the CDC as well... 
`Serial0` will limit it to UART only. 
Any opinion?

libraries/USB/examples/FirmwareMSC/FirmwareMSC.ino
SuGlider: That's something I was thinking about yesterday as well... 
I left it there because the user could change it to any other Print communication channel...
What do you say? We can change it to "PrintObj" instead of "HWSerial"

cores/esp32/USBCDC.h
```diff
@@ -141,8 +141,13 @@ class USBCDC: public Stream
...
 };
 
-#if ARDUINO_USB_CDC_ON_BOOT && !ARDUINO_USB_MODE //Serial used for USB CDC
-extern USBCDC Serial;
+#if !ARDUINO_USB_MODE        // Native USB CDC selected
+#if ARDUINO_USB_CDC_ON_BOOT  //Serial used for USB CDC
+// When using CDC on Boot, Arduino Serial is the USB device
+#define Serial USBSerial
+#endif
+// USBSerial is always available to be used
```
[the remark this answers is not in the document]
SuGlider: OK.

libraries/USB/examples/FirmwareMSC/FirmwareMSC.ino
me-no-dev: Looking at the code, I think that with those changes, it needs to just call Serial and not even check this define. If it's set to be UART, it will go there, else whatever CDC it's set to.

me-no-dev: Replace all HWSerial with Serial and delete references to USBSerial

SuGlider: OK.

SuGlider:
@me-no-dev - the changes shall be addressed now. I also added a way to run `SerialEvent()` like callbacks for HW and USB CDC.

PR #8813, opened 2023-10-27
RMT: Fix RMT TX interrupt level

lucasssvaz:
## Description of Change

This PR aims to fix RMT failing to start due to the uninitialized interrupt value. This would cause the interrupt to be initialized with garbage memory values:

```
E (18193) rmt: rmt_new_tx_channel(210): invalid interrupt priority:-2113922172
```

## Tests scenarios

Tested with ESP32-S3-DevKitC-1 using the BlinkRGB example.


P-R-O-C-H-Y (approved):
LGTM!

QnBarb:
Do you have a simple RMT example to pulse an infrared led and example for a receiver? I'm just trying to shoot infrared pulses between two mcu's. I've tried the nec_example by espressif, but no luck. I'm using arduino, given that, I have little experience at the moment.
Please point me to a good source if possible. Thank you in advance.

PR #5768, opened 2021-10-14
Add another Wire requestFrom function signature

torntrousers:
Arduino HardwareI2C defines a requestFrom function
```
    virtual size_t requestFrom(uint8_t address, size_t len, bool stopBit) = 0;
```
whereas the ESP32 Wire doesn't include any requestFrom using size_t types, so this can cause a compile error, eg when using the Arduino ATECC608 library with an ESP32, see arduino-libraries/ArduinoECCX08#25

PR #8817, opened 2023-10-30
Fix ambiguous for TwoWire::requestFrom() methods and align API with Arduino.cc

safocl:
## Description of Change
Remove (comment out) the TwoWire::requestFrom() method overloads.
The compiler says "warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second"

## Tests scenarios
[Example 1](https://godbolt.org/z/oMWWP9a49)
[Example 2](https://godbolt.org/z/9ojcfefoh)
[Example 3](https://godbolt.org/z/3bjf46j8a)
All compilers report this warning.

## Related links
[C++17 standard says](https://timsong-cpp.github.io/cppwp/n4659/over.match#best-2) ([over.match.best]): 
```
If there is exactly one viable function that is a better function than all other viable functions, then it is the one selected by overload resolution; otherwise the call is ill-formed.) [ Example:

void Fcn(const int*,  short);
void Fcn(int*, int);

int i;
short s = 0;

void f() {
  Fcn(&i, s);       // is ambiguous because &i → int* is better than &i → const int*
                    // but s → short is also better than s → int

  Fcn(&i, 1L);      // calls Fcn(int*, int), because &i → int* is better than &i → const int*
                    // and 1L → short and 1L → int are indistinguishable

  Fcn(&i, 'c');     // calls Fcn(int*, int), because &i → int* is better than &i → const int*
                    // and c → int is better than c → short
}
 — end example ]
```

There is no viable override for these methods that is better - any overrides require equal conversions of different arguments.

upd: this pull request now reflects the accumulated changes for all errors and warnings about ambiguous TwoWire method overloads. It was decided to make a HardwareI2C interface from Arduino.


safocl:
This is an error, not a warning: `g++` requires the `-pedantic` option to conform to the standard.

safocl:
I believe that additional method overloads are not the best solution for disambiguating overload resolution. I believe that requests https://github.com/espressif/arduino-esp32/pull/5764 and https://github.com/espressif/arduino-esp32/pull/5768 are erroneous. -- That requests are not solve issue https://github.com/arduino-libraries/ArduinoECCX08/issues/25

VojtechBartoska:
Thanks for the PR, same here please sign CLA.

safocl:
> Thanks for the PR, same here please sign CLA.

Oh yes. I didn’t understand at first that this needed to be done.
I did it.

lucasssvaz:
@safocl Please fix the CI errors

safocl:
> @safocl Please fix the CI errors

I did commits.
Need review for TwoWire::begin(uint8_t address) -- it is should be available without slave
support by SoC?

libraries/Wire/src/Wire.h
```diff
@@ -81,12 +83,45 @@ class TwoWire: public Stream
...
+	 * Fixme: it is necessary to determine whether it is possible to use without SOC_I2C_SUPPORT_SLAVE? !!!
+	 */
+    void begin(uint8_t address) override final
+    {
+#if SOC_I2C_SUPPORT_SLAVE
+        begin(address, -1, -1, 0);
+#else
+        log_e("I2C slave is not supported by you SoC!!!");
+#endif
+    }
```
lucasssvaz, line 108: This override in specific is only used with SOC_I2C_SUPPORT_SLAVE. For master mode the address is set in `beginTransmission`.

lucasssvaz (changes requested):
I don't think we should remove the error checking from the function return. It is a good way to check if everything inside the method was correctly checked.

The aim of this PR is fixing the ambiguity of the `requestFrom` methods. I think it is fine to create a `HardwareI2C` class here but removing the return values should be done in another PR so it can be discussed.

libraries/Wire/src/Wire.h
safocl: ok. 
Then it is necessary to change to :
```
#if SOC_I2C_SUPPORT_SLAVE
        begin(address, -1, -1, 0);
#else
        beginTransmission(address);
#endif
```
???

lucasssvaz: You can use `#if SOC_I2C_SUPPORT_SLAVE` around the entire function implementation and declaration. That way it won't be compiled if it does not have support for I2C slave mode.

lucasssvaz: You can put the check in the abstract class as well.

safocl: > You can put the check in the abstract class as well.

Then it will not be compatible with the Arduino protocol.

safocl: I thought that the whole point of this was to implement the Arduino interface

lucasssvaz: I still don't understand what you are trying to do. Your PR description only mentions fixing the requestFrom methods. What is your objective with all these changes in the PR ? Could you please update the description with your reasoning ?

safocl: https://github.com/espressif/arduino-esp32/pull/8818#issuecomment-1829914025

lucasssvaz: Okay, thanks for the clarification. You don't need to change anything. The way you did is correct. Just remove the fixme comment

lucasssvaz: So, just for me to confirm, you are changing everything to match the arduino API, correct ?

safocl: > So, just for me to confirm, you are changing everything to match the arduino API, correct ?

yes.

lucasssvaz (changes requested):
After discussing with the team we agreed that it is better to keep the return types in the abstract class and the implementation even if differ from the Arduino.cc API.

safocl:
> After discussing with the team we agreed that it is better to keep the return types in the abstract class and the implementation even if differ from the Arduino.cc API.

Do I need to return the return types as they were originally and change them in the abstract class itself?

me-no-dev:
Yes. change `HardwareI2C` and revert in `Wire`

libraries/Wire/src/Wire.h
```diff
@@ -81,12 +83,42 @@ class TwoWire: public Stream
...
+
+    bool begin(uint8_t address) override final
+    {
+#if SOC_I2C_SUPPORT_SLAVE
+        return begin(address, -1, -1, 0);
+#else
+        log_e("I2C slave is not supported by you SoC!!!");
+#endif
+    }
```
P-R-O-C-H-Y, line 105: CI is failing here as no return statement if slave is not supported.
```suggestion
    bool begin(uint8_t address) override final
    {
#if SOC_I2C_SUPPORT_SLAVE
        return begin(address, -1, -1, 0);
#else
        log_e("I2C slave is not supported on " CONFIG_IDF_TARGET);
        return false;
#endif
    }
```

P-R-O-C-H-Y (changes requested):
@safocl PTAL

libraries/Wire/src/Wire.cpp
```diff
@@ -606,70 +615,27 @@ int TwoWire::peek(void)
...
 }
 
-uint8_t TwoWire::endTransmission(void)
+// sets function called on slave read
+void TwoWire::onRequest( void (*function)(void) )
 {
-    return endTransmission(true);
+#if SOC_I2C_SUPPORT_SLAVE
+  user_onRequest = function;
+#endif
```
P-R-O-C-H-Y, line 638: Please fix indentation.
```suggestion
#if SOC_I2C_SUPPORT_SLAVE
    user_onRequest = function;
#endif
```

libraries/Wire/src/Wire.h
```diff
@@ -95,78 +128,24 @@ class TwoWire: public Stream
...
-    inline size_t write(unsigned int n)
-    {
-        return write((uint8_t)n);
-    }
-    inline size_t write(int n)
-    {
-        return write((uint8_t)n);
-    }
+
+
```
P-R-O-C-H-Y, line 140: Remove unnecessary empty lines.

P-R-O-C-H-Y (changes requested):
Just small formatting fixes, else looks good.
Tested i2c with OLED, works fine.

PR #8818, opened 2023-10-30
Remove TwoWire::begin(int) overload

safocl:
## Description of Change
Remove `TwoWire:: inline bool begin(int addr)` overload.
Inside the function, this overload truncated the data type to a shorter one. This could break some users' hopes.


safocl:
> maybe we need to drop to this: https://github.com/arduino/ArduinoCore-API/blob/master/api/HardwareI2C.h

It's a good idea to inherit from HardwareI2C

PR #8819, opened 2023-10-31
Rewrite IPAddress without union

safocl:
## Rewrite IPAddress without union
Using a union allowed undefined behavior for array-like INPUT and OUTPUT arguments through a 32-bit integer (and vice versa).

  C++ standard says (https://timsong-cpp.github.io/cppwp/n4659/class.union#1):
  "At most one of the non-static data members of an object of union type
  can be active at any time, that is, the value of at most one of the
  non-static data members can be stored in a union at any time. [ Note:
  One special guarantee is made in order to simplify the use of unions:
  If a standard-layout union contains several standard-layout structs
  that share a common initial sequence, and if a non-static data member
  of an object of this standard-layout union type is active and is one
  of the standard-layout structs, it is permitted to inspect the common
  initial sequence of any of the standard-layout struct members; see
  [class.mem].  — end note ]"

  and (https://timsong-cpp.github.io/cppwp/n4659/class.mem#22):
  "Two standard-layout unions are layout-compatible if they have the same number of non-static data members and corresponding non-static data members (in any order) have layout-compatible types."

you can't hope that compilers don't seem to do undefined behavior at the moment

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -20,62 +20,62 @@
 #include <Arduino.h>
 #include <IPAddress.h>
 #include <Print.h>
+#include <algorithm>
 
-IPAddress::IPAddress()
-{
-    _address.dword = 0;
-}
+// IPAddress::IPAddress()
+// {
+//     _address.dword = 0;
+// }
 
 IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet)
-{
-    _address.bytes[0] = first_octet;
-    _address.bytes[1] = second_octet;
-    _address.bytes[2] = third_octet;
-    _address.bytes[3] = fourth_octet;
-}
+	: _address({first_octet,
+				second_octet,
+    			third_octet,
+    			fourth_octet})
+{}
 
 IPAddress::IPAddress(uint32_t address)
 {
-    _address.dword = address;
+    uint32_t& addressRef = reinterpret_cast<uint32_t&>(_address.front());
+	addressRef = address;
 }
 
-IPAddress::IPAddress(const uint8_t *address)
-{
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
-}
+IPAddress::IPAddress(const uint8_t *address) : _address({address[0], address[1], address[2], address[3]})
+{}
 
 IPAddress& IPAddress::operator=(const uint8_t *address)
 {
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+	std::copy(address, address + _address.size(), _address.begin());
     return *this;
 }
 
 IPAddress& IPAddress::operator=(uint32_t address)
 {
-    _address.dword = address;
+    uint32_t& addressRef = reinterpret_cast<uint32_t&>(_address.front());
+	addressRef = address;
     return *this;
 }
 
 bool IPAddress::operator==(const uint8_t* addr) const
 {
-    return memcmp(addr, _address.bytes, sizeof(_address.bytes)) == 0;
+    return std::equal(_address.begin(), _address.end(), addr);
 }
 
 size_t IPAddress::printTo(Print& p) const
 {
     size_t n = 0;
     for(int i = 0; i < 3; i++) {
-        n += p.print(_address.bytes[i], DEC);
+        n += p.print(_address[i], DEC);
         n += p.print('.');
     }
-    n += p.print(_address.bytes[3], DEC);
+    n += p.print(_address[3], DEC);
     return n;
 }
 
 String IPAddress::toString() const
 {
     char szRet[16];
-    sprintf(szRet,"%u.%u.%u.%u", _address.bytes[0], _address.bytes[1], _address.bytes[2], _address.bytes[3]);
+    sprintf(szRet,"%u.%u.%u.%u", _address[0], _address[1], _address[2], _address[3]);
     return String(szRet);
 }
 
@@ -103,7 +103,7 @@ bool IPAddress::fromString(const char *address)
                 // Too much dots (there must be 3 dots)
                 return false;
             }
-            _address.bytes[dots++] = acc;
+            _address[dots++] = acc;
             acc = 0;
         }
         else
@@ -117,7 +117,7 @@ bool IPAddress::fromString(const char *address)
         // Too few dots (there must be 3 dots)
         return false;
     }
-    _address.bytes[3] = acc;
+    _address[3] = acc;
     return true;
 }
 
--- a/cores/esp32/IPAddress.h
+++ b/cores/esp32/IPAddress.h
@@ -23,16 +23,14 @@
 #include <stdint.h>
 #include <WString.h>
 #include <Printable.h>
+#include <array>
 
 // A class to make it easier to handle and pass around IP addresses
 
 class IPAddress: public Printable
 {
 private:
-    union {
-        uint8_t bytes[4];  // IPv4 address
-        uint32_t dword;
-    } _address;
+	std::array<uint8_t,4> _address{}; // IPv4 address
 
     // Access the raw byte array containing the address.  Because this returns a pointer
     // to the internal structure rather than a copy of the address this function should only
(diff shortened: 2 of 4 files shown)


VojtechBartoska:
Hello @safocl, please sign CLA. Thanks!

safocl:
> Hello @safocl, please sign CLA. Thanks!

I did it.

safocl:
[Example with UB](https://godbolt.org/z/sGWeGs1hx) -- from [cppreference](https://en.cppreference.com/w/cpp/language/union)

mrengineer7777:
Modifying TwoWire seems unrelated to this PR and may introduce breaking changes.

safocl:
> Modifying TwoWire seems unrelated to this PR and may introduce breaking changes.

yes, this is separate PR.

cores/esp32/IPAddress.cpp
```diff
@@ -20,62 +20,62 @@
...
 {
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+	std::copy(address, address + _address.size(), _address.begin());
     return *this;
 }
 
 IPAddress& IPAddress::operator=(uint32_t address)
 {
-    _address.dword = address;
+    uint32_t& addressRef = reinterpret_cast<uint32_t&>(_address.front());
```
TD-er, line 54: is `_address.front()` guaranteed to be 32-bit aligned?
If not, then this will cause undefined behavior.

cores/esp32/IPAddress.h
```diff
@@ -54,26 +52,27 @@ class IPAddress: public Printable
     bool fromString(const char *address);
     bool fromString(const String &address) { return fromString(address.c_str()); }
 
+
     // Overloaded cast operator to allow IPAddress objects to be used where a pointer
     // to a four-byte uint8_t array is expected
     operator uint32_t() const
     {
-        return _address.dword;
+        return reinterpret_cast<const uint32_t&>(_address.front());
```
TD-er, line 60: Idem, reinterpret_cast should be used with utmost care, especially with alignment.

cores/esp32/IPAddress.cpp
safocl: > is `_address.front()` guaranteed to be 32-bit aligned? If not, then this will cause undefined behavior.

It is guaranteed be aligned as structure -- array is once member in this structure -- it is has same address. It behavior same as union or c-style array (uint8_t a[4]).

Standart says (https://timsong-cpp.github.io/cppwp/n4659/array):

> class template for storing fixed-size sequences of objects. An array is a [contiguous container](https://timsong-cpp.github.io/cppwp/n4659/container.requirements.general#def:contiguous_container). An instance of array<T, N> stores N elements of type T, so that size() == N is an invariant.

> **An array is an [aggregate](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#def:aggregate) that can be list-initialized with up to N elements whose types are convertible to T.**

https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#1:

> An [aggregate](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#def:aggregate) is an array or a [class](https://timsong-cpp.github.io/cppwp/n4659/class) with
> 
> [(1.1)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#1.1) no user-provided, explicit, or inherited constructors ([[class.ctor]](https://timsong-cpp.github.io/cppwp/n4659/class.ctor)),
> [(1.2)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#1.2) no private or protected non-static data members (Clause [[class.access]](https://timsong-cpp.github.io/cppwp/n4659/class.access)),
> [(1.3)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#1.3) no [virtual functions](https://timsong-cpp.github.io/cppwp/n4659/class.virtual), and
> [(1.4)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#1.4) no virtual, private, or protected base classes ([[class.mi]](https://timsong-cpp.github.io/cppwp/n4659/class.mi)).
> [ Note: Aggregate initialization does not allow accessing protected and private base class' members or constructors.  — end note ]

https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#2:

> 2     The [elements](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#def:aggregate,elements) of an aggregate are:
> 
> [(2.1)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#2.1) for an array, the array elements in increasing subscript order, or
> [(2.2)](https://timsong-cpp.github.io/cppwp/n4659/dcl.init.aggr#2.2) for a class, the direct base classes in declaration order, followed by the direct non-static data members ([[class.mem]](https://timsong-cpp.github.io/cppwp/n4659/class.mem)) that are not members of an anonymous union, in declaration order.

and (https://timsong-cpp.github.io/cppwp/n4659/container.requirements#general-6):

> begin() returns an iterator referring to the first element in the container. end() returns an iterator which is the past-the-end value for the container. If the container is empty, then begin() == end().

https://timsong-cpp.github.io/cppwp/n4659/sequence.reqmts#tab:containers.sequence.optional:

> a.front() = *a.begin()

It mean that `&_address.front() == &(*_address.begin()) == &_address == &IPAddress`
and correspondingly an align of `_address.front()` depend on an align of `IPAddress` structure.

safocl: but [this example](https://godbolt.org/z/rehMs66We) shows that `alignof` = 1 for struct with `std::array<uint8_t,4>` member.
`alignas(4) `is required to achieve 32-bit alignment
```
struct alignas(4) Sarray { 
std::array<uint8_t,4> a{};
};
```

safocl: but [this example](https://godbolt.org/z/GGhKe65T6) shows that class `IPAddress` has 4 bytes of aligment because it inherits from an abstract type (with vtable).

in this case, the _address field still has 1 byte of alignment. -- need `alignas( alignof(uint32_t) )` for `IPAddress`.
```
struct IPAddress: public Printable
{
    alignas( alignof(uint32_t) ) std::array<uint8_t,4> _address{};
};
```
[as  here](https://godbolt.org/z/cfncPhefW)

I didn’t initially take into account that inheritance occurs

safocl: [i fix this omission.](https://github.com/espressif/arduino-esp32/pull/8819/commits/5d65a547906388e2dcb866549c8927d506513a8b)

safocl: here, of course, I would like to generally highlight that `uint32_t` may not exist at all on a specific platform.
Maybe `uint_least32_t` should be used for compatibility?

TD-er: `uint32_t` is known in all platforms this repository covers.
I don't know `uint_least32_t`, but it seems to be not guaranteed to be a fixed length type, which will cause issues when it is used in some struct of fixed length or offset of members (e.g. storing a struct or transferring it to another node)

TD-er:
> > Modifying TwoWire seems unrelated to this PR and may introduce breaking changes.
> 
> yes, this is separate PR.

But the Wire files are included in this PR.

cores/esp32/IPAddress.cpp
safocl: > I don't know `uint_least32_t`, but it seems to be not guaranteed to be a fixed length type, which will cause issues when it is used in some struct of fixed length or offset of members (e.g. storing a struct or transferring it to another node)

https://timsong-cpp.github.io/cppwp/n4659/cstdint#syn -- types that are at least N-bit

safocl:
> > > Modifying TwoWire seems unrelated to this PR and may introduce breaking changes.
> > 
> > 
> > yes, this is separate PR.
> 
> But the Wire files are included in this PR.

ok. I'll do a new PR.

safocl:
i recreate PR https://github.com/espressif/arduino-esp32/pull/8829

TD-er:
You could simply have removed the 2 Wire files from this PR by copying the currently existing code of the main branch and committing it to this PR.
There are very likely more elegant ways of doing this using Git, but this is by far the easiest way to suggest.

me-no-dev:
too many pull requests open. please fix them up and leave only minimal amount open (like one for Wire and one for IPAddress)

PR #8829, opened 2023-11-03
Rewrite IPAddress without union

safocl:
## Rewrite IPAddress without union
Using a union allowed undefined behavior for array-like INPUT and OUTPUT arguments through a 32-bit integer (and vice versa).

  C++ standard says (https://timsong-cpp.github.io/cppwp/n4659/class.union#1):
  "At most one of the non-static data members of an object of union type
  can be active at any time, that is, the value of at most one of the
  non-static data members can be stored in a union at any time. [ Note:
  One special guarantee is made in order to simplify the use of unions:
  If a standard-layout union contains several standard-layout structs
  that share a common initial sequence, and if a non-static data member
  of an object of this standard-layout union type is active and is one
  of the standard-layout structs, it is permitted to inspect the common
  initial sequence of any of the standard-layout struct members; see
  [class.mem].  — end note ]"

  and (https://timsong-cpp.github.io/cppwp/n4659/class.mem#22):
  "Two standard-layout unions are layout-compatible if they have the same number of non-static data members and corresponding non-static data members (in any order) have layout-compatible types."

you can't hope that compilers don't seem to do undefined behavior at the moment

note: this PR is recreated https://github.com/espressif/arduino-esp32/pull/8819 PR -- to fix https://github.com/espressif/arduino-esp32/pull/8819#issuecomment-1792010809

### DIFF
--- a/cores/esp32/IPAddress.h
+++ b/cores/esp32/IPAddress.h
@@ -23,16 +23,14 @@
 #include <stdint.h>
 #include <WString.h>
 #include <Printable.h>
+#include <array>
 
 // A class to make it easier to handle and pass around IP addresses
 
 class IPAddress: public Printable
 {
 private:
-    union {
-        uint8_t bytes[4];  // IPv4 address
-        uint32_t dword;
-    } _address;
+	alignas(alignof(uint32_t)) std::array<uint8_t,4> _address{}; // IPv4 address
 
     // Access the raw byte array containing the address.  Because this returns a pointer
     // to the internal structure rather than a copy of the address this function should only
@@ -40,12 +38,12 @@ private:
     // stored.
     uint8_t* raw_address()
     {
-        return _address.bytes;
+        return _address.data();
     }
 
 public:
     // Constructors
-    IPAddress();
+    IPAddress() = default;
     IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet);
     IPAddress(uint32_t address);
     IPAddress(const uint8_t *address);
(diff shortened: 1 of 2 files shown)


cores/esp32/IPAddress.h
```diff
@@ -23,29 +23,27 @@
...
 // A class to make it easier to handle and pass around IP addresses
 
 class IPAddress: public Printable
 {
 private:
-    union {
-        uint8_t bytes[4];  // IPv4 address
-        uint32_t dword;
-    } _address;
+	alignas(alignof(uint32_t)) std::array<uint8_t,4> _address{}; // IPv4 address
```
TD-er, line 33: Does `std::array` initialize with a zero value?
You did remove the default constructor, so it must be set to zero here.

me-no-dev:
Before we can do anything here, you need to first propose this upstream: https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.cpp

safocl:
> Before we can do anything here, you need to first propose this upstream: https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.cpp

[I created PR](https://github.com/arduino/ArduinoCore-API/pull/225)

me-no-dev:
Thanks! Keep us posted :)

me-no-dev:
@VojtechBartoska removed the milestone, because this is pending being merged upstream into the official Arduino API first. We will sync with what they currently have for 3.0.0 and later merge this, once updated upstream.

safocl:
> @VojtechBartoska removed the milestone, because this is pending being merged upstream into the official Arduino API first. We will sync with what they currently have for 3.0.0 and later merge this, once updated upstream.

They said that no std::* objects can be used to implement internal components.
I have rewritten it with a C-style array and am waiting for acceptance.

me-no-dev:
@safocl you will update this PR, once merged upstream, correct?

safocl:
> @safocl you will update this PR, once merged upstream, correct?

yes, i will...
All that remains is to find out which version they will accept the work on, then I will immediately update it to the same type.

me-no-dev:
because we are clearing PRs. I will close this one for now and @safocl can reopen it when main Arduino merges the proposal

PR #6331, opened 2022-02-22
Absolute mouse support

tobozo:
This PR adds `USBHIDAbsMouse` class with absolute mouse positioning support.

TinyUSB does not appear to provide (yet) the report struct for absolute positioning, only relative positioning.

Limitations: only one of `USBHIDMouse` / `USBHIDAbsMouse` should be used.

I'm still open to discussion as how to properly contribute the `abs_mouse_report_descriptor` and `struct abs_mouse_report_t` to the TinyUSB project so it eventually lands back in the espressif codebase.

PR #8831, opened 2023-11-03
Absolute mouse support (was #6331)

tobozo:
This PR adds USBHIDAbsMouse class with absolute mouse positioning support.

It is a followup of #6331

TinyUSB now provides the report struct for absolute positioning but isn't available yet in the current idf version.
A temporary conditional macro has been set in a separate file that will be safe to delete once the new tinyusb version is bundled with esp-idf.

Limitations: only one of USBHIDMouse / USBHIDAbsMouse should be used.


libraries/USB/src/USBHIDMouse.h
```diff
@@ -34,26 +34,68 @@
...
+
+class USBHIDMouseBase: public USBHIDDevice {
 public:
-    USBHIDMouse(void);
+    USBHIDMouseBase(HIDMouseType_t *type);
     void begin(void);
     void end(void);
-
-    void click(uint8_t b = MOUSE_LEFT);
-    void move(int8_t x, int8_t y, int8_t wheel = 0, int8_t pan = 0);
```
me-no-dev, line 49: if you add the common methods here as virtual, you could cast either mouse as `USBHIDMouseBase` and call it's methods, regardless which type it is. This is just a suggestion. Do not feel obligated to do it :)
```cpp
    virtual void move(int8_t x, int8_t y, int8_t wheel, int8_t pan) = 0;
    virtual void click(uint8_t b) = 0;
    virtual void buttons(uint8_t b) = 0;
```

tobozo: I can't do that with `::move()` as the data types are different for `x` and `y` (relative mouse uses `int8_t` while absolute mouse uses `int16_t`) but I'll add the `::click()` as virtual to the parent class and mark them as override in the child classes

P-R-O-C-H-Y (approved):

lucasssvaz (approved):
LGTM

PR #8839, opened 2023-11-06
Update/upgrade OTAWebUpdater.ino example

lbernstone:
Streamlined a lot of the code
HTML now sends file size
HTML has better progress graphics
HTML redirects back to root after completion
WebServer uses POST and GET to differentiate b/w status
Includes a lot more examples: generate passwords, functionalized setup, ticker, functionalized WebServer response, sending a const file
literal & C-style strings included in a separate header file
Moved to Update library- this does not use ArduinoOTA


libraries/Update/examples/OTAWebUpdater/OTAWebUpdater.ino
```diff
@@ -0,0 +1,102 @@
...
+
+void webServerInit() {
+  server.on("/update", HTTP_POST, [](){handleUpdateEnd();}, [](){handleUpdate();});
+  server.on("/favicon.ico", HTTP_GET, [](){
+     server.sendHeader("Content-Encoding", "gzip");
+     server.send_P(200, "image/x-icon", favicon_ico_gz, favicon_ico_gz_len);
+  });
+  server.onNotFound([]() {server.send(200, "text/html", indexHtml);});
+  server.begin();
+  Serial.printf("Web Server ready at http://esp32.local or http://%s\n", WiFi.softAPIP().toString());
```
me-no-dev, line 85:
```cpp
WiFi.softAPIP().toString().c_str()
```

libraries/Update/examples/OTAWebUpdater/OTAWebUpdater.ino
```diff
@@ -0,0 +1,102 @@
...
+void apMode() {
+  WiFi.mode(WIFI_AP);
+  char ssid[13];
+  char passwd[11];
+  uint32_t espmac = ESP.getEfuseMac() >> 24; 
+  snprintf(ssid, 13, SSID_FORMAT, espmac);
+#ifdef PASSWORD
+  snprintf(passwd, 11, PASSWORD);
+#else
+  snprintf(passwd, 11, "%s", generatePass(10).c_str());
```
me-no-dev, line 32: i'm not sure why the compiler is complaining about truncation, but I suggest you make `passwd` be 16 bytes long instead.

lbernstone: The size is actually fairly specific, as it allows you to do a wifi: QR code in a 64 bit square.

lbernstone: I'm not getting this error. Are you using some additional flags?

lbernstone: How about I just take out "%s"?

PR #8842, opened 2023-11-06
Fixes Rainmaker to be a  Managed Component

SuGlider:
## Description of Change
This PR changes `CMakLists.txt` and `idf_component.yml` to use most possible Managed Components, including RainMaker.
It also fixes RainMaker Arduino Layer to be compatible with the latest RainMaker release.

## Tests scenarios
It has been tested building the Arduino Lib Builder and also by building a RainMaker example using Arduino as IDF component.

## Related links
This shall be merged along with https://github.com/espressif/esp32-arduino-lib-builder/pull/145


idf_component.yml
```diff
@@ -27,3 +27,15 @@ dependencies:
   idf: ">=5.1"
   mdns: "^1.1.0"
   chmorgan/esp-libhelix-mp3: "1.0.3"
+  esp-dsp: "^1.3.4"
+  espressif/esp_rainmaker: "^1.0.0"
+  espressif/rmaker_common: "^1.4.3"
+  espressif/esp_insights:  "^1.0.1"
+  espressif/qrcode: "^0.1.0~1"
+  espressif/esp32-camera: "^2.0.5"
```
me-no-dev, line 35: because camera driver is not used in any code in the core, I would ask you to remove it from here and leave it being handled by the lib-builder instead.

SuGlider: Done

SuGlider:
@me-no-dev - This PR must be merged before fixing the issue in the Lib Builder with sr_models.bin because it adds ESP-SR to the components, which is necessary to run Lib Builder CI correctly in https://github.com/espressif/esp32-arduino-lib-builder/pull/145

Issue #8807, opened 2023-10-25
Support of Zigbee protocol


P-R-O-C-H-Y:
This issue tracks support of Zigbee protocol (supported by ESP32C6 and ESP32H2 chips)

Initial steps:

- [x] Investigate [ESP-IDF Zigbee library](https://components.espressif.com/components/espressif/esp-zigbee-lib) 
- [x] Make the Zigbee and Boss libs available - PR #8844
- [x] Followup [PR](https://github.com/espressif/esp32-arduino-lib-builder/pull/146) for lib-builder
- [x] Test IDF example running inside Arduino
- [x] Write easy to use examples

P-R-O-C-H-Y:
Tasks have been updated.

Hedda:
@P-R-O-C-H-Y Off-topic; just wondering if you might be interested in looking at Zigbee for Tasmota? -> https://github.com/arendst/Tasmota/discussions/20045

For reference, Tasmota's Zigbee2Tasmota (Z2T) implementation already supports Texas Instruments ZNP and Silicon Labs EZSP Zigbee stacks/interfaces with respective Zigbee Coordinator adapter running Zigbee NCP (Network Co-Processor) host firmware:

https://tasmota.github.io/docs/Zigbee/

Tasmota itself is a arduino-esp32 based open source firmware for DIY devices (especially popular for home automation solutions):

https://tasmota.github.io/docs/

PS: While Tasmota itself is mature, the ESP32-C6 support in Tasmota is still experimental, however you can follow its progress here:

https://github.com/arendst/Tasmota/discussions/19350

P-R-O-C-H-Y:
Update: Ported 2 simple Zigbee examples from ESP-Zigbee-SDK (esp-idf) #9024.
Creating a wrapper or a library is not planned for now, so after merging the examples we are closing this issue.

fillibar:
With the Arduino examples, are there any plans to make them able to connect to another hub/controller and not just another esp32c6 acting as a coordinator?

While it looks like that might be possible, at this time they just hang (monitor shows it is stuck at "Network steering was not successful"). This seems very similar to the issue here:
https://github.com/espressif/esp-idf/issues/10662

With the IDF example a heavily revised esp_zb_task seemed to help (also updating the dependencies since the ones for those examples are a bit behind).

Making a similar change to the esp_zb_task in the .ino file did not work even after copying that section from code that works when flashed with the idf.

[...]

PR #8844, opened 2023-11-07
Zigbee protocol support

P-R-O-C-H-Y:
## Description of Change
Adds menu in tools for selecting Zigbee mode to include proper libraries.

## Tests scenarios
Tested with reworked esp-idf zigbee on/off light example with manually build libraries by lib-builder with this [PR](https://github.com/espressif/esp32-arduino-lib-builder/pull/146) applied .

## Related links
Related #8807


boards.txt
```diff
@@ -17,6 +17,7 @@ menu.EventsCore=Events Run On
 menu.MemoryType=Memory Type
 menu.EraseFlash=Erase All Flash Before Sketch Upload
 menu.JTAGAdapter=JTAG Adapter
+menu.Zigbee=Zigbee Mode
```
me-no-dev, line 20: please rename the menu to `ZigbeeMode`. This is a bit too generic

boards.txt
```diff
@@ -187,6 +194,20 @@ esp32h2.menu.EraseFlash.none.upload.erase_cmd=
 esp32h2.menu.EraseFlash.all=Enabled
 esp32h2.menu.EraseFlash.all.upload.erase_cmd=-e
 
+esp32h2.menu.Zigbee.default=Disabled
+esp32h2.menu.Zigbee.default.build.zigbee_mode=
+esp32h2.menu.Zigbee.default.build.extra_libs=
+esp32h2.menu.Zigbee.ed=Zigbee ED (end device)
+esp32h2.menu.Zigbee.ed.build.zigbee_mode=-DZIGBEE_MODE_ED
+esp32h2.menu.Zigbee.ed.build.extra_libs=-lesp_zb_api_ed -lesp_zb_cli_command -lzboss_stack.ed.trace -lzboss_stack.ed -lzboss_port
```
me-no-dev, line 202: let's allow `build.extra_libs` to be available for something else that would be use-case specific. Add `build.zigbee_libs` instead. Makes it more consistent and prevents some other feature to overwrite the variable.

boards.txt
P-R-O-C-H-Y: Done.

boards.txt
P-R-O-C-H-Y: Done, good catch as that could collide with the PSRAM extra_libs. Renamed as suggested

PR #8849, opened 2023-11-08
WiFiSTA - new status constant WL_STOPPED 

JAndrassy:
WL_NO_SHIELD as initial status() value stops any generic Arduino WiFi example

Generic Arduino WiFi examples start with
```
  if (WiFi.status() == WL_NO_SHIELD) {
    Serial.println("Communication with WiFi module failed!");
    // don't continue
    while (true) {
      delay(1);
    }
  }
```


SuGlider:
@JAndrassy - I think that this change will cause issues. WL_IDLE is used for other cases.

I can see that WL_NO_SHILED is in the Arduino WiFi Shiled Library examples, but it is not the case for Arduino ESP32.
https://github.com/arduino-libraries/WiFi/blob/master/examples/WiFiWebClient/WiFiWebClient.ino

I also see that the Arduino WiFi Library has a Note: this library was retired and is no longer maintained.

Looking into the new UNO R4 WiFi, it uses other way to check the module: WL_NO_MODULE = WL_NO_SHIELD and it comes from ESP32 module.
https://github.com/arduino/ArduinoCore-renesas/blob/main/libraries/WiFi/examples/WiFiWebClient/WiFiWebClient.ino

In ESP32 Arduino when the WiFi Interface is suspended, it goes to WL_NO_SHIELD state, which makes sense because it would be like "removing the shield". When the code starts the WiFi is suspended, which is equivalent to WL_NO_SHIELD...

This status is changed in the WiFi Event Call back
https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiGeneric.cpp#L1036
https://github.com/espressif/arduino-esp32/blob/e2487b128d59a717ecada6cb36566addc3ddc769/libraries/WiFi/src/WiFiGeneric.cpp#L1051-L1053

SuGlider (commented):
I don't see a clear motivation to this change in the ESP32 Arduino scope.
@me-no-dev - what do you think?

me-no-dev:
@SuGlider agreed.

JAndrassy:
maybe be there should be a new constant in `wl_status_t` for stopped STA

SuGlider:
> maybe be there should be a new constant in `wl_status_t` for stopped STA

Let's say that we create a WL_STOPPED and this is the initial state. What would WL_NO_SHIELD be used for?
I only can find WL_NO_SHIELD in the ESP32 Arduino WiFi Code in 3 places:

1- in the declaration of the `wl_status_t` enumeration in `WiFiType.h`
2- for the initial value of `static wl_status_t _sta_status = WL_NO_SHIELD;` in `WiFiSTA.cpp`
3- when the the event `ARDUINO_EVENT_WIFI_STA_STOP` happens + `WiFiSTAClass::_setStatus(WL_NO_SHIELD);` in `WiFiGeneric.cpp`

A new constant for stopped STA would be used in (2) and (3). It should be added with a higher value (254) because the inital state must be higher than  `WL_DISCONNECT` to get it started in the code of `uint8_t WiFiSTAClass::waitForConnectResult(unsigned long timeoutLength)`.

This would make `WL_NO_SHIELD` useless and it would be there just for backward compatibility.

But ... this could be a way to make it compatible with Arduino Upstream. @me-no-dev PTAL.

JAndrassy:
> This would make WL_NO_SHIELD useless and it would be there just for backward compatibility.

as already the comment next to is states :-)

SuGlider:
> > This would make WL_NO_SHIELD useless and it would be there just for backward compatibility.
> 
> as already the comment next to is states :-)

OK, please proceed with the change for evaluation.

SuGlider (changes requested):
Let's create a WL_STOPPED = 254 and use it as inital value.

me-no-dev:
All those statuses are really there only for some sort of compatibility. Reality is that original WiFi library is STA only, external chip with a very different capabilities. I am generally all for compatibility, as long as it does not prevent us to progress and provide new and better things. As @SuGlider said, this is almost never used, so I am also fine with whatever you guys decide, as long as it does not break anything.

SuGlider (approved):
Thanks @JAndrassy - LGTM.

PR #8853, opened 2023-11-08
WiFi.BSSID and scan result BSSID with parameter as in WiFi libraries by Arduino

JAndrassy:
Arduino doc on BSSID getters: https://www.arduino.cc/reference/en/libraries/wifi/wifi.bssid/

I [compared more than a dozen of Arduino Networking libraries](https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md) to find the common API for Arduino networking. Then I wrote [a test sketch](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiTest/WiFiTest.ino) which checks how a library complies to this common API.

ESP32 Arduino WiFi library was based on ESP8266 Arduino WiFi library which was based on the first WiFi library by Arduino. Arduino later created more WiFi libraries and the API diverged, but at least with the most recent library they looked at the ESP32 WiFi library for new method names (for example dnsIP(n)).

While doing the research and tests I do PR in Arduino networking libraries repositories with in most cases simple modifications to unify the API where it doesn't break the library (for example I will not ask esp8266 and esp32 maintainers to make WiFi.begin blocking :-) ). 

Similar [PR](https://github.com/esp8266/Arduino/pull/9008) by me was recently merged in esp8266 Arduino repository.

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.h
+++ b/libraries/WiFi/src/WiFiSTA.h
@@ -98,7 +98,7 @@ public:
     String SSID() const;
     String psk() const;
 
-    uint8_t * BSSID();
+    uint8_t * BSSID(uint8_t* bssid = NULL);
     String BSSIDstr();
 
     int8_t RSSI();
--- a/libraries/WiFi/src/WiFiScan.h
+++ b/libraries/WiFi/src/WiFiScan.h
@@ -42,7 +42,7 @@ public:
     String SSID(uint8_t networkItem);
     wifi_auth_mode_t encryptionType(uint8_t networkItem);
     int32_t RSSI(uint8_t networkItem);
-    uint8_t * BSSID(uint8_t networkItem);
+    uint8_t * BSSID(uint8_t networkItem, uint8_t* bssid = NULL);
     String BSSIDstr(uint8_t networkItem);
     int32_t channel(uint8_t networkItem);
     static void * getScanInfoByIndex(int i) { return _getScanInfoByIndex(i); }; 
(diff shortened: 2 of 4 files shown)


SuGlider (changes requested):
@JAndrassy - It sounds good to make it compatible to upstream Arduino WiFi API.
I'd suggest to unify the APIs, instead of creating a second one.

Something like: 
unique API would be just `uint8_t * BSSID(uint8_t networkItem, uint8_t* bssid = NULL);` and `uint8_t * BSSID(uint8_t* bssid = NULL);` 
The code shall test `bssid != NULL` before copying the information into it.

libraries/WiFi/src/WiFiSTA.h
```diff
@@ -99,6 +99,7 @@ class WiFiSTAClass
     String psk() const;
 
     uint8_t * BSSID();
+    uint8_t * BSSID(uint8_t* bssid);
```
SuGlider, line 102: use just one API: `uint8_t * BSSID(uint8_t* bssid = NULL);`
and replace both
```
uint8_t * BSSID();
uint8_t * BSSID(uint8_t* bssid);
```

libraries/WiFi/src/WiFiScan.h
```diff
@@ -43,6 +43,7 @@ class WiFiScanClass
     wifi_auth_mode_t encryptionType(uint8_t networkItem);
     int32_t RSSI(uint8_t networkItem);
     uint8_t * BSSID(uint8_t networkItem);
+    uint8_t * BSSID(uint8_t networkItem, uint8_t* bssid);
```
SuGlider, line 46: use just one API: `uint8_t * BSSID(uint8_t networkItem, uint8_t* bssid = NULL);`
and replace both
```
uint8_t * BSSID(uint8_t networkItem);
uint8_t * BSSID(uint8_t networkItem, uint8_t* bssid);
```

SuGlider (approved):
Thanks! LGTM

PR #8854, opened 2023-11-08
WiFiSTA - method setDNS as in WiFi libraries by Arduino

JAndrassy:
Arduino doc: https://www.arduino.cc/reference/en/libraries/wifi/wifi.setdns/

I write about the motivation for this in the other [PR](https://github.com/espressif/arduino-esp32/pull/8853)

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.cpp
+++ b/libraries/WiFi/src/WiFiSTA.cpp
@@ -431,6 +431,19 @@ bool WiFiSTAClass::config(IPAddress local_ip, IPAddress gateway, IPAddress subne
 }
 
 /**
+ * Change DNS server for static IP configuration
+ * @param dns1       Static DNS server 1
+ * @param dns2       Static DNS server 2 (optional)
+ */
+bool WiFiSTAClass::setDNS(IPAddress dns1, IPAddress dns2)
+{
+    if(WiFiGenericClass::getMode() == WIFI_MODE_NULL)
+        return false;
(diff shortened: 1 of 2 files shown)


me-no-dev:
Thanks for the PR, but we will probably go another way. On our platform, the DNS servers are set for all client interfaces and are not specific to a given one, meaning, if you have ETH and WiFi together, they will use the same DNS servers for both connections. The API as such is and will be misleading.

JAndrassy:
the reason to do Arduino platform and libraries is not because it has good, clear and not confusing API

me-no-dev:
yes, now go explain that to the users who complain that we allow setting DNS to interfaces, while it actually sets it globally.

me-no-dev:
merging now, will add notes about compatibility as part of the rework

JAndrassy:
> On our platform, the DNS servers are set for all client interfaces and are not specific to a given one

so the DNS server returned with DHCP on some interface is added to global registry of DNS servers? even if it is 192.168.0.1 which is not valid for network interfaces connected to a different LAN?

me-no-dev:
I'm not sure I understood your question clearly, but to explain slightly better:

DNS and interface addresses are not necessarily connected. DNS servers only resolve names to addresses and are expected to always return the same addresses, regardless of your local network address. Google.com will always resolve to google's servers for example. From then on, it's up to your network routing to figure out which interface to use and how to get to the resolver IP address from DNS and reach the target server. So having the same DNS is nothing bad or wrong, as long as your ESP has a way to reach the DNS IP and then the resolved IPs

PR #8860, opened 2023-11-09
WiFiServer - deprecate available(), use accept() everywhere

JAndrassy:
In esp8266 and esp32 Arduino WiFi library server.available() was never implemented as documented. It is implemented like server.accept() is [documented by Arduino](https://www.arduino.cc/reference/en/libraries/ethernet/server.accept/).

server.accept() was an alias for the server.available() from the beginning here in esp32 Arduino platform.

I think with the new major version 3 this should be done. esp8266 Arduino did this change for version 3 too.


SuGlider (approved):
LGTM - It still is back compatible by keeping the `available()` as it is, but with a warning for being deprecated.

More information about the argument that @JAndrassy has made:

Based on the documentation from https://www.arduino.cc/reference/en/libraries/wifi/server.available/
The method `available()` inherits from the Stream utility class. Therefore it should tell how many bytes are ready for reading.

The PR is correct in changing it to `accept()` which is how it was implemented.
https://www.arduino.cc/reference/en/libraries/ethernet/server.accept/

SuGlider:
ESP8266 Arduino has done same changes:
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/WiFiServer.cpp#L129-L153
https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/WiFiServer.h#L84-L85

SuGlider:
Thanks @JAndrassy !!

mrengineer7777:
Beautiful work.  Well done!

JAndrassy:
@SuGlider 
> The method available() inherits from the Stream utility class. Therefore it should tell how many bytes are ready for reading.

Server inherits from Print. It doesn't inherit from Stream. 
server.available() is for a Server implementation which manages the connect clients. `available()` then returns first connected client which has data available. It can return the same client repeatedly. 

[documentation for `server.available()'](https://www.arduino.cc/reference/en/libraries/wifi/server.available/):
> Gets a client that is connected to the server **and** has data available for reading.

My NetApiHelpers library has ArduinoWiFiServer for ESP32. This has [correct `available()`](https://github.com/JAndrassy/NetApiHelpers/blob/6c43bc2f7f790bfdd13d6fb24b6c75d43e91ac8b/src/ServerTemplate.h#L53) and the print-to-all-clients function to really implement the Print base class. But most users don't need this. `accept()` is what they really need.

PR #8863, opened 2023-11-10
WiFiClients.setConnectionTimeout added

JAndrassy:
<del>`WiFiClient::setTimeout` now shadows `Stream::setTimeout` which has a different purpose and the unit of the parameter is different. Related [issue](https://github.com/espressif/arduino-esp32/issues/5558).</del>

<del>The solution is to rename `WiFiClient::setTimeout` to [Arduino standard](https://www.arduino.cc/reference/en/libraries/ethernet/client.setconnectiontimeout/) `WiFiClient::setConnectionTimeout`.</del>

EDIT: at the end most of this PR was done in https://github.com/espressif/arduino-esp32/pull/6676 and https://github.com/espressif/arduino-esp32/pull/8998

overview of Client API in Arduino networking libraries:
https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#client-getters-and-setters

### DIFF
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -348,6 +348,11 @@ int WiFiClient::getOption(int option, int *value)
     return res;
 }
 
+void WiFiClient::setConnectionTimeout(uint32_t milliseconds)
+{
+    _timeout = milliseconds;
+}
+
 int WiFiClient::setNoDelay(bool nodelay)
 {
     int flag = nodelay;
--- a/libraries/WiFi/src/WiFiClient.h
+++ b/libraries/WiFi/src/WiFiClient.h
@@ -33,6 +33,7 @@ class ESPLwIPClient : public Client
 public:
         virtual int connect(IPAddress ip, uint16_t port, int32_t timeout) = 0;
         virtual int connect(const char *host, uint16_t port, int32_t timeout) = 0;
+        virtual void setConnectionTimeout(uint32_t milliseconds) = 0;
 };
 
 class WiFiClient : public ESPLwIPClient
@@ -92,6 +93,7 @@ public:
     int getSocketOption(int level, int option, const void* value, size_t size);
     int setOption(int option, int *value);
     int getOption(int option, int *value);
+    void setConnectionTimeout(uint32_t milliseconds);
     int setNoDelay(bool nodelay);
     bool getNoDelay();
 


VojtechBartoska:
Thanks for the PR @JAndrassy.

This relates to https://github.com/espressif/arduino-esp32/pull/6676.

@P-R-O-C-H-Y PTAL

JAndrassy:
the difference is that https://github.com/espressif/arduino-esp32/pull/6676 doesn't define `setConnectioniTimeout` (the standard name in Arduino networking API)

JAndrassy:
after https://github.com/espressif/arduino-esp32/pull/6676 was merged WiFiClient.setConnectionTimeut now looks here like a new method, but it just the old setTimeout modified (same way this PR does for WiFiClientSecure)

me-no-dev:
Please revert any changes to WiFiClientSecure. It will inherit those methods from WiFiClient. That was the point of the previous PR that we merged.

libraries/WiFiClientSecure/src/WiFiClientSecure.h
```diff
@@ -80,7 +80,7 @@ class WiFiClientSecure : public WiFiClient
     void setAlpnProtocols(const char **alpn_protos);
     const mbedtls_x509_crt* getPeerCertificate() { return mbedtls_ssl_get_peer_cert(&sslclient->ssl_ctx); };
     bool getFingerprintSHA256(uint8_t sha256_result[32]) { return get_peer_fingerprint(sslclient, sha256_result); };
-    int setTimeout(uint32_t seconds);
+    int setConnectionTimeout(uint32_t milliseconds);
```
me-no-dev, line 83: @P-R-O-C-H-Y any clues why I still see setTimeout here?

P-R-O-C-H-Y: @me-no-dev probably the rebase went wrong. This should not be there. 
Lets open a new PR to remove it.

me-no-dev:
@P-R-O-C-H-Y PTAL if this makes sense now :)

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -333,6 +333,23 @@ int WiFiClient::getSocketOption(int level, int option, const void* value, size_t
...
+        if(setSocketOption(SO_RCVTIMEO, (char *)&tv, sizeof(struct timeval)) < 0) {
+            return -1;
+        }
+        return setSocketOption(SO_SNDTIMEO, (char *)&tv, sizeof(struct timeval));
+    }
+    else {
+        return 0;
+    }
+}
+
```
P-R-O-C-H-Y, line 352: As the timeout will be set in the `write()` and `read()` functions anyways, because the `_lastReadTimeout` and `_lastWriteTimeout` remains unchanged, and also the Arduino standart says to be void function, I recommend changing the function just to set the `_timeout` variable like this:
```suggestion
void WiFiClient::setConnectionTimeout(uint32_t milliseconds)
{
    _timeout = milliseconds;
}

```

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -1166,6 +1166,7 @@ bool HTTPClient::connect(void)
 
     // set Timeout for WiFiClient and for Stream::readBytesUntil() and Stream::readStringUntil()
     _client->setTimeout(_tcpTimeout);	
+    _client->setConnectionTimeout(_tcpTimeout);
```
P-R-O-C-H-Y, line 1169: This won't be necessary as it will set just the _timeout variable if changed according to my comment the same as the `setTimeout` function.

P-R-O-C-H-Y (changes requested):
@me-no-dev @JAndrassy please take a look on my comment, if it makes sense to you as well.

JAndrassy:
it is a timeout for `connect`

libraries/HTTPClient/src/HTTPClient.cpp
JAndrassy: setTimeout is Stream:setTimeout and sets the timeout field in Stream. setConnectionTimeout setts the timeout filed in WiFiClient

JAndrassy: I see

JAndrassy:
so at the end it is just a simple setter

P-R-O-C-H-Y (approved):
Just simple like that :) side question: Arduino API seems to be unit16_t, should we align or better will be to go with uint32_t as @JAndrassy did. 
@me-no-dev?

TD-er:
It isn't unthinkable someone wants to set it to >65 seconds.
Also it doesn't require a reference value, so existing code expecting `uint16_t` will still compile.

JAndrassy:
the uint32_t is a leftover from rename from `int setTimeout(uint32_t seconds);`
_timeout is `int`

TD-er:
> _timeout is `int`

What would be the meaning of a negative timeout?

PR #8864, opened 2023-11-10
Add platform specific flash speed to magicFlashChipSpeed

TD-er:
## Description of Change

Various ESP32-variants like the C6, H2, etc. differ in their used flash frequencies.

Also the maximum possible flash size wasn't handled in `magicFlashChipSize`

## Related links

Used the info from [esptool.py](https://github.com/search?q=repo%3Aespressif%2Fesptool%20FLASH_FREQUENCY%20&type=code)

### DIFF
--- a/cores/esp32/Esp.cpp
+++ b/cores/esp32/Esp.cpp
@@ -391,37 +391,108 @@ FlashMode_t EspClass::getFlashChipMode(void)
 
 uint32_t EspClass::magicFlashChipSize(uint8_t byte)
 {
+/*
+    FLASH_SIZES = {
+        "1MB": 0x00,
+        "2MB": 0x10,
+        "4MB": 0x20,
+        "8MB": 0x30,
+        "16MB": 0x40,
+        "32MB": 0x50,
+        "64MB": 0x60,
+        "128MB": 0x70,
+    }
+*/
     switch(byte & 0x0F) {
-    case 0x0: // 8 MBit (1MB)
-        return (1_MB);
-    case 0x1: // 16 MBit (2MB)
-        return (2_MB);
-    case 0x2: // 32 MBit (4MB)
-        return (4_MB);
-    case 0x3: // 64 MBit (8MB)
-        return (8_MB);
-    case 0x4: // 128 MBit (16MB)
-        return (16_MB);
-    default: // fail?
+      case 0x0: return (1_MB);   // 8 MBit (1MB)
+      case 0x1: return (2_MB);   // 16 MBit (2MB)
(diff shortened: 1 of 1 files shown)


Jason2866 (approved):

Jason2866:
@me-no-dev What do you think about adding C2 Arduino support? I have added in our fork and project Tasmota does run stable on the C2.
The needed changes are mostly just additional #ifdefs. The change in this PR is the only specific C2 related one and the changes are also needed for the H2 and the C6.

me-no-dev:
@Jason2866 feel free to add the C2 changes. We do not promise that it will be available through the board manager, but nothing wrong in having it supported "as-component" and even as an option through the lib-builder. Just do not want to add additional megabytes to download for rarely used chip.

TD-er:
Just curious, why do you think the C2 will only be used rarely?

It looks like it will be a preferred module/chip for those vendors of cheap "smart" devices, like the Sonoff Basic used to be back in the days.
It isn't the best ESP32 variant, but still it is quite capable and you really don't need lots of CPU power for switching simple stuff.

me-no-dev:
@TD-er what I mean is that I do not see many development boards popping out and many of our users going for such a limited chip, when C3 and C6 could do so much more for the same money. Just that C2 is a bit too limited for the general public. I agree that specific optimized designs can be achieved and costs can be saved in mass production, just not the home user. That is why we can offer option for those that want to venture, but not load the 99% of the users with additional 100+ MB that they will never need. There is some initiative to split the core into separate chips and have the users install only what they have, but Arduino's platform is not yet supporting everything necessary for that to work. If/when that happens, we will surely add full support for C2 and maybe more future budget chips. Then even things like ESP32-Solo could be supported. I know there are some devices that use it.

Jason2866:
@me-no-dev  I am perfectly fine with not adding to the boards manager, tbh i don't like it all. But that's not your thing, it is coming from ArduinoIDE. It would be a great step to have all needed C2 changes in source code. So everyone can build the framework with the Arduino Lib Builder (just a few changes there). BTW adding C2 does not add a lot of Mbytes. It is way less than Speech Recognition or Rainmaker does add. Our forked Framework with supporting additionally the C2 has 150Mbyte at all (stripped the fat SR and Rainmaker stuff).

Since extracting the C2 parts out of our Arduino fork is "some" work, i can and will only do a PR if it will be merged fast (with the things you mentioned) so that no rebase work needs to be done.

me-no-dev:
@Jason2866 I don't see a problem to merge C2 support fast, as long as you cover it fast enough here.

PR #8848, opened 2023-11-08
WiFiSTA - allow using DHCP again after disconnecting static IP

JAndrassy:
reset _useStaticIp to false in WiFi.disconnect()

PR #8873, opened 2023-11-12
Revert "WiFiSTA - allow using DHCP again after disconnecting static IP"

JAndrassy:
I want to revert my recent PR https://github.com/espressif/arduino-esp32/pull/8848

It turned out that all libraries I tested preserved the static IP settings over `disconnect()` (except my WiFiEspAT). [I learned](https://github.com/esp8266/Arduino/pull/9020#issuecomment-1806789080) that the right way to return to DHCP is  `config(INADDR_NONE, INADDR_NONE, INADDR_NONE)`. And this really worked in all WiFi libraries.
So the user of the library can `config` the static IP and then `disconnect` and later `begin` again with the same static IP.


JAndrassy:
No history. I had the wrong idea that disconnect() should clear static IP and my PR a few weeks ago was merged. It turned out that I was wrong. No library clears static IP in disconnect(). All use WiFi.config(INADDR_NONE) to return to DHCP.

VojtechBartoska:
This PR is ready to be merged as an Revert.

PR #8878, opened 2023-11-13
Add Arduino as Component examples

P-R-O-C-H-Y:
## Description of Change
Added new folder to the repository containing examples for running Arduino as an ESP-IDF component.
The examples will be also visible on [ESP Registry](https://components.espressif.com/components/espressif/arduino-esp32) and would be easy to set up the project with just one idf.py command to get all dependencies and have the proper SDK Config changes.

## Tests scenarios
Tested locally on all SOCs.


idf_component_examples/Hello_world/README.md
```diff
@@ -0,0 +1,63 @@
...
+See [Arduino-esp32](https://components.espressif.com/components/espressif/arduino-esp32) in ESP Registry.
+
+## How to use example
+
+To create a ESP-IDF project from this example with the latest relase of Arduino-esp32, you can simply run command: `idf.py create-project-from-example "espressif/arduino-esp32:hello_world"`.
+ESP-IDF will download all dependencies needed from the component registry and setup the project for you.
+
+If you want to use cloned Arduino-esp32 repository, you can build this example directly.
+Go to the example folder `arduino-esp32/idf_component_examples/Hello_world`.
+First you need to comment line 6 `pre_release: true` in examples `/main/idf_component.yml`.
```
SuGlider, line 16: I just tested it in a separated folder.
It is also necessary to comment out the line 5 `override_path: '../../../'` as it may not exist.

SuGlider (changes requested):
Good work! Just check the comment I made. I could be included in the README.md.

idf_component_examples/Hello_world/README.md
P-R-O-C-H-Y: Actually, it should be removed by the ESP Registry so if you get it from there as an example, this won't exist. On other components it is done same way, for development purposes. If I understood correctly :)

Jason2866:
Can you add an example where PSRAM is used? There will be probably an include error for `
esp_psram.h`

SuGlider (approved):
LGTM

PR #8881, opened 2023-11-13
Add ESP32-C2 support

Jason2866:
## Description of Change
Adding the needed code changes to support build of the ESP32-C2. Libs for the C2 are still needed. 
-> Built with Arduino Lib Builder 
PR https://github.com/espressif/esp32-arduino-lib-builder/pull/149 for the Lib builder to support the C2

## Tests scenarios
The changes have been tested in Tasmota Arduino fork and project Tasmota is successful built and running for the C2

@me-no-dev as discussed in https://github.com/espressif/arduino-esp32/pull/8864


cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -1,3 +1,6 @@
+#include "soc/soc_caps.h"
+#if SOC_RMT_SUPPORTED
+
 #include "esp32-hal-rgb-led.h"
```
igrr, line 4: Note, since arduino-esp32 can use managed components now: we can use [led_strip](https://components.espressif.com/components/espressif/led_strip) component here, which has API-compatible implementations based on RMT and SPI. Then the feature will be still available for C2, and there would be less code to maintain.

platform.txt
```diff
@@ -107,6 +108,7 @@ build.copy_jtag_files=0
 build.openocdscript.esp32=esp32-wrover-kit-3.3v.cfg
 build.openocdscript.esp32s2=esp32s2-kaluga-1.cfg
 build.openocdscript.esp32s3=esp32s3-builtin.cfg
+build.openocdscript.esp32c2=esp32c2-builtin.cfg
```
igrr, line 111: There is no such config file in OpenOCD, because C2 doesn't have a USB_SERIAL_JTAG peripheral.

cores/esp32/esp32-hal-rgb-led.c
me-no-dev: If you are suggesting to replace our RGB code with led_strip, then this would be a larger item than this PR should handle, because it involves Peripheral Manager. I would suggest to leave it as-is and note in the PR that we need to add led_strip support (or figure out something else that would make sense)

Jason2866: My goal was to do as less as changes as possible. The suggested change is great imho should be done after this PR has merged.

me-no-dev: My suggestion would be to wrap the contents of the function instead. You can even `log_e` in `#else` to prompt us that this function is not implemented yet.

platform.txt
```diff
@@ -80,6 +80,7 @@ compiler.libraries.ldflags=
 build.extra_flags.esp32=-DARDUINO_USB_CDC_ON_BOOT=0
 build.extra_flags.esp32s3=-DARDUINO_USB_MODE={build.usb_mode} -DARDUINO_USB_CDC_ON_BOOT={build.cdc_on_boot} -DARDUINO_USB_MSC_ON_BOOT={build.msc_on_boot} -DARDUINO_USB_DFU_ON_BOOT={build.dfu_on_boot}
 build.extra_flags.esp32s2=-DARDUINO_USB_MODE=0 -DARDUINO_USB_CDC_ON_BOOT={build.cdc_on_boot} -DARDUINO_USB_MSC_ON_BOOT={build.msc_on_boot} -DARDUINO_USB_DFU_ON_BOOT={build.dfu_on_boot}
+build.extra_flags.esp32c2=-DARDUINO_USB_MODE=1 -DARDUINO_USB_CDC_ON_BOOT=0
```
me-no-dev, line 83: You can remove `-DARDUINO_USB_MODE=1`, like it's done for ESP32

libraries/Wire/src/Wire.cpp
```diff
@@ -338,10 +340,12 @@ bool TwoWire::end()
         }
 #endif
         if(is_slave){
+#if SOC_I2C_SUPPORT_SLAVE
```
me-no-dev, line 343: why only this code and initializations above commented? If this does not cause issues, please leave it alone and add a note in the PR for me to implement it properly.

Jason2866: removed. Thx for looking into doing a real fix :-)

platform.txt
Jason2866: removed.

libraries/Wire/src/Wire.cpp
me-no-dev: I guess we just haven't had a chip without it yet :) no worries

cores/esp32/esp32-hal-rgb-led.c
Jason2866: added log_e

me-no-dev: Like this:
```cpp
void neopixelWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val){
#if SOC_RMT_SUPPORTED
    // ... the current code
#else
    log_e("RMT is not supported on " CONFIG_IDF_TARGET);
#endif /* SOC_RMT_SUPPORTED */
}
```

libraries/Wire/src/Wire.cpp
me-no-dev: https://github.com/espressif/arduino-esp32/pull/8882 created. Will merge it after CI and pull master here.

me-no-dev:
Thanks @Jason2866 :) I'll take over to finish support

cores/esp32/esp32-hal-spi.c
```diff
@@ -532,7 +539,7 @@ static void spiInitBus(spi_t * spi)
     spi->dev->slave.trans_done = 0;
 #endif
     spi->dev->slave.val = 0;
-#if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32H2
+#if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32H2
     spi->dev->misc.val = 0;
```
TD-er, line 543: Wouldn't it make much more sense to have 1 define right above where this is used to either define it as `misc` or `pin` ?

Jason2866 (commented):
Ah, sorry forgot about this one. I added the missing in IDF

me-no-dev:
@Jason2866 now it compiles as component. Will also look at the lib-builder PR

@P-R-O-C-H-Y @SuGlider @lucasssvaz PTAL

me-no-dev:
Board Test is expected to fail due to missing SDK for C2.

Jason2866:
@me-no-dev Thx, for the polishing and fixing.

me-no-dev:
@Jason2866 all is good now :) both lib-builder and arduino are ready. Sketches are compiling, but we will not have it through the board manager for now. The size of the libs is 100MB. If there is enough interest or we can have them as separate package, we can include them too

Jason2866:
@me-no-dev Big thx for the work you put in and the lightning speed!
Still curious why it is so big. Our complete framework for Arduino 3.0.0 including all MCUs and libs is around 150MB (stripped Rainmaker, SR and DSP).
The working framework for the solo1 is about 30MB https://github.com/tasmota/arduino-esp32/releases
So i expected that the C2 does add not more than 30MB.

me-no-dev:
it adds 30MB to the ZIP and 100MB extracted. Besides that RainMaker is currently not working on it. There is close to zero interest in providing Board Manager support for it and it also takes time in the cron job that we can not afford at the time. As I said a few times, we can look into adding full support if there is enough interest or when we can make it painless to others (solo support will probably also come then)

Jason2866:
Perfectly fine! And the size is the same the 30MB mentioned was zipped. We will see if users are asking for. I don't think Rainmaker will ever work well with the C2. The free heap is probably not enough. The C2 is not designed for heavy lifting, but it is a nice powerfull little MCU. For me it is an follow up of the ESP8266...

me-no-dev:
>  I don't think Rainmaker will ever work well with the C2.

it's mentioned on the product page, so it should work. I'll report the bug

cores/esp32/esp32-hal-rgb-led.c
igrr: Yes, this comment wasn't intended to be a PR blocker. I just asked to consider this and create a follow-up task if you think this is a good idea to consider.

me-no-dev: One of the main reasons Peripheral Manager was implemented is to keep track of used peripherals and allow us to pick an available one that can achieve the requested functionality. Ex. PWM can be done with LedC, RMT, etc. Same can be said for neopixels (RMT, I2S, SPI, etc.). Probably proper task would be to implement such functionality and expand the options for all chips

Jason2866: For sure there should be a follow up here, to provide an universal working solution for every MCU, choosen by default the "best" solution (with user override, when set).

me-no-dev (approved):

variants/esp32c2/pins_arduino.h
```diff
@@ -0,0 +1,29 @@
...
+static const uint8_t TX = 20;
+static const uint8_t RX = 19;
+
+static const uint8_t SDA = 8;
+static const uint8_t SCL = 9;
+
+static const uint8_t SS    = 4;
+static const uint8_t MOSI  = 7;
+static const uint8_t MISO  = 5;
+static const uint8_t SCK   = 6;
```
P-R-O-C-H-Y, line 21: Let SPI pins be the same as C3 has:
```suggestion
static const uint8_t SS    = 7;
static const uint8_t MOSI  = 6;
static const uint8_t MISO  = 5;
static const uint8_t SCK   = 4;
```

P-R-O-C-H-Y (approved):
Please match the SPI pins with the C3. 
Tested Wifi + SD-SPI and both works well.
Thanks @Jason2866 for PR!

cores/esp32/HardwareSerial.h
```diff
@@ -165,6 +171,8 @@ typedef enum {
       #define TX1 (gpio_num_t)27
     #elif CONFIG_IDF_TARGET_ESP32S2
       #define TX1 (gpio_num_t)5
+    #elif CONFIG_IDF_TARGET_ESP32C2
+      #define TX1 (gpio_num_t)10
```
lucasssvaz, line 175: Pin 9 is already used for I2C. Maybe we can set Serial1 default pins to 10 and 18 (that are free).

```suggestion
    #elif CONFIG_IDF_TARGET_ESP32C2
      #define TX1 (gpio_num_t)18
```

cores/esp32/HardwareSerial.h
```diff
@@ -149,6 +153,8 @@ typedef enum {
       #define RX1 (gpio_num_t)26
     #elif CONFIG_IDF_TARGET_ESP32S2
       #define RX1 (gpio_num_t)4
+    #elif CONFIG_IDF_TARGET_ESP32C2
+      #define RX1 (gpio_num_t)9
```
lucasssvaz, line 157: Pin 9 is already used for I2C. Maybe we can set Serial1 default pins to 10 and 18 (that are free).

```suggestion
    #elif CONFIG_IDF_TARGET_ESP32C2
      #define RX1 (gpio_num_t)10
```

lucasssvaz (approved):
LGTM. I only recommend switching the default Serial1 pins to avoid possible conflicts.

P-R-O-C-H-Y:
@Jason2866 Can you please change the Serial1 pins as suggested by @lucasssvaz? We will merge after that. Thanks

Jason2866:
@P-R-O-C-H-Y  Changed.

PR #8892, opened 2023-11-16
WiFi.config handle Arduino parameters ordering and auto dns,gw,mask

JAndrassy:
In WiFi libraries by Arduino `config` has parameters `ip, dnsIP, gatewayIP, subnet`. only ip is not optional. for unspecified subnet default is 255.255.255.0 and unspecified gateway and dns are based on IP with last number changed to 1.

the ESP8266WiFi library accepts the 'esp\ ordering ip, gw, subnet, dns and Arduino ordering ip, dns, gw, subnet. The detection is based on the position of the subnet mask which has 255 as first number.

This PR adds handling of Arduino ordering of parameters. Parameter subnet is on position of dns1 which is an optional parameter so default value is used if needed.

And this PR adds a second `config` with two parameters local_ip and dns IP. dns IP is optional. gateway IP is calculated, dns IP is calculated if not provided. then the other `config` is invoked.

this allows `config` invocations as in Arduino library
```
config(ip)
config(ip, dns)
config(ip, dns, gw)
config(ip, dns, gw, subnet)
```
it also allows `WiFi.config(INADDR_NONE)` to return to DHCP as in other WiFi libraries.

overview of WiFi object API in significant libraries: 
https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md#wifi-station-network-interface


me-no-dev:
When Arduino decided on the order of arguments, someone must have been drunk... ip,dns,gw,mask must be the worst option out there. We will probably accept this, but it's stupid and confusing as hell. And it will mess up any other attempt at making this API proper. I dislike it with passion :D

JAndrassy:
the Arduino ordering is good for the automatic values. dns is first where the automatic value is probably not right
```
config(ip)
config(ip, dns)
config(ip, dns, gw)
config(ip, dns, gw, subnet)
```

me-no-dev:
proper order is: ip, mask, gw, dns. IP is the address, mask tells who it can talk to, gw tells how it can reach things outside the mask and finally DNS resolves those external names and is the GW or behind the GW usually. And to be frank, even this is wrong. Since we have an actual network stack with multiple supported interfaces, DNS is global. GW should also be global, but our stack does not have that complicated of routing.
```cpp
WiFi.config(ip); // mask: 255.255.255.0, gw = dns = first-ip-in-mask
WiFi.config(ip, mask); // gw = dns = first-ip-in-mask
WiFi.config(ip, mask, gw); // gw = dns
WiFi.config(ip, mask, gw, dns);
```

JAndrassy:
yes. you are right from the networking point of view mask should be first after IP because without the mask it can't work while without gateway it just can't transfer outside of the local network.
that parameters ordering could be detected too :-)

VojtechBartoska:
Hello, 

we discussed this PR internally and proposed changes will break current projects. We are not able to merge those changes now. We will stick to current implementations and change the order later in new API, also contribute changed orders upstream to Arduino.cc WiFi APIs.

Thanks for the PR @JAndrassy.

PR #8893, opened 2023-11-16
IDF release/v5.1

me-no-dev:
esp-idf: release/v5.1 5c61c89308
tinyusb: master 0f187b4d1

### DIFF
--- a/package/package_esp32_index.template.json
+++ b/package/package_esp32_index.template.json
@@ -39,7 +39,7 @@
             {
               "packager": "esp32",
               "name": "esp32-arduino-libs",
-              "version": "idf-release_v5.1-5c61c89308"
+              "version": "idf-release_v5.1-3662303f31"
             },
             {
               "packager": "esp32",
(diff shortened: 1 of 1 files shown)


Magnatto:
Hello, on arduino-esp32/cores/esp32/esp32-hal-rmt.c on line 540:
rmt_rx_channel_config.t type no longer contains intr_priority, removing it solved for me when using the function rmtInit.
image
This was the error:
image

me-no-dev:
@Magnatto are you using Arduino as component? Because you must have not updated your IDF to the latest `release/v5.1` branch. The flag exists now and without it being set there, it would cause RMT not to start

Magnatto:
> @Magnatto are you using Arduino as component? Because you must have not updated your IDF to the latest `release/v5.1` branch. The flag exists now and without it being set there, it would cause RMT not to start

Pardon me, you are correct, indeed i'm using arduino as component, however i'm using esp-idf v5.1.1, it didn't have the interrupt priority but 5.2 now has it.

PR #8894, opened 2023-11-17
add tinyuf2 partitions csv

hathach:
## Description of Change
TinyUF2 is gettting a bit more popular with ESP32 S2/S3, however, tinyuf2 partition is currently treated as custom paritions and stored in `variants/{board}` folder. All Adafruit S2/S3 boards make use of tinyuf2, this causes quite a bit of csv file duplication within the repo.

This PR adding commonly used tinyuf2 partitions will help to reduce csv duplication and make it easier to add new tinyuf2-enabled boards. 

## Tests scenarios
Nothing to test, will update Adafruit to use these and removed all duplicated csv in adafruit boards as follow-up later on.

@ladyada


ladyada:
thanks!

PR #8907, opened 2023-11-21
Adding IPv6 Support for Arduino 3.0.0

Jason2866:
Follow up or better replaces PR https://github.com/espressif/arduino-esp32/pull/7722
Port is done by @s-hadinger 

fyi and review @P-R-O-C-H-Y

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -20,68 +20,87 @@
 #include <Arduino.h>
 #include <IPAddress.h>
 #include <Print.h>
+#include <StreamString.h>
 
-IPAddress::IPAddress()
-{
-    _address.dword = 0;
+IPAddress::IPAddress() {
+#if LWIP_IPV6
+  _ip = *IP6_ADDR_ANY;
+#else
+  _ip = *IP_ADDR_ANY;
+#endif
+    // _ip = *IP_ANY_TYPE; // lwIP's v4-or-v6 generic address
 }
 
-IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet)
+IPAddress::IPAddress(const IPAddress& from)
 {
-    _address.bytes[0] = first_octet;
-    _address.bytes[1] = second_octet;
-    _address.bytes[2] = third_octet;
-    _address.bytes[3] = fourth_octet;
+    ip_addr_copy(_ip, from._ip);
 }
 
-IPAddress::IPAddress(uint32_t address)
-{
-    _address.dword = address;
+IPAddress::IPAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet) {
+    uint8_t addr[] {
+        first_octet,
+        second_octet,
+        third_octet,
+        fourth_octet,
+    };
+    *this = &addr[0];
 }
 
-IPAddress::IPAddress(const uint8_t *address)
-{
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
+IPAddress::IPAddress(uint8_t o1, uint8_t o2, uint8_t o3, uint8_t o4, uint8_t o5, uint8_t o6, uint8_t o7, uint8_t o8, uint8_t o9, uint8_t o10, uint8_t o11, uint8_t o12, uint8_t o13, uint8_t o14, uint8_t o15, uint8_t o16) {
+    setV6();
+    (*this)[0] = o1;
+    (*this)[1] = o2;
+    (*this)[2] = o3;
+    (*this)[3] = o4;
+    (*this)[4] = o5;
+    (*this)[5] = o6;
+    (*this)[6] = o7;
+    (*this)[7] = o8;
+    (*this)[8] = o9;
+    (*this)[9] = o10;
+    (*this)[10] = o11;
+    (*this)[11] = o12;
+    (*this)[12] = o13;
+    (*this)[13] = o14;
+    (*this)[14] = o15;
+    (*this)[15] = o16;
 }
 
-IPAddress& IPAddress::operator=(const uint8_t *address)
-{
-    memcpy(_address.bytes, address, sizeof(_address.bytes));
-    return *this;
+IPAddress::IPAddress(IPType type, const uint8_t *address) {
+    IPAddress(type, address, 0);
 }
 
-IPAddress& IPAddress::operator=(uint32_t address)
-{
-    _address.dword = address;
-    return *this;
-}
-
-bool IPAddress::operator==(const uint8_t* addr) const
-{
-    return memcmp(addr, _address.bytes, sizeof(_address.bytes)) == 0;
-}
-
-size_t IPAddress::printTo(Print& p) const
-{
-    size_t n = 0;
-    for(int i = 0; i < 3; i++) {
-        n += p.print(_address.bytes[i], DEC);
-        n += p.print('.');
+IPAddress::IPAddress(IPType type, const uint8_t *address, uint8_t zone) {
+    if (type == IPv4) {
+        setV4();
+        memcpy(&this->_ip.u_addr.ip4, address, 4);
+    } else if (type == IPv6) {
+        setV6();
+        memcpy(&this->_ip.u_addr.ip6.addr[0], address, 16);
+        setZone(zone);
+    } else {
+#if LWIP_IPV6
+  _ip = *IP6_ADDR_ANY;
+#else
+  _ip = *IP_ADDR_ANY;
+#endif
     }
-    n += p.print(_address.bytes[3], DEC);
-    return n;
+
 }
 
-String IPAddress::toString() const
-{
-    char szRet[16];
-    sprintf(szRet,"%u.%u.%u.%u", _address.bytes[0], _address.bytes[1], _address.bytes[2], _address.bytes[3]);
-    return String(szRet);
+bool IPAddress::fromString(const char *address) {
+    if (!fromString4(address)) {
+#if LWIP_IPV6
+        return fromString6(address);
+#else
+        return false;
+#endif
+    }
+    return true;
 }
 
-bool IPAddress::fromString(const char *address)
-{
-    // TODO: add support for "a", "a.b", "a.b.c" formats
+bool IPAddress::fromString4(const char *address) {
+    // TODO: (IPv4) add support for "a", "a.b", "a.b.c" formats
 
     uint16_t acc = 0; // Accumulator
     uint8_t dots = 0;
@@ -103,7 +122,7 @@ bool IPAddress::fromString(const char *address)
                 // Too much dots (there must be 3 dots)
                 return false;
             }
-            _address.bytes[dots++] = acc;
+            (*this)[dots++] = acc;
             acc = 0;
         }
         else
@@ -117,9 +136,122 @@ bool IPAddress::fromString(const char *address)
         // Too few dots (there must be 3 dots)
         return false;
     }
-    _address.bytes[3] = acc;
+    (*this)[3] = acc;
+
+    setV4();
     return true;
 }
 
+IPAddress& IPAddress::operator=(const uint8_t *address) {
+    uint32_t value;
+    memcpy_P(&value, address, sizeof(value));
+    *this = value;
+    return *this;
+}
+
+IPAddress& IPAddress::operator=(uint32_t address) {
+    setV4();
+    v4() = address;
+    return *this;
+}
+
+bool IPAddress::operator==(const uint8_t* addr) const {
+    if (!isV4()) {
+        return false;
+    }
+
+    uint32_t value;
+    memcpy_P(&value, addr, sizeof(value));
+
+    return v4() == value;
+}
+
+size_t IPAddress::printTo(Print& p) const {
+    size_t n = 0;
+
+#if LWIP_IPV6
+    if (isV6()) {
+        int count0 = 0;
+        for (int i = 0; i < 8; i++) {
+            uint16_t bit = PP_NTOHS(raw6()[i]);
+            if (bit || count0 < 0) {
+                n += p.printf("%x", bit);
+                if (count0 > 0)
+                    // no more hiding 0
+                    count0 = -8;
+            } else
+                count0++;
+            if ((i != 7 && count0 < 2) || count0 == 7)
+                n += p.print(':');
+        }
+        // add a zone if zone-id si non-zero
+        if (_ip.u_addr.ip6.zone) {
+            n += p.print('%');
+            char if_name[NETIF_NAMESIZE];
+            netif_index_to_name(_ip.u_addr.ip6.zone, if_name);
+            n += p.print(if_name);
+        }
+        return n;
+    }
+#endif
+
+    for(int i = 0; i < 4; i++) {
+        n += p.print((*this)[i], DEC);
+        if (i != 3)
+            n += p.print('.');
+    }
+    return n;
+}
+
+String IPAddress::toString() const
+{
+    StreamString sstr;
+#if LWIP_IPV6
+    if (isV6())
+        sstr.reserve(44); // 8 shorts x 4 chars each + 7 colons + nullterm + '%' + zone-id
+    else
+#endif
+        sstr.reserve(16); // 4 bytes with 3 chars max + 3 dots + nullterm, or '(IP unset)'
+    printTo(sstr);
+    return sstr;
+}
+
+bool IPAddress::isValid(const String& arg) {
+	return IPAddress().fromString(arg);
+}
+
+bool IPAddress::isValid(const char* arg) {
+	return IPAddress().fromString(arg);
+}
+
+const IPAddress INADDR46_ANY; // generic "0.0.0.0" for IPv4 & IPv6
+const IPAddress INADDR46_NONE(255,255,255,255);
+
+void IPAddress::clear() {
+    (*this) = INADDR46_ANY;
+}
+
+/**************************************/
+
+#if LWIP_IPV6
+
+bool IPAddress::fromString6(const char *address) {
+    ip6_addr_t ip6;
+    if (ip6addr_aton(address, &ip6)) {
+        setV6();
+        memcpy(&this->_ip.u_addr.ip6.addr[0], ip6.addr, 16);
+        // look for '%' in string
+        const char *s = address;
+        while (*s && *s != '%') { s++; }
+        if (*s == '%') {
+            // we have a zone id
+            setZone(netif_name_to_index(s + 1));
+        }
+        return true;
+    }
+    return false;
+}
+#endif // LWIP_IPV6
+
 // declared one time - as external in IPAddress.h
-IPAddress INADDR_NONE(0, 0, 0, 0);
+IPAddress INADDR_NONE(0, 0, 0, 0);      // TODO
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -28,7 +28,6 @@ IPAddress::IPAddress() {
 #else
   _ip = *IP_ADDR_ANY;
 #endif
-    // _ip = *IP_ANY_TYPE; // lwIP's v4-or-v6 generic address
 }
 
 IPAddress::IPAddress(const IPAddress& from)
@@ -80,9 +79,9 @@ IPAddress::IPAddress(IPType type, const uint8_t *address, uint8_t zone) {
         setZone(zone);
     } else {
 #if LWIP_IPV6
-  _ip = *IP6_ADDR_ANY;
+        _ip = *IP6_ADDR_ANY;
 #else
-  _ip = *IP_ADDR_ANY;
+        _ip = *IP_ADDR_ANY;
 #endif
     }
 
--- a/libraries/WiFi/src/WiFiClient.cpp
+++ b/libraries/WiFi/src/WiFiClient.cpp
@@ -23,6 +23,11 @@
 #include <lwip/netdb.h>
 #include <errno.h>
 
+#define IN6_IS_ADDR_V4MAPPED(a) \
+        ((((__const uint32_t *) (a))[0] == 0) \
+         && (((__const uint32_t *) (a))[1] == 0) \
+         && (((__const uint32_t *) (a))[2] == htonl (0xffff)))
+
 #define WIFI_CLIENT_DEF_CONN_TIMEOUT_MS  (3000)
 #define WIFI_CLIENT_MAX_WRITE_RETRY      (10)
 #define WIFI_CLIENT_SELECT_TIMEOUT_US    (1000000)
@@ -217,22 +222,33 @@ int WiFiClient::connect(IPAddress ip, uint16_t port)
 {
     return connect(ip,port,_timeout);
 }
+
 int WiFiClient::connect(IPAddress ip, uint16_t port, int32_t timeout_ms)
 {
+    struct sockaddr_storage serveraddr = {};
     _timeout = timeout_ms;
-    int sockfd = socket(AF_INET, SOCK_STREAM, 0);
+    int sockfd = -1;
+
+    if (ip.type() == IPv6) {
+        struct sockaddr_in6 *tmpaddr = (struct sockaddr_in6 *)&serveraddr;
+        sockfd = socket(AF_INET6, SOCK_STREAM, 0);
+        tmpaddr->sin6_family = AF_INET6;
+        memcpy(tmpaddr->sin6_addr.un.u8_addr, &ip[0], 16);
+        tmpaddr->sin6_port = htons(port);
+        tmpaddr->sin6_scope_id = ip.zone();
+    } else {
+        struct sockaddr_in *tmpaddr = (struct sockaddr_in *)&serveraddr;
+        sockfd = socket(AF_INET, SOCK_STREAM, 0);
+        tmpaddr->sin_family = AF_INET;
+        tmpaddr->sin_addr.s_addr = ip;
+        tmpaddr->sin_port = htons(port);
+    }
     if (sockfd < 0) {
         log_e("socket: %d", errno);
         return 0;
     }
     fcntl( sockfd, F_SETFL, fcntl( sockfd, F_GETFL, 0 ) | O_NONBLOCK );
 
-    uint32_t ip_addr = ip;
-    struct sockaddr_in serveraddr;
-    memset((char *) &serveraddr, 0, sizeof(serveraddr));
-    serveraddr.sin_family = AF_INET;
-    memcpy((void *)&serveraddr.sin_addr.s_addr, (const void *)(&ip_addr), 4);
-    serveraddr.sin_port = htons(port);
     fd_set fdset;
     struct timeval tv;
     FD_ZERO(&fdset);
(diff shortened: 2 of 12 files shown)


VojtechBartoska:
Thanks @Jason2866 for this PR. 

Just writing a note to close/process other Ipv6 PRs if this one will be merged.

libraries/WiFi/src/WiFiUdp.cpp
```diff
@@ -62,12 +66,30 @@ uint8_t WiFiUDP::begin(IPAddress address, uint16_t port){
...
+    memset((char *) tmpaddr, 0, sizeof(struct sockaddr_in));
+    tmpaddr->sin6_family = AF_INET6;
+    tmpaddr->sin6_port = htons(server_port);
+    tmpaddr->sin6_scope_id = ip_addr->u_addr.ip6.zone;
+    inet6_addr_from_ip6addr(&tmpaddr->sin6_addr, ip_2_ip6(ip_addr));
+    tmpaddr->sin6_flowinfo = 0;
+    sock_size = sizeof(sockaddr_in6);
+  } else
+#endif
+  if (1) {
```
me-no-dev, line 84: you can remove `if (1)`

libraries/WiFi/src/WiFiUdp.cpp
```diff
@@ -111,19 +157,37 @@ void WiFiUDP::stop(){
...
+      // iterate on each interface
+      for (netif* intf = netif_list; intf != nullptr; intf = intf->next) {
+        mreq.ipv6mr_interface = intf->num + 1;
+        if (intf->name[0] != 'l' || intf->name[1] != 'o') {   // skip 'lo' local interface
+          setsockopt(udp_server, IPPROTO_IPV6, IPV6_LEAVE_GROUP, &mreq, sizeof(mreq));
+        }
+      }
+    } else
+#endif
+    if (1) {
```
me-no-dev, line 176: same about `if (1)`

libraries/WiFi/src/WiFiUdp.cpp
```diff
@@ -77,23 +99,47 @@ uint8_t WiFiUDP::begin(IPAddress address, uint16_t port){
...
+          }
+        }
+        if (!joined) {
           log_e("could not join igmp: %d", errno);
           stop();
           return 0;
+        }
+      } else
+#endif
+      if (1) {
```
me-no-dev, line 130: same about `if (1)`

me-no-dev:
Why not use the official `IPAddress` class from Arduino.cc? https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.h

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -217,22 +222,33 @@ int WiFiClient::connect(IPAddress ip, uint16_t port)
...
+
 int WiFiClient::connect(IPAddress ip, uint16_t port, int32_t timeout_ms)
 {
+    struct sockaddr_storage serveraddr = {};
     _timeout = timeout_ms;
-    int sockfd = socket(AF_INET, SOCK_STREAM, 0);
+    int sockfd = -1;
+
+    if (ip.type() == IPv6) {
+        struct sockaddr_in6 *tmpaddr = (struct sockaddr_in6 *)&serveraddr;
```
me-no-dev, line 233: other IPv6 code is guarded. Shouldn't this check if v6 is enabled too?

libraries/WiFi/src/WiFiClient.cpp
```diff
@@ -552,8 +581,24 @@ IPAddress WiFiClient::remoteIP(int fd) const
...
-    return IPAddress((uint32_t)(s->sin_addr.s_addr));
+
+    // IPv4 socket, old way
+    if (((struct sockaddr*)&addr)->sa_family == AF_INET) {
+        struct sockaddr_in *s = (struct sockaddr_in *)&addr;
+        return IPAddress((uint32_t)(s->sin_addr.s_addr));
+    }
+
+    // IPv6, but it might be IPv4 mapped address
+    if (((struct sockaddr*)&addr)->sa_family == AF_INET6) {
```
me-no-dev, line 592: same question about v6 guards

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1590,6 +1609,37 @@ int WiFiGenericClass::hostByName(const char* aHostname, IPAddress& aResult)
...
 }
 
+/**
+ * Resolve the given hostname to an IP6 address.
+ * @param aHostname     Name to be resolved
+ * @param aResult       IPv6Address structure to store the returned IP address
+ * @return 1 if aHostname was successfully converted to an IP address,
+ *          else error code
+ */
+int WiFiGenericClass::hostByName6(const char* aHostname, ip_addr_t& aResult)
```
me-no-dev, line 1619: this should accept `IPAddress` instead. We use Arduino APIs if we can. Also this function is not guarded if v6 is enabled.

libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -196,6 +202,7 @@ class WiFiGenericClass
     static const char * getHostname();
     static bool setHostname(const char * hostname);
     static bool hostname(const String& aHostname) { return setHostname(aHostname.c_str()); }
+    static int hostByName6(const char *aHostname, ip_addr_t& aResult);
```
me-no-dev, line 205: This method should be private and user should need to call only `hostByName`. You can add extra argument to `hostByName` that can force v6 answer, but if not set, there should be a logic to pick which result to give.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -160,6 +161,8 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout)
             log_i("[WIFI] Connecting BSSID: %02X:%02X:%02X:%02X:%02X:%02X SSID: %s Channel: %d (%d)", bestBSSID[0], bestBSSID[1], bestBSSID[2], bestBSSID[3], bestBSSID[4], bestBSSID[5], bestNetwork.ssid, bestChannel, bestNetworkDb);
 
             WiFi.begin(bestNetwork.ssid, bestNetwork.passphrase, bestChannel, bestBSSID);
+            if (ipv6_support == true)
+                WiFi.IPv6(true);
```
me-no-dev, line 165: can't you just proxy `WiFiMulti::IPv6` to `WiFi.IPv6` and not need to keep a redundant variable?
```cpp
void WiFiMulti::IPv6(bool state) {
     WiFi.IPv6(state);
}
```

libraries/WiFi/src/WiFiServer.cpp
```diff
@@ -76,14 +76,23 @@ void WiFiServer::begin(uint16_t port, int enable){
...
   if (sockfd < 0)
     return;
   setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(int));
-  server.sin_family = AF_INET;
-  server.sin_addr.s_addr = _addr;
-  server.sin_port = htons(_port);
+  server.sin6_family = AF_INET6;
+  if (_addr.type() == IPv4) {
+    // log_e("---------------- IPv4");
+    memcpy(server.sin6_addr.s6_addr+11, (uint8_t*)&_addr[0], 4);
```
me-no-dev, line 87: this should probably be `+12`?

libraries/WiFi/src/WiFiServer.cpp
```diff
@@ -76,14 +76,23 @@ void WiFiServer::begin(uint16_t port, int enable){
...
+    server.sin6_addr.s6_addr[10] = 0xFF;
+    server.sin6_addr.s6_addr[11] = 0xFF;
+  } else {
+    // log_e("---------------- IPv6");
+    memcpy(server.sin6_addr.s6_addr, (uint8_t*)&_addr[0], 16);
+  }
+  memset(server.sin6_addr.s6_addr, 0x0, 16);
```
me-no-dev, line 94: After copying the IPs in the above lines, you then overwrite it with zero? Maybe this memset was supposed to be above the if/else?

libraries/WiFi/src/WiFiUdp.cpp
s-hadinger: will do, and leave only `{`

libraries/WiFi/src/WiFiUdp.cpp
s-hadinger: will do, and leave only `{`

libraries/WiFi/src/WiFiUdp.cpp
s-hadinger: will do, and leave only `{`

s-hadinger:
> Why not use the official `IPAddress` class from Arduino.cc? https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.h

The official is missing too many features. This one should be an superset, and it is well aligned with ESP8266

libraries/WiFi/src/WiFiUdp.cpp
s-hadinger: Will do

libraries/WiFi/src/WiFiClient.cpp
s-hadinger: Good catch, will fix

libraries/WiFi/src/WiFiClient.cpp
s-hadinger: Same

libraries/WiFi/src/WiFiGeneric.cpp
s-hadinger: I will check this one, I think it comes from ESP8266

libraries/WiFi/src/WiFiGeneric.h
s-hadinger: Will check

libraries/WiFi/src/WiFiMulti.cpp
s-hadinger: Will check

libraries/WiFi/src/WiFiServer.cpp
s-hadinger: I'm not sure where this code comes from, but indeed it should be 12

libraries/WiFi/src/WiFiServer.cpp
s-hadinger: Same comment, I'm not sure where this code comes from, will check.

libraries/WiFi/src/WiFiMulti.cpp
s-hadinger: I think the main reason is that when calling `WiFiMulti::IPv6()`, the `WiFi` object is not yet available. The whole creation of `WiFi` must probably be done in the right sequence, and `WiFiMulti` needs to keep track if it enables IPv6 or not.

libraries/WiFi/src/WiFiServer.cpp
s-hadinger: Fixed

libraries/WiFi/src/WiFiServer.cpp
s-hadinger: Fixed

libraries/WiFi/src/WiFiGeneric.h
s-hadinger: Actually I agree, this code comes from another source (can't remember which) and we don't use it in Tasmota.

me-no-dev:
> The official is missing too many features. This one should be an superset, and it is well aligned with ESP8266

My issue is mainly that it depends on esp-idf as is the code currently now. I would like those classes to have nothing to do with the underlying OS (just like they are in mainstream Arduino).

libraries/WiFi/src/WiFiMulti.cpp
me-no-dev: The `WiFi` object is available before even `setup()` is called and the user has a chance to call `WiFi.IPv6()` so it should be OK. Can you please confirm?

cores/esp32/IPAddress.cpp
```diff
@@ -117,9 +135,122 @@ bool IPAddress::fromString(const char *address)
...
+    return v4() == value;
+}
+
+size_t IPAddress::printTo(Print& p) const {
+    size_t n = 0;
+
+#if LWIP_IPV6
+    if (isV6()) {
+        int count0 = 0;
+        for (int i = 0; i < 8; i++) {
```
sgryphon, line 174: Is this supposed to output the canonical format, i.e. with consecutive zeros of the longest run (not necessarily the first) combined?

sgryphon: There is a ip6addr_ntoa() that will do the convert for us (and we then copy to the print buffer).

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -1561,8 +1561,27 @@ static void wifi_dns_found_callback(const char *name, const ip_addr_t *ipaddr, v
...
+    }
+    xEventGroupSetBits(_arduino_event_group, WIFI_DNS_DONE_BIT);
+}
+
+/**
+ * Resolve the given hostname to an IP address.
+ * @param aHostname     Name to be resolved
  * @param aResult       IPAddress structure to store the returned IP address
  * @return 1 if aIPAddrString was successfully converted to an IP address,
  *          else error code
```
sgryphon, line 1587: With IPv6 enabled, this doesn't do what it says -- it does not get back the IPAddress of the host name (which might be IPv6 only, or dual stack) -- it gives back the IPv4 address only.

libraries/WiFi/src/WiFiGeneric.cpp
sgryphon, line 1619: Why is this a separate function from hostByName() ???    If I have enabled IPv6, then I just want a function to pass in a host name and get back the IPAddress (any type).  I can check the type if I want, but I should just call one function and get back the address irrespective of whether the target is IPv6 only, dual stack, or IPv4.  

The IPAddress result can handle all of these cases.

libraries/WiFi/src/WiFiGeneric.h
sgryphon: Host are more likely to have multiple IPv6 addresses than multiple IPv4, but either is possible, so rather than one hostByName() with one address results, there probably should be a way to get multiple results.

For a function that only returns one result, it should kind of follow the logic of RFC 3484. Usually that logic is used to select both the source and destination address, given a list of both. e.g. if both have a local scope use that.

However in the case of hostByName we only have the destination address (not the source); as it is a DNS lookup, it is probably also a global address, and for safety we should assume that, i.e. return a global address in preference to a local scope one.

Generally then the algoritm would preference IPv6 over IPv4 over deprecated addresses.

sgryphon: I have a very rough algorithm `GsmModemCommon::compareIPAddress` at https://github.com/sgryphon/AdvancedGsmClient/blob/main/src/AdvancedGsm/GsmModemCommon.cpp

Note that it was written pre-support for IPv6 and so operates on raw strings. This means scopes like link-local are determined by "fe80:" prefix, rather than actual bytes. But it could be converted relatively easily.

libraries/WiFi/src/WiFiSTA.h
```diff
@@ -92,6 +92,7 @@ class WiFiSTAClass
     uint8_t subnetCIDR();
     
     bool enableIpV6();
+    bool IPv6(bool state);
     IPv6Address localIPv6();
```
sgryphon, line 96: Is this now to be marked deprecated (i.e. use IPAddress in preference to IPv6Address).

Also, will the localIP() function be updated to return IPv6 addresses, e.g. if IPv6-only or dual stack?

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -796,6 +796,19 @@ bool WiFiSTAClass::enableIpV6()
...
+ */
+bool WiFiSTAClass::IPv6(bool state)
+{
+   if (state)
+       WiFiGenericClass::setStatusBits(WIFI_WANT_IP6_BIT);
+   else
+       WiFiGenericClass::clearStatusBits(WIFI_WANT_IP6_BIT);
+   return true;
+}
+
```
sgryphon, line 811: There is also a bunch of code, like localIP(...) that only returns the IPv4 version. If IPAddress now supports both, then it should return both types of values as appropriate, e.g. if IPv6 only or dual stack.

There should also be tests to ensure things like config(IPAddress local_ip, ...) support configuration with a static IPv6.

cores/esp32/IPAddress.cpp
sgryphon: There is also an `IPAddress:printTo()` function in https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.cpp which has the correct IPv6 IETF canonical format to compress the left-most longest run of two or more zero fields.

It does two passes... first start from the left to find a run of two or more zero fields, only replacing if a later run is longer (not the same). This gives a single location of the left-most longest run. The second pass then outputs, adding the compression at the detected location.

... but if we have access to ip6addr_ntoa() it might be better to just use that.

sgryphon: I also suggest adding an automated test project. The ArduinoCore-API has automated tests for a whole bunch of corner cases of canonical to string, at https://github.com/arduino/ArduinoCore-API/blob/master/test/src/IPAddress/test_printTo6.cpp

e.g. where there is a run of 2 zeros followed by a later run of 3 zeros.

me-no-dev:
@sgryphon there is a lot more work that will be done. Not just IPv6 but the whole network stack. Many things will change places. We still unfortunately need to keep compatible with original Arduino WiFi lib, ESP8266 and so on.

libraries/WiFi/src/WiFiMulti.cpp
s-hadinger: This is fixed now, as you suggested, and tested: confirmed to work fine.

libraries/WiFi/src/WiFiGeneric.h
s-hadinger: To be fair, I don't use this one anymore in Tasmota. I call directly `dns_gethostbyname_addrtype()` with some custom logic to favor IPv4 or IPv6 depending on configuration of options.
I'm ok with any implementation.

s-hadinger:
> > The official is missing too many features. This one should be an superset, and it is well aligned with ESP8266
> 
> My issue is mainly that it depends on esp-idf as is the code currently now. I would like those classes to have nothing to do with the underlying OS (just like they are in mainstream Arduino).

I will at least check that the public interface is the same

cores/esp32/IPAddress.cpp
s-hadinger: This is a cut and paste from ESP8266. I have not experienced any issue in IPv6 formatting. Do you have a breaking example?

libraries/WiFi/src/WiFiGeneric.cpp
s-hadinger: Same comment as above, maybe we should think about unifying both to a single call

libraries/WiFi/src/WiFiGeneric.cpp
s-hadinger: True, this was also a cut and paste, we should revise

libraries/WiFi/src/WiFiSTA.h
s-hadinger: True.

s-hadinger: True that we don't need `IPv6Address` anymore.

For `localIP()`, we had already the discussion with Core2. For compatibility, I think that `localIP()` should only return IPv4, otherwise a lot of code will break. We need a separate function to get local IPv6

libraries/WiFi/src/WiFiSTA.cpp
s-hadinger: See above, if legacy `localIP()` returns IPv6, a lot of legacy code will break

cores/esp32/IPAddress.cpp
sgryphon: A tricky example, where the last run of zeros is longest is: ip(0,0, 0,1, 0,0, 0,0, 0,2, 0,0, 0,0, 0,0). This should output: "0:1:0:0:2::"

Executing the current algorithm by hand gives ":1:0:0:2:0:0:0" (I think)

Some other tricky ones to check

ip(0,0, 0,1, 0,0, 0,0, 0,0, 0,2, 0,0, 0,0); => "0:1::2:0:0"
ip(0,0, 0,2, 0,3, 0,4, 0,5, 0,6, 0,7, 0,8); => "0:2:3:4:5:6:7:8"
ip(0,1, 0,2, 0,3, 0,4, 0,5, 0,6, 0,7, 0,0); => "1:2:3:4:5:6:7:0"
ip(0,0, 0,0, 0,3, 0,4, 0,5, 0,6, 0,7, 0,0); => "::3:4:5:6:7:0"
ip(0,0, 0,2, 0,3, 0,4, 0,5, 0,6, 0,0, 0,0); => "0:2:3:4:5:6::"
ip(0,1, 0,0, 0,0, 0,4, 0,5, 0,0, 0,0, 0,8); => "1::4:5:0:0:8" (both the same length, so compress the left-most)

If you have automated tests, just copy the cases from ArduinoCore

libraries/WiFi/src/WiFiSTA.h
sgryphon: Should you be able to guard with `LWIP_IPV6` or maybe some additional attributes?

Legacy code that isn't recompiled should be fine. Legacy code with LWIP_IPV6 will also be fine (previous behaviour).

Even with LWIP_IPV6 turned on, if you don't enable WiFiSTA IPv6 then it won't get a v6 address, i.e. localIP() will only have an IPv4 to return.

At the point you call IPv6(true) to enable IPv6, I don't think it is legacy code anymore, and you shouldn't be surprised if legacy handling of localIP() fails after that.

What would be weird is if you turned on all the IPv6 stuff, e.g. called IPv6(true), and then localIP() did not return IPv6.

Maybe you do need a way to show multiple local addressess, i.e. pick to get only IPv4, but while there might be one IPv4, often there are multiple IPv6.

SuGlider (changes requested):
> > The official is missing too many features. This one should be an superset, and it is well aligned with ESP8266
> 
> My issue is mainly that it depends on esp-idf as is the code currently now. I would like those classes to have nothing to do with the underlying OS (just like they are in mainstream Arduino).

After reviewing this PR, I agree to @me-no-dev.
The IPV6 support should use not just `IPAdrress`, but also `IPv6Adress` -- maybe unify both...

The WiFi/Eth and other classes shall identify which address type (v4|v6) is used and execute the IDF Calls related to it. We shall make it as transparent as possible and correctly organized in software layers, trying to unify the methods.

There may be some work to be done here.

s-hadinger:
> There may be some work to be done here.

Agree with this, but I'm not equipped to make such big changes. Are you taking over from now?

SuGlider:
> > There may be some work to be done here.
> 
> Agree with this, but I'm not equipped to make such big changes. Are you taking over from now?

Thanks @s-hadinger, @sgryphon and @Jason2866 

We will discuss it here and take over based on your good work!

TD-er:
> After reviewing this PR, I agree to @me-no-dev. The IPV6 support should use not just `IPAdrress`, but also `IPv6Adress` -- maybe unify both...
> 

I am in favour of making it separate classes.
Not in the least because lots of legacy code, settings etc. assume 4 bytes for an IP-address.
But also because it isn't mutual exclusive IPv4 xor IPv6.
For example:
- result from a DNS lookup can have both an IPv4 and IPv6 address)
- IP address of an interface can have IPv4 and IPv6 addresses (even multiple and also with different scope)

Also if you have a single class for both IPv4 and IPv6, you still need quite a lot of checking in user code.
And IPv6 does have quite a few features completely unknown to IPv4, so it makes no sense to have functions for those in the class which also needs to support IPv4.

You could consider making a single wrapping class or helper to keep the result of a DNS lookup or "the IP-address" of an interface.
For example with a pointer to an IPv4 and a pointer to an array of IPv6 addresses.
Preferrably with some kind of smart pointer concept to make sure the user doesn't have to worry about deleting stuff.

But please keep IPv4 and IPv6 as separated classes.
This will keep functions that really need only IPv4 or IPv6 much simpler.



> The WiFi/Eth and other classes shall identify which address type (v4|v6) is used and execute the IDF Calls related to it. We shall make it as transparent as possible and correctly organized in software layers, trying to unify the methods.
> 
> There may be some work to be done here.

Not sure whether WiFi/Eth etc should be aware of layer-3 (and above) stuff.
So in principle such interfaces do not need to know about IPv4 or IPv6.
They only handle layer-2 stuff.

me-no-dev:
@TD-er Mainstream Arduino has one class for both v4 and v6 IPs and we need to stick to that. We could probably extend the IPAddress class into two new classes IPv4Address and IPv6Address to be used where we need a specific type of address, or just blankly accept IPAddress if it does not matter. In any case, we need to pick a route now and stick to it.

If we properly integrate this and make sure that all code does not care about the address type, we can make it work. The network itself does not care much what address it would connect to, as long as it can reach it.

Since we do not now support IPv6 properly, we could ease into it a bit, use a common IPAddress class and see how thing will go, then expand from there.

cores/esp32/IPAddress.cpp
```diff
@@ -20,68 +20,86 @@
...
 
-IPAddress::IPAddress()
-{
-    _address.dword = 0;
+IPAddress::IPAddress() {
+#if LWIP_IPV6
+  _ip = *IP6_ADDR_ANY;
+#else
+  _ip = *IP_ADDR_ANY;
+#endif
```
TD-er, line 30: Default constructor does not set all bytes to 0.
Thus when setting an IPv4 address only using `operator[]`, the IPv6 bytes will be uninitialized and thus any compare (`operator==` not using an `uint8_t` pointer) will fail.

sgryphon:
> @me-no-dev Do you have a link to this Arduino class implementation you're referring to?
> Is it this one? https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.h 

Yes, I think that is the one.

> And even functions like [`operator==`] do not even seem to work for IPv6 addresses.

Comparing, via ===, to a raw pointer is a bit of an uncommon thing to do. I mean you can't even assume the pointer has at least 4 bytes, but if it didn't then why would you even be comparing.

> Also the default constructor doesn't seem to set all bytes to 0
> bool test() {
>   IPAddress ip1, ip2;
> Not tested, but above function will probably return false most of the times

There is a test suite, so if you are concerned you could add your test there, or raise an issue. 

There is already a test for the default constructor, but you could add more if you like (and test them): https://github.com/arduino/ArduinoCore-API/blob/1cec094bddec923824f48a756455d5a75215102f/test/src/IPAddress/test_IPAddress.cpp#L17

Also, the default constructor does appear to do a memset on all bytes to zero: https://github.com/arduino/ArduinoCore-API/blob/1cec094bddec923824f48a756455d5a75215102f/api/IPAddress.cpp#L30

ArduinoCore API is used by several implementations - megaavr, mbed, samd, renesas - and other uses are encouraged.

It does not use a specific networking library, so you may get some benefit from going your own way, but I suggest not straying too far, so that generic "Arduino" code will also work on your platform.

libraries/WiFi/src/WiFiSTA.cpp
sgryphon: > See above, if legacy localIP() returns IPv6, a lot of legacy code will break

No it won't.

If you have IPv6 turned off (e.g. WIFI_WANT_IP6_BIT) then you will be backwards compatible, e.g. comparing strings, storing in 15 chars, display on screen, user input, etc. will all work. In an IPv4 only network, localIP() will only ever return an IPv4 address.

If you have legacy code that does not support IPv6, then don't turn IPv6 on.

However, once you have updated all your code to support IPv6, then setting the flag should work. You don't want to have to do additional work to go through your entire code base and convert a single call to localIP() to a complex set of checking two separate functions and then deciding which is best.

sgryphon:

me-no-dev:
Arduino's version of IPAddress has been merged into master. Please rebase this PR to not modify that code.

Jason2866:
@me-no-dev A simple merge will break the function of the PR.

SuGlider (approved):

Jason2866:
@me-no-dev  Don't want to do different. We have no resources to implement every needed! change here to get a nice and Arduino compatible IPv6 integration. You wrote earlier you take over this PR. So i am a bit surprised.

me-no-dev:
@Jason2866 closing this in favor of https://github.com/espressif/arduino-esp32/pull/9016 let's discuss what you might be missing there. We can not edit this PR, because of permissions :)

PR #8917, opened 2023-11-25
WiFiClient copy constructor

JAndrassy:
WiFiClient has assignment operator defined and no copy constructor. Then I get a warning
```
implicitly-declared 'WiFiClient::WiFiClient(const WiFiClient&)' is deprecated [-Wdeprecated-copy]
```
on returning an existing client instance from a function as a return value.

This PR adds the copy constructor.

The copy-ability of implementations of the Client class is a feature of the 'Arduino language' to avoid use of pointers.


SuGlider:
@JAndrassy - I can't reproduce the message `implicitly-declared 'WiFiClient::WiFiClient(const WiFiClient&)' is deprecated [-Wdeprecated-copy]`. The complier should use the implicit copy construcutor whenever necessary.

Please provide more information:
- Arduino Core used
- Compiler Xtensa, Risc-V (SoC? ESP32, C3)
- The simpler Sketch example that can reproduce the message.

SuGlider (changes requested):
Need more information about the use case that this PR solves.

SuGlider:
This is interesting... I'm using exactly the same IDE version and C3 with the same configuration.
But I get no error message.
The difference is that I'm using Windows 11.

JAndrassy:
> But I get no error message.

maybe
image
in IDE Preferences?

PR #8926, opened 2023-11-27
Docs: Add ESP32-C2 to readme

lucasssvaz:
## Description of Change

This PR aims to add ESP32-C2 to the README.md file as support for this SoC was added in #8881.


P-R-O-C-H-Y (commented):
@lucasssvaz @pedrominatel Should we add a note at least, that you cannot find the C2 in the board list without commenting the hide option in `boards.txt` file and also you need to build the libs manually to the docs somewhere?

PR #8930, opened 2023-11-28
WiFiServer - don't inherit from Print and Server

JAndrassy:
Arduino has a base class Server. It is inherited from the Print class. The official base class Server has only `virtual void begin()=0;`. The Server class can't declare other required server methods `available()` and `accept()`, because they have the return type of the specific Client implementation of the library (for example EthernetClient in the Ethernet library). In ESP32 core the Server class is even modified (so external networking libraries produce an error).. **It is not possible to use the base class Server to work with an instance of an inherited class.**

Class Server inherits from Print for the print-to-all-clients functionality. This functionality is tied with a proper `available()` method implementation. The implementation of the Server class should register all connected clients for correct implementation of `available()` and for print-to-all-clients functionality. **The ESP32 WiFiServer doesn't manage clients for available() and print-to-all-clients. The write method is unimplemented.**

**This PR removes the inheritance from Server and Print. ESP8266 did it almost two years ago.** For ESP8266 I wrote [ArduinoWiFiServer](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/ArduinoWiFiServer.h) which inherits from WiFiServer and Print and manages the clients for proper `available()` and print-to-all-clients. But I doubt somebody uses it.  My NetApiHelpers library has ArduinoWiFiServer for ESP32, older version of esp8266 and MbedCore


SuGlider:
@JAndrassy - I understand that this PR wants to make `WiFiServer` class to expose the same APIs as Arduino ESP8266 `WiFiServer` does.

But looking into Arduino.cc implementation for `WiFiServer` I see exactly the same as we do in ESP32 Arduino.
https://github.com/arduino-libraries/WiFi/blob/master/src/WiFiServer.h

I'm not sure why we should follow ESP8226 and not the UpStream Arduino implementation.

Could you please elaborate more about it?
Thanks!

SuGlider (changes requested):
Please explain more about the change.

JAndrassy:
this PR a cleanup suggestion. removing code which never worked. what is unclear in the description of the PR in the first comment? 
the esp32 WiFiServer is a copy of the esp8266 WiFiServer. It doesn't work like WiFiServer in Arduino's WiFi libraries. 
and users request the esp8266 way, not the Arduino way.

you can read more about the 'Arduino'  server.available() and print-to-all-clents [here](https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPIDev.md#clients). The esp8266 and esp32 WiFiServer don't implememt that.

JAndrassy:
I was thinking about better explanation: By inheriting from Server base class the WiFiServer class declares that it works as Arduino Server. But it doesn't. It doesn't manage clients for proper `available()` and print-to-all-clients (for which Server inherits from Print). 
Changing the implementation of esp8266/esp32 WiFiServer would be a big breaking change without good reason. It would change how `server.available()` in 'esp' WiFiServer works.

VojtechBartoska:
Same here, we are not able to merge. We need to keep align with Arduino.cc and this introduce a breaking change, closing. Thanks for understanding.

JAndrassy:
do you plan to implement the print-to-all-clents function? make the `writ` work? he WiFiServer then has to mange the connected clients to print to all of them.

this example must work https://github.com/JAndrassy/Arduino/blob/d5eb265f78bff9deb7063d10030a02d021c8c66c/libraries/ESP8266WiFi/examples/PagerServer/PagerServer.ino#L61

in esp8266 core https://github.com/JAndrassy/Arduino/blob/master/libraries/ESP8266WiFi/src/ArduinoWiFiServer.h

PR #8950, opened 2023-12-01
[CI] Add to Boards Test check if build.board is uppercase

P-R-O-C-H-Y:
## Description of Change
Added automated check if boards_name.build.board is uppercase.

## Tests scenarios
Tested locally

## Related links
#7942

### DIFF
--- a/.github/workflows/boards.yml
+++ b/.github/workflows/boards.yml
@@ -45,6 +45,16 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Check if build.board is uppercase
+        run: |
+          board_name=$(echo "espressif:pltf:esp32" | awk -F':' '{print $NF}')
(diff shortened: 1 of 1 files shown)


SuGlider (approved):
Nice check!

SuGlider:
I found a board with '-' symbol too
`ttgo-t-oi-plus.build.board=TTGO-T-OI-PLUS_DEV`

P-R-O-C-H-Y:
> I found a board with '-' symbol too `ttgo-t-oi-plus.build.board=TTGO-T-OI-PLUS_DEV`

I see but almost all boards use only uppercase chars + numbers + _ 
I think we should just stick to it and don't allow any other symbols.

P-R-O-C-H-Y:
> Will we change all board to upper case?

For keeping compatibility we will leave old boards as is and only check this for new incoming boards.
Probably will change only some Heltec boards which were causing some issue. Need to double check it first.

PR #8433, opened 2023-07-19
2.0.10 escaping fix

pillo79:
## Description of Change
Fix escaping issues on 2.0.10 (Windows path and IDE 1.x). 

## Tests scenarios
Tested on clean Windows 11 box, Arduino IDE 2.1.1, username with spaces, 2.0.10. The issue is present and is fixed by these commits.
Includes patch by @pennam fixing #8424 for IDE 1.8.x.  

## Related links
Closes #8424 and fixes Windows path issues reported in Arduino forums e.g. [here](https://forum.arduino.cc/t/compilation-error-exit-status-1/1149342) and [here](https://forum.arduino.cc/t/compiling-for-esp32-fails-access-is-denied/1149574)

Issue #8424, opened 2023-07-18
Core Compilation failure on Linux in v2.0.10


josesnchz:
### Board

Adafruit ESP32 Feather - HUZZAH32

### Device Description

Plain Module in breadboard

### Hardware Configuration

Nothing attached to the dev board. Using USB power. 

### Version

v2.0.10

### IDE Name

Arduino IDE 

### Operating System

Ubuntu 22.04 with Kernel 5.15.0-76

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Running version v2.0.10, using any sketch, when it comes the moment to compile the core, I always get the same error. Downgrading to v2.0.9 solves the problem.

### Sketch

```cpp
/*
  Blink

  Turns an LED on for one second, then off for one second, repeatedly.

  Most Arduinos have an on-board LED you can control. On the UNO, MEGA and ZERO
  it is attached to digital pin 13, on MKR1000 on pin 6. LED_BUILTIN is set to
  the correct LED pin independent of which board is used.
  If you want to know what pin the on-board LED is connected to on your Arduino
  model, check the Technical Specs of your board at:
  https://www.arduino.cc/en/Main/Products

  modified 8 May 2014
  by Scott Fitzgerald
  modified 2 Sep 2016
  by Arturo Guadalupi
  modified 8 Sep 2016
  by Colby Newman

  This example code is in the public domain.

  https://www.arduino.cc/en/Tutorial/BuiltInExamples/Blink
*/

// the setup function runs once when you press reset or power the board
void setup() {
  // initialize digital pin LED_BUILTIN as an output.
  pinMode(LED_BUILTIN, OUTPUT);
}

// the loop function runs over and over again forever
void loop() {
  digitalWrite(LED_BUILTIN, HIGH);   // turn the LED on (HIGH is the voltage level)
  delay(1000);                       // wait for a second
  digitalWrite(LED_BUILTIN, LOW);    // turn the LED off by making the voltage LOW
  delay(1000);                       // wait for a second
}
```

### Debug Message

```plain
Compiling core...
bash -c "echo '-DARDUINO_CORE_BUILD' > /tmp/arduino_build_925905/file_opts"
>: -c: line 1: unexpected EOF while looking for matching `''
>: -c: line 2: syntax error: unexpected end of file
exit status 2
Error compiling for board Adafruit ESP32 Feather.
```

### Other Steps to Reproduce

[...]

PR #8968, opened 2023-12-07
3.0.0-alpha3 escaping fix

lyusupov:
## Description of Change

On a Linux host the **alpha3** gives a build error like this:

```
>: -c: line 1: unexpected EOF while looking for matching `''
>: -c: line 2: syntax error: unexpected end of file
```

**alpha2** was fine.

The issue looks very similar to #8424

The idea of this fix was taken from that PR #8433 .


lucasssvaz:
What is the IDE version you are using ? We were having this issue on the CI but it was due to an old IDE version being used. It was fixed in #8961.

lyusupov:
> What is the IDE version you are using ? 

**IDE1**  ( ARDUINO IDE VERSION: 1.8.13 )

me-no-dev:
@pillo79 could you please comment on this. It is indeed the case that older IDE1 versions have issue with escaping that line.

pillo79:
Thanks for the heads-up! We don't support the old IDE anymore so I had not checked that... but it's such an easy fix to make everyone happy, I'm all for it!

Issue #8952, opened 2023-12-03
Crash in WiFiScanClass::_scanDone() with negative scan result


TD-er:
### Board

Any

### Device Description

Not related to device, but crash happens on ESP32-C2 and -C3

### Hardware Configuration

Not HW related

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

I get crashes related to the new-operator when processing results from a WiFi scan
It seems to be caused by the code in this function:
https://github.com/espressif/arduino-esp32/blob/b811ea40875bbfbf8bd62b4038879ac7be2c2e2b/libraries/WiFi/src/WiFiScan.cpp#L109-L121

As can be seen, there is no check for negative scan results as the type of `_scanCount` is an `uint16_t`.
However the result of a scan can be negative, so maybe there is some conversion somewhere to this unsigned value and thus resulting in an attempt to allocate 65k elements of `wifi_ap_record_t`, which does fail at least on a C2.

### Sketch

```cpp
-
```

### Debug Message

```plain
abort() was called at PC 0x420e944d on core 0
=> 0x420e944d: __wrap__Unwind_RaiseException at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/cxx/cxx_exception_stubs.cpp:156
Core  0 register dump:
MEPC    : 0x40381ce8  RA      : 0x40385308  SP      : 0x3fcbb380  GP      : 0x3fca8e60
=> 0x40381ce8: panic_abort at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp_system/panic.c:471
=> 0x40385308: __ubsan_include at /home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/esp_system/ubsan.c:313
TP      : 0x3fc92268  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130
S0/FP   : 0x3fcbb3ac  S1      : 0x3fcbb390  A0      : 0x3fcbb3ac  A1      : 0x3fcbb38e
A2      : 0x00000000  A3      : 0x3fcbb3d9  A4      : 0x00000001  A5      : 0x3fcb3000
[...]

PR #8981, opened 2023-12-11
Wifi async scan example and fix

SuGlider:
## Description of Change
This PR adds a new example for using WiFi Scan in Asynchronous mode.
It also fixes a problem with Scan Status return related to Scan Timeout.

## Tests scenarios
Used the example with ESP32 and ESP32-C3.

## Related links

issue #8952 may benefit from it.


SuGlider:
This PR depends on #8982 due to a documentation CI error.

Issue #8999, opened 2023-12-14
analogReadMilliVolts() does not handle changes in resolution made by analogReadResolution()


bkari02:
### Board

ESP32-S2

### Device Description

Tested with Unexpected Maker FeatherS2 & senseBox MCU-S2 ESP32-S2

### Hardware Configuration

Any analog sensor (e.g. Truebner SMT50 or SparkFun SEN-13322) is connected.

### Version

v2.0.14

### IDE Name

ArduinoIDE

### Operating System

macOS 14.1.2

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Expected behaviour:
When using ESP32-S2 and `analogReadMilliVolts(pin)` I should always get the measured voltage in milliVolts. 

Weird behaviour: When changing the default 13 bit resolution of the ADC1 to e.g. 12bit using `analogReadResolution(12);` the outputs of analogReadMilliVolts(pin) are incorrect (half in size in case of 12bit). 

Tried to track it down and found this to be happening (in esp32-hal-adc.c):
The `__analogRead()` function uses the 12 bit resolution (as expected because of `analogReadResolution(12)`). But `__analogReadMilliVolts()` internally also uses `__analogRead()` which returns a 12bit ADC value. But during transformation in `esp_adc_cal_raw_to_voltage((uint32_t)adc_reading, &chars)` the adc_reading is always expected to be the raw 13bit value in case of the ESP32-S2. There it assumes the 12 bit value to be 13 bit and the resulting value in milliVolts half of the correct value. 

I assume this also happens with other chips than the S2, but haven't tested it so far. 

### Sketch

```cpp
// Define the pin that the analog sensor is connected to (in my case A2 of the UM FeatherS2)
#define analogSensorPin A2

void setup() {
  // initialize serial communication at 115200 bits per second:
  Serial.begin(115200);
  delay(1000); 
  
  // set the resolution to the 13 bits (0-8192)  --> this is default, but just to make sure
  analogReadResolution(13);
  int analogValue = analogRead(analogSensorPin);
  int analogVolts = analogReadMilliVolts(analogSensorPin);
  Serial.printf("13 bit - ADC analog value = %d\n",analogValue);
  Serial.printf("13 bit - ADC millivolts value = %d\n",analogVolts);

  // set the resolution to 12 bits (0-4096)
  analogReadResolution(12);
  analogValue = analogRead(analogSensorPin);
[...]

PR #9006, opened 2023-12-15
fix(adc): Always use default read resolution in __analogReadMilliVolts

bkari02:
## Description of Change
1. Add new __analogReadRaw function and move code from __analogRead without mapResolution part to __analogReadRaw.
2. Refactor __anlogRead to use analogReadRaw (and mapResolution).
3. Refactor __analogReadMilliVolts to always use default read resolution when reading adc value, as expected input by esp_adc_cal_raw_to_voltage is in default resolution (means replacing all calls of __analogRead with __analogReadRaw).

This shall avoid that a change of the resolution of analog readings leads to incorrect values when using the analogReadMilliVolts function, as this function always expects the resolution to be the default one. For more information see description in #8999. 
## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.14 with UM FeatherS2 (ESP32-S2) board with an analog sensor (Truebner  SMT50).

## Related links
Related to #8999.

### DIFF
--- a/cores/esp32/esp32-hal-adc.c
+++ b/cores/esp32/esp32-hal-adc.c
@@ -149,7 +149,7 @@ void __analogReadResolution(uint8_t bits)
 #endif
 }
 
-uint16_t __analogRead(uint8_t pin)
+uint16_t __analogReadRaw(uint8_t pin)
 {
     int8_t channel = digitalPinToAnalogChannel(pin);
     int value = 0;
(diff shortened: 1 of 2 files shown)


bkari02:
@P-R-O-C-H-Y @me-no-dev 
Not sure about this being the best way to solve the issue, just let me know if you have better ideas how to fix it.

me-no-dev:
Looks sound to me. @P-R-O-C-H-Y PTAL

P-R-O-C-H-Y (approved):
LGTM! Thanks for the fix :)

PR #9011, opened 2023-12-17
Improve Log Messages in GPIO HAL

### LINKED ISSUE #9008: log expansion
### Related area

logging

### Hardware specification

any

low logging info

line _167_ and line _177_ of [https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-gpio.c]arduino-esp32/blob/master/cores/esp32/esp32-hal-gpio.c(url)
`log_e("Pin is not set as GPIO.");`
change to 
`log_e("Pin %i is not set as GPIO.", pin);`

_No response_

### Additional context

_No response_

SuGlider:
## Description of Change
Improves the log messages by adding the GPIO number that has triggered the error message for better debugging.

## Tests scenarios
Tested with ESP32-C3.

## Related links
Fixes #9008

### DIFF
--- a/cores/esp32/esp32-hal-gpio.c
+++ b/cores/esp32/esp32-hal-gpio.c
@@ -104,14 +104,14 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
 #endif
 
     if (pin >= SOC_GPIO_PIN_COUNT) {
-        log_e("Invalid pin (%i) selected", pin);
+        log_e("Invalid IO %i selected", pin);
         return;
     }
 
     if(perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) == NULL){
         perimanSetBusDeinit(ESP32_BUS_TYPE_GPIO, gpioDetachBus);
         if(!perimanClearPinBus(pin)){
-            log_e("Deinit of previous bus from pin %i failed", pin);
+            log_e("Deinit of previous bus from IO %i failed", pin);
             return;
         }
     }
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-gpio.c
```diff
@@ -104,14 +104,14 @@ extern void ARDUINO_ISR_ATTR __pinMode(uint8_t pin, uint8_t mode)
...
-        log_e("Invalid pin selected");
+        log_e("Invalid pin (%i) selected", pin);
         return;
     }
 
     if(perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) == NULL){
         perimanSetBusDeinit(ESP32_BUS_TYPE_GPIO, gpioDetachBus);
         if(!perimanClearPinBus(pin)){
-            log_e("Deinit of previous bus failed");
+            log_e("Deinit of previous bus from GPIO %i failed", pin);
```
P-R-O-C-H-Y, line 114: Pin is not set as GPIO yet, so better to use "pin" instead of "GPIO".

P-R-O-C-H-Y (approved):
Just suggested small change, else looks good. :)

cores/esp32/esp32-hal-gpio.c
me-no-dev: Pin: usually the physical pin number on a chip or module
GPIO: usually the hardware IO+IOBUS number inside the chip. Do not apply for special purpose IOs
IO: usually wider sense, which includes GPIO and special purpose IOs

BUT because we are special and have IO mode = GPIO and all our IOs can be GPIOs, let's use IO from now on for when we want to specify something on the hardware level that might not be in GPIO mode.

If we are in `pinMode` or going through the matrix, then we are in GPIO mode. If we are using something analog (ADC, DAC, Touch, etc.) we are in RTC mode. Else we are in special mode (I'm pretty sure we don't currently use it, but UART0 starts that way, Flash/PSRAM are that way also and if IO numbers match SPI, we could use it there too for better 80MHz support)

lucasssvaz (approved):
LGTM

SuGlider:
@me-no-dev - Ready for final review.

P-R-O-C-H-Y (approved):
LGTM!

PR #9018, opened 2023-12-18
Set TFT_BACKLIGHT pin for adafruit_camera_esp32s3 variant

caternuson:
Just updating since pin is known now for this board.


caternuson:
I've had PR's merged before. Do I need to agree to the CLA with every PR?

caternuson:
Closing. This is already taken care of in #9021

PR #9027, opened 2023-12-20
undeprecate WiFiServer::available()

SuGlider:
## Description of Change
Removes depricated message for WiFiServer::available() to follow Arduino upstream API.

## Tests scenarios
None

## Related links
None

### DIFF
--- a/libraries/WiFi/src/WiFiServer.h
+++ b/libraries/WiFi/src/WiFiServer.h
@@ -45,7 +45,7 @@ class WiFiServer : public Server {
       log_v("WiFiServer::WiFiServer(addr=%s, port=%d, ...)", addr.toString().c_str(), port);
     }
     ~WiFiServer(){ end();}
-    WiFiClient available() __attribute__((deprecated("Renamed to accept().")));
+    WiFiClient available();
     WiFiClient accept();
     void begin(uint16_t port=0);
     void begin(uint16_t port, int reuse_enable);


lucasssvaz (approved):

JAndrassy:
do you want to implement the `available()` as it should work. the WiFiServer than has to mange the connected clients and return with available() only a client which has data available().
then this example must work https://github.com/JAndrassy/Arduino/blob/d5eb265f78bff9deb7063d10030a02d021c8c66c/libraries/ESP8266WiFi/examples/PagerServer/PagerServer.ino#L61

in esp8266 core https://github.com/JAndrassy/Arduino/blob/master/libraries/ESP8266WiFi/src/ArduinoWiFiServer.h

SuGlider:
@JAndrassy - We have discussed it and decided to follow WiFi Arduino API as much as possible. That's just it.
You are always welcome to add PRs and improvements.

If you consider worthy, please submit it upstream Arduino in their Github. We will follow!

JAndrassy:
the Arduino Core API doesn't have anything about `server.available()`. the Server class is
```
class Server : public Print {
  public:
    virtual void begin() = 0;
};
```

SuGlider:
> the Arduino Core API doesn't have anything about `server.available()`. the Server class is
> 
> ```
> class Server : public Print {
>   public:
>     virtual void begin() = 0;
> };
> ```

https://www.arduino.cc/reference/en/libraries/wifi/server.available/

JAndrassy:
> https://www.arduino.cc/reference/en/libraries/wifi/server.available/

"Gets a client that is connected to the server AND **has data available for reading**"

SuGlider:
This is the [Arduino example](https://www.arduino.cc/reference/en/libraries/wifi/server.available/) that closes the client connection:
``` cpp
void loop() {
  // listen for incoming clients
  WiFiClient client = server.available();
  if (client) {

    if (client.connected()) {
      Serial.println("Connected to client");
    }

    // close the connection:
    client.stop();
  }
}
```

PR #9028, opened 2023-12-20
WiFiServer - implementation aligned with Arduino.cc. With example. (breaking change)

JAndrassy:
implements `available()` as documented here https://www.arduino.cc/reference/en/libraries/wifi/server.available/

implements print-to-all-clients as documented here https://www.arduino.cc/reference/en/libraries/wifi/server.print/

adds WiFiPagerServer example to demonstrate how server.available() and  server.print() work. 

Requires #9029

### DIFF
--- a/libraries/WiFi/src/WiFiServer.h
+++ b/libraries/WiFi/src/WiFiServer.h
@@ -24,6 +24,10 @@
 #include "WiFiClient.h"
 #include "IPAddress.h"
 
+#ifndef SERVER_MAX_MONITORED_CLIENTS
+#define SERVER_MAX_MONITORED_CLIENTS CONFIG_LWIP_MAX_SOCKETS
+#endif
+
 class WiFiServer : public Server {
   private:
     int sockfd;
@@ -33,17 +37,23 @@ class WiFiServer : public Server {
     uint8_t _max_clients;
     bool _listening;
     bool _noDelay = false;
+    
+    WiFiClient connectedClients[SERVER_MAX_MONITORED_CLIENTS];
+    uint8_t index = 0;
+
+#if !CONFIG_DISABLE_HAL_LOCKS
+    SemaphoreHandle_t _lock;
+#endif
+
+    void acceptClients();
 
   public:
     void listenOnLocalhost(){}
 
-    WiFiServer(uint16_t port=80, uint8_t max_clients=4):sockfd(-1),_accepted_sockfd(-1),_addr(),_port(port),_max_clients(max_clients),_listening(false),_noDelay(false) {
-      log_v("WiFiServer::WiFiServer(port=%d, ...)", port);
-    }
-    WiFiServer(const IPAddress& addr, uint16_t port=80, uint8_t max_clients=4):sockfd(-1),_accepted_sockfd(-1),_addr(addr),_port(port),_max_clients(max_clients),_listening(false),_noDelay(false) {
-      log_v("WiFiServer::WiFiServer(addr=%s, port=%d, ...)", addr.toString().c_str(), port);
-    }
-    ~WiFiServer(){ end();}
+    WiFiServer(uint16_t port=80, uint8_t max_clients=4);
+    WiFiServer(const IPAddress& addr, uint16_t port=80, uint8_t max_clients=4);
+    ~WiFiServer();
+
     WiFiClient available();
     WiFiClient accept();
     void begin(uint16_t port=0);
--- a/libraries/WiFi/examples/WiFiPagerServer/WiFiPagerServer.ino
+++ b/libraries/WiFi/examples/WiFiPagerServer/WiFiPagerServer.ino
@@ -1,0 +1,59 @@
+/*
+  WiFi Pager Server - server.available
+  and print-to-all-clients example
+
+  The example is a simple server that echoes any incoming
+  messages to all connected clients. Connect two or more
+  telnet sessions to see how server.available() and
+  server.print() work.
+
+  created in September 2020
+  by Juraj Andrassy
+ */
+
+#include <WiFi.h>
+
+WiFiServer server(2323);
+
+#include "arduino_secrets.h"
+///////please enter your sensitive data in the Secret tab/arduino_secrets.h
+char ssid[] = SECRET_SSID;        // your network SSID (name)
+char pass[] = SECRET_PASS;    // your network password (use for WPA, or use as key for WEP)
+
+void setup() {
+
+  Serial.begin(115200);
+  while (!Serial);
+
+  Serial.print("Attempting to connect to SSID \"");
+  Serial.print(ssid);
+  Serial.println("\" with DHCP ...");
+  WiFi.begin(ssid, pass);
+  while (WiFi.status() != WL_CONNECTED) {
+    Serial.print('.');
+    delay(1000);
+  }
+  Serial.println();
+
+  server.begin();
+
+  IPAddress ip = WiFi.localIP();
+  Serial.println();
+  Serial.println("Connected to WiFi network.");
+  Serial.print("To access the server, connect with Telnet client to ");
+  Serial.print(ip);
+  Serial.println(" 2323");
+}
+
+void loop() {
+
+  WiFiClient client = server.available();     // returns first client which has data to read or a 'false' client
+  if (client) {                               // client is here true only if it is connected and has data to read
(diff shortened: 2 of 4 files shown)


VojtechBartoska:
Thanks @JAndrassy for the PR, we will review it soon. Probably in January due to Christmas and New Years Eve vacations.

We are trying to find the best approach for Wifi refactoring so hopefully we will end up in some reasonable solutions which will be also acceptable for you. :) It is a bit challenging.

JAndrassy:
The server.available() method was what started me to evaluate differences between Arduino networking libraries. Almost two years ago we found a solution for the esp8266 WiFi library: deprecating `available()` and providing a separate ArduinoWiFiServer class for `available()` and print-to-all-clients.

In August this year I stared to write down all my experience with creating and maintaining Arduino networking libraries, because new libraries appear and have the same mistakes I try to eliminate in old libraries. But writing the Guide let to necessity of defining the networking API by analyzing the significant libraries. I found many often simple differences so I started to do PR in all repositories. Then I wrote test sketches for basic functions and those found more errors in implementations than I would expect (even in my libraries.). So more PR.

VojtechBartoska:
I see, I have noticed you contributed to ESP8266 repository, I wasn`t aware about the contributions to other libraries as well. Good to know, we will check it.

lucasssvaz (commented):
Please fix the errors shown by the CI

JAndrassy:
@lucasssvaz this PR fixes it https://github.com/espressif/arduino-esp32/pull/9029

me-no-dev:
@JAndrassy
- what happens to clients that are connected, but have not sent any data?
- how would you implement a protocol that needs to send "HELLO" to every new connection?

JAndrassy:
> what happens to clients that are connected, but have not sent any data?

they stay connected. but it is possible that there is a timeout set in IDF LwIP. the AT firmware has it.

> how would you implement a protocol that needs to send "HELLO" to every new connection?

using server.accept(), not server.available(). 

The server.available() originates in Processing. It is a strange idea to simplify a TCP server for artist. I don't like it. But my [PR](https://github.com/espressif/arduino-esp32/pull/9028) to deprecate the name 'available' was [reverted](https://github.com/espressif/arduino-esp32/pull/9028).

me-no-dev:
> they stay connected. but it is possible that there is a timeout set in IDF LwIP. the AT firmware has it.

This is not a nice solution. Lingering resources with no way to clear them is not good.

> using server.accept(), not server.available().

But then the client will not be put in the list and you will not be able to further query it with `available()`?

I do understand very well your point, but I must be sure that this will cause more good than harm if merged :)

libraries/WiFi/src/WiFiServer.h
```diff
@@ -24,6 +24,10 @@
 #include "WiFiClient.h"
 #include "IPAddress.h"
 
+#ifndef SERVER_MAX_MONITORED_CLIENTS
+#define SERVER_MAX_MONITORED_CLIENTS 5
```
me-no-dev, line 28: `CONFIG_LWIP_MAX_SOCKETS` is `16`, meaning there can be more than 5 monitored clients. What happens with the rest?

JAndrassy:
> I do understand very well your point, but I must be sure that this will cause more good than harm if merged :)

I would prefer deprecation. as I already wrote on multiple places, nobody expects what available() does in Arduino libraries. basically developers at Arduino  don't know it and do it wrong in new libraries (https://github.com/arduino/uno-r4-wifi-usb-bridge/pull/20, https://github.com/arduino/ArduinoCore-mbed/pull/793)

the link to deprecation  PR was wrong in my previous comment. it is https://github.com/espressif/arduino-esp32/pull/8860

me-no-dev:
Ok :) before we can consider this for merging, here are some suggestions:
- Track all clients, not just the ones you accept through `available`
- Access to the list of clients must be guarded with semaphore (read/write)
- Add documentation explaining the changes and some small examples to show what to update in the code
- Make sure that all relevant examples are updated

JAndrassy:
examples are already updated to accept(). that was not reverted from the deprecation PR. this PR adds an example for server.available().
users will have to change their code from server.available() to server.accept() to have the same functionality as with 'old' available(). we have no way to warn them about that (as the deprecation would do)

JAndrassy:
now I realized there is a problem. Arduino doc instructs to stop() the connection returned by the server when it is handled. but client.stop() in this library does nothing and the connection is closed from local side only when the last copy of WiFiClient is destroyed. So the copy in WiFiServer will hold the connection.

can we please deprecate server.available and be done. nobody will miss it

me-no-dev:
so you are good to close this if we deprecate available?

JAndrassy:
> so you are good to close this if we deprecate available?

I already did the deprecation. it was reverted with the argument about "aligned with Arduino.cc"

me-no-dev:
I just need the full picture. Nobody has complained about this in the last 10 years, so if we are to break something, there ought to be a good reason. Since `available` is just an alias to `accept`, deprecating it will only force people to use `accept`, which will bring nothing to either party. Can't we just document that both do the same and that `available` is there for compatibility?

JAndrassy:
and https://github.com/espressif/arduino-esp32/pull/9028 ?

Jason2866:
Maybe i misunderstood the goal of the PR completely wrong. Drop `available` and replace with `accept` and break every existing code which uses `available`?
Doing this "just" to be compatible?! Nobody had an problem the last year's doing this way. Imho a really not needed breaking change, a less invasive solution should be found.

JAndrassy:
> Maybe i misunderstood the goal of the PR completely wrong

the goal of this PR was to show what does it mean to implement proper available() and print-to-all-clients
In my view it is the only alternative to available() marked as deprecated to clearly say that it is not the same as in other Arduino libraries.

me-no-dev:
> show what does it mean to implement proper available() and print-to-all-clients

I agreed with that and listed the required changes necessary to make it mergeable. We are on FreeRTOS with multiple cores and threads, so safety is important. Current code will not do. Also some connections are saved and others not is also not a solution. Either all or none.

Jason2866:
@JAndrassy Thx, for clarification.

JAndrassy:
> some connections are saved and others not is also not a solution. 

using CONFIG_LWIP_MAX_SOCKETS for SERVER_MAX_MONITORED_CLIENTS would solve that.  semaphore could be done too. 

but what with `stop()`? why `stop()` doesn't close the connection?

I coded the available() and print-to-all-clients enhancement of WiFiServer added in this PR 3 years ago for my TelnetStream library. Now it is in my [NetApiHelpers library](https://github.com/JAndrassy/NetApiHelpers#servertemplate). It doesn't have to be in the specific WiFi or Ethernet library.

me-no-dev:
@JAndrassy if you are willing to implement what is outlined [here](https://github.com/espressif/arduino-esp32/pull/9028#issuecomment-1893429233), then we can discuss merging :)

JAndrassy:
> Add documentation explaining the changes and some small examples to show what to update in the code

to migration guide?

how to fix the failing check? why does it compile a WiFi example for H2?

me-no-dev:
> how to fix the failing check? why does it compile a WiFi example for H2?

You add a file named `.skip.esp32h2` in the sketch folder.

> to migration guide?

And to the library documentation (if there is such)

JAndrassy:
> You add a file named .skip.esp32h2 in the sketch folder.

now I see it. I had .* files hidden in Eclipse, while I was looking at the other examples

> And to the library documentation (if there is such)

There isn't. Only WiFi STA and WiFi SoftAP are documented. WiFiClient, WiFiUDP and WiFiServer are not documented in ESP32 Arduino doc

me-no-dev:
there is still the issue of clients acquired by `accept()` are not added to the list. I suggest you rename `accept()` to `_accept()`, set it also as `protected` and use it instead of `accept()` internally. Then create a new `accept()` , that will call `_accept()` and also add to the list and can be mass-written to.

JAndrassy:
directly accepted clients should not be added to the list
https://www.arduino.cc/reference/en/libraries/ethernet/server.accept/

me-no-dev:
> so then again. why not just deprecate available() and remove inheritance from Print (disguised as Server)?

OK, you made your point :) Let's go with that instead

me-no-dev:
Please open a new one with the deprecation

PR #9029, opened 2023-12-20
WiFiClient - assignment shouldn't stop connection

JAndrassy:
assignment shouldn't stop connection in the left client. there may be other copy of WiFiClient working with that connection. 
let shared_ptr stop the connection when it is not referred anymore.
the default assignment operator works ok.


VojtechBartoska:
Same here, thanks for the PR, awaiting review.

lucasssvaz (approved):
LGTM

TD-er:
In the current implementation, the call to `stop()` does clear the last read and write timeout.
https://github.com/espressif/arduino-esp32/blob/cc6da052c21312588e5e3502cdc88dcd435a11fa/libraries/WiFi/src/WiFiClient.cpp#L200-L207
Those are not copied from the `other` WiFiClient. (in the current situation)
No idea if `Client` from which `WiFiClient` inherits has any members which need to be copied, but those are also not copied.

No idea yet whether that's important or not.

In your PR you let the compiler generate this function , which essentially does the same as the `stop()` call except for the reset of the timeout members.

I can't oversee yet whether this is a problem, but what I do see is that this change does not match your description of **_why_** you need this change as it still does reassign a clientSocketHandle and buffer.

The only thing I can imagine is that maybe the `shared_ptr` count does not reach 0 when you assign it to itself? (or indirectly assign it to itself)

So my question is what actual problem or use case do you try to fix here?

JAndrassy:
> So my question is what actual problem or use case do you try to fix here?

the actual problem is a warning (which causes PR tests to end with error). the warning says that a copy constructor must exist if assignment operator is defined. but the assignment operator is only defined to stop the connection. and it shouldn't do that. so instead of adding a copy constructor we delete the assignment operator.

TD-er:
Sounds like you're also testing using PVS Studio? :)

Or is there a compiler flag which can also tell me this, then I would really like to know (and use) it. :)

Anyway, about this PR, I don't know whether this actually may change behavior as it depends on the implementation of the shared_ptr and what these timeout values are used for.

JAndrassy:
I use Eclipse Sloeber.
The default copy constructor and assignment operator are well defined. They copy all fields the same way parameters of a function are copied. 
shared_ptr is designed exactly for this case. to count copies and delete the managed object when the last copy is destructed.

TD-er:
> shared_ptr is designed exactly for this case. to count copies and delete the managed object when the last copy is destructed.

I know, but like I said the only difference here is that the shared_ptr is assigned a `nullptr` inbetween (current implementation with call to `stop()`.
Thus if you are not assigning the same connection to it, then it makes no difference whether you call `stop()` inbetween or not. (apart from the already mentioned differences of not copying the timeout vars)
In either case any reference count to the existing connection is decreased by 1.

So the only difference here is there will be an assignment of a nullptr inbetween. (current implementation)
Now that you're removing this extra assignment, the only use case I can imagine which will be different is when you're assigning the same connection to this WiFiClient.
If the assignment operator of shared_ptr does check whether it already has the same pointer address, then there is a change in behavior.
If it doesn't, then it is exactly the same behavior (regarding the shared_ptr variables) as before.
In any other use case I don't see how you're preventing the current connection to be closed when assigning the `other` client. (N.B. the connection of `other` is not closed)

JAndrassy:
sorry. yes. I forgot that in esp8266 the connection closes from local side only if the last copy is destroyed. that is a way to prevent invalid copies. already two or three time in last years I wondered why a connection continuous to work when I invoked stop()

JAndrassy:
> so what does this PR improve in reality?

removes code which causes trouble and confusion. for example the WiFiServer PR https://github.com/espressif/arduino-esp32/pull/9028 fails checks because of this

me-no-dev:
PR is failing, because you are compiling for H2, which does not have WiFi at all. But yes, this also gives warning.

PR #9031, opened 2023-12-21
LEDC - Allow custom channel selection

### LINKED ISSUE #8967: New LEDC auto channel function needs ability to manual adjust channel.
### Board

ESP32

### Device Description

ESP32

### Hardware Configuration

transducer and backlight

### Version

latest development Release Candidate (RC-X)

### IDE Name

Arduino IDE

### Operating System

Windows 10

### Flash frequency

40Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

The new function got rid of the setup call that allowed you to manually set your channel.  The problem is my project has a audio transducer and a lcd with backlight.  When using ledc for tones and controlling the backlight brightness the two interfere.  This is because the esp groups two sets of eight channels and the frequency is shared amongst the set.  Since it now auto adds the channels its grouping both of my devices and when the frequency changes for tone control the backlight is then off and you can't fix this without a blinking in the backlight while they are grouped.  Previously I could set channel 1 for backlight and 9 for the speaker and everything was happy since the frequencies were seperated.

### Sketch

```cpp
ledcAttach(buzzer,2000,8);
ledcAttach(TFT_BACKLIGHT,2000,8);
```

### Debug Message

```plain
No debug messages
```

### Other Steps to Reproduce

_No response_

P-R-O-C-H-Y:
## Description of Change
This PR adds a new API `ledcAttachChannel` to LEDC peripheral, so the LEDC channel can be specified and won't be automatically selected, like is done by the `ledcAttach` function.

`bool ledcAttachChannel(uint8_t pin, uint32_t freq, uint8_t resolution, uint8_t channel);`

Also added info log message what channel is used when LEDC is attached to pin.

## Tests scenarios
Tested on ESP32-C6 with edited LEDC Software fade example.

## Related links
Closes #8967


lucasssvaz (approved):
LGTM

me-no-dev:
what happens on ESP32 where we have 2x8 channels?

P-R-O-C-H-Y:
> what happens on ESP32 where we have 2x8 channels?

You can select channel 0-15.

P-R-O-C-H-Y:
> Excuse me, I just want to ask if the above modification to the ledc library has been added to the 3.0.0-Alpha3 version? Can I download the newest code if I download 3.0.0-Alpha3 from the Arduino IDE?
> 
> Thank you.

Hello @Hs-Cheng, this change was not merged yet. If you want to test/use this modification you need to manually update the source code with changes from this PR or if you use manual installation instead of board manager in Arduino IDE, you can simply checkout this PR.

PR #9034, opened 2023-12-22
SD library: Fix format

prplz:
## Description of Change
SD card formatting in the SD library was broken by #6745. The work buffer for `f_mkfs` was moved from the stack to the heap, but the length argument was not updated, instead it currently passes the size of a pointer (32 bits).

## Tests scenarios
Before my change, `SD.begin` with `format_if_empty` set to true and an erased SD card will always fail with the error `[sd_diskio.cpp:824] sdcard_mount(): f_mkfs failed: (17) LFN working buffer could not be allocated`.
After my change `SD.begin` succeeds and creates a FAT filesystem on the card.


P-R-O-C-H-Y (approved):

P-R-O-C-H-Y:
Nice catch @prplz! Thanks for the fix 👍

lucasssvaz (approved):
LGTM

PR #9062, opened 2024-01-03
WiFi.disconnect() "aligned with Aduino.cc". waiting for status change.

JAndrassy:
WiFi.disconnect() "aligned with Aduino.cc". 
`WiFi.disconnect` renamed to `disconnectAsync`
new `WiFi.disconnect` waits for `status` change

the WiFi library with async disconnect behaves very strange. after disconnect() and immediate WiFi.begin it can't connect. I had to add a wait for status change after disconnect in my [test sketch](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiTest/WiFiTest.ino).
So I think the blocking version of disconnect() is a better option for everybody. It is quick.

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.cpp
+++ b/libraries/WiFi/src/WiFiSTA.cpp
@@ -367,7 +367,7 @@ bool WiFiSTAClass::reconnect()
  * @param eraseap `true` to erase the AP configuration from the NVS memory.
  * @return `true` when successful.
  */
-bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap)
+bool WiFiSTAClass::disconnectAsync(bool wifioff, bool eraseap)
 {
     wifi_config_t conf;
     wifi_sta_config(&conf);
@@ -389,6 +389,31 @@ bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap)
     }
 
     return false;
+}
+
+/**
+ * Disconnect from the network.
+ * @param wifioff `true` to turn the Wi-Fi radio off.
+ * @param eraseap `true` to erase the AP configuration from the NVS memory.
+ * @param timeoutLength timeout to wait for status change
+ * @return `true` when successful.
+ */
+bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap, unsigned long timeoutLength)
+{
+    if (!disconnectAsync(wifioff, eraseap)) {
+        return false;
+    }
+    if (!timeoutLength) {
+        return true;
+    }
+    const unsigned long start = millis();
+    while ((WiFiGenericClass::getStatusBits() & STA_CONNECTED_BIT) != 0) {
+        if((millis() - start) >= timeoutLength){
+            return false;
+        }
+        delay(2);
+    }
+    return true;
 }
 
 /**
--- a/libraries/WiFi/src/WiFiSTA.h
+++ b/libraries/WiFi/src/WiFiSTA.h
@@ -61,7 +61,8 @@ public:
     bool bandwidth(wifi_bandwidth_t bandwidth);
 
     bool reconnect();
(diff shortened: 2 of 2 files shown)


libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -391,6 +391,25 @@ bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap)
...
+ * @return `true` when successful.
+ */
+bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap, unsigned long timeoutLength)
+{
+  bool rv = disconnectAsync(wifioff, eraseap);
+  if (!rv)
+      return false;
+  unsigned long start = millis();
+  while (status() == WL_CONNECTED && (millis() - start) < timeoutLength) {
+      delay(100);
```
TD-er, line 408: Shouldn't this `100` be something like `timeoutLength < 100 ? timeoutLength : 100` ?

TD-er:
Just something to consider.
`wl_status_t WiFiSTAClass::begin(const char* ssid, const char *passphrase, int32_t channel, const uint8_t* bssid, bool connect)` is also calling `esp_wifi_disconnect()`.
So instead of letting the user take care of async vs "waiting for' disconnect, shouldn't the begin() call take care of this?

Also all `begin()` calls do check for quite a lot of things but the only return state in case of an error is `WL_CONNECT_FAILED`.
Not really informative.

libraries/WiFi/src/WiFiSTA.cpp
JAndrassy: it is copied from `waitForConnectResult`

me-no-dev:
```cpp
uint32_t delay_time = timeoutLength < 10 ? timeoutLength : 10;
while ((WiFiGenericClass::getStatusBits() & STA_CONNECTED_BIT) != 0 && (millis() - start) < timeoutLength) {
  delay(delay_time);
}
```

JAndrassy:
the state changes under 1 millisecond.
disconnectAsync takes.2 milliseconds

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -391,6 +391,31 @@ bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap)
...
+ */
+bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap, unsigned long timeoutLength)
+{
+    if (!disconnectAsync(wifioff, eraseap)) {
+        return false;
+    }
+    if (!timeoutLength) {
+        return true;
+    }
+    unsigned long start = millis();
```
TD-er, line 409: Please make this const.

TD-er: Well it is more like I trust the compiler to make further optimizations when I tell it some values are not supposed to change.
And when reading `const` in the code, I (as a person) can read code much faster as I don't have to keep track of it in my mind to see where it might be causing issues.
So it is just a habbit of me to use const whenever possible.

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -391,6 +391,31 @@ bool WiFiSTAClass::disconnect(bool wifioff, bool eraseap)
...
+    }
+    if (!timeoutLength) {
+        return true;
+    }
+    const unsigned long start = millis();
+    while ((WiFiGenericClass::getStatusBits() & STA_CONNECTED_BIT) != 0) {
+        if((millis() - start) >= timeoutLength){
+            return false;
+        }
+        delay(1);
```
me-no-dev, line 414: delay at least 2ms (more than one RTOS tick) to ensure that you allow other tasks to run

PR #9063, opened 2024-01-03
WiFi.begin "aligned with Arduino.cc". blocking (breaking change)

JAndrassy:
change `WiFi.begin` to blocking to be "aligned with Arduino.cc". 

PR adds `beginTimeout` property used in `begin` with `waitForConnectResult`

with `beginTimeout` 0 `begin` is async. 
default `beginTimeout` is a minute as in `waitForConnectResult`

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.h
+++ b/libraries/WiFi/src/WiFiSTA.h
@@ -108,12 +108,20 @@ public:
 
     static void _setStatus(wl_status_t status);
     
+    void setBeginTimeout(unsigned long timeout) {
+        beginTimeout = timeout;
+    }
+    void setBeginAsync() {
+        beginTimeout = 0;
+    }
+
 protected:
     static bool _useStaticIp;
     static bool _autoReconnect;
     static wifi_auth_mode_t _minSecurity;
     static wifi_scan_method_t _scanMethod;
     static wifi_sort_method_t _sortMethod;
+    unsigned long beginTimeout = 60000;
 
 public: 
     bool beginSmartConfig(smartconfig_type_t type = SC_TYPE_ESPTOUCH, char* crypt_key = NULL);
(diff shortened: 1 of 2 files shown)


me-no-dev:
I do not agree with this. Not to mention how many projects will be unnecessarily broken by such change. Why cater to the 0.01% of people who might be still using the old and outdated WiFi shields?

me-no-dev:
Only if it makes sense and will not otherwise cause issues. Changing the behavior in such drastic way on something that has been used for a decade at this point, especially something as critical as `WiFi.begin()`, should not be done. Maybe if you make the default timeout to be `0`, then we could add it. Will not break anything and will also allow for blocking calls, if one so desires.

libraries/WiFi/src/WiFiSTA.h
```diff
@@ -108,12 +108,20 @@ class WiFiSTAClass
...
+        beginTimeout = 0;
+    }
+
 protected:
     static bool _useStaticIp;
     static bool _autoReconnect;
     static wifi_auth_mode_t _minSecurity;
     static wifi_scan_method_t _scanMethod;
     static wifi_sort_method_t _sortMethod;
+    unsigned long beginTimeout = 60000;
```
TD-er, line 124: 1 minute timeout is absolutely unusable.
10% of it might be something to consider, but a default timeout of 60 seconds is hardly every usable in the world of microcontrollers.

As was already mentioned, a default of 0 would be at least non-breaking, but then you explicitly need to set the timeout and if you would change your code to do so, then you probably already have a similar loop right after the call to `WiFi.begin()`

Jason2866:
Imho not every bad incomplete Arduino code should be implemented. Arduino code is not the holey gral and everything else than perfect.
To get core 3.0.0 going let us focus on the needed! changes. Fine tuning is the step after that.

PR #9075, opened 2024-01-06
sd_diskio.cpp: return 0 instead of false when return type != bool

everslick:
Fixes bad coding style, possibly due to cut'n paste.


me-no-dev:
Thanks!

Issue #6563, opened 2022-04-11
Investigate supporting power management in arduino-esp32 (DFS + tickless idle)


igrr:
### Related area

Power management

### Hardware specification

All ESP32 series chips

### Is your feature request related to a problem?

Recently a community member has mentioned that the current consumption of ESP32 connected to Wi-Fi is too high. While it is possible to reduce current consumption with ESP-IDF, this option is not accessible to Arduino users.

### Describe the solution you'd like

To reduce current consumption in Wi-Fi connected state, several things are required:

- Power management should be enabled (CONFIG_PM_ENABLE).
  - [ ]  Investigate what are the downsides enabling it if it is not turned on at run time.
- FreeRTOS tickless idle should be enabled (CONFIG_FREERTOS_USE_TICKLESS_IDLE).
  - [ ]  Investigate what are the downsides enabling it if the DFS and automatic light sleep are not enabled at runtime.
- Enabling DFS and automatic light sleep at run time via some API.
- An easy-to-understand example showing how to achieve lower current consumption when Wi-Fi is connected.
- Documentation explaining the caveats (mostly related to peripherals and timers) of enabling the above mentioned features.

### Describe alternatives you've considered

Keep not supporting DFS and tickless idle in Arduino. But then everyone keeps thinking that it's not possible to have Wi-Fi connection with lower average current.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

our30K:
just want to pump this up , is there anyone working on this ? it has been a long while :)

tshcherban:
i can take a look at this. recently i've been playing around with IDF and power management API, can try applying this to the arduino core.
@igrr @our30K can i sometimes ping you here if any questions?

our30K:
that would be great, I am interested in both low power WIFI and BLE via arudino.

tshcherban:
I've added arduino core and some of the arduino libs (OneWire, DallasTemperature) as a components to ESP-IDF.
Currently enabled `CONFIG_PM_ENABLE`, `CONFIG_FREERTOS_USE_TICKLESS_IDLE` and used automatic light-sleep with `esp_wifi_set_ps(wifi_ps_type_t::WIFI_PS_MAX_MODEM)` and `esp_pm_configure`.
Used a simple HttpClient request and OneWire library to test if it's alive. `millis` works as expected since it relies on `esp_timer_get_time`.
[...]

PR #9083, opened 2024-01-09
PowerManagement support

tshcherban:
## Description of Change
Enables Power Management support for arduino sketches.

## Tests scenarios
Tested WiFi connectivity, BLE scanning, OneWire, TwoWire, SPI, Serial on ESP32, ESP32-S3, ESP32-C3 boards.

## Related links
Solves [#6563](https://github.com/espressif/arduino-esp32/issues/6563) partially

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -110,4 +110,5 @@ jobs:
           idf.py create-project test
           echo CONFIG_FREERTOS_HZ=1000 > test/sdkconfig.defaults
           echo CONFIG_PM_ENABLE=y >> test/sdkconfig.defaults
+          echo CONFIG_FREERTOS_USE_TICKLESS_IDLE=y >> test/sdkconfig.defaults
           idf.py -C test -DEXTRA_COMPONENT_DIRS=$PWD/components build
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -109,4 +109,5 @@ jobs:
           . ${IDF_PATH}/export.sh
           idf.py create-project test
           echo CONFIG_FREERTOS_HZ=1000 > test/sdkconfig.defaults
+          echo CONFIG_PM_ENABLE=y >> test/sdkconfig.defaults
           idf.py -C test -DEXTRA_COMPONENT_DIRS=$PWD/components build


tshcherban:
@SuGlider @VojtechBartoska is it possible to migrate from patching SDK-Config in github actions to just copy corresponding sdkconfig.defaults.<target>? that would be a bit more easy editing defaults, separate them for different targets and reuse for some other purposes (i.e. copy for examples, use in CMake build checks).

SuGlider:
> @SuGlider @VojtechBartoska is it possible to migrate from patching SDK-Config in github actions to just copy corresponding **sdkconfig.defaults._target_**? that would be a bit more easy editing defaults, separate them for different targets and reuse for some other purposes (i.e. copy for examples, use in CMake build checks).

Actually we need to add this IDF sdkconfig changes to the Lib Builder scripts.
sdkconfig is used by Lib Builder to compile and build all IDF libraries, therefore, when using Arduino, all settings from sdkconfig are setled in the static IDF libs.

The only way to changed sdkconfig and test it is by using Arduino as IDF Component and then build some examples that will verify how PM and light sleep will work.

tshcherban:
Found an issue with IDF's UART implementation [IDFGH-11870](https://github.com/espressif/esp-idf/issues/12957)

tshcherban:
@igrr @VojtechBartoska @SuGlider for a BLE part - i dont have ESP32 board with an external crystal, so for that part I would need your help in testing. On an ESP32S3 I'm planning to use Main XTAL as low power source by default.

SuGlider:
@tshcherban - 
On top of this PR, I'd like to work on an API that will allow the user to set the source clock too.

This is part of another Task which also wants to add Power Management feature based on automatic Dynamic Frequency Scaling (DFS). The proposal is that the user will be able to set the Clock Source for each peripheral and also set it in a Global way whenever necessary in order to increase DFS applicability.

This should be ready for Arduino 3.0.0 final release.
Is that OK for you?

tshcherban:
yes, for sure
do i need to grant you an access to my repository? so all the changes come from the one codebase

libraries/WiFi/src/WiFiGeneric.h
```diff
@@ -103,6 +103,16 @@ class WiFiGenericClass {
   bool setSleep(wifi_ps_type_t sleepType);
   wifi_ps_type_t getSleep();
 
+  /**
+  * @brief Set modem sleep state
+  * @param sleepType Modem sleep type, one of WIFI_PS_NONE, WIFI_PS_MIN_MODEM, WIFI_PS_MAX_MODEM
+  * @param listenInterval Listen interval for ESP32 station to receive beacon when WIFI_PS_MAX_MODEM is set. Units: AP beacon intervals. Defaults to 3 if set to 0.
```
atanisoft, line 109: it would be good to indicate the units of this parameter. As written it could be interpreted as milliseconds, seconds, minutes, etc.

safocl:
https://github.com/espressif/arduino-esp32/pull/9083/files#diff-148dc0a05cfb442a25f8653a18647250c2fe811144f2a424149f127c7eb0f488R298
I suggest making the variable a constant:
```
const uint32_t cpuFreq = getCpuFrequencyMhz();
```
instead of
```
uint32_t cpuFreq = getCpuFrequencyMhz();
```

safocl:
API and its comments are not compatible:
```
bool setAutomaticLightSleep(bool enabled);
```
and
```
/**
 * @brief Set automatic light sleep state. CPU will go into light sleep if no ongoing activity (active task, peripheral activity etc.)
 * @param enabled true to enable automatic lightSleep
 * @return
 *      - ESP_OK on success
 *      - ESP_ERR_NOT_SUPPORTED if CONFIG_PM_ENABLE is not enabled in sdkconfig
 */
```
return type is bool (not [esp_error_type])

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -621,6 +621,27 @@ bool WiFiGenericClass::setSleep(wifi_ps_type_t sleepType) {
   return false;
 }
 
+bool WiFiGenericClass::setSleep(wifi_ps_type_t sleepType, uint16_t listenInterval)
+{
+    if (!setSleep(sleepType))
+        return false;
+    
+    wifi_config_t current_conf;
+    if(esp_wifi_get_config((wifi_interface_t)ESP_IF_WIFI_STA, &current_conf) != ESP_OK){
```
safocl, line 630: from 
```(wifi_interface_t)ESP_IF_WIFI_STA```
to
```wifi_interface_t::WIFI_IF_STA```
?

File: cores/esp32/esp32-hal-cpu.c [hunk 3]
```diff
@@ -291,3 +292,16 @@ uint32_t getApbFrequency() {
   rtc_clk_cpu_freq_get_config(&conf);
   return calculateApb(&conf);
 }
+
+bool setAutomaticLightSleep(bool enabled)
+{
+    uint32_t cpuFreq = getCpuFrequencyMhz();
```
safocl, line 298: from
```
uint32_t cpuFreq = getCpuFrequencyMhz();
```
to
```
const uint32_t cpuFreq = getCpuFrequencyMhz();
```
?

libraries/WiFi/src/WiFiGeneric.cpp
```diff
@@ -621,6 +621,27 @@ bool WiFiGenericClass::setSleep(wifi_ps_type_t sleepType) {
...
+        log_e("setSleep(wifi_ps_type_t sleepType, uint16_t listenInterval) failed: get current config failed!");
+        return false;
+    }
+
+    current_conf.sta.listen_interval = listenInterval;
+
+    if(esp_wifi_set_config((wifi_interface_t)ESP_IF_WIFI_STA, &current_conf) != ESP_OK){
```
safocl, line 637: from 
```(wifi_interface_t)ESP_IF_WIFI_STA```
to
```wifi_interface_t::WIFI_IF_STA```
?

libraries/WiFi/src/WiFiGeneric.h
tshcherban: Unit is specified here (AP beacon intervals). It is determined by the access point parameters, I've often seen a 102400us value during testing with different routers, however it can differ.

francisgariepy:
Anyone to resolve conflict and merge anytime soon? It's been more than 6 months and automatic light sleep is not working.

Issue #9020, opened 2023-12-18
Serial2 breaks Serial1


terrafirma2021:
### Board

esp32-s3

### Device Description

wokwi S3
Feather-S3

### Hardware Configuration

No

### Version

latest master (checkout manually)

### IDE Name

2.2.1

### Operating System

win 10 x64

### Flash frequency

240

### PSRAM enabled

yes

### Upload speed

115200

### Description

// works fine
void sendHOverSerial1() {
  Serial1.begin(115200, SERIAL_8N1, 38, 3); // Re-initialize Serial1 with different pins
  Serial1.write("H"); 
  delay(1000); 
 Serial1.begin(115200, SERIAL_8N1, 38, 1); // Initialize Serial1 back to stock pins
}
// Causes panic 
void sendIOverSerial2() {
  Serial2.begin(115200, SERIAL_8N1, 47, 21);  // Initialize Serial2 with pin swap 
  Serial2.write("I"); 
  delay(1000); 
  // Serial2.begin(115200, SERIAL_8N1, 11, 10); // <------- causes panic
}

Also when 
void sendIOverSerial2() {

is enabled 

void sendHOverSerial1() { // breaks

### Sketch

```cpp
//https://wokwi.com/projects/384481453752614913 attached for easy viewing

void setup() {
  Serial.begin(115200);                
  Serial2.begin(115200, SERIAL_8N1, 11, 10);
  Serial1.begin(115200, SERIAL_8N1, 38, 1);
}

void loop() {
  // Check if Serial1 has received any data
  if (Serial1.available()) {
    char byteReceived1 = Serial1.read();
    Serial.print("Byte received on Serial1: ");
    Serial.println(byteReceived1);
  }

  // Check if Serial2 has received any data
  if (Serial2.available()) {
    char byteReceived2 = Serial2.read();
    Serial.print("Byte received on Serial2: ");
    Serial.println(byteReceived2);
  }

  // Call the functions to send "H" and "I"
  sendHOverSerial1();
  sendIOverSerial2(); // Stops sendHOverSerial1() working
}

void sendHOverSerial1() {
  Serial1.begin(115200, SERIAL_8N1, 38, 3); // Re-initialize Serial1 with different pins
  Serial1.write("H"); 
  delay(1000); 
 Serial1.begin(115200, SERIAL_8N1, 38, 1); // Initialize Serial1 back to stock pins
}

void sendIOverSerial2() {
[...]

PR #9095, opened 2024-01-12
UART: fixes begin() after a previous begin() :: deleting previous RX/TX buffers and its data 

SuGlider:
## Description of Change
`HardwareSerial::begin()` has a bug that by always restaring the IDF UART driver, the currently running UART driver will release all the RX and TX ringbuffers causing it to drop all data waiting to be read.

This is very easy to verify by cross connecting RX/TX running `begin()`, `write(byte)` and then `begin()` + `read()`. No data is read.

The second issue is relatet to changing the RX Pin using `begin()` or `setPins()`.
This causes a BREAK in the line and stops the driver, unless it is caught.

Both issues are corrected in this PR.

`HardwareSerial::end(bool)` has been changed to the standard `HardwareSerial::end()`

#### Important note:

No data is lost when changing the Arduino standard parameters, such as baud rate, uart frame format and pins.
But if the user changes IDF driver internal settings, like Line Inverting, RX/TX buffer size or RX FIFO IRQ Threshold, the driver must be restarted and it will cause the RX/TX buffer to be released with its data.

## Tests scenarios
Tested with ESP32-S3 using UART0, UART1 and UART3

``` cpp
//https://wokwi.com/projects/384481453752614913 attached for easy viewing
// pins 3 and 38 are connected (RX1-TX1) || pins 47 and 21 are connected (RX2-TX2)

void setup() {
  Serial.begin(115200);                
  Serial2.begin(115200, SERIAL_8N1, 11, 10);
  Serial1.begin(115200, SERIAL_8N1, 38, 1);
}

void loop() {
  // Check if Serial1 has received any data
  if (Serial1.available()) {
    char byteReceived1 = Serial1.read();
    Serial.print("Byte received on Serial1: ");
    Serial.println(byteReceived1);
  }

  // Check if Serial2 has received any data
  if (Serial2.available()) {
    char byteReceived2 = Serial2.read();
    Serial.print("Byte received on Serial2: ");
    Serial.println(byteReceived2);
  }

  // Call the functions to send "H" and "I"
  sendHOverSerial1();
  sendIOverSerial2(); // Stops sendHOverSerial1() working
}

void sendHOverSerial1() {
  Serial1.begin(115200, SERIAL_8N1, 38, 3); // Re-initialize Serial1 with different pins
  Serial1.write("H"); 
  delay(1000); 
 Serial1.begin(115200, SERIAL_8N1, 38, 1); // Initialize Serial1 back to stock pins <---- causes data loss
}

void sendIOverSerial2() {
  Serial2.begin(115200, SERIAL_8N1, 47, 21);  // Initialize Serial2 with pin swap 
  Serial2.write("I"); 
  delay(1000); 
  // Serial2.begin(115200, SERIAL_8N1, 11, 10); // <------- causes BREAK
}
```

## Related links
#9020


cores/esp32/HardwareSerial.cpp
```diff
@@ -389,22 +398,17 @@ void HardwareSerial::updateBaudRate(unsigned long baud)
   uartSetBaudRate(_uart, baud);
 }
 
-void HardwareSerial::end(bool fullyTerminate)
+void HardwareSerial::end()
```
lucasssvaz, line 401: This will break the UART CI test as it uses the fullyTerminate parameter

lucasssvaz (changes requested):

cores/esp32/HardwareSerial.cpp
SuGlider: The point is that `HardwareSerial::end(bool fullyTerminate)` is not necessary anymore, doesn't make sense with the changes and it is not Arduino HardwareSerial standard API.

Would it be possible to change the CI Test Code?
Or is it dependent from the API?

SuGlider: @lucasssvaz - I changed the CI sketch. It is not breaking the UART CI anymore. Please review these changes.
I tested it here and all of them passed.

SuGlider:
@me-no-dev | @lucasssvaz - It seems that HIL isn't working... CI is faling due to some Python failure.

PR #9097, opened 2024-01-13
[release/2.x] Set Default Parameter for `edge` in timerAttachInterupt(...) Function

hasenradball:
# Set Default Parameter for edge in timerAttachInterupt
The actual Documentation for the funtion ` timerAttachInterupt(...)` looks like:

image

But the actual code look like:
image

To be consistent with the documentation, and `edge`is not supportet:
see:
```
void timerAttachInterruptFlag(hw_timer_t *timer, void (*fn)(void), bool edge, int intr_alloc_flags){
    if(edge){
        log_w("EDGE timer interrupt is not supported! Setting to LEVEL...");
    }
    timer_isr_callback_add(timer->group, timer->num, timerFnWrapper, fn, intr_alloc_flags);
}
```
A default parameter for `edge = false ` would be a good change to be consistent with the documentation.

### DIFF
--- a/cores/esp32/esp32-hal-timer.h
+++ b/cores/esp32/esp32-hal-timer.h
@@ -37,7 +37,7 @@ void timerSetConfig(hw_timer_t *timer, uint32_t config);
 uint32_t timerGetConfig(hw_timer_t *timer);
 
 void timerAttachInterruptFlag(hw_timer_t *timer, void (*fn)(void), bool edge, int intr_alloc_flags);
-void timerAttachInterrupt(hw_timer_t *timer, void (*fn)(void), bool edge);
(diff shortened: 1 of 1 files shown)


me-no-dev (approved):

P-R-O-C-H-Y (approved):

me-no-dev:
https://github.com/espressif/arduino-esp32/actions/runs/7511742476/job/20531529273?pr=9097#step:5:104

Default values can be given only for C++ functions/methods. In C that is not allowed.

me-no-dev:
so... the docs you are looking at are for the 3.0.0 version. In 2.x the parameter exists and should exist to not break compatibility.

Closing

PR #9117, opened 2024-01-16
SPI - Allow detaching of some SPI pins without stopping SPI

### LINKED ISSUE #9036: SPI.end() prints removeApbChangeCallback error
### Board

ESP32 Dev Module

### Device Description

ESP32-S3-DevKitC-1 v1.1

### Hardware Configuration

No specific configuration

### Version

### IDE Name

Arduino IDE

### Operating System

Windows 11

### Flash frequency

240

### PSRAM enabled

yes

### Upload speed

921600

### Description

Starting from Arduino ESP32 3.0.0-alpha3 SPI.end() calls spiStopBus twice
- From spiDetachSCK (Probably via spiDetachBus that is set as perimanSetBusDeinit)
- And spiStopBus itself

The second call tries to invoke removeApbChangeCallback which is already removed by then.

### Sketch

```cpp
#include <SPI.h>

void setup() {
  SPI.begin();
  SPI.end();
}

void loop() {
  yield();
}
```

### Debug Message

```plain
ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0x1 (POWERON),boot:0x18 (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
Octal Flash Mode Enabled
For OPI Flash, Use Default Flash Boot Mode
mode:SLOW_RD, clock div:1
load:0x3fce3818,len:0x438
load:0x403c9700,len:0x4
load:0x403c9704,len:0x9ac
load:0x403cc700,len:0x2758
entry 0x403c9868
[   134][E][esp32-hal-cpu.c:140] removeApbChangeCallback(): not found func=0x42002a24 arg=0x3fc93c00
```

### Other Steps to Reproduce

_No response_

### LINKED ISSUE #9066: Change SPI to allow detaching some pins, like MISO or CS
> Wonderful. Purely for my own understanding, why would configuring the pin with pin mode cause the SPI to break? Does it disable the SPI interface?

In Arduino 3.0.0 there is a new feature called Peripheral Manager.
In the specific case of SPI, when any of its 4 pins (MOSI, MISO, CLK and CS) are used for any other function, for instance, with `pinMode()`, the whole SPI is detached. Therefore, SPI stops working, being necessary to restart it with `SPI.begin()`.

I think that your example brings a good questioning because most displays will only care about using MOSI and not MISO.
I could also use CS pin  for something else and only set it as CS whenever it is necessary.

Therefore, I would say that we may review it in order to allow using SPI pins partly.

_Originally posted by @SuGlider in https://github.com/espressif/arduino-esp32/issues/9040#issuecomment-1876012166_

P-R-O-C-H-Y:
## Description of Change
This PR adds a feature to allow detach of SS, MISO or MOSI without stopping the SPI.
This PR fixes issue, when spiStopBus() was called twice (by SPI.end and spiDetachBus), printing an error. 

Also removed pin parameter from detaching functions, as the pins are now stored in the SPI structure. 
Without this change, peripheral manager was able to set to INIT pin, which was already detached from SPI.

## Tests scenarios
Tested on esp32-c3.

## Related links
Closes #9036 
Closes #9066


cores/esp32/esp32-hal-spi.c
```diff
@@ -578,10 +627,10 @@ spi_t * spiStartBus(uint8_t spi_num, uint32_t clockDiv, uint8_t dataMode, uint8_
...
     }
 
-    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_SCK, spiDetachBus);
-    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_MISO, spiDetachBus);
-    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_MOSI, spiDetachBus);
-    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_CS, spiDetachBus);
+    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_SCK, spiDetachBus_SCK);
+    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_MISO, spiDetachBus_MISO);
+    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_MOSI, spiDetachBus_MOSI);
+    perimanSetBusDeinit(ESP32_BUS_TYPE_SPI_MASTER_CS, spiDetachBus_SS);
```
lucasssvaz, line 633: For the sake of consistency I would change `spiDetachBus_SS` to `spiDetachBus_CS`

cores/esp32/esp32-hal-spi.c
```diff
@@ -177,36 +177,73 @@ static spi_t _spi_bus_array[] = {
...
+        }
+        if(spi->miso != -1){
+            spiDetachMISO(spi);
+        }
+        if(spi->mosi != -1){
+            spiDetachMOSI(spi);
+        }
+        if(spi->ss != -1){
+            spiDetachSS(spi);
+        }
```
lucasssvaz, line 200: The checks for -1 can be removed as they are already done inside the function.
```suggestion
        spiDetachSCK(spi);
        spiDetachMISO(spi);
        spiDetachMOSI(spi);
        spiDetachSS(spi);
```

cores/esp32/esp32-hal-spi.c
lucasssvaz: Just saw that the pin in the internal structure is SS and the enum is CS 🤷🏼‍♂️

libraries/SPI/src/SPI.cpp
```diff
@@ -124,15 +124,17 @@ void SPIClass::end()
...
-    spiDetachSCK(_spi, _sck);
+    spiDetachSCK(_spi);
     if(_miso >= 0){
-        spiDetachMISO(_spi, _miso);
+        spiDetachMISO(_spi);
     }
     if(_mosi >= 0){
-        spiDetachMOSI(_spi, _mosi);
+        spiDetachMOSI(_spi);
     }
```
lucasssvaz, line 133: Same here, the pin check is not needed as it is also done inside the function

cores/esp32/esp32-hal-spi.c
P-R-O-C-H-Y: I was not sure, in some places its SS and in others its CS 🤷🏼‍♂️. Lets discuss it today.

P-R-O-C-H-Y: Done

lucasssvaz (approved):

me-no-dev:
tested on all chips with DC = MISO

PR #9139, opened 2024-01-19
Improves WiFiMulti

SuGlider:
## Description of Change
This PR is based on PR #2775 from @Daemach, with some changes.

WiFiMulti has now new methods that improves its ways to connect to a WiFi AP.
It adds a new example to the ESP32 Core called `WiFiMultiAdvanced.ino`.

An important improvement is that the WiFiMulti `run()` will mark the failed WiFi APs in order to do not include them in the next `run()` execution. This avoids that it fails all the time in the very same AP with the highest RSSI. After failing with all possible APs of the list, it will reset the "failed marks" and try again from fresh.

#### New APIs:
- `void setStrictMode(bool bStrict)`
Default is `true`: It forces to only keep connected or to connect to an AP from the provided WiFiMulti list.
When bStrict is `false`, it will keep the last/current connected AP even if not in the WiFiMulti List.
    
- `void setAllowOpenAP(bool bAllowOpenAP)`
Default is `false`: doesn't add open AP networks to the WiFiMulti list.
When bAllowOpenAP is `true`, it will add ANY scanned open open AP to the WiFiMulti list, even if not in the initial user list

-  `void APlistClean(void)`
Clears the current list of Multi APs and frees the memory

- `void setConnectionTestCallbackFunc(ConnectionTestCB_t cbFunc)`
Default is NULL - no validation callback function
It allows the user to define a callback function [`bool cb_func(void)`] that will validate the connection to the Internet.
If the callback returns true, the connection is considered valid and the AP will added to the validated AP list.
Set the callback to NULL to disable the feature and validate any SSID that is in the list.

## Tests scenarios
Using 3 APs, 1 Open AP.
Used the new `WiFiMultiAdvanced.ino` example to test many combinations in Verbose Mode.

## Related links
Closes #2775


P-R-O-C-H-Y (approved):
LGTM!

lucasssvaz (approved):

libraries/WiFi/examples/WiFiMultiAdvanced/WiFiMultiAdvanced.ino
```diff
@@ -0,0 +1,113 @@
...
+// This is used to test the Internet connection of connected the AP
+// Use a non-302 status code to ensure we bypass captive portals.  Can be any text in the webpage. 
+String _testResp = "301 Moved"; // usually http:// is moves to https:// by a 301 code
+// You can also set this to a simple test page on your own server to ensure you can reach it, 
+// like "http://www.mysite.com/test.html"
+String _testURL = "http://www.espressif.com"; // Must include "http://" if testing a HTTP host
+const int _testPort = 80;  // HTTP port
+
+bool testConnection(){
+    //parse url
```
me-no-dev, line 21: why not use HTTP client directly?

me-no-dev:
Except the example, LGTM too

libraries/WiFi/examples/WiFiMultiAdvanced/WiFiMultiAdvanced.ino
SuGlider: Done.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -47,17 +46,22 @@ WiFiMulti::~WiFiMulti()
...
 
-    if(!ssid || *ssid == 0x00 || strlen(ssid) > 31) {
+    if(!ssid || *ssid == '\0' || strlen(ssid) > 31) {
         // fail SSID too long or missing!
         log_e("[WIFI][APlistAdd] no ssid or ssid too long");
         return false;
     }
 
-    if(passphrase && strlen(passphrase) > 64) {
+    if(passphrase && strlen(passphrase) > 63) {
```
thijstriemstra, line 64: 👁️‍🗨️

PR #9151, opened 2024-01-22
Update LittleFS Platformio example

Jason2866:
Actual Platformio Platform espressif32 do not need anymore the Python script to build the LittleFS image. 
Use custom partition scheme to guarantee to have a SPIFFS partition which works with the example code.


3110:
I think `board_build.filesystem = littlefs` is necessary in `platformio.ini`.

PR #9190, opened 2024-01-30
Add more keys in USBHIDKeyboard

andreock:
## Description of Change
I added some keys code in USBHIDKeyboard header file so it can be used in Keyboard.press()

## Tests scenarios
I have tested the following change with framework arduino-esp32 2.0.11 using an ESP32-S3. The PC that I used for test, has installed nixOS 23.11 with kernel 6.1.75. All the keys added in this PR works in my test.

### DIFF
--- a/libraries/USB/src/USBHIDKeyboard.h
+++ b/libraries/USB/src/USBHIDKeyboard.h
@@ -63,6 +63,7 @@ typedef union {
 #define KEY_DOWN_ARROW  0xD9
 #define KEY_LEFT_ARROW  0xD8
 #define KEY_RIGHT_ARROW 0xD7
+#define KEY_MENU        0xFE
 #define KEY_SPACE       0x20
 #define KEY_BACKSPACE   0xB2
 #define KEY_TAB         0xB3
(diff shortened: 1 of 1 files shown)


libraries/USB/src/USBHIDKeyboard.h
```diff
@@ -63,6 +63,7 @@ typedef union {
 #define KEY_DOWN_ARROW  0xD9
 #define KEY_LEFT_ARROW  0xD8
 #define KEY_RIGHT_ARROW 0xD7
+#define KEY_MENU        0xFE
```
P-R-O-C-H-Y, line 66: @andreock I am not sure about this one. Can you please post any docs where you get this from? Thanks

andreock: I got it from here: https://github.com/msommacal/duckuino/blob/master/parser.y#L47

me-no-dev: have you tested them?

andreock: yes

me-no-dev: @P-R-O-C-H-Y  changes are OK with me

P-R-O-C-H-Y (approved):

PR #9216, opened 2024-02-05
SPI (fix): Adds SPI 3 to the ESP32-S2 and adds comments about it

### LINKED ISSUE #9210: FSPI and ESP32-S2
### Board

ESP32-S2FN4R2

### Device Description

none

### Hardware Configuration

none

### Version

### IDE Name

none

### Operating System

none

### Flash frequency

none

### PSRAM enabled

yes

### Upload speed

115200

### Description

looks like [this](https://github.com/espressif/arduino-esp32/blob/d14873ab8a95b86c8f38ceaed01642dd2909dcb3/cores/esp32/esp32-hal-spi.h#L35) is not vaild for ESP32S2.

when i look at the ESP32S2 documentation:
Ports of embedded flash correspond to pins of ESP32-S2FH2 and ESP32-S2FH4 as follows:
–CS# = SPICS0
–DI = SPID
–DO = SPIQ
–CLK = SPICLK
–WP# = SPIWP
–HOLD# = SPIHD
Ports of embedded PSRAM correspond to pins of ESP32-S2FN4R2 and ESP32-S2R2 as follows:
–CE# = SPICS1
–SI/SIO0 = SPID
–SO/SIO1 = SPIQ
–SCLK = SPICLK
–SIO2 = SPIWP
–SIO3 = SPIHD
These pins are not recommended for other uses

could you please fix this ?
thanks,
vincent

### Sketch

```cpp
none
```

### Debug Message

```plain
none
```

### Other Steps to Reproduce

_No response_

SuGlider:
## Description of Change
ESP32-S2 has 3 SPI buses available.
It fixes names and definitons in order to add the 3rd SPI to the list.
It also adds some commentaries in order to help users.

## Tests scenarios
CI Only

## Related links
Closes #9210

### DIFF
--- a/cores/esp32/esp32-hal-spi.h
+++ b/cores/esp32/esp32-hal-spi.h
@@ -31,12 +31,15 @@ extern "C" {
 #if CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32H2 || CONFIG_IDF_TARGET_ESP32S3
 #define FSPI  0
 #define HSPI  1
-#else
-#define FSPI  1 //SPI bus attached to the flash (can use the same data lines but different SS)
-#define HSPI  2 //SPI bus normally mapped to pins 12 - 15, but can be matrixed to any pins
-#if CONFIG_IDF_TARGET_ESP32
-#define VSPI  3 //SPI bus normally attached to pins 5, 18, 19 and 23, but can be matrixed to any pins
-#endif
+#elif CONFIG_IDF_TARGET_ESP32S2
+#define FSPI  1 //SPI 1 bus. ESP32S2: for external memory only (can use the same data lines but different SS)
+#define HSPI  2 //SPI 2 bus. ESP32S2: external memory or device  - itcan be matrixed to any pins
+#define SPI2  2 // Another name for ESP32S2 SPI 2
+#define SPI3  3 //SPI 3 bus. ESP32S2: device only - itcan be matrixed to any pins
+#elif CONFIG_IDF_TARGET_ESP32
+#define FSPI  1 //SPI 1 bus attached to the flash (can use the same data lines but different SS)
+#define HSPI  2 //SPI 2 bus normally mapped to pins 12 - 15, but can be matrixed to any pins
+#define VSPI  3 //SPI 3 bus normally attached to pins 5, 18, 19 and 23, but can be matrixed to any pins
 #endif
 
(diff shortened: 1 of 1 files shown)


TD-er:
While you're at it, can you perhaps also add (or link to) some info about how the naming schema for all FSPI/VSPI/HSPI is intended?
I didn't yet find any logic to it.

If FSPI is always connected to the flash, then it would make sense to me.
And if VSPI vs HSPI is about max speed (or bus width), then it also makes sense.

However it seems to be a bit random how it is used among ESP32-variants.

me-no-dev:
I'm kinda leaning towards switching to a different way to identify the buses. Maybe by a custom ID (because otherwise they will have to start from 2), so `0` is the first available port. No names.

TD-er:
> I'm kinda leaning towards switching to a different way to identify the buses. Maybe by a custom ID (because otherwise they will have to start from 2), so `0` is the first available port. No names.

Maybe it can even be extended to the peripheral manager to ask for a bus used for peripheral type X
This also eliminates the need for code to start the SPI bus or checking if it has been started.
And when you want to "join" a SPI bus, you can hand over some identifier which you later can use to get the SPI bus.

This removes the need to know about all platforms how they all use the SPI bus.
And you can still get some access to busses you shouldn't use to get some info from them like pins used, frequency, etc.

me-no-dev:
> Maybe it can even be extended to the peripheral manager to ask for a bus used for peripheral type X

This can't really work in Arduino. Such buses use multiple pins and so on. As sketch developer, it's your job to know which bus is used for which device. Sometimes this also needs to be defined on board level, when you have screen on a separate bus. Question is rather semantics. in ESP the SPI buses that are available to the users do not start from 0, which is confusing. It worked somewhat using names, but with more chips added, this also proves not perfect. Having all start from 0 would make it more intuitive and easier to handle. We could add name definitions for backward compatibility.

me-no-dev:
> I know asking for used pins isn't really in the scope of Arduino.

We do provide the used pins from peripheral manager
```
GPIO Info:
------------------------------------------
  GPIO : BUS_TYPE[bus/unit][chan]
  --------------------------------------  
     1 : UART_TX[0]
     3 : UART_RX[0]
     4 : ETHERNET_SPI
     5 : ETHERNET_SPI
    12 : SPI_MASTER_MISO[3]
    13 : SPI_MASTER_MOSI[3]
    14 : SPI_MASTER_SCK[3]
    15 : ETH_CS
```

lucasssvaz (approved):
LGTM

TD-er:
@SuGlider Thanks, I will have another look at this list and try to find a proper way to represent the available SPI options in the ESPEasy UI.

PR #9224, opened 2024-02-07
Fix AdvancedWebServer.ino uptime conversion

cydanil:
## Description of Change
This PR resolves an insignificant bug in the `AdvancedWebServer.ino` example, where the uptime is converted from seconds to hours, minutes, seconds format.  
The changes do not affect the overall sketch nor its intended usage.  

Prior to the changes, the reported uptime would be incorrectly going over 60 seconds a minute, and minutes would increase incorrectly.  
For examples:  
```text
0:0:59
0:1:0

0:1:60
0:1:61

0:1:118
0:1:119
0:2:120
0:2:121
```

After these changes, minutes and seconds update as expected.

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core `v2.0.14` with ESP32 on an `ESP32-Ethernet-Kit_A_V1.2` development board.

Thank you for your time,
Cyril :slightly_smiling_face:

### DIFF
--- a/libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
+++ b/libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
@@ -44,8 +44,9 @@ void handleRoot() {
   digitalWrite(led, 1);
   char temp[400];
   int sec = millis() / 1000;
-  int min = sec / 60;
-  int hr = min / 60;
+  int hr = sec / 3600;
+  int min = (sec / 60) % 60;
+  sec = sec % 60
 
   snprintf(temp, 400,
 
--- a/libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
+++ b/libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
@@ -46,7 +46,7 @@ void handleRoot() {
   int sec = millis() / 1000;
   int hr = sec / 3600;
   int min = (sec / 60) % 60;
-  sec = sec % 60
+  sec = sec % 60;
 
   snprintf(temp, 400,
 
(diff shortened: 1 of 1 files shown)


libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
```diff
@@ -44,8 +44,10 @@ void handleRoot() {
   digitalWrite(led, 1);
   char temp[400];
   int sec = millis() / 1000;
-  int min = sec / 60;
+
   int hr = min / 60;
```
me-no-dev, line 48: here `hr` is set to be `min` divided by 60, but `min` is defined on the next line. This should do the trick:
```cpp
int sec = millis() / 1000;
int hr = sec / 3600;
int min = (sec / 60) % 60;
sec = sec % 60;
```

libraries/WebServer/examples/AdvancedWebServer/AdvancedWebServer.ino
```diff
@@ -44,8 +44,9 @@ void handleRoot() {
   digitalWrite(led, 1);
   char temp[400];
   int sec = millis() / 1000;
-  int min = sec / 60;
-  int hr = min / 60;
+  int hr = sec / 3600;
+  int min = (sec / 60) % 60;
+  sec = sec % 60
```
me-no-dev, line 49: missing `;` at the end of the line

me-no-dev: BTW looking at the code below, both `min` and `sec` are `% 60`, which makes the new code in effect the same as the old one. Also the old one looks OK, so I wonder how come you got the wrong values?

PR #9228, opened 2024-02-09
Fix C2 compilation for Updater.cpp

lucasssvaz:
Fixes issue reported in https://github.com/espressif/arduino-esp32/pull/5807#issuecomment-1935770156

This will work exactly the same for all chips except C2, where it will use the mbedtls functions. In the future we will need to replace mbedtls with something simpler.

### DIFF
--- a/libraries/Update/src/Update.h
+++ b/libraries/Update/src/Update.h
@@ -5,7 +5,6 @@
 #include <MD5Builder.h>
 #include <functional>
 #include "esp_partition.h"
-#include "aes/esp_aes.h"
 
 #define UPDATE_ERROR_OK                 (0)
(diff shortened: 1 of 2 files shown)


lucasssvaz:
@Jason2866 Could you check please if it fixes the issue ? I don't currently have the C2 environment configured on my computer.

Jason2866:
@lucasssvaz The PR solves the issue. The "price" (used additionally resources) for the encrypt feature which is probably used very seldom is imho to high.

lucasssvaz:
There only will be additional uses of resources for C2 as the functions from mbedtls are replaced with esp ones when hardware acceleration is available. We discussed it in the meeting and it is an acceptable solution for now. In the future we will remove the mbedtls dependency.

Issue #9246, opened 2024-02-13
[W5500] M5Stack AtomPoE doesn't have INT/RST connected


TD-er:
### Board

M5Stack AtomS3 + ATOM_PoE

### Device Description

See [schematic of AtomPoE](https://docs.m5stack.com/en/atom/atom_poe)
[AtomS3](https://docs.m5stack.com/en/core/AtomS3)
N.B. the pins on the AtomPoE board refer to an ESP32 Atom module.
The AtomS3 uses an ESP32-S3 and thus has different GPIO pins.

### Hardware Configuration

- SPI CLK: GPIO-5
- W5500 CS: GPIO-6
- SPI MISO: GPIO-7
- SPI MOSI: GPIO-8

INT & RST  set to -1

5-pin row on the Atom PoE board:
- 3V3
- CLK
- CS
- MISO
- MOSI

![image](https://github.com/espressif/arduino-esp32/assets/3751318/46d477d3-bdba-4c49-b1ec-815d6f0be8ab)
![image](https://github.com/espressif/arduino-esp32/assets/3751318/ff9d74b3-9309-4ad8-9fda-e6c3234467db)

N.B. Other boards with W5500 and no INT connected:
- [PoECAM](https://docs.m5stack.com/en/unit/poe_cam)
- [M5PoECAM-W](https://docs.m5stack.com/en/unit/PoECAM-W)

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

There are boards with a W5500 SPI Ethernet controller which do not have the INT and RST pin connected.
However in the `ETH::beginSPI` function, this will be rejected as it checks for those pins.

To overcome this check, this function can be changed to something like this:
```c++
bool ETHClass::beginSPI(eth_phy_type_t type, uint8_t phy_addr, int cs, int irq, int rst, 
#if ETH_SPI_SUPPORTS_CUSTOM
    SPIClass *spi, 
#endif
    int sck, int miso, int mosi, spi_host_device_t spi_host, uint8_t spi_freq_mhz){
    esp_err_t ret = ESP_OK;

    if(_eth_started || _esp_netif != NULL || _eth_handle != NULL){
        log_w("ETH Already Started");
        return true;
    }
    if(cs < 0 || 
	   (
#if CONFIG_ETH_SPI_ETHERNET_W5500
[...]

PR #9251, opened 2024-02-15
[Enhancement] Support no-IRQ for SPI Ethernet W5500

TD-er:
Requires this commit from ESP-IDF to be cherry picked: https://github.com/espressif/esp-idf/commit/fd0a1dc53c4cff99f04dfe658047f7c2e3e6d8ad

Related discussion: https://github.com/espressif/esp-idf/issues/12682#issuecomment-1914807052

Fixes: https://github.com/espressif/arduino-esp32/issues/9246

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -351,8 +351,13 @@ bool ETHClass::beginSPI(eth_phy_type_t type, uint8_t phy_addr, int cs, int irq, 
         log_w("ETH Already Started");
         return true;
     }
+#if ETH_SPI_SUPPORTS_NO_IRQ
     if(cs < 0){
         log_e("CS pin must be defined!");
+#else
+    if(cs < 0 || irq < 0){
+        log_e("CS and IRQ pins must be defined!");
+#endif
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


Jason2866:
@TD-er  This is not in official version (Tasmota fork only)
```
case ETH_PHY_JL1101:
            phy = esp_eth_phy_new_jl1101(&phy_config);
            break;
```
Needs to be removed.

me-no-dev:
this will have to first be backported from IDF.

libraries/Ethernet/src/ETH.cpp
```diff
@@ -351,7 +351,7 @@ bool ETHClass::beginSPI(eth_phy_type_t type, uint8_t phy_addr, int cs, int irq,
         log_w("ETH Already Started");
         return true;
     }
-    if(cs < 0 || irq < 0){
+    if(cs < 0 /*|| irq < 0*/){
         log_e("CS and IRQ pins must be defined!");
```
me-no-dev, line 355: remove the commented check and fix the error message

TD-er:
> this will have to first be backported from IDF.

Absolutely, but when that's done it is very easy to implement :)

me-no-dev:
@TD-er updated the PR to be able to merge it now and enable the feature with define, once it's available from IDF

TD-er:
Ah good to know, so I should not forget to add those defines in my own builds or else it won't work with my own builds when updating the Arduino code :)

me-no-dev:
@TD-er you can add `#define ETH_SPI_SUPPORTS_NO_IRQ 1` to `ETH.h` or globally define it in the build to enable the feature

TD-er:
@TD-er Yep, already add it to my platformIO config or else I'm sure I will forget it later :)
Thanks for merging.

PR #9253, opened 2024-02-15
WebServer: Ignore extra headers within multipart forms

tcsullivan:
## Description of Change
This Pull Request provides a fix for the WebServer library to ignore additional headers that may appear in `multipart/form-data` uploads. The current code assumes that `Content-Disposition` headers can only possibly be followed by a `Content-Type` header, and while that is what the standard (RFC7578) intends, it is also required that any other headers be ignored (see related link).

Currently, when extra header(s) are present the WebServer library will fail to properly locate the start of the submitted data. To fix this, the `Content-Type` check is simply wrapped in a `while` loop to skip over any additional headers.

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v2.0.11 with an ESP32C3 Dev Module with this scenario:

1. Create a file upload form with the WebServer.
2. Send a POST request (in my case, using [okhttp](https://github.com/square/okhttp)) to the WebServer that embeds additional headers, for example with `Content-Length`:
```
--f9426897-2492-4877-b202-7f7703681600
content-disposition: form-data; name="firmware"; filename="update.bin"
Content-Type: application/octet-stream
Content-Length: 399366

<data>
```
3. The current WebServer code will fail to acknowledge the `Content-Length` header and assume that the file data begins on the empty line that ends the header section, i.e. the WebServer will prepend the newline sequence `0x0D 0x0A` to the received file data.

With the provided fix, the header is ignored and the file upload is successful.

## Related links
RFC 7578's requirement to ignore other header fields: [https://www.rfc-editor.org/rfc/rfc7578#section-4.8](https://www.rfc-editor.org/rfc/rfc7578#section-4.8)

okhttp's fix, not yet in a stable release: [https://github.com/square/okhttp/issues/2604#issuecomment-1150938218](https://github.com/square/okhttp/issues/2604#issuecomment-1150938218)

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -368,10 +368,12 @@ bool WebServer::_parseForm(WiFiClient& client, String boundary, uint32_t len){
           argType = FPSTR(mimeTable[txt].mimeType);
           line = client.readStringUntil('\r');
           client.readStringUntil('\n');
-          if (line.length() > 12 && line.substring(0, 12).equalsIgnoreCase(FPSTR(Content_Type))){
-            argType = line.substring(line.indexOf(':')+2);
-            //skip next line
-            client.readStringUntil('\r');
+          while (line.length() > 0) {
+            if (line.length() > 12 && line.substring(0, 12).equalsIgnoreCase(FPSTR(Content_Type))){
+              argType = line.substring(line.indexOf(':')+2);
+            }
+            //skip over any other headers
+            line = client.readStringUntil('\r');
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y:
@tcsullivan Can you please retest with 3.0.0-aplha3 version or latest master? Just to be sure :)

tcsullivan:
@P-R-O-C-H-Y I just retested with 3.0.0-alpha3, the issue is still present.

lucasssvaz (approved):

P-R-O-C-H-Y (approved):

PR #9257, opened 2024-02-16
feature(ledc): Add output invert option for LEDC pin + minor fixes

### LINKED ISSUE #7501: ledcAttachPin shoud support output_invert param (IDFGH-8781)

### LINKED ISSUE #9212: ledcFade can lead to dead-lock on ESP32-classic
### Board

ESP32 (classic)

### Device Description

ESP32-classic

### Hardware Configuration

Nothing attached to the board

### Version

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

40MHz

### PSRAM enabled

yes

### Upload speed

115200

### Description

Several calls to `ledcAttach` followed by `ledcFade` on the same pin without calling `ledcDetach` for that pin will lead to a completely frozen system on ESP32.

N.B. the same code on ESP32-C2 does not show this behavior. Not tested on other ESP32-variants.

Since there doesn't seem to be a simple check to see if a pin is already attached, I added a simple `ledcWrite` check first to set the start duty cycle for the upcoming fade.
If this returns false, I run `ledcAttach` on that pin.

So either of these 2 variants do work:

```c++
    if (!ledcWrite(gpio, start_duty)) {
      // Pin not yet attached
      if (!ledcAttach(gpio, frequency, resolution)) {
        return false;
      }
    }

    if (!ledcFade(gpio, start_duty, target_duty, fadeDuration_ms)) {
      return false;
    }
```

```c++
    ledcDetach(gpio);
    if (!ledcAttach(gpio, frequency, resolution)) {
      return false;
    }

    if (!ledcFade(gpio, start_duty, target_duty, fadeDuration_ms)) {
      return false;
    }
```

### Sketch

```cpp
// Not working code
while (true) {
    if (ledcAttach(gpio, frequency, resolution)) {
      ledcFade(gpio, start_duty, target_duty, fadeDuration_ms);
    }
    delay(2*fadeDuration_ms);
}
```

### Debug Message

```plain
Node just freezes after a few calls
Probably due to the semaphore code in ledcFadeConfig
```

### Other Steps to Reproduce

_No response_

### LINKED ISSUE #9213: [LedC] ledc_handle.used_channels is never initialized
### Board

ESP32 (classic)

### Device Description

-

### Hardware Configuration

-

### Version

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

40MHz

### PSRAM enabled

yes

### Upload speed

115200

### Description

`ledcAttach` uses `ledc_handle.used_channels` to pick an unused channel.
However this is never initialized to 0.

This can lead to nearly impossible to reproduce issues where per build (or run?) the number of usable ledC channels can differ.

Suggested change:
```c++
bool ledcAttach(uint8_t pin, uint32_t freq, uint8_t resolution)
{
    if (!fade_initialized) {
        ledc_handle.used_channels = 0;
    }
...
```

Slightly related issue: #9212 

### Sketch

```cpp
-
```

### Debug Message

```plain
-
```

### Other Steps to Reproduce

_No response_

P-R-O-C-H-Y:
## Description of Change

- Added `ledcOutputInvert(uint8_t pin, bool out_invert)` function, to enable setting inverting LEDC output for selected pin. Can be called after `ledcAttach()` anytime.
- Fixing `ledc_handle.used_channels` never initialised before use.
- Documented all function the the header file.

## Tests scenarios
Tested on ESP32. On screen below `ledcOutputInvert` on LED2 was called on the black line, and LED2 output was immediately inverted.
Snímek obrazovky 2024-02-16 v 13 54 14

## Related links
Closes #7501 
Closes #9213 
Closes #9212


cores/esp32/esp32-hal-ledc.c
```diff
@@ -120,11 +132,11 @@ bool ledcAttachChannel(uint8_t pin, uint32_t freq, uint8_t resolution, uint8_t c
...
 {
     uint8_t free_channel = ~ledc_handle.used_channels & (ledc_handle.used_channels+1);
-    if (free_channel == 0 || resolution > LEDC_MAX_BIT_WIDTH){
-        log_e("No more LEDC channels available! (maximum %u) or bit width too big (maximum %u)", LEDC_CHANNELS, LEDC_MAX_BIT_WIDTH);
+    if (free_channel == 0){
+        log_e("No more LEDC channels available! (maximum is %u channels)", LEDC_CHANNELS);
         return false;
     }
-    int channel = log2(free_channel & -free_channel);
+    uint8_t channel = log2(free_channel & -free_channel);
```
TD-er, line 139: (I think) I understand the logic behind this, but isn't it a bit too much to use such a heavy function to find a specific bit not yet set to 0?
After all this log function does convert to double types and if this is the only time you use this log function it will add quite a bit of code.
And given the hardly readable bit operations which look really (a bit too) smart and fast, using this log2 function will get rid of all the gained speed and small code.

TD-er: Hmm I didn't know the CTZ. Not sure how standard C++ this is, but it will probably create the smallest code.
The while loop is the most readable one.

Not 100% sure if there should be an extra check to make sure we're not returning any channel higher than the max? (are there ESP's with less than 8 channels?)

P-R-O-C-H-Y: I will run some more test to check if the there is no risk of getting channel higher than max. But there is already a check at the top of `ledcAttachChannel()` which is called by ledcAttach:
```
if(channel >= LEDC_CHANNELS || ledc_handle.used_channels & (1UL << channel)){
        log_e("Channel %u is not available (maximum %u) or already used!", channel, LEDC_CHANNELS);
        return false;
}
```

I think as the `__buildin_ctz()` is standard and useful function in GCC, it can be used. Also can add a simple comment what it does for better readability.

TD-er: I read the `__buildin_ctz(0)` is undefined, but since you already checked for it, that's fine here.
Including some comment, it is probably the most clean code indeed.

P-R-O-C-H-Y: Call __buildin_ctz(0) will never happen, as before that there is a check for `free_channel`:
```
if (free_channel == 0){
        log_e("No more LEDC channels available! (maximum is %u channels)", LEDC_CHANNELS);
        return false;
}
```
So we have winner here :)

TD-er: Not only do we have a winner, I also learned about a new function in C/C++ :)

docs/en/api/ledc.rst
lucasssvaz, line 1: I think this broke some things in the documentation:

https://github.com/espressif/arduino-esp32/blob/651d19fc49da4c95bfc901eaeb5ce3c4fcb2fb1c/docs/en/api/ledc.rst

lucasssvaz (changes requested):

cores/esp32/esp32-hal-ledc.h
```diff
@@ -45,20 +45,150 @@ typedef struct {
...
+ * @return frequency configured for the LEDC channel.
+ *         If ``0`` is returned, error occurs and LEDC pin was not configured.
+ */
+uint32_t ledcChangeFrequency(uint8_t pin, uint32_t freq, uint8_t resolution);
+
+/**
+ * @brief Sets inverting of the output signal for a given LEDC pin.
+ * 
+ * @param pin GPIO pin
+ * @param out_invert select, if output should be inverted (1 = inverting output).
```
SuGlider, line 148:
```suggestion
 * @param out_invert select, if output should be inverted (true = inverting output).
```
maybe it is better using bool value?

docs/en/api/ledc.rst
```diff
@@ -171,6 +168,21 @@ This function is used to set frequency for the LEDC pin.
...
+****************
+
+This function is used to set inverting output for the LEDC pin.
+
+.. code-block:: arduino
+
+    bool ledcOutputInvert(uint8_t pin, bool out_invert);
+
+* ``pin`` select LEDC pin.
+* ``out_invert`` select, if output should be inverted (1 = inverting output).
```
SuGlider, line 181:
```suggestion
* ``out_invert`` select, if output should be inverted (true = inverting output).
```
The same. Using bool values may be better.

cores/esp32/esp32-hal-ledc.c
```diff
@@ -58,15 +60,25 @@ static bool ledcDetachBus(void * bus){
...
 {
-    if (channel >= LEDC_CHANNELS || resolution > LEDC_MAX_BIT_WIDTH)
-    {
-        log_e("Channel %u is not available! (maximum %u) or bit width too big (maximum %u)", channel, LEDC_CHANNELS, LEDC_MAX_BIT_WIDTH);
+    if(channel >= LEDC_CHANNELS || ledc_handle.used_channels & (1UL << channel)){
+        log_e("Channel %u is not available (maximum %u) or already used!", channel, LEDC_CHANNELS);
+        return false;
+    }
+    
+    if (resolution > LEDC_MAX_BIT_WIDTH){
```
SuGlider, line 68:
```suggestion
    if (resolution == 0 || resolution > LEDC_MAX_BIT_WIDTH){
```
Resolution can't be zero, right?

cores/esp32/esp32-hal-ledc.c
```diff
@@ -58,15 +60,25 @@ static bool ledcDetachBus(void * bus){
...
+        log_e("LEDC resolution too big (maximum %u)", LEDC_MAX_BIT_WIDTH);
```
SuGlider, line 69:
```suggestion
        log_e("LEDC resolution is zero or it is too big (maximum %u)", LEDC_MAX_BIT_WIDTH);
```
Message about zero resolution.

cores/esp32/esp32-hal-ledc.c
SuGlider: It may also be necessary in `uint32_t ledcChangeFrequency()`.

SuGlider (changes requested):
A minor suggestion for the function parameter description.
A suggestion about testing resolution for Zero (which may be not allowed).
Check if this makes sense. Overall it is good.

cores/esp32/esp32-hal-ledc.c
```diff
@@ -58,15 +60,25 @@ static bool ledcDetachBus(void * bus){
...
 bool ledcAttachChannel(uint8_t pin, uint32_t freq, uint8_t resolution, uint8_t channel)
 {
-    if (channel >= LEDC_CHANNELS || resolution > LEDC_MAX_BIT_WIDTH)
-    {
-        log_e("Channel %u is not available! (maximum %u) or bit width too big (maximum %u)", channel, LEDC_CHANNELS, LEDC_MAX_BIT_WIDTH);
+    if(channel >= LEDC_CHANNELS || ledc_handle.used_channels & (1UL << channel)){
+        log_e("Channel %u is not available (maximum %u) or already used!", channel, LEDC_CHANNELS);
+        return false;
+    }
+
```
SuGlider, line 67:
```suggestion
    if (freq == 0) {
        log_e("LEDC pin %u - frequency can't be zero.", pin);
    }
```
frequency can't be zero as well.  IDF will fail with BAD PARAMETER error.

SuGlider: This verification (freq != 0 && resoution != 0) shall also be done in
`uint32_t ledcChangeFrequency(uint8_t pin, uint32_t freq, uint8_t resolution)`

TD-er:
Just as a suggestion about resolution being 0.
For some functions, there are multiple parameters given which also depend on eachother, like frequency and resolution.
There are some configurations possible where frequency and resolution exceed the max. clock frequency they depend on.

So a resolution of 0 could perhaps be interpreted like "leave as it is or lower resolution if needed".
Is this something that's useful? Or will it result in confusion to the user as the max. PWM duty then also changes?

If this is not practical, then maybe this class could have some static function to compute the max. frequency given some resolution and vice verse?
Especially if the max. depends on platform and/or the max. frequency can change at runtime for example in low power mode or when the CPU clock has changed.

cores/esp32/esp32-hal-ledc.c
SuGlider: in function `ledcChangeFrequency()`, it may better to change/add this:

``` cpp
        if (resolution == 0 || resolution > LEDC_MAX_BIT_WIDTH){  
            log_e("LEDC pin %u - resolution is zero or it is too big (maximum %u)", pin, LEDC_MAX_BIT_WIDTH);
            return 0;
        }
        if (freq == 0) {
            log_e("LEDC pin %u - frequency can't be zero.", pin);
            return 0;
        }
```

P-R-O-C-H-Y:
> Just as a suggestion about resolution being 0. For some functions, there are multiple parameters given which also depend on eachother, like frequency and resolution. There are some configurations possible where frequency and resolution exceed the max. clock frequency they depend on.
> 
> So a resolution of 0 could perhaps be interpreted like "leave as it is or lower resolution if needed". Is this something that's useful? Or will it result in confusion to the user as the max. PWM duty then also changes?
> 
> If this is not practical, then maybe this class could have some static function to compute the max. frequency given some resolution and vice verse? Especially if the max. depends on platform and/or the max. frequency can change at runtime for example in low power mode or when the CPU clock has changed.

I think this is not ideal, as the resolution will be kind of "unknown" to the user, so it will be hard for them to set right duty for the PWM. It will make sense, if we also add a function like `ledcWritePercentage(uint8_t percentage)`. So the correct duty will be calculated according to the resolution set and the percentage automatically.

docs/en/api/ledc.rst
P-R-O-C-H-Y: fixed :)

SuGlider (approved):
Very good!

lucasssvaz (approved):
LGTM

cores/esp32/esp32-hal-ledc.c
```diff
@@ -119,12 +134,12 @@ bool ledcAttachChannel(uint8_t pin, uint32_t freq, uint8_t resolution, uint8_t c
 
 bool ledcAttach(uint8_t pin, uint32_t freq, uint8_t resolution)
 {
-    uint8_t free_channel = ~ledc_handle.used_channels & (ledc_handle.used_channels+1);
-    if (free_channel == 0 || resolution > LEDC_MAX_BIT_WIDTH){
-        log_e("No more LEDC channels available! (maximum %u) or bit width too big (maximum %u)", LEDC_CHANNELS, LEDC_MAX_BIT_WIDTH);
+    int free_channel = ~ledc_handle.used_channels & (ledc_handle.used_channels+1);
```
TD-er, line 137: If I'm not mistaken, ESP32-C3 and -C2 do have 6 channels.
Thus if all 6 are occupied, `free_channel` will be `0b01000000`
And even with 8 channels you will end up with `0b100000000`, right?

Shouldn't this check be more like this:
```c++
int free_channel = (~ledc_handle.used_channels & (ledc_handle.used_channels+1)) & ((1<<LEDC_CHANNELS)-1));
```

P-R-O-C-H-Y: Works fine for C3 with 6 channels only. There is no need for any changes.

SuGlider: I agree that the free channel calculation must take LEDC_CHANNEL into consideration.

P-R-O-C-H-Y: If bit is 1, channel is used. 0 if channel is free.

SuGlider: @P-R-O-C-H-Y - PTAL.

P-R-O-C-H-Y: The LEDC_CHANNEL is taken in consideration in the structure -> ledc_handle.used_channel variable.
```
typedef struct {
    int used_channels : LEDC_CHANNELS;              // Used channels as a bits
} ledc_periph_t;
```

TD-er:
Should this also fix https://github.com/espressif/arduino-esp32/issues/9212 ?
Or couldn't you reproduce that one?

Anyway, see my comment above as your changes still have a bug in the computation of calculating the free channel.

cores/esp32/esp32-hal-ledc.c
P-R-O-C-H-Y: Just tested that for C3 (6 channels) and ESP32 (16 channels). Both working well :)

TD-er: > The LEDC_CHANNEL is taken in consideration in the structure -> ledc_handle.used_channel variable.
> 
> ```
> typedef struct {
>     int used_channels : LEDC_CHANNELS;              // Used channels as a bits
> } ledc_periph_t;
> ```

But if you copy it to an int and invert it, when is the inversion done?
Thus is the result like this `~((int)used_channels)` or `(int)(~used_channels)` ?

SuGlider: Yes, the logic works percfetly for any number of bits. Incluing empty channels in any place.
Example: mask = 0x101 ==> (NOT mask) = 0x010 ==> (mask +1) = 0x110 :: NOT & (+1) = 0x010 & 0x110 = 0x010 (channel is free).

Solved.

TD-er: My point was that there is a check for being 0 after this assignment and if you're not masking it with nr of bits, then the free_channel is not 0.

TD-er: `ledcAttachChannel` does check for `LEDC_CHANNELS`, so it will not fail indeed (apart from the possible concurrency mentioned by SuGlider)
So then it only is a matter of (possibly) never reaching the error here: `if (free_channel == 0){`

P-R-O-C-H-Y: A simple sketch for testing on C3 (6 channels):
```
#define LEDC_RESOLUTION  8
#define LEDC_FREQ     5000

void setup() {
    ledcAttach(4, LEDC_FREQ, LEDC_RESOLUTION);
    ledcAttach(5, LEDC_FREQ, LEDC_RESOLUTION);
    ledcAttach(6, LEDC_FREQ, LEDC_RESOLUTION);
    ledcAttach(7, LEDC_FREQ, LEDC_RESOLUTION);
    ledcAttach(8, LEDC_FREQ, LEDC_RESOLUTION);
    ledcAttach(9, LEDC_FREQ, LEDC_RESOLUTION);

    //No more channels available
    ledcAttach(10, LEDC_FREQ, LEDC_RESOLUTION);

    // Set the LED pin as an output
    ledcDetach(7);
    ledcAttach(10, LEDC_FREQ, LEDC_RESOLUTION);
}

void loop() {
}
```

I have added debug logs to the `ledcAttach()` function:
```
bool ledcAttach(uint8_t pin, uint32_t freq, uint8_t resolution)
{
    log_d("Used channels: %X", ledc_handle.used_channels);
    int free_channel = ~ledc_handle.used_channels & (ledc_handle.used_channels+1);
    log_d("Free channel: %u", free_channel);
    if (free_channel == 0){
        log_e("No more LEDC channels available! (maximum is %u channels)", LEDC_CHANNELS);
        return false;
    }
    uint8_t channel = __builtin_ctz(free_channel); // Convert the free_channel bit to channel number
    log_d("Channel num: %u", channel);
    return ledcAttachChannel(pin, freq, resolution, channel);
}
```

Debug output:
```
[   401][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: 0
[   407][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 1
[   412][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 0
[   418][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 4 (channel 0, resolution 8)
[   427][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: 1
[   433][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 2
[   439][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 1
[   445][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 5 (channel 1, resolution 8)
[   454][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: 3
[   460][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 4
[   465][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 2
[   471][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 6 (channel 2, resolution 8)
[   480][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: 7
[   486][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 8
[   492][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 3
[   498][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 7 (channel 3, resolution 8)
[   507][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: F
[   513][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 16
[   518][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 4
[   524][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 8 (channel 4, resolution 8)
[   533][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: 1F
[   539][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 32
[   545][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 5
[   551][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 9 (channel 5, resolution 8)
[   560][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: FFFFFFFF
[   566][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 0      <------ free channel gets to 0
[   572][E][esp32-hal-ledc.c:141] ledcAttach(): No more LEDC channels available! (maximum is 6 channels)
[   581][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: FFFFFFF7
[   588][D][esp32-hal-ledc.c:139] ledcAttach(): Free channel: 8
[   593][D][esp32-hal-ledc.c:145] ledcAttach(): Channel num: 3
[   599][I][esp32-hal-ledc.c:131] ledcAttachChannel(): LEDC attached to pin 10 (channel 3, resolution 8)
```

You can see that after using all 6 channels (0-5), the free channel is correctly calculated as 0, so the error message is printed and returned correctly.

P-R-O-C-H-Y:
> Should this also fix #9212 ? Or couldn't you reproduce that one?

I was not albe to reproduce that, but as I have fixed part, where if the pin was already attached to the LEDC was wrongly reattached. Now if the pin is already attached to the LEDC, the `ledcAttach()` will return false and print info of how the pin is set up (channel, frequency and resolution).

So you are not able to call `ledcAttach()` multiple times on same pin, without previously detaching the pin from LEDC.

TD-er:
That sounds good.

Just a question, is there a generic/preferred function to clear any assigned bus/peripheral from a pin, regardless of what was using it?
If there is one, then I could use it as a generic way to make sure a pin will not be assigned to the same peripheral without 'unregistering' it first.
Or maybe even better, to check if a pin is set to some bus/peripheral before trying to assign it to another one and just present the user with some error including how that pin was already assigned.

cores/esp32/esp32-hal-ledc.c
TD-er: Now I see where I got it wrong with the bitwise algebra...
See the log line:
```
[   560][D][esp32-hal-ledc.c:137] ledcAttach(): Used channels: FFFFFFFF
```

The N-bit value is of type `int`, not `unsigned int`.
```c++
typedef struct {
    int used_channels : LEDC_CHANNELS;              // Used channels as a bits
} ledc_periph_t;
```
This thus sets it to a negative value when the "most significant" bit is set.
If you then negate it, those `F`'s become `0`'s

I've been bitten by those signed/unsigned in structs with specific bit-size quite a lot and probably will be a lot more in the future :(

SuGlider:
> That sounds good.
> 
> Just a question, is there a generic/preferred function to clear any assigned bus/peripheral from a pin, regardless of what was using it? If there is one, then I could use it as a generic way to make sure a pin will not be assigned to the same peripheral without 'unregistering' it first. Or maybe even better, to check if a pin is set to some bus/peripheral before trying to assign it to another one and just present the user with some error including how that pin was already assigned.

Arduino Core 3.0.0 has the Peripheral Manager API. 
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-periman.h
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-periman.c

The way it work when a pin is already attached to a peripheral, like PWM (LEDC), and then it is used as RMT, the RMT call to attahc it will call first the LEDC detaching and then run the RMT attachment routine.
Therefore, it will work correctly, turning off all the drivers and setup of previous peripheral before starting the same pin in the new peripheral.

P-R-O-C-H-Y:
> That sounds good.
> 
> Just a question, is there a generic/preferred function to clear any assigned bus/peripheral from a pin, regardless of what was using it? If there is one, then I could use it as a generic way to make sure a pin will not be assigned to the same peripheral without 'unregistering' it first. Or maybe even better, to check if a pin is set to some bus/peripheral before trying to assign it to another one and just present the user with some error including how that pin was already assigned.

@TD-er To check what bus(peripheral) is assigned to a pin you can use `peripheral_bus_type_t perimanGetPinBusType(uint8_t pin);` which will return a bus type assigned to a pin. If the pin has no bus(peripheral) attached to it, you will get `ESP32_BUS_TYPE_INIT` as returned value.

There is no need to clear the bus first as @SuGlider mentioned. The detaching is automatic before attaching new bus(peripheral). But if you want to clear the pins bus(peripheral) without knowing, what peripheral is attached to the pin, you can use `perimanClearPinBus(PIN);`.

SuGlider:
@P-R-O-C-H-Y - I found an issue... aparently with LEDC detaching, I think.

Try running the example from https://github.com/espressif/arduino-esp32/pull/9257#issuecomment-1954879245
I have a LED attached to GPIO2 (ESP32).
When LEDC runs before SigmaDelta, the LED doesn't light up with SigmaDelta.
But if I invert the order and run SigmaDelta before LEDC, both work fine.

P-R-O-C-H-Y:
> @P-R-O-C-H-Y - I found an issue... aparently with LEDC detaching, I think.
> 
> Try running the example from [#9257 (comment)](https://github.com/espressif/arduino-esp32/pull/9257#issuecomment-1954879245) I have a LED attached to GPIO2 (ESP32). When LEDC runs before SigmaDelta, the LED doesn't light up with SigmaDelta. But if I invert the order and run SigmaDelta before LEDC, both work fine.
> 
> ~--- NOTE: no... the issue is in SigmaDelta...~ it need more debuging.

@SuGlider Will do some testing tomorrow.

SuGlider:
> > @P-R-O-C-H-Y - I found an issue... aparently with LEDC detaching, I think.
> > Try running the example from [#9257 (comment)](https://github.com/espressif/arduino-esp32/pull/9257#issuecomment-1954879245) I have a LED attached to GPIO2 (ESP32). When LEDC runs before SigmaDelta, the LED doesn't light up with SigmaDelta. But if I invert the order and run SigmaDelta before LEDC, both work fine.
> > ~--- NOTE: no... the issue is in SigmaDelta...~ it need more debuging.
> 
> @SuGlider Will do some testing tomorrow.

Issue was in SigmaDelta. Solved with #9268 
LEDC is fine. Ready for Merging.

TD-er:
Both thanks for the extensive explanation and apparently answering this led to another bug :)

I will have another close look at those periman functions to see how I can use them in ESPEasy.

PR #7782, opened 2023-02-01
feat(esp32-s3): Add support for auto-baudrate-detection on S3

imwhocodes:
Add support for auto-baudrate-detection on S3 variant of esp32, with different clock support (could be extended to C3 variant)

-----------
## Description of Change
Add support (fix `hal` wrapper) for auto-baudrate-detection to the esp32-S3
Currently there was no support for it and correct codepath was missing (and not even hinting to the missing support like done for the C3)

Added `#ifdef` for enabling the baudrate detection hardware
Added codepath for correctly computing baudrate given the clock source used for the uart hardware block

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core master with ESP32-S3 Boards

## Related links
Close https://github.com/espressif/arduino-esp32/issues/7718
Progress https://github.com/orgs/espressif/projects/3/views/15?pane=issue&itemId=19207281

## Note about calculations and ESP32-C3
I cannot test because i have no board with this chip (esp32-C3) but I'm supposing that the same code could be used also for fixing this:
https://github.com/espressif/arduino-esp32/blob/23d715af1b185fac21558c907af4ff5b86526aa4/cores/esp32/esp32-hal-uart.c#L670-L674

In particular S3 and C3 variant support other clock sources other than **APB** and they are in fact defaulted to XTAL clock as source by Arduino, so the previous problem could be by this line not tacking that in account:
https://github.com/espressif/arduino-esp32/blob/23d715af1b185fac21558c907af4ff5b86526aa4/cores/esp32/esp32-hal-uart.c#L725

PR #9261, opened 2024-02-17
feat: adds baudrate detection to S3 and C3 using LL API for all SoC

SuGlider:
## Description of Change
Adds Baud Rate detection into V2.0.15 for the ESP32-C3 and ESP32-S3.
Those 2 SoC have a maximum resolution of 12 bits, while ESP32 and ESP32-S2 have a resolution of 20 bits.
The implication is that ESP32-S3 and ESP32-C3 can not detect any baud rate under 9600 baud.
This is because the calculation of the minimum baud = 40MHz / 4096 =~ 9600

After testing all the 3 possible Clock Source for UART, it is clear that the most precise is when using XTAL as CLK SOURCE.
Using APB or RTC do not produce good results for the C3 or S3.

In case that the calculate baud rate uses the maximum 12 bits resolution (4095), the code will issue a warning message (log_w).
This is because it may be wrong as it can be in the range of 300 to 9600 baud, but the calculation will always be 9600.

This PR shall replace #7782 

## Tests scenarios
Using the Serial example in https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/Serial/BaudRateDetect_Demo/BaudRateDetect_Demo.ino

## Related links
Closes #7782

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -664,49 +664,15 @@ unsigned long uartBaudrateDetect(uart_t *uart, bool flg)
  * detected calling uartBadrateDetect(). The raw baudrate is computed using the UART_CLK_FREQ. The raw baudrate is 
  * rounded to the closed real baudrate.
  * 
- * ESP32-C3 reports wrong baud rate detection as shown below:
- * 
- * This will help in a future recall for the C3.
- * Baud Sent:          Baud Read:
- *  300        -->       19536
- * 2400        -->       19536
- * 4800        -->       19536 
- * 9600        -->       28818 
- * 19200       -->       57678
- * 38400       -->       115440
- * 57600       -->       173535
- * 115200      -->       347826
- * 230400      -->       701754
- * 
- * 
 */
 void uartStartDetectBaudrate(uart_t *uart) {
     if(uart == NULL) {
         return;
     }
 
-#ifdef CONFIG_IDF_TARGET_ESP32C3
-    
-    // ESP32-C3 requires further testing
-    // Baud rate detection returns wrong values 
-   
-    log_e("ESP32-C3 baud rate detection is not supported.");
-    return;
-
-    // Code bellow for C3 kept for future recall
-    //hw->rx_filt.glitch_filt = 0x08;
-    //hw->rx_filt.glitch_filt_en = 1;
-    //hw->conf0.autobaud_en = 0;
-    //hw->conf0.autobaud_en = 1;
-#elif CONFIG_IDF_TARGET_ESP32S3
-    log_e("ESP32-S3 baud rate detection is not supported.");
-    return;
-#else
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-uart.c
```diff
@@ -732,37 +703,27 @@ uartDetectBaudrate(uart_t *uart)
...
     
-    //log_i("APB_FREQ = %d\nraw baudrate detected = %d", getApbFrequency(), baudrate);
-
-    static const unsigned long default_rates[] = {300, 600, 1200, 2400, 4800, 9600, 19200, 38400, 57600, 74880, 115200, 230400, 256000, 460800, 921600, 1843200, 3686400};
-
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
+    unsigned long baudrate = getApbFrequency() / divisor;  // ESP32 and S2 APB Freq
+#else
+    unsigned long baudrate = (getXtalFrequencyMhz() * 1000000) / divisor;  // S3 and C3 XTAL Frequency
+#endif
```
lucasssvaz, line 714: What do you think of doing something like:

```suggestion
#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
    unsigned long baudrate = getApbFrequency() / divisor;  // ESP32 and S2 APB Freq
#elif CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C3
    unsigned long baudrate = (getXtalFrequencyMhz() * 1000000) / divisor;  // S3 and C3 XTAL Frequency
#else
    log_e("Baud rate detection not supported.");
    return 0;
#endif
```

SuGlider: This is for Arduino 2.0.x - there no other SoC beside ESP32, ESP32-2, ESP32-S3 and ESP32-C3.
The code works the same way for S3 and C3. It was already working fine for the ESP32 and S2.

cores/esp32/esp32-hal-uart.c
```diff
@@ -664,61 +673,23 @@ unsigned long uartBaudrateDetect(uart_t *uart, bool flg)
...
-
-    // Code bellow for C3 kept for future recall
-    //hw->rx_filt.glitch_filt = 0x08;
-    //hw->rx_filt.glitch_filt_en = 1;
-    //hw->conf0.autobaud_en = 0;
-    //hw->conf0.autobaud_en = 1;
-#elif CONFIG_IDF_TARGET_ESP32S3
-    log_e("ESP32-S3 baud rate detection is not supported.");
-    return;
-#else
```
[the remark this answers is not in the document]
SuGlider: It works fine for ESP32-S3 and ESP32-C3. There is no `#else`. This is for Arduino Core 2.0.15 only.

P-R-O-C-H-Y (approved):

lucasssvaz (approved):

Issue #8661, opened 2023-09-22
assertion "Invalid mbox" failed


Abhay-2412:
### Board

ESP32 wrover

### Device Description

In my project, I successfully utilized several ICs that communicate via I2C, as well as a camera. Additionally, the project incorporated both Wi-Fi and Bluetooth functionality.

### Hardware Configuration

In my project, I successfully utilized several ICs that communicate via I2C, as well as a camera. Additionally, the project incorporated both Wi-Fi and Bluetooth functionality.

### Version

v2.0.3

### IDE Name

Arduino IDE

### Operating System

windows 11

### Flash frequency

80Mhz

### PSRAM enabled

yes

### Upload speed

921600

### Description

im trying to implement a project using BLE  and WIFI features and some ICs i encounterd following error 

```
14:21:45.597 -> E (2504) cam_hal:  not in cam_obj->jpeg_mode
14:21:45.597 -> E (2504) esp32 ll_cam: node_size: 1280, nodes_per_line: 4, lines_per_node: 1, dma_half_buffer_min:  5120, dma_half_buffer: 40960,lines_per_half_buffer:  8, dma_buffer_size: 81920, image_size: 307200
14:21:45.597 -> E (2511) cam_hal: buffer_size: 81920, half_buffer_size: 40960, node_buffer_size: 1280, node_cnt: 64, total_cnt: 7
14:21:45.644 -> OV5640 camera is connected
14:22:01.373 -> assertion "Invalid mbox" failed: file "/home/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/esp-idf/components/lwip/lwip/src/api/tcpip.c", line 374, function: tcpip_send_msg_wait_sem
14:22:01.406 -> abort() was called at PC 0x400f9e1f on core 1
14:22:01.406 -> 
14:22:01.406 -> ELF file SHA256: 0000000000000000
14:22:01.406 -> 
[...]

PR #9262, opened 2024-02-17
Fixed "Invalid mbox" error in SimpleTime example

kinoh:
## Description of Change

As stated in https://github.com/espressif/arduino-esp32/issues/8661, we have to call `WiFi.mode` before networking.

## Tests scenarios

I've tested my Pull Request on arduino-esp32 v2.0.14 with ESP32-DevKitC-32E Board.

## Related links

- https://github.com/espressif/arduino-esp32/issues/8661

### DIFF
--- a/libraries/ESP32/examples/Time/SimpleTime/SimpleTime.ino
+++ b/libraries/ESP32/examples/Time/SimpleTime/SimpleTime.ino
@@ -33,8 +33,8 @@ void setup()
 {
   Serial.begin(115200);
 
-  // configure WiFi
-  WiFi.mode(WIFI_STA);
+  // configure WiFi STA and connect in order to get the current time and date.
+  WiFi.begin(ssid, password);
 
   // set notification call-back function
   sntp_set_time_sync_notification_cb( timeavailable );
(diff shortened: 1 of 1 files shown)


SuGlider:
@me-no-dev - This should also be merged to the `release/v2.x` branch.

lucasssvaz (approved):

P-R-O-C-H-Y:
@SuGlider is it ok now to call Wifi.begin() twice?

SuGlider:
> @SuGlider is it ok now to call `Wifi.begin()` twice in the example?

You are right, it is also called latter. I must call it at first and then use IDF calls.

P-R-O-C-H-Y (approved):

vortigont:
Does mentioned issue (invalid mbox) really affects this ntp example? It is related to TCP socket, but NTP does not use TCP at all.
What this change is definitely breaks is NTPoDHCP feature, because dhcp client option should be set _before_ acquiring IP address as stated in comment below.

https://github.com/espressif/arduino-esp32/blob/50ef6f4369fb85139f000f7bbc5a9f9d5bc02b9f/libraries/ESP32/examples/Time/SimpleTime/SimpleTime.ino#L45-L53

PR #9264, opened 2024-02-19
Added support for 2 new Olimex boards and updated 3 existing ones

Stanimir-Petev:
Hello!
Olimex employee here.

I have added support for the boards: Olimex ESP32-H2-DevKit-LiPo and ESP32-SBC-FabGL. Added a "pins_arduino.h" files in the "variants" folder and in the board.txt. The boards are open-source hardware and info can be found at our website.

Also for 3 of our existing in your package boards (ESP32-POE, ESP32-POE-ISO, ESP32-Gateway) inside the pins_arduino.h files I added the ethernet macros to match the fix from https://github.com/espressif/arduino-esp32/pull/9242 . 

For ESP32-POE and ESP32-POE-ISO added extra menus and menu options for: PSRAM, Flash Size, Flash mode and Partition scheme.

I tested what I could but let me know if I have missed something and needs further changes.

Thanks espressif and Arduino team.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -16261,6 +16261,371 @@ esp32c6-evb.menu.EraseFlash.all.upload.erase_cmd=-e
 
 ##############################################################
 
+esp32h2-devkitlipo.name=OLIMEX ESP32-H2-DevKit-LiPo
+esp32h2-devkitlipo.vid.0=0x303a
+esp32h2-devkitlipo.pid.0=0x1001
+
+esp32h2-devkitlipo.bootloader.tool=esptool_py
+esp32h2-devkitlipo.bootloader.tool.default=esptool_py
+
+esp32h2-devkitlipo.upload.tool=esptool_py
+esp32h2-devkitlipo.upload.tool.default=esptool_py
+esp32h2-devkitlipo.upload.tool.network=esp_ota
+
+esp32h2-devkitlipo.upload.maximum_size=1310720
+esp32h2-devkitlipo.upload.maximum_data_size=327680
+esp32h2-devkitlipo.upload.flags=
+esp32h2-devkitlipo.upload.extra_flags=
+esp32h2-devkitlipo.upload.use_1200bps_touch=false
+esp32h2-devkitlipo.upload.wait_for_upload_port=false
+
+esp32h2-devkitlipo.serial.disableDTR=false
+esp32h2-devkitlipo.serial.disableRTS=false
+
+esp32h2-devkitlipo.build.tarch=riscv32
+esp32h2-devkitlipo.build.target=esp
+esp32h2-devkitlipo.build.mcu=esp32h2
+esp32h2-devkitlipo.build.core=esp32
+esp32h2-devkitlipo.build.variant=esp32h2-devkit-lipo
+esp32h2-devkitlipo.build.board=ESP32H2_DEVKIT_LIPO
+esp32h2-devkitlipo.build.bootloader_addr=0x0
+
+esp32h2-devkitlipo.build.cdc_on_boot=0
+esp32h2-devkitlipo.build.f_cpu=96000000L
+esp32h2-devkitlipo.build.flash_size=4MB
+esp32h2-devkitlipo.build.flash_freq=64m
+esp32h2-devkitlipo.build.img_freq=48m
+esp32h2-devkitlipo.build.flash_mode=qio
+esp32h2-devkitlipo.build.boot=qio
+esp32h2-devkitlipo.build.partitions=default
+esp32h2-devkitlipo.build.defines=
+
+esp32h2-devkitlipo.menu.JTAGAdapter.default=Disabled
+esp32h2-devkitlipo.menu.JTAGAdapter.default.build.copy_jtag_files=0
+esp32h2-devkitlipo.menu.JTAGAdapter.builtin=Integrated USB JTAG
+esp32h2-devkitlipo.menu.JTAGAdapter.builtin.build.openocdscript=esp32h2-builtin.cfg
+esp32h2-devkitlipo.menu.JTAGAdapter.builtin.build.copy_jtag_files=1
+esp32h2-devkitlipo.menu.JTAGAdapter.external=FTDI Adapter
+esp32h2-devkitlipo.menu.JTAGAdapter.external.build.openocdscript=esp32h2-ftdi.cfg
+esp32h2-devkitlipo.menu.JTAGAdapter.external.build.copy_jtag_files=1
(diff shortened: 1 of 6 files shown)


P-R-O-C-H-Y (approved):

Jason2866:
@Stanimir-Petev The psram fixes are only needed when you use the very old esp32 revision 0 and 1. Since the OLIMEX ESP32-SBC-FABGL is a new design the can be removed.

boards.txt
```diff
@@ -16261,6 +16373,371 @@ esp32c6-evb.menu.EraseFlash.all.upload.erase_cmd=-e
...
+esp32-sbc-fabgl.menu.JTAGAdapter.external.build.copy_jtag_files=1
+esp32-sbc-fabgl.menu.JTAGAdapter.bridge=ESP USB Bridge
+esp32-sbc-fabgl.menu.JTAGAdapter.bridge.build.openocdscript=esp32-bridge.cfg
+esp32-sbc-fabgl.menu.JTAGAdapter.bridge.build.copy_jtag_files=1
+
+esp32-sbc-fabgl.menu.PSRAM.disabled=Disabled
+esp32-sbc-fabgl.menu.PSRAM.disabled.build.defines=
+esp32-sbc-fabgl.menu.PSRAM.disabled.build.extra_libs=
+esp32-sbc-fabgl.menu.PSRAM.enabled=Enabled
+esp32-sbc-fabgl.menu.PSRAM.enabled.build.defines=-DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue -mfix-esp32-psram-cache-strategy=memw
```
Jason2866, line 16603: Psram fix not needed. Can be removed.

Stanimir-Petev: Hello @Jason2866 and thanks for the suggestion.

When I created the board entry in the board.txt I just copied that part from the generic ESP32 board (lines 1131-1318) because I have tested that it worked with it being selected. And after that I just changed the name name, build.variant and build.board.
I wasn't aware of what these mfix options does so I thought they are there for a reason and I just let them stay there.
If you are suggesting that they are unnecessary and can be safely removed that's all good.

The question I have is should I remove them from my forked repository and make a new pull request or you can do it directly on your repository?

Jason2866: Dont have the needed rights to change. Not a member -> Collaborator

me-no-dev: @Stanimir-Petev just remove `-mfix-esp32-psram-cache-issue -mfix-esp32-psram-cache-strategy=memw` from all non-ESP32 boards. Those flags are valid only on the original ESP32

boards.txt
```diff
@@ -15071,17 +15071,82 @@ esp32-poe.build.variant=esp32-poe
...
 esp32-poe.build.partitions=default
 esp32-poe.build.defines=
+esp32-poe.build.loop_core=
+esp32-poe.build.event_core=
+
+esp32-poe.menu.PSRAM.disabled=Disabled (WROOM)
+esp32-poe.menu.PSRAM.disabled.build.defines=
+esp32-poe.menu.PSRAM.disabled.build.extra_libs=
+esp32-poe.menu.PSRAM.enabled=Enabled (WROVER)
+esp32-poe.menu.PSRAM.enabled.build.defines=-DBOARD_HAS_PSRAM
```
me-no-dev, line 15087: here you need the workaround flags, because the board is ESP32 based

boards.txt
```diff
@@ -15152,17 +15208,82 @@ esp32-poe-iso.build.variant=esp32-poe-iso
...
 esp32-poe-iso.build.partitions=default
 esp32-poe-iso.build.defines=
+esp32-poe-iso.build.loop_core=
+esp32-poe-iso.build.event_core=
+
+esp32-poe-iso.menu.PSRAM.disabled=Disabled (WROOM)
+esp32-poe-iso.menu.PSRAM.disabled.build.defines=
+esp32-poe-iso.menu.PSRAM.disabled.build.extra_libs=
+esp32-poe-iso.menu.PSRAM.enabled=Enabled (WROVER)
+esp32-poe-iso.menu.PSRAM.enabled.build.defines=-DBOARD_HAS_PSRAM
```
me-no-dev, line 15224: flags also needed here

boards.txt
me-no-dev: Sorry on my end... @Jason2866 this is ESP32 based board. Why did you ask the flags to be removed?

Jason2866: @me-no-dev Since it is clearly documented that **only** old revisions of the esp32 do need this fix. https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/external-ram.html#chip-revisions

Jason2866: Sadly many boards have this absolutely not needed entry, since the are newer than revision 1.

Jason2866: @me-no-dev We are checking in project Tasmota the esp32 revision and search for PSRAM at init phase and enable it when revision >1. This works perfect.
When the revision check identifies esp32 with the bug PSRAM is not enabled and a message is printed, for PSRAM usage the fix files needs to be used during compile.

me-no-dev: @Jason2866 so how do you detect that here? Do we know for sure that all boards are ECO3?

Jason2866: @me-no-dev Just recommended to remove from the *new* board "esp32-sbc-fabgl"

me-no-dev: @Stanimir-Petev can you confirm that you are using only V3 ESP32 modules?

me-no-dev: If that is true, then I can agree, but then why is PSRAM not enabled by default?

Jason2866: True, PSRAM should be always enabled!

DanKoloff: No way to confirm if all ESP32-POE-WROVER and ESP32-POE-ISO-WROVER variants had version 3 modules. But these were made starting 2020 so good chance they are all version 3.

If you ask for all ESP32-POE and ESP32-POE-ISO boards - these use WROOM modules and had been in production since the early days of the ESP32.

We would enable the PSRAM for the ESP32-SBC_FABGL, for the rest the variants with WROOM module are still more popular and it makes more sense to have the PSRAM disabled. We don't want to create addition entries for each variant of the board, since some boards would need to have 3 entries like ESP32-POE, ESP32-POE-WROVER, ESP32-POE-16MB all the same design with different module soldered (one with WROVER module and PSRAM, and the one with bigger flash).

Stanimir-Petev: With the latest commit PSRAM enabled should be the default option on ESP32-SBC-FabGL. Is there anything else that has to be done?

PR #9294, opened 2024-02-27
Enable USB Serial only if selected

me-no-dev:
USB Serial seems to get enabled even if we do not select USB CDC ON BOOT. This should fix it


SuGlider (approved):
LGTM.

SuGlider:
@me-no-dev - please merge it soon because I will rebase the PR about HWCDC/SOF with this changes, before releasing it.

me-no-dev:
@SuGlider done :)

Issue #9043, opened 2023-12-27
Delay with ESP32-S3 USB OTG CDC if no Serial Monitor is open.


SuGlider:
related to Testing ESP-S3-BOX3 GFX/TouchScreen Arduino drivers:

I also just realised that there is an effect with `Serial.print()` in the ESP-BOX-3 when the GFX demo sketch is run with no open `Serial Monitor`. This is a sort of delay in every `Serial.print()` from loop, related to some issue with the USB CDC driver.

In order to avoid it, it is necessary to disable `Serial`, by commenting out the `Serial` initialization:

``` cpp
void setup() {
  //Serial.begin(115200);
  //Serial.setDebugOutput(true);
  //while (!Serial) delay(100);

```

_Originally posted by @SuGlider in https://github.com/espressif/arduino-esp32/issues/8978#issuecomment-1870448731_

SuGlider:
The S3, C3, C6 and H2 that have HW CDC JTAG USB port may cause a delay when using HWCDC in `Serial` by activating `CDC on Boot` and selecting the USB Mode for `Hardware CDC and JTAG`. This happens when no USB application in USB Host side is started to receive the data sent by the ESP32xx.

The USB CDC buffer gets full and the Arduino HW CDC layer will timeout, by default 100ms, until give up trying to send the CDC data.

As a workaround, it is necessary to use `HWCDC::setTxTimeoutMs(timeout_ms)` and set it to zero.

Example:
``` cpp
void setup() {
  Serial.begin(); // USB HW CDC doesn't need any baudrate
  Serial.setDebugOutput(true);  // sends all log_e(), log_i() messages to USB HW CDC
  Serial.setTxTimeoutMs(0);       // sets no timeout when trying to write to USB HW CDC
}
```

TD-er:
@SuGlider Can this also cause similar issues on the USB CDC layer used on the ESP32-S2?

SuGlider:
> @SuGlider Can this also cause similar issues on the USB CDC layer used on the ESP32-S2?

It shoud not affect the ESP32-S2 because it uses OTG USB based on TinyUSB software layer.
S2 has a real full USB interface. I tested the S3 with OTG and this "delay issue" didn't happen.

The S3 has both, the OTG USB and the HW JTAG CDC.
The C3 and C6 has only the HW JTAG CDC.

SuGlider:
@TD-er @softhack007 @djelibeybi

[...]

PR #9307, opened 2024-02-28
fixes default 100ms delay with HWCDC write() is CDC is not connected 

SuGlider:
## Description of Change
Improves the functionality of HW CDC (HW Serial) when CDC is not connected to the Host side. At this point it solves the delay issue only when it is not plugged to the USB port. But even when plugged and the host enumerates it, it still delays when CDC is not connected, causing some side effect.

This PR fixes it and also adds 2 new APIs:
- HWCDC::isConnected() will return true when HWCDC has a CDC connection to an application in the host side (like a serial terminal) and false otherwise.
- HWCDC::isPlugged() will return true when USB is plugged to a USB host and it has enumarated the CDC class.

## Tests scenarios
``` cpp
void setup() {
  Serial0.begin(115200);
  HWCDCSerial.begin();

  Serial0.println("\n ==> Connect to the CDC Serial Terminal\n");
  while(!HWCDCSerial) delay(100);

  Serial0.println("CDC is connected. It will write to the CDC and delay 1s.\nCheck the millisecond counter.");
  Serial0.println("After checking that it is steady in 1,000 milliseconds, unplug the USB cable. Check the time.");
  Serial0.println("Latter plug back the USB cable. Check the time. Do not open the CDC terminal.");
}

void loop() {
  Serial0.print("Time (ms): ");
  Serial0.println(millis());
  HWCDCSerial.print("Time (ms): ");
  HWCDCSerial.println(millis());

  // when USB is plugged and CDC terminal is not connected, it loop execution will take 1.1 seconds instead of 1.0 seconds.
  delay(1000);
}
```

### Bad Output (from UART0):

```
ESP-ROM:esp32h2-20221101
Build:Nov  1 2022
rst:0x1 (POWERON),boot:0xc (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x4083cfd0,len:0xb68
load:0x4083efd0,len:0x2718
load:0x408460e8,len:0x590
entry 0x4083cfd0

 ==> Connect to the CDC Serial Terminal

CDC is connected. It will write to the CDC and delay 1s.
Check the millisecond counter.
After checking that it is steady in 1,000 milliseconds, unplug the USB cable. Check the time.
Latter plug back the USB cable. Check the time. Do not open the CDC terminal.
Time (ms): 12049             <<<<<======== USB plugged and CDC connected
Time (ms): 13049
Time (ms): 14054
Time (ms): 15054 <<<<<< average is + 1,000 ms
Time (ms): 16059
Time (ms): 17059
Time (ms): 18064             <<<<<======== USB NOT plugged and CDC NOT connected
Time (ms): 19064
Time (ms): 20069
Time (ms): 21069  <<<<<< average is + 1,000 ms
Time (ms): 22074
Time (ms): 23074
Time (ms): 24079
Time (ms): 25079             <<<<<======== USB plugged and CDC NOT connected :: PROBLEM
Time (ms): 26184
Time (ms): 27284
Time (ms): 28389
Time (ms): 29489
Time (ms): 30594 <<<<<< average is + 1,100 ms
Time (ms): 31694
Time (ms): 32799
Time (ms): 33899
Time (ms): 35004
```

## Related links
#9043

### DIFF
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -37,7 +37,7 @@ static QueueHandle_t rx_queue = NULL;
 static uint8_t rx_data_buf[64] = {0};
 static intr_handle_t intr_handle = NULL;
 static SemaphoreHandle_t tx_lock = NULL;
-static volatile bool isConnected = false;
+static volatile bool connected = false;
 
 // timeout has no effect when USB CDC is unplugged
 static uint32_t requested_tx_timeout_ms = 100;
@@ -79,12 +79,12 @@ static void hw_cdc_isr_handler(void *arg) {
     if (usbjtag_intr_status & USB_SERIAL_JTAG_INTR_SERIAL_IN_EMPTY) {
         // Interrupt tells us the host picked up the data we sent.
         if(!usb_serial_jtag_is_connected()) {
-            isConnected = false;
+            connected = false;
             usb_serial_jtag_ll_clr_intsts_mask(USB_SERIAL_JTAG_INTR_SERIAL_IN_EMPTY);
             // USB is unplugged, nothing to be done here
             return;
         } else {
-            isConnected = true;
+            connected = true;
         }
         if (usb_serial_jtag_ll_txfifo_writable() == 1) {
             // We disable the interrupt here so that the interrupt won't be triggered if there is no data to send.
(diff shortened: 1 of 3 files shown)


lucasssvaz (approved):
LGTM

PR #9319, opened 2024-03-01
HardwareSerial Available For Write

### LINKED ISSUE #9317: Rethink the result of HardwareSerial::availableForWrite() and setTxBufferSize
I think we were expecting setTxBufferSize and  availableForWrite() to be reporting on a single "TX buffer size", and not that one was "ADDING" to the hardware buffer in reports.

I think there is an argument that says that the setTxBufferSize is buggy or at the least, confusing. 

Certainly setTxBufferSize(N) does nothing if N<128.   Which seems reasonable. and makes one expect that (say)
setTxBufferSize(200) would set a total size of 200.. not 328 as it does. 

It also means its impossible to set a (total size) buffer of (say) 200!  as setTxBufferSize(72) will be rejected. 

Its very complex and I have been getting lost in this.. but does that make sense?

_Originally posted by @dagnall53 in https://github.com/espressif/arduino-esp32/issues/6998#issuecomment-1973427912_
           
           
### Potential Solution:

 Considering that IDF driver (mandatory for our Arduino implementation) only allows to set an additional Tx (or Rx) Buffer when its size is over 128 bytes, which is the HW TX FIFO size:

setTxBufferSize(tx_size) would behave in this way:
if tx_size is lower than 128, it will do nothing and return 128, as the size set.
if tx_size is higher than 128, it will set, internally, the IDF Driver buffer with size = tx_size - 128. It will return tx_size as the size set.
availableForWrite() will return the total available space, including the HW TX FIFO space of 128 bytes.

SuGlider:
## Description of Change
When setting RX buffer, the `setRxBufferSize()` method would fail with values lower than 128 bytes. 
Although it seems what IDF requests, it is not the best way to execute it, because the return value would be zero and the buffer size would be kept the same, by default 256. Now it returns the size it has been set, defining a minimum of 128+1 bytes, as required by IDF UART driver.

By other hand, for TX buffer, the  `setTxBufferSize()` method also fails with less than 128 bytes, returning also 0.
With the change, it will return and set it to 128 (minimum), as requeired by IDF. 
It also changes `availableForWrite()` in order to return a value that matches `setTxBufferSize()` when there is no data to be written, making it possible for applications to better compare and decide when to write to the UART.

## Tests scenarios
Tested with ESP32 and ESP32-S3.

``` cpp
void setup() {
  size_t Ts = Serial.setTxBufferSize(100);
  size_t Rs = Serial.setRxBufferSize(100);
  Serial.begin(115200);
  delay(1000);
  
  Serial.printf("TX Buffer (%d) Size is %d\n", Ts, Serial.availableForWrite());
  Serial.printf("RX Buffer Size is %d\n", Rs);
}

void loop() {
  if (Serial.available()) {
    char c = Serial.read();
    Serial.write(c);
  }
}
```

## Related links
Closes #9317

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -537,35 +537,38 @@ bool HardwareSerial::setMode(SerialMode mode)
     return uartSetMode(_uart, mode);
 }
 
+// minimum total RX Buffer size is the UART FIFO space (128 bytes for most SoC) + 1. IDF imposition.
 size_t HardwareSerial::setRxBufferSize(size_t new_size) {
 
     if (_uart) {
-        log_e("RX Buffer can't be resized when Serial is already running.\n");
+        log_e("RX Buffer can't be resized when Serial is already running. Set it before calling begin().");
         return 0;
     }
 
     if (new_size <= SOC_UART_FIFO_LEN) {
-        log_e("RX Buffer must be higher than %d.\n", SOC_UART_FIFO_LEN);  // ESP32, S2, S3 and C3 means higher than 128
-        return 0;
+        log_w("RX Buffer set to minimum value: %d.", SOC_UART_FIFO_LEN + 1);  // ESP32, S2, S3 and C3 means higher than 128
+        new_size = SOC_UART_FIFO_LEN + 1;
     }
 
     _rxBufferSize = new_size;
     return _rxBufferSize;
 }
 
+// minimum total TX Buffer size is the UART FIFO space (128 bytes for most SoC).
 size_t HardwareSerial::setTxBufferSize(size_t new_size) {
 
     if (_uart) {
-        log_e("TX Buffer can't be resized when Serial is already running.\n");
+        log_e("TX Buffer can't be resized when Serial is already running. Set it before calling begin().");
         return 0;
     }
 
-    if (new_size <= SOC_UART_FIFO_LEN) {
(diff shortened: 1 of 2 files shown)


cores/esp32/HardwareSerial.cpp
```diff
@@ -537,35 +537,37 @@ bool HardwareSerial::setMode(SerialMode mode)
...
-        return 0;
+        log_w("TX Buffer set to minimum value: %d.", SOC_UART_FIFO_LEN);  // ESP32, S2, S3 and C3 means higher than 128
+        _txBufferSize = 0; // it will use just UART FIFO with SOC_UART_FIFO_LEN bytes (128 for most SoC)
+        return SOC_UART_FIFO_LEN;
     }
-
+    // if new_size is higher than SOC_UART_FIFO_LEN, TX Ringbuffer will be active and it will be used to report back "availableToWrite()"
     _txBufferSize = new_size;
-    return _txBufferSize;
+    return new_size;
```
lucasssvaz, line 572: Isn't this exactly the same as it was before ? `_txBufferSize` and `new_size` should have the same value.

SuGlider: yes, it makes no difference... it is a left over along the implementation. In the testing it was not the same in some older code version.

lucasssvaz (approved):

P-R-O-C-H-Y (approved):

PR #9052, opened 2023-12-31
ESP32-S3 PowerFeather Fixes

powerfeatherdev:
## Description of Change

Follow up to https://github.com/espressif/arduino-esp32/pull/8889.

- Adds missing pin definitions for D7 and D8.
- Fixes pin definitions for ALARM and EN
- Makes CDC on boot enabled default.
- Remove partition schemes for flash sizes other than what is on PowerFeather (8MB)
- Add other 8MB schemes: default FATFS variation, large FATFS and SPIFFS

## Tests scenarios

- Internal tests of all publicly exposed pin definitions are ok.
- CDC on boot is default for PowerFeather confirmed.
- Partition schemes build and flash, including newly added ones. The flashed test firmware lists partitions using `esp_partition_xxx` APIs, verified it matches the csv files.

## Related links

First PR: https://github.com/espressif/arduino-esp32/pull/8889
Website: https://www.powerfeather.dev/

PR #8889, opened 2023-11-16
Add support for ESP32-S3 PowerFeather

powerfeatherdev:
## Description of Change

Adds a new `variant` directory for `esp32s3_powerfeather`, under which the `pins_arduino.h` is created.
Entries for `esp32s3_powerfeather` in `boards.txt` are added.

## Tests scenarios

- [x] `pins_arduino.h` is ok
- [x] `boards.txt` entries are ok

## Related links
Website: https://www.powerfeather.dev/
Allocation for PID under Espressif VID used in this PR: https://github.com/espressif/usb-pids/pull/123

PR #9325, opened 2024-03-03
ESP32-S3 PowerFeather Support for v2.x

powerfeatherdev:
Cherry-picked commits from #8889 and #9052, but for release/v2.x.


powerfeatherdev:
Hello @P-R-O-C-H-Y, in https://github.com/espressif/arduino-esp32/pull/8889#discussion_r1401878889 you mentioned that these definitions are no longer necessary. I thought it was only for v3, but is it also the case for v2 (since apparently all checks have passed?).

P-R-O-C-H-Y:
> Hello @P-R-O-C-H-Y, in https://github.com/espressif/arduino-esp32/pull/8889#discussion_r1401878889 you mentioned that these definitions are no longer necessary. I thought it was only for v3, but is it also the case for v2 (since apparently all checks have passed)?

Hello @powerfeatherdev, our core is not using those definitions anywhere, so as you said all check passed without that. But there are lot of 3rd party libraries and examples which may use those definitions. So if they are not defined for the board, the user may get in trouble :)

powerfeatherdev:
@P-R-O-C-H-Y Thanks for the clarification.

powerfeatherdev:
@SuGlider @me-no-dev @P-R-O-C-H-Y Please take a look.

P-R-O-C-H-Y (approved):
Looks good :)

PR #9400, opened 2024-03-21
Add MacAddress string constructor

lucasssvaz:
## Description of Change
Adds MacAddress object constructor from String.

## Tests scenarios
Tested in future ESP NOW example

### DIFF
--- a/cores/esp32/MacAddress.h
+++ b/cores/esp32/MacAddress.h
@@ -54,6 +54,7 @@ public:
     MacAddress(const uint8_t *macbytearray) : MacAddress(MAC6, macbytearray) {}
 
     MacAddress(const char *macstr);
+    MacAddress(const String &macstr) : MacAddress(macstr.c_str()) {}
 
     virtual ~MacAddress() {}
 
--- a/cores/esp32/MacAddress.h
+++ b/cores/esp32/MacAddress.h
@@ -54,7 +54,7 @@ public:
     MacAddress(const uint8_t *macbytearray) : MacAddress(MAC6, macbytearray) {}
 
     MacAddress(const char *macstr);
-    MacAddress(const String &macstr) : MacAddress(macstr.c_str()) {}
+    MacAddress(const String &macstr);
 
     virtual ~MacAddress() {}
 
--- a/cores/esp32/MacAddress.cpp
+++ b/cores/esp32/MacAddress.cpp
@@ -28,6 +28,10 @@ MacAddress::MacAddress(const char *macstr){
     fromString(macstr);
 }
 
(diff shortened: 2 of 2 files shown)


cores/esp32/MacAddress.h
```diff
@@ -54,6 +54,7 @@ class MacAddress : public Printable {
     MacAddress(const uint8_t *macbytearray) : MacAddress(MAC6, macbytearray) {}
 
     MacAddress(const char *macstr);
+    MacAddress(const String &macstr) : MacAddress(macstr.c_str()) {}
```
TD-er, line 57: Why not do the actual implementation in the .cpp file?
By implementing this in the .h file, you will add about 20 (??) bytes for each call made to this constructor.

lucasssvaz: I just made it this way to be similar to the other constructors. For me it doesn't matter if it is implemented in the header or the file.

lucasssvaz: Also, genuine question, Why would the size of the object created increase if the constructor is defined in the header ?

TD-er: When implemented in a .h file, it will always be inlined by the compiler.

For very simple stuff like returning a (private) member value, this is absolutely perfect as it is in fact less "expensive" compared to making another function call, putting stuff on the stack, etc.
But when you do slightly more complex stuff then it will add up.

I had noticed this when playing around with some functions of the `String` class.
There are several functions using a flash string as an argument.
If you simply call those functions and the only thing they do is wrapping a `String()` constructor around it, then this is done in each call to this function.
Similar to when you have your own functions with a `const String&` argument.
If you call those with a flash string, the compiler will wrap this argument with an extra call to the string constructor each time you call the function.

However when you implement your code in the .cpp then it will be compiled only once and your code just makes a function call to it.
Drawback is that it will add to the stack usage, but it will result in smaller binary code.
So it is a trade-off which one is preferred, based on how often the function is called.

There is another drawback to implementing it in the .cpp and that is when there are several matching constructors for your type of object you need to 'convert' to your new class.
It can sometimes lead to strange issues. Similar issues have been reported for the String class, after which some of these optimizations have been reverted.
But I guess the String class is a bit of an odd one anyway with this rather strange object type as flash string which is essentially a const char pointer but should be read using specific functions (at least on ESP8266)

lucasssvaz: Thanks for the explanation. As I said either way for me is fine, so I changed the PR to add the implementation to the cpp file.

P-R-O-C-H-Y (approved):

PR #9401, opened 2024-03-21
fix: HW CDC write()

SuGlider:
## Description of Change
When writing a stream of data to the HW CDC of S3/C3/C6/H2, the SoC looses some bytes and don't send them all.

This fix makes it work fine.

## Tests scenarios
Using an Arduino Sketch and a Python script to test the data flow.
Run the sketch in the S3/C3/C6/H2 with HWCDC on Boot enabled.
Run the script in the USB Host computer. It will open the CDC port and run the test.

### Arduino side:
``` cpp
// ESP32-C3 using HW Serial CDC on Boot enabled

bool testCase1(const String &data) {
  size_t receivedLen = data.length();
  size_t sendLen = 0;

  while (sendLen < receivedLen) {
    if (!Serial.write(data[sendLen++])) break;   // <<< if it fails sending a single byte, it will fail the whole test case
  }
  Serial.println("===");
  if (sendLen != receivedLen) {
    Serial0.printf("\r\n ===> T1::Error -- didn't send all data [%d of %d]\r\n", sendLen, receivedLen);
    return false;
  } else {
    return true;
  }
}

bool testCase2(const String &data) {
  size_t receivedLen = data.length();

  size_t sent = Serial.write(data.c_str(), receivedLen);
  Serial.println("===");
  if (sent != receivedLen) {
    Serial0.printf("\r\n ===> T2::Error -- didn't send all data [%d of %d]\r\n\r\n", sent, receivedLen);
    return false;
  }  else {
    return true;
  }
}

void setup() {
  Serial0.begin(115200);

  Serial.begin();
  Serial0.println("\r\nStarting... open Serial Monitor with CDC port.");
  while(!Serial) delay(100);
  Serial.println("\r\nCDC Started.");

  bool result = true;
  String testString = "0";
  for (int x = 0; x < 100; x++) {  // runs the test cases 100 times
    String testString = "0";
    for (int i = 0; i < 302; i++) {  // sends from 1 to 301 bytes per test
      result &= testCase1(testString);
      result &= testCase2(testString);
      if (!result) {
        Serial0.printf("Failed with %d bytes\r\n", i);
        break;
      }
      char c = testString.charAt(testString.length() - 1);
      c = c == '9' ? '0' : c + 1;
//      if(i > 22) c = '7'; inserts an error to test the python script
      testString += c;
    }
    if (!result) break;
    Serial0.print(x+1);
    Serial0.print(" ");
    if (((x+1) % 10) == 0) Serial0.println();
  }
[...]

PR #9406, opened 2024-03-22
Comment usb_serial_jtag_ll_txfifo_flush()

Jason2866:
Patch PR #9401 with the change from https://github.com/espressif/arduino-esp32/issues/9378#issuecomment-2013740414

@SuGlider


SuGlider (changes requested):
It fails with the test case that I have provided in the PR.
S3 and C3 are hanging after about 50 transmissions.

Jason2866:
Closing since does introduce new issues.

PR #9410, opened 2024-03-24
Fix matching requests made from browsers in WiFi examples

pzygielo:
## Description of Change
I propose to change matching the request line to determine if it is `GET` for `H` or `L`.

## Tests scenarios
With example running - when accessed via HTTP client - currently the request is not recognized as `GET /(H|L)` due to additional content in line (like HTTP protocol version) _following_ the method.

With this fix it's possible to interact with the example as advertised:
https://github.com/espressif/arduino-esp32/blob/f9bb62a91fc5b5886ab5188f56fbca40a63b1464/libraries/WiFi/examples/SimpleWiFiServer/SimpleWiFiServer.ino#L6

## Related links
- https://www.rfc-editor.org/rfc/rfc9110.html#section-3.9


pzygielo:
> * footer must have leading blank line

https://github.com/conventional-changelog/commitlint/issues/896?

P-R-O-C-H-Y:
Hello @pzygielo, I have tested the example, and for me it's working with both versions. With the current code and with your changes the same. Even though in the header these is another content like the HTTP, the current line is headed by each char, so at the end with your change, it will trigger the ON/FF multiple times.

```
13:47:03.219 -> G
13:47:03.219 -> GE
13:47:03.219 -> GET
13:47:03.219 -> GET 
13:47:03.219 -> GET /
13:47:03.219 -> GET /H
13:47:03.219 -> GET /H 
13:47:03.219 -> GET /H H
13:47:03.219 -> GET /H HT
13:47:03.219 -> GET /H HTT
13:47:03.219 -> GET /H HTTP
13:47:03.219 -> GET /H HTTP/
13:47:03.219 -> GET /H HTTP/1
13:47:03.219 -> GET /H HTTP/1.
13:47:03.219 -> GET /H HTTP/1.1
```
We can close this PR, as it won't be probably merged. @me-no-dev

pzygielo:
> I have tested the example

Just to make sure - with the (real) browser?

P-R-O-C-H-Y:
> > I have tested the example
> 
> Just to make sure - with the (real) browser?

Yes, and what I posted is the log from the ESP32, when I added a line to print `currentLine` before the check.

pzygielo:
I'll re-try the example on my side again then. Thanks for checking.

pzygielo:
> with [this] change, it will trigger the ON/FF multiple times

Yeah... that would be terrible.


I can't explain/remember the error I was getting initially (in January).

But I confirm that now (w/o my change it is), on the next try - everything works as expected with both examples. :tada:

Issue #6102, opened 2022-01-06
how to use esp32-hal in arduino,example esp32-hal-uart.h  for serial interrupt  in arduino?


zouzhe1:
### Board

esp32

### Device Description

esp32

### Hardware Configuration

esp32

### Version

latest master

### IDE Name

platformio

### Operating System

win10

### Flash frequency

40

### PSRAM enabled

no

### Upload speed

9600

### Description

i want to use serial hardware interrupt in arduino,but i found the arduino api _serialEvent()_ is not really hardware serial interrupt.so,i fonud some in _.platformio\packages\framework-arduinoespressif32\cores\esp32_   ,**esp32-hal-uart.c** and **esp32-hal-uart.h**
i think this two file can help me. but i dont find any help document about this HAL file.

//example this function
uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rxPin, int8_t txPin, uint16_t queueLen, bool inverted);

what uart_nr?   what config?  what queueLen?i dontknow rev data len. what inverted?

### Sketch

```cpp
//example this function
uart_t* uartBegin(uint8_t uart_nr, uint32_t baudrate, uint32_t config, int8_t rxPin, int8_t txPin, uint16_t queueLen, bool inverted);

what uart_nr?   what config?  what queueLen?i dontknow rev data len. what inverted?
```

### Debug Message

```plain
....
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

SuGlider:
Hi @zouzhe1 
Currently Arduino Core version 2.0.2 uses UART IRQ to fill up buffer in Arduino Serial. I guess that what you are looking is some sort of a callback function that will react immediately when a byte arrives in the UART - which you describe as Serial Interrupt.

This can be achieved using a Free RTOS Task that keeps waiting for an event to arrive in its queue, based on IDF call uart_driver_install(), as described in [IDF Ref - uart_driver_install](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/uart.html#_CPPv419uart_driver_install11uart_port_tiiiP13QueueHandle_ti)

[...]

PR #9415, opened 2024-03-25
Adds a new example: onReceiveExample.ino

SuGlider:
Adds a new example that uses HardwareSerial::onReceive(). The example demosntrates how to read all the data sent to UART0, considering that the end of transmission is defined by a period of time with UART in idle state.

## Description of Change
This PR adds a new example that uses HardwareSerial::onReceive().
The example demosntrates how to read all the data sent to UART0, considering that the end of transmission is defined by a period of time with UART in idle state.

This is based on a question from a user.

## Tests scenarios
Tested using the code from the example with the ESP32, ESP32-S3 and ESP32-C3.

## Related links
https://github.com/espressif/arduino-esp32/issues/6102#issuecomment-2009207450

### DIFF
--- a/libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
+++ b/libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
@@ -1,3 +1,54 @@
+/*
+
+   This Sketch demonstrates how to use onReceive(callbackFunc) with HardwareSerial
+
+   void HardwareSerial::onReceive(OnReceiveCb function, bool onlyOnTimeout = false)
+
+   It is possible to register an UART callback function that will be called
+   every time that UART receives data and an associated interrupt is generated.
+
+   In summary, HardwareSerial::onReceive() works like an RX Interrupt callback, that can be adjusted
+   using HardwareSerial::setRxFIFOFull() and HardwareSerial::setRxTimeout().
+
+   OnReceive will be called, while receiving a stream of data, when every 120 bytes are received (default FIFO Full),
+   which may not help in case that the application needs to get all data at once before processing it.
+   Therefore, a way to make it work is by detecting the end of a stream transmission. This can be based on a protocol
+   or based on timeout with the UART line in idle (no data received - this is the case of this example). 
+
+   In some cases, it is necessary to wait for receiving all the data before processing it and parsing the
+   UART input. This example demonstrates a way to create a String with all data received from UART0 and
+   signaling it using a Mutex for another task to process it. This example uses a timeout of 500ms as a way to
+   know when the reception of data has finished.
+
+   The onReceive() callback is called whenever the RX ISR is triggered.
+   It can occur because of two possible events:
+
+   1- UART FIFO FULL: it happens when internal UART FIFO reaches a certain number of bytes.
+      Its full capacity is 127 bytes. The FIFO Full threshold for the interrupt can be changed
+      using HardwareSerial::setRxFIFOFull(uint8_t fifoFull).
+      Default FIFO Full Threshold is set in the UART initialization using HardwareSerial::begin()
+      This will depend on the baud rate used when begin() is executed.
+      For a baud rate of 115200 or lower, it it just 1 byte, mimicking original Arduino UART driver.
+      For a baud rate over 115200 it will be 120 bytes for higher performance.
+      Anyway, it can be changed by the application at any time.
+
+   2- UART RX Timeout: it happens, based on a timeout equivalent to a number of symbols at
+      the current baud rate. If the UART line is idle for this timeout, it will raise an interrupt.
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y (approved):
LGTM! Tested on ESP32S2 👍

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
lucasssvaz, line 1: Please add a header to the beginning of the file explaining what this is supposed to do.

lucasssvaz (commented):
LGTM. Just a small suggestion.

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -0,0 +1,65 @@
...
+      UART0.print("[");
+      UART0.print(counter++);
+      UART0.print("] [");
+      UART0.print(uart_buffer.length());
+      UART0.print(" bytes] ");
+      UART0.println(uart_buffer);
+      uart_buffer = "";  // reset uart_buffer for the next UART reading
+    }
+    // releases the mutex for more data to be received
+    xSemaphoreGive(uart_buffer_Mutex);
```
me-no-dev, line 61: you should give the semaphore only if you have taken it. Move that line in the `if`

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
SuGlider: done

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
SuGlider: correct! done. thanks!

lucasssvaz (approved):

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -70,7 +70,7 @@ SemaphoreHandle_t uart_buffer_Mutex = NULL;
 // task created when onReceive() is used
 void UART0_RX_CB() {
   // take the mutex, waits forever until loop() finishes its processing
-  if (xSemaphoreTake(uart_buffer_Mutex, portMAX_DELAY)) {
+  if (uart_buffer_Mutex != NULL && xSemaphoreTake(uart_buffer_Mutex, portMAX_DELAY)) {
```
lucasssvaz, line 73: I think this will fail either way as `xSemaphoreTake` will be called with NULL pointer. I think it's better to lock the sketch if the creation of the mutex fails

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// setup() and loop() are functions executed by a low priority task
+// Therefore, there are 2 tasks running when using onReceive()
+void setup() {
+  // creates a mutex object to control access to uart_buffer
+  uart_buffer_Mutex = xSemaphoreCreateMutex();
+  if(uart_buffer_Mutex == NULL) {
+     log_e("Error creating Mutex. Sketch will fail.");
+     delay(1000);
```
lucasssvaz, line 93:
```suggestion
     while(1);
```

me-no-dev: if it should wait forever, do not `while(1);` on ESPs or you might/should trigger WDT. `while(1) delay(1000);` is much better @lucasssvaz .

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
me-no-dev: it will not call `xSemaphoreTake` is `uart_buffer_Mutex` is `NULL`.

lucasssvaz: Yes, my bad. I forgot about the short circuit evaluation.

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
SuGlider: Ok. done. I also always think that most users don't turn on DEBUG, so as an example, I also print to a regular message.
Fixed.

SuGlider: After this change the mutex won't be NULL. ==> Removing the test for NULL.

SuGlider:
@lucasssvaz @me-no-dev - it shall be ready now.

lucasssvaz (approved):

P-R-O-C-H-Y (approved):

Issue #9082, opened 2024-01-09
esp32c3 SPI SD can't not work 


TasLR:
### Board

esp32-c3-wroom-02

### Device Description

esp32-c3-wroom-02-N4

### Hardware Configuration

#define PIN_NUM_MISO  10
#define PIN_NUM_MOSI  19
#define PIN_NUM_CLK   18
#define PIN_NUM_CS    3

### Version

v2.0.14

### IDE Name

platformIO

### Operating System

windows10

### Flash frequency

40

### PSRAM enabled

yes

### Upload speed

115200

### Description

I want to run spi sd test with esp32c3, but It can't work under arduino framework,and It can work under the esp idf framework. I use the same ports on these example. under arduino framework, the log is sd drive fail, can you tell how to do slove this problem? thank you !

### Sketch

```cpp
/*
 * pin 1 - not used          |  Micro SD card     |
 * pin 2 - CS (SS)           |                   /
 * pin 3 - DI (MOSI)         |                  |__
 * pin 4 - VDD (3.3V)        |                    |
 * pin 5 - SCK (SCLK)        | 8 7 6 5 4 3 2 1   /
 * pin 6 - VSS (GND)         | ▄ ▄ ▄ ▄ ▄ ▄ ▄ ▄  /
 * pin 7 - DO (MISO)         | ? ? █ ? █ ? ? ? |
 * pin 8 - not used          |_________________|
 *                             ║ ║ ║ ║ ║ ║ ║ ║
 *                     ╔═══════╝ ║ ║ ║ ║ ║ ║ ╚═════════╗
 *                     ║         ║ ║ ║ ║ ║ ╚══════╗    ║
 *                     ║   ╔═════╝ ║ ║ ║ ╚═════╗  ║    ║
 * Connections for     ║   ║   ╔═══╩═║═║═══╗   ║  ║    ║
 * full-sized          ║   ║   ║   ╔═╝ ║   ║   ║  ║    ║
 * SD card             ║   ║   ║   ║   ║   ║   ║  ║    ║
 * Pin name         |  -  DO  VSS SCK VDD VSS DI CS    -  |
 * SD pin number    |  8   7   6   5   4   3   2   1   9 /
 *                  |                                  █/
 *                  |__▍___▊___█___█___█___█___█___█___/
 *
 * Note:  The SPI pins can be manually configured by using `SPI.begin(sck, miso, mosi, cs).`
[...]

PR #9422, opened 2024-03-26
Matched pin numbers in doco to numbers in variants files

lbernstone:
## Description of Change
Matched pin numbers in doco to numbers in variants files.  This should make it easier to use the examples as is. Also added entries for the new variants. Changed SD examples to make REASSIGN_PINS more transparent. Fixes https://github.com/espressif/arduino-esp32/issues/9082

## Tests scenarios
Tested on esp32-c3.


P-R-O-C-H-Y (approved):
LGTM!

PR #9427, opened 2024-03-28
Handle large octet-stream

nathannau:
When body of request is too large (like many application/octet-stream), RAM is too small and malloc fail.

Body can be handle by many callback call like with upload file.

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -34,7 +34,7 @@
 #define _STR(a) __STR(a)
 const char * _http_method_str[] = {
 #define XX(num, name, string) _STR(name),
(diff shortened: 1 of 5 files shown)


lucasssvaz:
please remove the formatting changes of the PR

nathannau:
> please remove the formatting changes of the PR

Sorry, I couldn't find the clear rules for formatting as a Prettier.
Where can I find them, please ?

me-no-dev:
@nathannau even if you do not find rules, that does not mean that it's OK to change the formatting of multiple files. Now we can not even see what your real changes are. Please re-submit the PR with only the changes you need to have this feature implemented.

nathannau:
Sorry about that. My IDE hides indentation changes. 
I'll clean it up.

PR #9430, opened 2024-03-28
Handle large octet-stream

nathannau:
### Issue

In scenarios where the request body size exceeds the available RAM, the ESP fails to allocate sufficient memory to process the request in one go.

### Change

Now, when the "Content-Type" of the request is "application/octet-stream", it will attempt to call the "ufn" function of the handler, similar to how it handles multipart requests.

**Code Example:**

```cpp
server.on(uri, HTTP_PUT, finalCallback, [](){
  HTTPRaw raw = server.raw();
  if (raw.status == RAW_START) {
    ...
  } else if (raw.status == RAW_WRITE) {
    ...
  } else if (raw.status == RAW_END) {
    ...
  }
});
```

### DIFF
--- a/libraries/WebServer/examples/UploadHugeFile/UploadHugeFile.ino
+++ b/libraries/WebServer/examples/UploadHugeFile/UploadHugeFile.ino
@@ -1,0 +1,132 @@
+#include <WiFi.h>
+#include <WiFiClient.h>
+#include <WebServer.h>
+#include <uri/UriRegex.h>
+#include <SD.h>
+
+const char* ssid = "**********";
+const char* password = "**********";
+
+WebServer server(80);
+
+File rawFile;
+void handleCreate() {
+  server.send(200, "text/plain", "");
+}
+void handleCreateProcess() {
+  String path = server.pathArg(0);
+  HTTPRaw& raw = server.raw();
+  if (raw.status == RAW_START) {
+    if (SD.exists((char *)path.c_str())) {
+      SD.remove((char *)path.c_str());
+    }
+    rawFile = SD.open(path.c_str(), FILE_WRITE);
+    Serial.print("Upload: START, filename: "); 
+    Serial.println(path);
+  } else if (raw.status == RAW_WRITE) {
+    if (rawFile) {
+      rawFile.write(raw.buf, raw.currentSize);
+    }
+    Serial.print("Upload: WRITE, Bytes: "); 
+    Serial.println(raw.currentSize);
+  } else if (raw.status == RAW_END) {
+    if (rawFile) {
+      rawFile.close();
+    }
+    Serial.print("Upload: END, Size: "); 
+    Serial.println(raw.totalSize);
+  }
+}
+
+void returnFail(String msg) {
+  server.send(500, "text/plain", msg + "\r\n");
+}
+
+void printDirectory() {
+  if (!server.hasArg("dir")) {
+    return returnFail("BAD ARGS");
+  }
+  String path = server.arg("dir");
+  if (path != "/" && !SD.exists((char *)path.c_str())) {
+    return returnFail("BAD PATH");
+  }
+  File dir = SD.open((char *)path.c_str());
+  path = String();
+  if (!dir.isDirectory()) {
+    dir.close();
+    return returnFail("NOT DIR");
+  }
+  dir.rewindDirectory();
+  server.setContentLength(CONTENT_LENGTH_UNKNOWN);
+  server.send(200, "text/json", "");
+  WiFiClient client = server.client();
+
+  server.sendContent("[");
+  for (int cnt = 0; true; ++cnt) {
+    File entry = dir.openNextFile();
+    if (!entry) {
+      break;
+    }
+
+    String output;
+    if (cnt > 0) {
+      output = ',';
+    }
+
+    output += "{\"type\":\"";
+    output += (entry.isDirectory()) ? "dir" : "file";
+    output += "\",\"name\":\"";
+    output += entry.path();
+    output += "\"";
+    output += "}";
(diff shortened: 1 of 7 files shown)


me-no-dev:
This is much better now :)

me-no-dev:
BTW I see that you target the 2.x branch. Do you plan to make PR to master also? Else this feature would exist only on 2.0.x versions

nathannau:
It would be nice to do the same PR for the master, but I'm using PlatformIO and I can't target the master. And I didn't want to work on the master if this PR wasn't validated. 😄

me-no-dev:
Code looks good. Can you please add slightly better description and also a small example to show how this feature can be used.

nathannau:
I added an exemple and improved the description.

me-no-dev:
@P-R-O-C-H-Y could you give this a shot and also evaluate the port to 3.0.0?

me-no-dev:
because it's changing the current behavior, I lean towards having this for 3.0.0 and not for 2.0.15. We do not want to break people's code in a bugfix release. Major v3 is a much better fit

me-no-dev:
@Jason2866 could you help @nathannau try this on 3.0.0 and PIO?

Jason2866:
@nathannau  To test a PR with Platformio against latest master. -> Make a branch with your changes from branch master and do the following setup in Platformio
```
[env:esp32]
board = esp-wrover-kit
platform = https://github.com/platformio/platform-espressif32.git
platform_packages =
	platformio/framework-arduinoespressif32 @ https://github.com/nathannau/arduino-esp32.git#your_branch_name
	platformio/framework-arduinoespressif32-libs @ https://github.com/espressif/esp32-arduino-libs.git#idf-release/v5.1
```

nathannau:
I'm currently testing whether the "Content-Type" is "Application/octet-stream".
It might be better to simply test that it's not a form. This would make it possible to generalise the processing of request bodies as you go along, when you don't want to load everything into memory.

me-no-dev:
@nathannau it's important that current behavior is not changed much for the average folk. Simple request should go fine as they do right now and use raw only when wanted to.

nathannau:
@me-no-dev  Exactly. This patch would retain backwards compatibility while adding the ability to do more things if needed.

libraries/WebServer/examples/UploadHugeFile/UploadHugeFile.ino
```diff
@@ -0,0 +1,132 @@
...
+    output += "\",\"name\":\"";
+    output += entry.path();
+    output += "\"";
+    output += "}";
+    server.sendContent(output);
+    entry.close();
+  }
+  server.sendContent("]");
+  dir.close();
+}
```
lucasssvaz, line 87: Function not used anywhere in code
```suggestion
```

libraries/WebServer/examples/UploadHugeFile/UploadHugeFile.ino
```diff
@@ -0,0 +1,132 @@
...
+const char* password = "**********";
+
+WebServer server(80);
+
+File rawFile;
+void handleCreate() {
+  server.send(200, "text/plain", "");
+}
+void handleCreateProcess() {
+  String path = server.pathArg(0);
```
lucasssvaz, line 17: Open fails and nothing is written to the SD card.

```suggestion
  String path = "/" + server.pathArg(0);
```

lucasssvaz (changes requested):

lucasssvaz (changes requested):
Same as 2.x

P-R-O-C-H-Y (commented):
Tested with @lucasssvaz fix which is needed. After the fix we can merge.

me-no-dev:
@nathannau this is currently blocking 2.0.15 release. Is there any chance that you take care of the notes from @lucasssvaz ? If you do not have the time, we could do that for you as well and merge the PR

libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -53,6 +59,13 @@ class FunctionRequestHandler : public RequestHandler {
             _ufn();
     }
 
+    void raw(WebServer& server, String requestUri, HTTPRaw& raw) override {
+        (void)server;
+        (void)raw;
+        if (canRaw(requestUri))
+            _ufn();
```
lucasssvaz, line 66:
```suggestion
        if (canRaw(requestUri)) {
            _ufn();
        }
```

libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -36,6 +36,12 @@ class FunctionRequestHandler : public RequestHandler {
 
         return true;
     }
+    bool canRaw(String requestUri) override {
+        if (!_ufn || _method == HTTP_GET)
+            return false;
```
lucasssvaz, line 41:
```suggestion
        if (!_ufn || _method == HTTP_GET) {
            return false;
        }
```

nathannau: Why ? The rest of the code does not have this formatting.

libraries/WebServer/src/detail/RequestHandlersImpl.h
nathannau: Why ? The rest of the code does not have this formatting.

lucasssvaz: It's better to avoid bracketless ifs. The rest of the code is old and shouldn't be like this. Either way all the code base will be reformatted soon and all short ifs will be removed.

P-R-O-C-H-Y (approved):

PR #9434, opened 2024-03-29
WiFi BSSID getter fix

JAndrassy:
I have tests as examples in the NetAPIHelpers library. You can run them too to avoid these kind of simple bugs.
The NetAPIHelpers library is in the Library Manager.


me-no-dev:
yup :) nice catch

PR #9440, opened 2024-03-31
Handle large octet-stream (master branch)

nathannau:
### Issue

In scenarios where the request body size exceeds the available RAM, the ESP fails to allocate sufficient memory to process the request in one go.

### Change

Now, when the "Content-Type" of the request is "application/octet-stream", it will attempt to call the "ufn" function of the handler, similar to how it handles multipart requests.

**Code Example:**

```cpp
server.on(uri, HTTP_PUT, finalCallback, [](){
  HTTPRaw raw = server.raw();
  if (raw.status == RAW_START) {
    ...
  } else if (raw.status == RAW_WRITE) {
    ...
  } else if (raw.status == RAW_END) {
    ...
  }
});
```

### DIFF
--- a/libraries/WebServer/src/detail/RequestHandlersImpl.h
+++ b/libraries/WebServer/src/detail/RequestHandlersImpl.h
@@ -38,6 +38,12 @@ public:
 
         return true;
     }
+    bool canRaw(String requestUri) override {
+        if (!_ufn || _method == HTTP_GET)
+            return false;
+
+        return true;
+    }
 
     bool handle(WebServer& server, HTTPMethod requestMethod, String requestUri) override {
         (void) server;
@@ -52,6 +58,13 @@ public:
         (void) server;
         (void) upload;
         if (canUpload(requestUri))
+            _ufn();
+    }
+
+    void raw(WebServer& server, String requestUri, HTTPRaw& raw) override {
+        (void)server;
+        (void)raw;
+        if (canRaw(requestUri))
             _ufn();
     }
 
--- a/libraries/WebServer/examples/UploadHugeFile/README.md
+++ b/libraries/WebServer/examples/UploadHugeFile/README.md
@@ -1,0 +1,13 @@
+# Upload Huge File To SD Over Http
+
+This project is an example of an HTTP server designed to facilitate the transfer of large files using the PUT method, in accordance with RFC specifications.
+
+### Example cURL Command
+
+```bash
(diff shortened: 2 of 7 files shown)


lucasssvaz:
@nathannau Please pull the commits and don't force push over them. I had already fixed the CI errors for the example.

lucasssvaz:
Following the instruction on the readme causes the error `curl: (55) Send failure: Broken pipe`.

lucasssvaz (changes requested):
Example doesn't work as expected: `curl: (55) Send failure: Broken pipe`

me-no-dev:
in the example folder you need to add an empty file with name `.skip.esp32h2`

lucasssvaz (changes requested):

libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -38,6 +38,12 @@ class FunctionRequestHandler : public RequestHandler {
 
         return true;
     }
+    bool canRaw(String requestUri) override {
+        if (!_ufn || _method == HTTP_GET)
+            return false;
```
lucasssvaz, line 43:
```suggestion
        if (!_ufn || _method == HTTP_GET) {
            return false;
        }
```

libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -55,6 +61,13 @@ class FunctionRequestHandler : public RequestHandler {
             _ufn();
     }
 
+    void raw(WebServer& server, String requestUri, HTTPRaw& raw) override {
+        (void)server;
+        (void)raw;
+        if (canRaw(requestUri))
+            _ufn();
```
lucasssvaz, line 68:
```suggestion
        if (canRaw(requestUri)) {
            _ufn();
        }
```

libraries/WebServer/src/detail/RequestHandlersImpl.h
me-no-dev: What's the problem with that request @nathannau ?

lucasssvaz (changes requested):
Please add an empty `.skip.esp32h2` file alongside your example

P-R-O-C-H-Y (approved):
Working as expected, tested also SDWebServer example with no issues.
@nathannau  Please add the skip file for H2. Thanks

Issue #9435, opened 2024-03-29
WiFi static IP configuration doesn't work after networking refactoring


JAndrassy:
### Description

WiFi doesn't use static IP configuration after the networking refactoring. Settings are not applied in NetworkInterface.config because esp_netif is NULL. In `begin` then DHCP is used.
https://github.com/espressif/arduino-esp32/blob/7b29bac625ee63f5140a5468cbbd7368879dec16/libraries/Network/src/NetworkInterface.cpp#L330

### Sketch

WiFiClientStaticIP example

---

If I first `begin` with DHCP then setting static IP works for next `begin`. But following `begin` again uses DHCP. This is not right. Already configured static IP configuration has to be used. Static IP should be only cleared with `config(INADDR_NONE)` (or `end()`).  All WiFi libraries have it that way including the esp8266 WiFi.
The static IP configuration is cleared because `ESP_NETIF_HAS_IP_BIT` is cleared so `config()` with default parameters in invoked.
https://github.com/espressif/arduino-esp32/blob/7b29bac625ee63f5140a5468cbbd7368879dec16/libraries/WiFi/src/STA.cpp#L354.

me-no-dev:
@JAndrassy please try https://github.com/espressif/arduino-esp32/pull/9445

JAndrassy:
output from my test
```
ttempting to connect to SSID "M20" with static IP 192.168.1.177
.
	...success

MAC: 30:AE:A4:08:B2:70
SSID: M20
BSSID: C0:25:E9:C3:2F:8A
RSSI:-56 dBm
IP Address: 192.168.1.177
gateway IP Address: 192.168.1.1
subnet IP mask: 255.255.255.0
DNS server: 192.168.1.1

Attempt to connect to port 80 on 192.168.1.1
	...success

[  2890][W][STA.cpp:157] _onStaArduinoEvent(): Reason: 8 - ASSOC_LEAVE
Attempting to connect without resetting static IP configuration

	...success

MAC: 30:AE:A4:08:B2:70
SSID: 
BSSID: 00:00:00:00:00:00
RSSI:0 dBm
IP Address: 192.168.1.177
gateway IP Address: 192.168.1.1
subnet IP mask: 255.255.255.0
DNS server: 192.168.1.1
DNS server2: 8.8.8.8

Attempt to connect to port 80 on 192.168.1.1
[...]

PR #9445, opened 2024-04-02
Ensure that Static IP configuration for network interfaces is kept until STOP

me-no-dev:
Closes: https://github.com/espressif/arduino-esp32/issues/9435


lucasssvaz (approved):
Works as expected. Tested with the Static IP example. Makes it work again after network refactoring.

PR #9446, opened 2024-04-02
Set Apache license to Update Library

SuGlider:
## Description of Change
Changes the Update Library license to Apache, fillowing all the other Espressif source code.

## Tests scenarios
N/A

## Related links
Internal.

### DIFF
--- a/libraries/Update/src/HttpsOTAUpdate.cpp
+++ b/libraries/Update/src/HttpsOTAUpdate.cpp
@@ -1,6 +1,5 @@
-/* OTA task
-
-   This example code is in the Public Domain (or CC0 licensed, at your option.)
+/* 
+   This code is in the Public Domain (or CC0 licensed, at your option.)
 
    Unless required by applicable law or agreed to in writing, this
    software is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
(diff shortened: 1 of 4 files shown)


SuGlider:
@me-no-dev - please take a look and let me know if you agree to this change.

libraries/Update/src/HttpsOTAUpdate.cpp
```diff
@@ -1,11 +1,9 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
 
-   This example code is in the Public Domain (or CC0 licensed, at your option.)
-
-   Unless required by applicable law or agreed to in writing, this
-   software is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
-   CONDITIONS OF ANY KIND, either express or implied.
-*/
```
me-no-dev, line 8: revert this change please.

SuGlider: @me-no-dev - reverted.
HttpsOTAUpdate.cpp/.h are marked as CC0
Update.h/Updater.cpp are marked as Apache-2.0

Is it correct?

me-no-dev: I wrote Updater, not HttpsOTAUpdate, so the user who wrote it, put that license there :)

PR #9447, opened 2024-04-02
core Server class back to Arduino Core API version

JAndrassy:
the Server class is now not used in NetworkServer. it is in the core only for external libraries like the Arduino Ethernet library, so it should be compatible with them. (optionality of the parameter doesn't apply when matching implementation of virtual function )

https://github.com/arduino-libraries/Ethernet/issues/88

https://github.com/arduino/ArduinoCore-API/blob/master/api/Server.h

### DIFF
--- a/cores/esp32/Server.h
+++ b/cores/esp32/Server.h
@@ -25,7 +25,7 @@
 class Server: public Print
 {
 public:
-    virtual void begin(uint16_t port=0) =0;
+    virtual void begin() =0;
 };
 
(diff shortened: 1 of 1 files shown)


lucasssvaz (approved):
LGTM. Is there any problem in going back to the Arduino API @me-no-dev ?

me-no-dev:
@lucasssvaz we are not extending it anyway, so it's fine

PR #9453, opened 2024-04-03
WiFiClient - rename flush() to clear()

JAndrassy:
this is a repeat of https://github.com/espressif/arduino-esp32/pull/8871 after networking refactoring.

With a major version release the `flush()` dilemma could be resolved for WiFiClient & co. as last on this platform.

I renamed the flush which cleared the rx buffer to `clear()`. but flush is pure virtual in Client so I had to add empty `flush()`
The position of the function in file is to group Print (write) related and Stream (read) related functions.

If Arduino adds to Stream class a method to clear the input buffer, it will be most probably named clear() because Stream is taken from Processing and there it has clear().

I [analyzed the API](https://github.com/JAndrassy/Arduino-Networking-API/blob/main/ArduinoNetAPILibs.md) of 18 Arduino networking libraries. ESP32 is the only one with wrong implementation of flush().
There is a very bad consequence of using rx clearing flush() instead of proper flush(). It means RX data loss.

### DIFF
--- a/docs/en/migration_guides/2.x_to_3.0.rst
+++ b/docs/en/migration_guides/2.x_to_3.0.rst
@@ -197,3 +197,11 @@ Functional changes
 * Any pin set as -1 in ``begin()`` or ``setPins()`` won't be changed nor detached.
 * ``begin(baud)`` will not change any pins that have been set before this call, through a previous ``begin(baud, rx, tx)`` or ``setPin()``.
 * If the application only uses RX or TX, ``begin(baud, -1, tx)`` or ``begin(baud, rx)`` will change only the assigned pin and keep the other unchanged.
+
+WiFi
+----
+
+Functional changes
+******************
+
+* In Arduino (and other frameworks) the method named ``flush()`` is intended to send out the transmit buffer content. ``WiFiClient`` and ``WiFiUDP`` method ``flush()`` won't clear the receive buffer anymore. A new method called ``clear()`` is now used for that. Currently ``flush()`` does nothing in ``WiFiClient``, ``WiFiClientSecure`` and ``WiFiUDP``.
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -159,7 +159,7 @@ public:
         return _fill - _pos + r_available();
     }
 
-    void flush(){
+    void clear(){
         if(r_available()){
             fillBuffer();
         }
@@ -391,6 +391,11 @@ int NetworkClient::read()
     return data;
 }
 
+void NetworkClient::flush()
+{
+
+}
+
 size_t NetworkClient::write(const uint8_t *buf, size_t size)
 {
     int res =0;
@@ -534,11 +539,9 @@ int NetworkClient::available()
     return res;
 }
 
-// Though flushing means to send all pending data,
-// seems that in Arduino it also means to clear RX
-void NetworkClient::flush() {
+void NetworkClient::clear() {
     if (_rxBuffer != nullptr) {
-        _rxBuffer->flush();
+        _rxBuffer->clear();
     }
 }
 
--- a/libraries/Network/src/NetworkClient.h
+++ b/libraries/Network/src/NetworkClient.h
@@ -59,11 +59,12 @@ public:
     size_t write(const uint8_t *buf, size_t size);
     size_t write_P(PGM_P buf, size_t size);
     size_t write(Stream &stream);
+    void flush(); // Print::flush tx
     int available();
     int read();
     int read(uint8_t *buf, size_t size);
     int peek();
-    void flush();
+    void clear(); // clear rx
     void stop();
     uint8_t connected();
     void setSSE(bool sse);
--- a/libraries/Network/src/NetworkUdp.cpp
+++ b/libraries/Network/src/NetworkUdp.cpp
@@ -288,6 +288,11 @@ size_t NetworkUDP::write(const uint8_t *buffer, size_t size){
   return i;
 }
 
+void NetworkUDP::flush()
+{
+
+}
+
 int NetworkUDP::parsePacket(){
   if(rx_buffer)
     return 0;
@@ -374,7 +379,7 @@ int NetworkUDP::peek(){
   return rx_buffer->peek();
 }
 
-void NetworkUDP::flush(){
+void NetworkUDP::clear(){
   if(!rx_buffer) return;
   cbuf *b = rx_buffer;
   rx_buffer = 0;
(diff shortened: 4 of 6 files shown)


lucasssvaz (changes requested):
Looks reasonable to me. IMO it just needs a better wording on the documentation. @me-no-dev Do you think we can add this to the breaking changes of 3.0.0 ?

lucasssvaz (approved):

TD-er:
There should have been a compile-time warning whenever those now empty `flush()` calls are being called.
Now we have to go through all code to see what has been **broken** by this change (and yes it does have consequences, this change)

So for any similar changes in future PRs, please make compile time warnings whenever those are being called as it is obviously a legacy mistake when you make a call to a function which is now empty and did do stuff before.
Those warnings can be temporary until the new major release of the SDK is being released, but it would prevent lots of lost hours in searching why stuff is now no longer working like before.

For example in my project I have searched for **countless hours** to see why some of my boards suddenly refused to respond to new calls via the webserver and no idea what caused this as it was nearly impossible to reproduce.

JAndrassy:
@TD-er 
the unread chars were in the way?
you can add the warning in your installation of the platform, until you resolve the occurrences

TD-er:
In my specific use case I have a wrapper for handling chunked transfers when sending replies to HTTP requests.
And this used to work very well, also when handling several quick requests in a row like the browser asking for the CSS while loading the page.
However with updating my code to comply with this change (I changed the code to use clear() instead of flush() for the webserver.client()) it is now working as it used to do.

Without my change, the device would appear to be inaccessible for quite a while to just about any HTTP request when it had to serve multiple requests in a burst.

The obvious symptoms were:
- CSS not loading
- JavaScript not loading
- JSON's were not fetched.

But my main issue with this is that it is a very significant change, which should have had some compile-time warnings about the API changes.
So for other similar changes, please consider adding such warnings as this really has taken quite a lot of time for me to search for the cause as I really had not thought of this change.

JAndrassy:
in this case it was not possible to add a warning as the flush() function is a valid function (pure virtual in the Client class) and it is not possible to know when it is used wrong

TD-er:
@JAndrassy 
Just to let you know the utter frustration I have from this PR being merged as it keeps biting me over and over again.

> in this case it was not possible to add a warning as the flush() function is a valid function (pure virtual in the Client class) and it is not possible to know when it is used wrong

**This** is exactly why it is a real pain in the *ss to have this PR being merged.

You simply don't know when it is being used wrong, not at compile time, not at runtime. You only keep seeing those symptoms.

In the platform packages provided by Jason for me to be able to implement support for ESP-IDF5.x this PR was reverted as it is nearly impossible to find all occurences where this is used incorrectly.
However when using 'vanilla' Arduino code I keep running into these really _really_ annoying issues caused by this PR where you run into vague and hard to reproduce timeouts, failed connections, hanging device as clients used up all available connections, stalls, etc.

I do understand that maybe in the past there may have been a poor choice in function naming. But the way this is now 'fixed' is NOT how it should have done as it is simply breaking stuff and you can't tell where it is all broken unless spending countless hours fixing stuff and with any new lib that may be relying on the older implementation you can do it all over again.
And worst of all as I need to keep multiple versions of the SDK compilable it really messes up the code which should not even have to worry about this at all.

All those hours needlessly spent on these kind of issues cannot be spent to make sure I can get rid of those older SDK requirements or even implement nice new features.

TD-er:
IMHO the best way forward would be to generally revert this PR and if there really is a need to make similar changes, there should be found a way to make it work so devs can see at compile-time what needs to be changed.

JAndrassy:
why is it hard to find occurrences of "flush()" in the source code?

TD-er:
You just can't replace `flush()` with `clear()` in the code.
If it was as simple like that, then there wouldn't be an empty flush() function in the code, right?
There is apparently a need for both functions in some base class.

As I also need to have my code to work with SDK versions not using this PR.
So this would then result in some global define where based on the platform code you define it either as `flush()` or `clear()`.
This can also be required in library code, which must then be adapted too and made sure to keep those changes when updating such a library.

So this is only enforcing the point I tried to make before as why this PR is causing a mess for developers.
You need to keep track of stuff where you should not need to worry about it.

TD-er:
While you're at it, can you also take a look at these:

https://github.com/espressif/arduino-esp32/blob/cd3d0bf569f8c0ec1d9ed549c649b9d237d10e28/libraries/WebServer/src/Parsing.cpp#L265

https://github.com/espressif/arduino-esp32/blob/cd3d0bf569f8c0ec1d9ed549c649b9d237d10e28/libraries/ArduinoOTA/src/ArduinoOTA.cpp#L379

That's exactly in the area where multiple reports have been filed lately, so I guess it might be related.

JAndrassy:
in Parsing.cpp the client.flush() is at the end of the processing of the HTTP request. the response is already sent. If the connection is not kept alive for the next request, then it is closed so why clearing it. If the connection is kept alive for the following HTTP request, then clear() could delete beginning of the following request request. 

In ArduinoOTA the UDP object is used to receive and send messages. The function `_onRx` (invoked before that flush()) reads a message and then sends a response. So it is unclear what the flush() was meant to do. If espota.py uploads work then everything is OK. But it would fail if _onRx doesn't read the whole incoming message because then all following parsePacket will always return 0 (and this is not a good behavior of the NetworkUDP class).

TD-er:
One of the git blame of these 2 was about a commit of 6+ years ago where the last commit text mentioned something about it being added intentionally. (some commit inbetween about code formatting)
I'm on an extremely slow laptop right now, so I can't get these git blames to load here.

Regardless if it might be some code smell of a problem elsewhere, it does seem like those 2 had to be changed too when this PR was merged to keep at least the functionality the same.
Do you agree those two occurences should be changed to .clear() too?
If so, then I will make a PR (or you can do one) later this evening.

Jason2866:
Still the best solution would to revert this PR completely. The regressions caused by this PR are hughe. Even two follow up PRs are already needed to fix the core itself! How many will be needed at the just for the core itself?
The situation is worse with the many many libs out there. The are all incompatible now.
Please @me-no-dev revert this PR!!

me-no-dev:
How about instead of reverting it, we have `flush()` to call `clear()`? Would that help you all?

TD-er:
I think that's a nice compromise.
Maybe also add some deprecated attribute if possible, so whenever it is used the compiler will warn?

me-no-dev:
PR with clear and deprecation created: https://github.com/espressif/arduino-esp32/pull/10242

Jason2866:
@me-no-dev Good solution, no silent breaking change anymore. Now there is an info that a mandatory change has be done and still all code is working as expected.

PR #9484, opened 2024-04-11
LWIP_PPP: Fix compilation 

Jason2866:
when no PPP support is enabled.
@me-no-dev


libraries/PPP/src/ppp.c
```diff
@@ -1,3 +1,4 @@
+#if CONFIG_LWIP_PPP_SUPPORT
```
me-no-dev, line 1: add `#include "sdkconfig.h"` before this line

Jason2866: yeah, needed. Did work since compiled in Arduino Lib Builder...

lucasssvaz (approved):

PR #9522, opened 2024-04-16
Ethernet - Arduino Ethernet API compatibility layer

JAndrassy:
This PR adds legacy Arduino Ethernet API as a wrapper class for ETH object.

The filename EthernetCompat.h is the same as in the esp8266 Arduino.

In legacy Arduino Ethernet API static IP parameters are parameters of `begin`. To set pins Arduino uses optional `init` method.

Default setting in the wrapper is W5500 on default SPI pins with SS as CS. Default IRQ pin is 4 as in ETH (conflicting with SCK for C2 and C3).

Instance named `Ethernet` is created. It wraps the ETH instance. It is possible to create additional instances.

EthernetClient, EthernetServer and EthernetUdp are preprocessor macros (same way as their WiFi counterparts).

The example is taken from the Arduino Ethernet library. Only the comment about `init` is modified.

Tested with 'classic' ESP32 with W5500 on SPI with default pins and alternative pins and with spi_host2 on default and alternative pins. I will try to test with LAN8720 and maybe with a C3 with W5500.

Test sketch was LegacyEthernetTest from my NetAPIHelpers library and the new example.


JAndrassy:
what is an "external library"?

me-no-dev:
External Library means that is hosted somewhere else. Usually github and is registered with Arduino libraries, so people can find it.

JAndrassy:
the changes in ETH.h start with `friend class` to access private `beginSPI`

me-no-dev:
You can keep all changes in ETH, including that.

JAndrassy:
new PR https://github.com/espressif/arduino-esp32/pull/9539

JAndrassy:
@me-no-dev I published the 'external library'. It uses IDF directly to add support for enc28j60.
https://github.com/Networking-for-Arduino/EthernetESP32

Implementation diffrences:
* I made an OOP way for drivers. 
* I register the event callback for every instance separately
* I left out periman

me-no-dev:
@JAndrassy periman is a good thing to have :) maybe add the library to the list of tested external libs here?

PR #9526, opened 2024-04-17
refactor(style): Change some style options

lucasssvaz:
## Description of Change
Updated Clang format to v18.1.3 and refactored the style options to reflect more closely our regular coding style.

Bracket wrapping is now the same as Linux + Bracket wrapping for case labels. This means that bracket wrapping will happen in class implementations, function implementations, namespaces and case labels only.
Also adds column limit of 120 characters.

You can compare the changes in settings in [this mock file](https://www.diffchecker.com/XPcWR8Pe/)

Please provide feedback so we can find a middle ground between everyone's opinions.
All the options are documented with examples in [Clang's docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html). Feel free to recommend settings to be applied.

## Tests scenarios
Tested locally and with CI.


me-no-dev:
we need to wait for PP to get in. do not want to have to redo it again

PR #9531, opened 2024-04-18
ci(pre-commit): Fix PR checks and cache

lucasssvaz:
## Description of Change
Makes the pre-commit status check more elegant and properly work when the workflows need approval.
Also fix the caching by running the pre-commit workflow in the master branch on pushes (without automatic changes).

## Tests scenarios
Tested in my fork


me-no-dev (approved):

Jason2866:
The memory usage test has a glitch. There is no code change and the size checker shows some (small) changes.

P-R-O-C-H-Y:
> The memory usage test has a glitch. There is no code change and the size checker shows some (small) changes.

Already on it @Jason2866, but unfortunately the sizes are from arduino cli compilation. So it's a glitch there. But I can do a small workaround to not show those fake small changes.

PR #9542, opened 2024-04-20
NetworkClient - close the connection in stop() method

JAndrassy:
this PR allows to close the TCP connection in `client.stop()` for all copies of NetworkClient which wrap the same TCP connection.

The existence of multiple copies of WiFiClient for the same TCP connection is a feature of the 'Arduino language' and the networking libraries must deal with it.

Destroy of a NetworkClient instance should not close the TCP connection unless it is the last copy handling that connection. Thanks to shared_ptr NetworkClientSocketHandle instance closes the connection as it is destroyed when the last copy of NetworkClient for that connection is destroyed.

tested with [WiFiPagerServer](https://github.com/JAndrassy/NetApiHelpers/blob/master/examples/WiFiPagerServer/WiFiPagerServer.ino) example from my NetAPIHelpers library with this added:
```
    if (s.equals("C")) {
      client.stop();
    }
```


VojtechBartoska:
Thanks for PR @JAndrassy, we will review it soon.

me-no-dev:
@JAndrassy can you come up with a test case for this? Think of a sketch that would use multiple copies of the client and close when it should. The sketch should not require external libs

JAndrassy:
the best example is an attempt to implement server.available() and print-to-all-clients in 'external library'.
here is a simple version of the class put directly into the sketch. the rest of the sketch is my PagerServer example which is available in multiple Arduino networking libraries.
without this PR it is not possible to stop the client from the sketch. 
send upper case C from telnet to invoke client.stop from server side.
I use ArduinoWiFiServer in my popular TelnetStream library and users don't like that they can't stop the client from server side on esp32. https://github.com/JAndrassy/TelnetStream/issues/23
```
#include <WiFi.h>

#include "arduino_secrets.h"
///////please enter your sensitive data in the Secret tab/arduino_secrets.h
char ssid[] = SECRET_SSID;
char pass[] = SECRET_PASS;

#define MAX_MONITORED_CLIENTS 5

class ArduinoWiFiServer : public WiFiServer, public Print {
public:

  ArduinoWiFiServer(uint16_t port) : WiFiServer(port) {}

  // https://www.arduino.cc/en/Reference/WiFiServerAvailable
  WiFiClient available() {

    // update connected clients
    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
      if (!connectedClients[i]) {
        connectedClients[i] = accept();
      }
    }
    // find next client with data available
    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
      if (index == MAX_MONITORED_CLIENTS) {
        index = 0;
      }
      WiFiClient& client = connectedClients[index];
      index++;
      if (client.available())
        return client;
    }
    return WiFiClient(); // no client with data found
  }

  virtual size_t write(uint8_t b) override {
    return write(&b, 1);
  }

  virtual size_t write(const uint8_t *buf, size_t size) override {
    size_t ret = 0;
    for (uint8_t i = 0; i < MAX_MONITORED_CLIENTS; i++) {
      WiFiClient& client = connectedClients[i];
      if (client) {
        ret = client.write(buf, size);
      }
    }
    return ret;
  }

  using Print::write;

private:
  WiFiClient connectedClients[MAX_MONITORED_CLIENTS];
  uint8_t index = 0;
};

ArduinoWiFiServer server(2323);

void setup() {

  Serial.begin(115200);
  while (!Serial);

  Serial.print("Attempting to connect to SSID \"");
  Serial.print(ssid);
  Serial.println("\" with DHCP ...");
  WiFi.begin(ssid, pass);
    Serial.print('.');
    delay(1000);
  }
  Serial.println();

  server.begin();

  IPAddress ip = WiFi.localIP();
  Serial.println();
  Serial.print("To access the server, connect with Telnet client to ");
  Serial.print(ip);
  Serial.println(" 2323");
}

void loop() {

  WiFiClient client = server.available();     // returns first client which has data to read or a 'false' client
  if (client) {                               // client is here true only if it is connected and has data to read
    String s = client.readStringUntil('\n');  // read the message incoming from one of the clients
    s.trim();                                 // trim eventual \r
    if (s == "C") {
      client.println("Bye");
      client.stop();                          // stop the client
    } else {
      Serial.println(s);                        // print the message to Serial Monitor
      client.print("echo: ");                   // this is only for the sending client
      server.println(s);                        // send the message to all connected clients
    }
  }
  delay(1); // avoid watchdog reset with 3.0.0
}
```

VojtechBartoska:
**Status update:**
- CI test will be added @lucasssvaz/ @P-R-O-C-H-Y

JAndrassy:
the sketch above is an almost real world example. of course for a CI test a much simpler example would work

JAndrassy:
a simple test:
```
  if (client.connect(server, 80)) {

    WiFiClient client2 = client;
    client2.stop();
    delay(3000);
    if (client.connected()) {
      Serial.println("still connected");
    }
  }
```

P-R-O-C-H-Y:
> a simple test:
> 
> ```
>   if (client.connect(server, 80)) {
>     Serial.println("connected to server");
> 
>     WiFiClient client2 = client;
>     client2.stop();
>     delay(3000);
>     if (client.connected()) {
>       Serial.println("still connected");
>     }
>   }
> ```

Hello @JAndrassy, can you please provide full small example which fails without your changes and work afterwards?
I can't make it work with the changes from your PR with the simple test you commented. Thanks

JAndrassy:
```
#include <WiFi.h>

const char* ssid = "yourssid";
const char* password = "yourpasswd";

const char* server = "arduino.tips";

WiFiClient client;

void setup() {
  Serial.begin(115200);

  WiFi.begin(ssid, password);
    delay(1000);
    Serial.print('.');
  }
  Serial.println();

  if (client.connect(server, 80)) {

    WiFiClient client2 = client;
    client2.stop();
    delay(3000);
    if (client.connected()) {
      Serial.println("still connected");
    }
  }
}

void loop() {
}
```

P-R-O-C-H-Y:
Did you test the sketch you posted? We are still getting same result with/without PR changes.

JAndrassy:
tested and retested now. the difference is, that without the PR it prints "still connected"

P-R-O-C-H-Y:
> tested and retested now. the difference is, that without the PR it prints "still connected"

My bad, I have my environment for testing set wrongly. Its fine now

me-no-dev (approved):

PR #9543, opened 2024-04-20
NetworkClient - don't stop() in destroy

JAndrassy:
Destroy of a NetworkClient instance should not close the TCP connection unless it is the last copy handling that connection. Thanks to shared_ptr NetworkClientSocketHandle instance closes the connection as it is destroyed when the last copy of NetworkClient for that connection is destroyed.


JAndrassy:
sorry. it only makes sense as part of PR https://github.com/espressif/arduino-esp32/pull/9542

PR #9554, opened 2024-04-23
Fix hal-uart truncated tx/rx buffer size (uint16_t -> uint32_t)

Unbinilium:
## Description of Change

Changed the parameter types of `rx_buffer_size` and `tx_buffer_size` in the incoming parameters of the `uartBegin` function from `uint16_t` to `uint32_t`, which solved 2 issues:
- When setting a tx / rx buffer larger than 65535 bytes, the function returns normally without warning, but the actual buffer is not the same as the one set by the user.
- Be able to use tx / rx buffer which larger than 65535 bytes.

## Tests scenarios

- Arduino IDE 2.3.2 Linux_64
- ESP32S3

## Related links

The `uart_driver_install` function accepts incoming tx / rx buffer size parameters of type `int`: https://github.com/espressif/esp-idf/blob/636ff35b52f10e1a804a3760a5bd94e68f4b1b71/components/esp_driver_uart/include/driver/uart.h#L113

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -375,7 +375,7 @@ bool uartSetHwFlowCtrlMode(uart_t *uart, uart_hw_flowcontrol_t mode, uint8_t thr
 
 // This helper function will return true if a new IDF UART driver needs to be restarted and false if the current one can continue its execution
 bool _testUartBegin(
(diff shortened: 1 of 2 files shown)


SuGlider (approved):
Thanks! LGTM.

SuGlider:
This PR has been backported into 2.0.16 by #9561

lucasssvaz (approved):

P-R-O-C-H-Y (approved):

PR #9565, opened 2024-04-24
feat (hwcdc): ports changes made in 2.0.15

### LINKED ISSUE #9545: ESP32 C3 Supermini only boots when USB is connected and Serial is used.
### Board

ESP32 C3 Supermini

### Device Description

ESP32C3 Dev Module 

### Hardware Configuration

No pins connected, just the build-in LED is used.

### Version

### IDE Name

Arduino IDE

### Operating System

Windows-10

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

115200

### Description

Hi all, I was used that when I use a Serial.begin(115200) in Setup() and I have USB CDC on Boot enabled I can see any Serial.print messages in the Serial Monitor of the IDE.
And yes, that works.
However, like since one week or so, when I flash the code that still works, but now the ESP32C3 no longer boots when the board is NOT connected to a usb port of my PC.
Even with the IDE not started it will not boot on pressing reset or at power on.
But, as soon as I start the IDE or open the serial port (COM12 in my case) eg. by using Putty on that serial port, at that moment YES, the board boots and will start to output the Serial.print messages.

I was used that a board would run anyway, even with no serial port connected.
If I flash with USB CDC on boot Disabled, it won't run either.

I can make it run only when I remove all Serial.xxxx statements from my code before flashing??

I have the feeling this started about a week ago, because I had code running that I flashed a week ago that was running fine when using an external 5V supply. (Yes, I checked the voltages)

I changed some small stuff in the code and flashed again this week, and now it won't boot anymore, only when connected to either the IDE or Putty using the proper COM-port, or commenting out ALL Serial.xxxx statements?
When I use Putty to see the output of the Serial.xxx statements the ESP32 stops working when I disconnect Putty...

Is this because the ESP32C3 always expects a serial-port active when declared as Serial.begin(115200)??

Is there a way of keeping the Serial.xxxx statements in there but setting the serial port to ignore when not there?

I am stumped, but it does reproduce 100%.

Un saludo,

Leo

### Sketch

```cpp
void setup() {
  // initialize digital pin LED_BUILTIN as an output.
  pinMode(8, OUTPUT);
  Serial.begin(115200);
}

// the loop function runs over and over again forever
void loop() {
  digitalWrite(8, HIGH);  // turn the LED on (HIGH is the voltage level)
  delay(1000);                      // wait for a second
  digitalWrite(8, LOW);   // turn the LED off by making the voltage LOW
  delay(1000);                      // wait for a second
  Serial.println("ESP32 is running");
```

### Debug Message

```plain
15:57:39.843 -> ESP32 is running
15:57:41.808 -> ESP32 is running
15:57:43.797 -> ESP32 is running
15:57:45.847 -> ESP32 is running
15:57:47.809 -> ESP32 is running
15:57:49.834 -> ESP32 is running
15:57:51.869 -> ESP32 is running
15:57:53.855 -> ESP32 is running
15:57:55.883 -> ESP32 is running
```

### Other Steps to Reproduce

Flased without any Serial.xxxxx statements it will boot and run.
Also when flashed with the Serial- statements connected to PC, but with Putty active instead of the Arduino IDE, it will boot and run too.

SuGlider:
## Description of Change
Ports many changes, fixes and improvements made in 2.0.15:
- correct use of timeout
- avoids problems with CDC ISR not reading data
- fixes problems with transmiting many bytes to USB Host
- changes how USB SOF and CDC connection is detected
- improves flushing TX data, forcing a FIFO policy when overfow is detected

## Tests scenarios
TBD 

## Related links
Fixes #9545

### DIFF
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -213,7 +213,7 @@ static void flushTXBuffer(const uint8_t *buffer, size_t size) {
     if (queued_size && queued_buff != NULL) {
       vRingbufferReturnItem(tx_ring_buf, (void *)queued_buff);
     }
-    // now add the new data that fits to the ring buffer
+    // now add the new data that fits into the ring buffer
     uint8_t *bptr = (uint8_t *)buffer;
     if (size >= ringbufferLength) {
       size = ringbufferLength;
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -29,7 +29,6 @@
 #include "hal/usb_serial_jtag_ll.h"
 #pragma GCC diagnostic warning "-Wvolatile"
 #include "rom/ets_sys.h"
-#include "driver/usb_serial_jtag.h"
 
 ESP_EVENT_DEFINE_BASE(ARDUINO_HW_CDC_EVENTS);
 
@@ -40,8 +39,11 @@ static intr_handle_t intr_handle = NULL;
 static SemaphoreHandle_t tx_lock = NULL;
 static volatile bool connected = false;
 
+static volatile unsigned long lastSOF_ms;
+static volatile uint8_t SOF_TIMEOUT;
+
 // timeout has no effect when USB CDC is unplugged
-static uint32_t requested_tx_timeout_ms = 100;
+static uint32_t tx_timeout_ms = 100;
 
 static esp_event_loop_handle_t arduino_hw_cdc_event_loop_handle = NULL;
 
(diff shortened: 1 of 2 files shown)


P-R-O-C-H-Y (approved):

cores/esp32/HWCDC.cpp
```diff
@@ -155,21 +170,71 @@ bool HWCDC::isCDC_Connected() {
...
+    size_t to_flush = size - freeSpace;
+    if(to_flush > ringbufferLength) {
+      to_flush = ringbufferLength;
+    }
+    size_t queued_size = 0;
+    uint8_t *queued_buff = (uint8_t *)xRingbufferReceiveUpTo(tx_ring_buf, &queued_size, 0, to_flush);
+    if (queued_size && queued_buff != NULL) {
+      vRingbufferReturnItem(tx_ring_buf, (void *)queued_buff);
+    }
+    // now add the new data that fits to the ring buffer
```
SuGlider, line 216:
```suggestion
    // now add the new data that fits into the ring buffer
```

lucasssvaz (approved):

PR #9575, opened 2024-04-29
Update variant unphone9 with digital pin mappings

garethhcoleman:
## Description of Change
This change only affects a single board variant, the unPhone. As co-designer of the board I confirm that this change will allow users to work with our hardware using this codebase more easily. No restrictions are imposed on the use of this code.

## Tests scenarios
We have sucessfully tested this change on an unPhone9.

## Related links
unphone description can be found at [unphone.net](https://unphone.net/)


garethhcoleman:
will redo with correct commit message format

PR #9594, opened 2024-05-06
fix(RAM): define HTTP_METHOD_STR as static const to save RAM

craiglink:
## Description of Change
defines constant array as to strings as static const so that it moves from RAM to Flash

## Tests scenarios
I have tested this using Arduino-esp32 core v2.0.16 running FluidNC - http://wiki.fluidnc.com/

This should be cherry-picked into the v2.0.x branch


me-no-dev (approved):

me-no-dev:
@craiglink please sign the CLA in order to merge your changes.

lucasssvaz (approved):

PR #9612, opened 2024-05-10
NetworkClientSecure made copyable

JAndrassy:
The PR makes NetworkClentSecure copyable as required by Arduino networking API.

I checked with Wireshark that the connection is closed using stop() or when the last copy is destroyed.

A test sketch
```
#include <WiFi.h>
#include <WiFiClientSecure.h>

const char *ssid = "yourssid";
const char *password = "yourpasswd";

const char* server = "example.com";

WiFiClientSecure connect() {
  WiFiClientSecure client;
  client.setInsecure();
  if (client.connect(server, 443)) {
    client.println("GET / HTTP/1.1");
    client.print("Host: ");
    client.println(server);
    client.println();
  }
  return client;
}

void stop(WiFiClientSecure client) {
  client.stop();
}

void setup() {
  Serial.begin(115200);
  while (!Serial);

  WiFi.begin(ssid, password);
    delay(1000);
    Serial.print('.');
  }
  Serial.println();

  WiFiClientSecure client = connect();
  client.find("\r\n\r\n");
  while (client.available()) {
    char c = client.read();
    Serial.write(c);
  }
//  stop(client);
//  delay(3000);
}

void loop() {
  delay(1);
}
```

### DIFF
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -32,8 +32,12 @@ NetworkClientSecure::NetworkClientSecure() {
   _connected = false;
   _timeout = 30000;  // Same default as ssl_client
 
-  sslclient = new sslclient_context;
-  ssl_init(sslclient);
+  sslclient.reset(new sslclient_context, [](struct sslclient_context *sslclient) {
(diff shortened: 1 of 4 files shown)


P-R-O-C-H-Y (approved):
Tested and working as expected.
Without PR changes it crashes.

PR #9626, opened 2024-05-14
ci(optimization): Run workflows only when related files change

lucasssvaz:
## Description of Change
Speed up the CI in some cases by making it run only the workflows related to the change.
This is the first step towards improving CI speed. Later the workflows will only compile the files that were changed.

## Tests scenarios
CI


lucasssvaz:
Closing as I will include this in the workflows refactoring

PR #9628, opened 2024-05-14
fix(cdc): Disable SOF interrupt and CDC reset on begin()

me-no-dev:


Jason2866:
@me-no-dev Did a few tests with this version and Tasmota. Does work.
When merged no need to use a modified version anymore :-)

PR #9631, opened 2024-05-14
Fix class used for web client in WiFiClientEnterprise.ino

cgraun:
Fix error as NetworkClient class is not found in the context. Used WiFiClient class instead.

## Description of Change
Replaced NetworkClient class with WiFiClient one as NetworkClient was not declared.

## Tests scenarios
Successfully compiled after change on Arduino IDE 2.3.2, platform 2.0.16, and PlatformIO in VSCode, platform 6.1.0, in both cases ESP32 Dev Module selected. Uploaded to an ESP32 Mini D1 by AZ-Delivery.

## Related links
No related issue found.


me-no-dev:
NetworkClient exists on version 3.0.0, which is the current master. This code is not meant to be compatible with Arduino 2.x. There is a branch that has the 2.x support. Closing as invalid

PR #9632, opened 2024-05-14
NetworkClientSecure - copyability improvements and _timeout shadowing fixed

JAndrassy:
The PR has 3 fixes for NetworkClientSecure

1) buffer to store the peeked byte is moved into the shared objects to be shared between NetworkClientSecure copies
2) last error code is moved into the shared objects to be available in all copies
3) NetworkClientSecure::_timeout shadowed NetworkClient::_timeout and then NetworkClient::setConnectionTimeout didn't set the timeout for NetworkClientSecure

test (to test the last error change port to 80)
```
#include <WiFi.h>
#include <WiFiClientSecure.h>

const char *ssid = "yourssid";
const char *password = "yourpasswd";

const char* server = "example.com";

WiFiClientSecure connect() {
  WiFiClientSecure client;
  client.setInsecure();
  if (client.connect(server, 443)) {
    client.println("GET / HTTP/1.1");
    client.print("Host: ");
    client.println(server);
    client.println();
    client.find("\r\n\r\n"); // skip headers
  }
  return client;
}

char peekTest(WiFiClientSecure client) {
  return client.peek();
}

void setup() {
  Serial.begin(115200);
  while (!Serial);

  WiFi.begin(ssid, password);
    delay(1000);
    Serial.print('.');
  }
  Serial.println();

  WiFiClientSecure client = connect();
  if (client) {

  char peek = peekTest(client);
  if (client.peek() != peek) {
      Serial.println("ERROR peek result doesn't match!");
      Serial.print("peek 1: ");
      Serial.println(peek);
      Serial.print("peek 2: ");
      Serial.println((char) client.peek());
      Serial.println();
    }

    while (client.available()) {
      char c = client.read();
      Serial.write(c);
    }
  } else {

    char buff[100];
    int errCode = client.lastError(buff, sizeof(buff));
    Serial.println();
    Serial.print("Connect result code: ");
    Serial.println(errCode);
    if (errCode < 0) {
      Serial.print("\tmessage: ");
      Serial.println(buff);
    }
  }
}

void loop() {
  delay(1);
}
```

### DIFF
--- a/libraries/NetworkClientSecure/src/ssl_client.h
+++ b/libraries/NetworkClientSecure/src/ssl_client.h
@@ -26,6 +26,10 @@ typedef struct sslclient_context {
 
   unsigned long socket_timeout;
   unsigned long handshake_timeout;
+
+  int last_error;
+  int peek_buf;
+
 } sslclient_context;
 
 void ssl_init(sslclient_context *ssl_client);
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -83,7 +83,7 @@ void NetworkClientSecure::stop() {
   stop_ssl_socket(sslclient.get());
 
   _connected = false;
-  _peek = -1;
+  sslclient->peek_buf = -1;
   _lastReadTimeout = 0;
   _lastWriteTimeout = 0;
 }
@@ -134,7 +134,7 @@ int NetworkClientSecure::connect(IPAddress ip, uint16_t port, const char *host, 
     log_i("Actual TLS start postponed.");
   }
 
-  _lastError = ret;
+  sslclient->last_error = ret;
 
   if (ret < 0) {
     log_e("start_ssl_client: connect failed: %d", ret);
@@ -175,7 +175,7 @@ int NetworkClientSecure::connect(const char *host, uint16_t port, const char *ps
   }
 
   int ret = start_ssl_client(sslclient.get(), address, port, host, _timeout, NULL, false, NULL, NULL, pskIdent, psKey, _use_insecure, _alpn_protos);
-  _lastError = ret;
+  sslclient->last_error = ret;
   if (ret < 0) {
     log_e("start_ssl_client: connect failed %d", ret);
     stop();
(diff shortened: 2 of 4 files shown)


me-no-dev:
Hi @JAndrassy :) This PR is failing CI. Please fix it and make sure that you add all you might need as breaking changes. Further such PRs will not be accepted until we get to next major version.

JAndrassy:
this PR is my last for 3.0.0. I don't consider is a breaking change.

the problem with checks is with existing code, not with my change: and I don't know how to fix it the right way.
> error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'sslclient_context'

```
C:\Users\runneradmin\Documents\Arduino\hardware\espressif\esp32\libraries\NetworkClientSecure\src\ssl_client.cpp:374:9: error: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'sslclient_context' {aka 'struct sslclient_context'}; use assignment or value-initialization instead [-Werror=class-memaccess]
  374 |   memset(ssl_client, 0, sizeof(sslclient_context));
      |   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\Users\runneradmin\Documents\Arduino\hardware\espressif\esp32\libraries\NetworkClientSecure\src\ssl_client.h:15:16: note: 'sslclient_context' {aka 'struct sslclient_context'} declared here
   15 | typedef struct sslclient_context {
      |                ^~~~~~~~~~~~~~~~~
```

libraries/NetworkClientSecure/src/ssl_client.h
```diff
@@ -26,6 +26,10 @@ typedef struct sslclient_context {
 
   unsigned long socket_timeout;
   unsigned long handshake_timeout;
+
+  int last_error = 0;
+  int peek_buf = -1;
```
me-no-dev, line 31: you are setting here default value `-1` and the code calls `memset(ssl_client, 0, sizeof(sslclient_context));` which would set it to 0. That is why you are getting the error. Do not add default value like that and  make sure that when cleared it is then set properly.

JAndrassy:
so the default values made the struct "non-trivial". thank you

PR #9687, opened 2024-05-26
feat(eth): Ethernet - enc28j60 driver added

JAndrassy:
## Description of Change

Adds enc28j60 support to Ethernet library. The driver is from IDF examples. I updated it for custom SPI support and use without the INT pin wired (polling) 

## Tests scenarios

ETH_W5500_Arduino_SPI and ETH_W5500_IDF_SPI examples with ETH_PHY_TYPE replaced to ETH_PHY_ENC28J60


JAndrassy:
I switched it to draft because suddenly the custom SPI doesn't work in my test and right now I don't have time to debug it.

me-no-dev:
We would rather not add such support in Arduino, because the chip and driver are quite buggy. That is the reason why it's just example in IDF. They were having many issues with modules, power, etc. and we do not want to transfer those problems here.

JAndrassy:
ok. I add it to the 'external library'

PR #9698, opened 2024-05-27
fix(release): install missing 'packaging' package

me-no-dev:
Necessary to combine the json files

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,6 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: '3.x'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -16,6 +16,8 @@ jobs:
     - uses: actions/setup-python@v5
       with:
         python-version: '3.x'
+        cache: 'pip' # caching pip dependencies
```
lucasssvaz, line 19: This only works if using a `requirements.txt` file as stated [here](https://github.com/actions/setup-python#caching-packages-dependencies)

```suggestion
```

lucasssvaz (approved):

PR #9717, opened 2024-05-29
fix(boards): Remove duplicate VID/PID from boards.txt and add default CDC/JTAG board

me-no-dev:
Adds a hidden board with the default VID/PID of USB CDC/JTAG in order to always show that in IDE when ESP32 CDC/JTAG device is connected and removes all other mentions of that VID/PID

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -33,6 +33,13 @@ menu.NetworkLogLevel=Network Log Level
 ### DO NOT PUT BOARDS ABOVE THE OFFICIAL ESPRESSIF BOARDS! ###
 ##############################################################
 
+esp32_family.name=ESP32 Family Device
+esp32_family.hide=true
+esp32_family.vid.0=0x303a
+esp32_family.pid.0=0x1001
+
+##############################################################
+
 esp32c2.name=ESP32C2 Dev Module
 esp32c2.hide=true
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
@P-R-O-C-H-Y we need to enforce that this VID/PID combo is not added to other boards

boards.txt
```diff
@@ -38,6 +38,8 @@ esp32_family.name=ESP32 Family Device
 esp32_family.hide=true
 esp32_family.vid.0=0x303a
 esp32_family.pid.0=0x1001
+esp32_family.upload_port.0.vid=0x303a
+esp32_family.upload_port.0.pid=0x1001
```
me-no-dev, line 42: @ladyada in order for discovery to work properly with IDE2 and CLI, it is necessary to add the `.upload_port` lines. Maybe someone there can update your board definitions?

cc: @UnexpectedMaker

P-R-O-C-H-Y (approved):

boards.txt
ladyada: thanks
@hathach can you add to your ToDo list :)
@me-no-dev is there a date you need this by?

me-no-dev: @ladyada we plan on pushing bugfix releases fairly often, so it's not a hurry.

hathach: @ladyada sure, I will follow this up

UnexpectedMaker:
Hey @me-no-dev - Thanks for pinging me as well...

So you mean we need to add this to each of our boards, using the same PID/VID as the board?
```
um_bling.name=UM BLING
um_bling.vid.0=0x303a
um_bling.pid.0=0x817F
um_bling.upload_port.0.vid=0x303a
um_bling.upload_port.0.pid=0x817F
```

me-no-dev:
> So you mean we need to add this to each of our boards, using the same PID/VID as the board?

Yes. First two lines are for IDE1 and second lines are for IDE2 and CLI

UnexpectedMaker:
Wicked, thanks, doing a PR now :)

PR #9722, opened 2024-05-29
fix(variant): Arduino Nano ESP32 variant fix for core 3

JAndrassy:
I don't know what hack that is including cpp files, but without this PR every build for this board fails with core 3. linker can't find `bytes2hex` from HEXBuilder.

I just unpacked my new Nano ESP32. I wonder why nobody cares about this board with core 3.


me-no-dev:
@pillo79 PTAL

pillo79:
Sorry, my bad for not checking it after the betas. I confirm this is the proper fix, thanks @JAndrassy!

PR #9725, opened 2024-05-29
fix(net): Fix IPv4 address construction from ip_addr_t and comparison (#9724)

### LINKED ISSUE #9724: IPAddress construction and comparison are buggy on Arduino 3
### Board

N/A

### Device Description

N/A

### Hardware Configuration

N/A

### Version

v3.0.0

### IDE Name

N/A

### Operating System

N/A

### Flash frequency

N/A

### PSRAM enabled

no

### Upload speed

N/A

### Description

The origin of this bug request comes for a bug report in my ESPAsyncWebServer maintained fork here: https://github.com/mathieucarbou/ESPAsyncWebServer/issues/26

The root of the problem lies in the fact that when connected to a WiFi, both should be true.

```
WiFi.localIP().toString() == request->client()->localIP().toString()
WiFi.localIP() == request->client()->localIP()
```

But the second one is false.

The issue lies at several places:

1) the comparison operator

```
bool IPAddress::operator==(const IPAddress &addr) const {
  return (addr._type == _type) && (memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) == 0);
}
```

For IPv4, it should only compare `_address.dword[IPADDRESS_V4_DWORD_INDEX]`, not the full set of bytes

2) the constructor:

```
IPAddress &IPAddress::from_ip_addr_t(const ip_addr_t *addr) {
  if (addr->type == IPADDR_TYPE_V6) {
    _type = IPv6;
    _address.dword[0] = addr->u_addr.ip6.addr[0];
    _address.dword[1] = addr->u_addr.ip6.addr[1];
    _address.dword[2] = addr->u_addr.ip6.addr[2];
    _address.dword[3] = addr->u_addr.ip6.addr[3];
#if LWIP_IPV6_SCOPES
    _zone = addr->u_addr.ip6.zone;
#endif /* LWIP_IPV6_SCOPES */
  } else {
    _type = IPv4;
    _address.dword[IPADDRESS_V4_DWORD_INDEX] = addr->u_addr.ip4.addr;
  }
  return *this;
}
```

Which is not "clearing" the indexes 0, 1 and 2 to 0, to ensure the comparison operator works for IPv4

Any of this fix would work. I tested this one:

```
IPAddress &IPAddress::from_ip_addr_t(const ip_addr_t *addr) {
  if (addr->type == IPADDR_TYPE_V6) {
    _type = IPv6;
    _address.dword[0] = addr->u_addr.ip6.addr[0];
    _address.dword[1] = addr->u_addr.ip6.addr[1];
    _address.dword[2] = addr->u_addr.ip6.addr[2];
    _address.dword[3] = addr->u_addr.ip6.addr[3];
#if LWIP_IPV6_SCOPES
    _zone = addr->u_addr.ip6.zone;
#endif /* LWIP_IPV6_SCOPES */
  } else {
    _type = IPv4;
    _address.dword[0] = 0;
    _address.dword[1] = 0;
    _address.dword[2] = 0;
    _address.dword[IPADDRESS_V4_DWORD_INDEX] = addr->u_addr.ip4.addr;
  }
  return *this;
}
```

But I think both should be applied (also making sure that the compare operator is comparing the relevant information depending on the IP type)

### Sketch

```cpp
N/A
```

### Debug Message

```plain
N/A
```

### Other Steps to Reproduce

_No response_

mathieucarbou:
This PR fixes #9724 and has been tested with the little sample below in the project where the bug report originated from:

https://github.com/mathieucarbou/ESPAsyncWebServer/blob/main/examples/Filters/Filters.ino

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -266,7 +266,7 @@ IPAddress &IPAddress::operator=(const IPAddress &address) {
 }
 
 bool IPAddress::operator==(const IPAddress &addr) const {
-  return (addr._type == _type) && (memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) == 0);
+  return (addr._type == _type) && (_type == IPType::IPv4 ? addr._address.dword[IPADDRESS_V4_DWORD_INDEX] == _address.dword[IPADDRESS_V4_DWORD_INDEX] : memcmp(addr._address.bytes, _address.bytes, sizeof(_address.bytes)) == 0);
 }
 
 bool IPAddress::operator==(const uint8_t *addr) const {
@@ -395,6 +395,7 @@ IPAddress &IPAddress::from_ip_addr_t(const ip_addr_t *addr) {
 #endif /* LWIP_IPV6_SCOPES */
   } else {
     _type = IPv4;
+    memset(_address.bytes, 0, sizeof(_address.bytes));
     _address.dword[IPADDRESS_V4_DWORD_INDEX] = addr->u_addr.ip4.addr;
   }
   return *this;


cores/esp32/IPAddress.cpp
```diff
@@ -395,6 +395,9 @@ IPAddress &IPAddress::from_ip_addr_t(const ip_addr_t *addr) {
 #endif /* LWIP_IPV6_SCOPES */
   } else {
     _type = IPv4;
+    _address.dword[0] = 0;
+    _address.dword[1] = 0;
+    _address.dword[2] = 0;
```
me-no-dev, line 400: It's better to fill the buffer to zeroes, in case that the index changes in the future
```suggestion
    memset(_address.bytes, 0, sizeof(_address.bytes));
```

mathieucarbou: Resolved :-)

me-no-dev (approved):

Issue #9744, opened 2024-06-01
Error compiling, works good on 2.0.14 but cannot compile and terminates on 3.0.0


TobbeG:
### Board

ESP32-S3 (Elecrow ESP32-Terminal)

### Device Description

Elecrow ESP32-Terminal

### Hardware Configuration

Standard device Elecrow ESP32-S3 with TFT display and Touch

### Version

v3.0.0

### IDE Name

Arduino IDE

### Operating System

Windows 11

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Cannot compile - Compilation terminates

### Sketch

```cpp
The error message indicates that the compiler is unable to find the bits/c++allocator.h header file, which is part of the standard C++ library implementation.
```

### Debug Message

```plain
cmd /c type nul > "C:\\Users\\THORBJ~1\\AppData\\Local\\Temp\\arduino_build_956191/file_opts"
Detecting libraries used...
[...]

PR #9760, opened 2024-06-03
fix(release): Shorten all toolchain names/versions

me-no-dev:
Helps to overcome the limit of windows path, when the usernames are long.

Fixes: https://github.com/espressif/arduino-esp32/issues/9744


SuGlider:
@me-no-dev - Where can I find the generated windows artifacts in order to test it in my machine?

lucasssvaz:
@SuGlider I think this only affects compilation. If the files were compiled successfully it should run exactly as it is supposed to.

We just need to check if that user's error is solved with this patch

me-no-dev:
@SuGlider generation happens on release. Will `release` an RC to confirm that all is properly renamed. This is the same we did for RiscV and shorting the package name and version

Issue #9756, opened 2024-06-03
PSRAM not working for esp32s3? project ignoring sdkconfig?


anthonyhoegberg:
### Board

ESP32s3

### Device Description

ESP32s3

### Hardware Configuration

FLASH 16MB DIO 80MHZ
OPI PSRAM

### Version

v3.0.0

### IDE Name

ESP-IDF vscode

### Operating System

Windows 11

### Flash frequency

80

### PSRAM enabled

yes

### Upload speed

default

### Description

So i cant figure out how to enable psram for arduino components 
My sdkconfig has been setup correctly with all the correct settings, however i feel like arduino components are ignoring some settings?

Like i cant get serial to work or my device unless i add definitions for USB CDC and USB MODE
But i think that should be automatic based on sdkconfig no? 
Also in older versions of arduino component the BOARD_HAS_PSRAM definition made it enable psram for me
However in this version 3.0.0 it no longer works and gives me the error 
project_folder/managed_components/espressif__arduino-esp32/cores/esp32/esp32-hal-psram.c:22:10: fatal error: esp_psram.h: No such file or directory      
   22 | #include "esp_psram.h"

Is arduino components somehow ignoring my sdkconfig and not applying settings correctly since i have to add definitions or is something wrong here? (When i remove the definition for BOARD HAS PSRAM I get no error but sketch shows no psram)

Project root cmake config:
```cmake
# The following lines of boilerplate have to be in your project's
# CMakeLists in this exact order for cmake to work correctly
cmake_minimum_required(VERSION 3.5)

add_definitions(-DBOARD_HAS_PSRAM -mfix-esp32-psram-cache-issue)
add_definitions(-DARDUINO_USB_CDC_ON_BOOT=1)
add_definitions(-DARDUINO_USB_MODE=1)

include($ENV{IDF_PATH}/tools/cmake/project.cmake)
project(amcamiot3)
```
[sdkconfig](https://github.com/user-attachments/files/15527561/sdkconfig.txt)

### Sketch

```cpp
#include "Arduino.h"

void setup() {
    Serial.begin(115200);
    Serial.println("Serial works!");

    Serial.printf("Total heap: %li\n", ESP.getHeapSize());
    Serial.printf("Free heap: %li\n", ESP.getFreeHeap());
    Serial.printf("Total PSRAM: %li\n", ESP.getPsramSize());
    Serial.printf("Free PSRAM: %li\n", ESP.getFreePsram());
}

void loop() {
}
```

[...]

PR #9762, opened 2024-06-03
feat(idf): Require esp_psram when as component

me-no-dev:
Fixes: #9756

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,8 +287,8 @@ endforeach()
 set(includedirs variants/${CONFIG_ARDUINO_VARIANT}/ cores/esp32/ ${ARDUINO_LIBRARIES_INCLUDEDIRS})
 set(srcs ${CORE_SRCS} ${ARDUINO_LIBRARIES_SRCS})
 set(priv_includes cores/esp32/libb64)
-set(requires spi_flash esp_partition mbedtls wifi_provisioning wpa_supplicant esp_adc esp_eth http_parser esp_psram)
-set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support bt esp_hid usb ${ARDUINO_LIBRARIES_REQUIRES})
+set(requires spi_flash esp_partition mbedtls wifi_provisioning wpa_supplicant esp_adc esp_eth http_parser)
+set(priv_requires fatfs nvs_flash app_update spiffs bootloader_support bt esp_hid usb esp_psram ${ARDUINO_LIBRARIES_REQUIRES})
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -287,7 +287,7 @@ endforeach()
 set(includedirs variants/${CONFIG_ARDUINO_VARIANT}/ cores/esp32/ ${ARDUINO_LIBRARIES_INCLUDEDIRS})
 set(srcs ${CORE_SRCS} ${ARDUINO_LIBRARIES_SRCS})
 set(priv_includes cores/esp32/libb64)
-set(requires spi_flash esp_partition mbedtls wifi_provisioning wpa_supplicant esp_adc esp_eth http_parser)
+set(requires spi_flash esp_partition mbedtls wifi_provisioning wpa_supplicant esp_adc esp_eth http_parser esp_psram)
```
igrr, line 290: Should it rather be in priv_requires? Seems like the public headers of arduino-esp32 don't include headers from esp_psram.

me-no-dev: I'm just not sure if users will not need to include anything from it

Jason2866:
Great fix. This bug annoyed me since a long time. Thought it was an issue in my fork of Platformio, since it worked before with Core 2.x and IDF 4.4.x.
Just tested my before failing example with the S3 and PSRAM and this change.
It does work now with this change!

CMakeLists.txt
igrr: In that case they should add `esp_psram` to the private requires list of their component. `idf.py` will also show a build-time hint if they miss this dependency and try to use the header file.

You only need to add components which provide header files which arduino-esp32 public headers include. If something is only used in a .c/.cpp file, or in a private header file, then it is sufficient to add that to the private requirements, instead.

(Of course, removing things from public requirements list is a breaking change, but since esp_psram wasn't there to begin with, I think it can go into the private requirements list.)

SuGlider: @igrr - This also happens when using Arduino as IDF component. Maybe that fix that you and @me-no-dev were working to add a sort of ARDUINO_LIB() to the project/user CMakeLists.txt could help.

Is it something that you guys are still working on?

me-no-dev:
@Jason2866 will your firmware compile if it's a private require?

Jason2866:
@me-no-dev Let me test. Will know in around a hour.

Jason2866:
@me-no-dev Does work too when added to private require.

CMakeLists.txt
igrr: No, I don't think the issue discussed here is related to CMake helper for registering Arduino libraries. The issue here is simply that the build system didn't include esp_psram component into the build, because there was no component which added it as a requirement.

PR #9796, opened 2024-06-06
feat(sdmmc): Add RAW disk functions

lbernstone:
## Description of Change
Added SD_MMC.ReadRAW and SD_MMC.WriteRAW. Added example to demonstrate USBMSC accessing SDMMC card.

## Tests scenarios
Tested on ESP32-S3 (the only supported platform ATM), v3.0.0

### DIFF
--- a/libraries/SD_MMC/examples/SD2USBMSC/SD2USBMSC.ino
+++ b/libraries/SD_MMC/examples/SD2USBMSC/SD2USBMSC.ino
@@ -15,31 +15,39 @@ int d0 = 37;
 int d1 = 38;
 int d2 = 33;
 int d3 = 34;
-bool onebit = true; // set to false for 4-bit. 1-bit will ignore the d1-d3 pins (but d3 must be pulled high)
+bool onebit = true;  // set to false for 4-bit. 1-bit will ignore the d1-d3 pins (but d3 must be pulled high)
 
-static int32_t onWrite(uint32_t lba, uint32_t offset, uint8_t* buffer, uint32_t bufsize){
+static int32_t onWrite(uint32_t lba, uint32_t offset, uint8_t *buffer, uint32_t bufsize) {
   uint32_t secSize = SD_MMC.sectorSize();
-  if (!secSize) return false; // disk error
+  if (!secSize) {
+    return false;  // disk error
+  }
   log_v("Write lba: %ld\toffset: %ld\tbufsize: %ld", lba, offset, bufsize);
-  for (int x=0; x< bufsize/secSize; x++) {
+  for (int x = 0; x < bufsize / secSize; x++) {
     uint8_t blkbuffer[secSize];
-    memcpy(blkbuffer, (uint8_t*)buffer + secSize*x, secSize);
-    if (!SD_MMC.writeRAW(blkbuffer, lba + x)) return false;
+    memcpy(blkbuffer, (uint8_t *)buffer + secSize * x, secSize);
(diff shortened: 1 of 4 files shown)


lucasssvaz (approved):
LGTM

libraries/SD_MMC/src/SD_MMC.h
```diff
@@ -16,7 +16,9 @@
 
 #include "sdkconfig.h"
 #include "soc/soc_caps.h"
-#ifdef SOC_SDMMC_HOST_SUPPORTED
+#ifndef SOC_SDMMC_HOST_SUPPORTED
+#error The SDMMC library requires a device with an SDIO Host
+#else
```
me-no-dev, line 21: the way it was, was the better way. IDF compiles everything, regardless of the target, so this will break compilation for chips other than S3

me-no-dev: @lucasssvaz I suggest the IDF CI runs any time lib or core is changed (not examples)

libraries/SD_MMC/examples/SD2USBMSC/ci.json
```diff
@@ -0,0 +1,9 @@
+{
+  "targets": {
+    "esp32": false,
```
me-no-dev, line 3: what is the reason to skip ESP32? It also has SDMMC, though it has fixed pins.

me-no-dev: ah nevermind... USB

libraries/SD_MMC/src/SD_MMC.h
lucasssvaz: @me-no-dev Ok, i'm currently in the process of fixing some issues I found with the CI. I can include this.

lbernstone: Is there a way to have IDF compile cleanly yet have a arduino compile time warning? This ends up giving a somewhat confusing error that the functions are not available if a newbie decides to try this on an unsupported platform.

lbernstone:
If someone has a S3 rig set up able to do 4-bit SDIO, I'd be interested to know what kind of transfer rate you can get vs 1-bit.

libraries/SD_MMC/src/SD_MMC.h
me-no-dev: it can be a warning. A bit annoying, but will work. You can probably detect if it's Arduino by some define too

lbernstone: Better?

libraries/SD_MMC/examples/SD2USBMSC/SD2USBMSC.ino
```diff
@@ -0,0 +1,102 @@
+#ifndef ARDUINO_USB_MODE
```
me-no-dev, line 1:
```cpp
#if !SOC_USB_OTG_SUPPORTED || ARDUINO_USB_MODE
#error Device does not support USB_OTG or native USB CDC/JTAG is selected
#endif
```

libraries/SD_MMC/src/SD_MMC.h
me-no-dev: better :)

libraries/SD_MMC/examples/SD2USBMSC/SD2USBMSC.ino
me-no-dev: @lbernstone would you please fix this?

PR #9798, opened 2024-06-07
change (3rd_party_board): XIAO_C3 delete the definition of pin A3

Lesords:
### Description of Change
Delete the definition of pin A3

### Reason
Because A3 corresponds to GPIO5, which happens to be ADC2, and ADC2 is not currently supported, so A3 is currently useless.


lucasssvaz:
Could you please fill the PR description for us to know why the pin is being removed ?

Lesords:
> Could you please fill the PR description for us to know why the pin is being removed ?

Okay, I updated it now

P-R-O-C-H-Y (approved):

Issue #9822, opened 2024-06-11
WIFI HTTP download file speed is slow


Llgok:
### Board

ESP32-S3-DevKitC-1

### Device Description

nothing

### Hardware Configuration

nothing

### Version

v3.0.1

### IDE Name

Arduino IDE

### Operating System

Windows 10

### Flash frequency

QIO 80Mhz

### PSRAM enabled

no

### Upload speed

115200

### Description

I called the function `stream->readBytes();` in the `HTTPClient.h` library to download resources from the internet, but found that the download speed is extremely slow, averaging between 100kb/s and 200kb/s. It has been confirmed that the current network status is good, and there is no speed limit on the resource download. Although the ESP32S3 has been tested with iperf and can achieve a download speed of around 5Mb, using the `HTTPClient.h` library for internet resource downloads has become very slow. Is this due to the inherent limitations of the ESP32S3, or is there something happening in the data acquisition function `stream->readBytes()` that is causing the download speed to be so slow?

### Sketch

```cpp
#include <WiFi.h>
#include <HTTPClient.h>

const char *ssid = "mywifi";
const char *password = "mywifi";

// const char *fileDownloadUrl = "https://code.visualstudio.com/docs/?dv=win64user";//vscode

static size_t CycleTime = 0;

void setup()
{
    Serial.begin(115200);

    WiFi.begin(ssid, password);
    while (WiFi.status() != WL_CONNECTED)
    {
        delay(1000);
        Serial.println("Connecting to WiFi...");
    }
    Serial.println("Successfully connected to WiFi");

    HTTPClient http;
    http.begin(fileDownloadUrl);

    const char *headerKeys[] = {"Location"};
    http.collectHeaders(headerKeys, 1);

    size_t startTime = millis();

    size_t uselessTime = 0;

    int httpCode = http.GET();

    while (httpCode == HTTP_CODE_MOVED_PERMANENTLY || httpCode == HTTP_CODE_FOUND)
    {
        String newUrl = http.header("Location");
        Serial.printf("Redirecting to: %s\n", newUrl.c_str());
        http.end(); 

        http.begin(newUrl);
        httpCode = http.GET();
    }

    if (httpCode == HTTP_CODE_OK)
    {
        size_t fileSize = http.getSize();
        Serial.printf("Starting file download...\n");
        Serial.printf("file size: %f Mb\n", fileSize / 1024.0 / 1024.0);

        WiFiClient *stream = http.getStreamPtr();

[...]

Issue #9823, opened 2024-06-11
HTTPUpdate read Timeout


hitecSmartHome:
### Board

ESP32-Wrover

### Device Description

-

### Hardware Configuration

-

### Version

v2.0.17

### IDE Name

PlatformIO

### Operating System

Windows10

### Flash frequency

80

### PSRAM enabled

yes

### Upload speed

115200

### Description

I want to perform a HTTPUpdate which was working fine before but suddenly I got read Timeout.

### Sketch

```cpp
WiFiClientSecure secureClient;
HTTPUpdate Updater;

secureClient.setInsecure();
secureClient.setTimeout(35);

Updater.rebootOnUpdate(false);
Updater.setFollowRedirects(HTTPC_FORCE_FOLLOW_REDIRECTS);

Updater.onStart([this]() { updateStart(); });
Updater.onEnd([this]() { updateSuccess(); });
Updater.onError([this](int err) { updateError(err); });
Updater.onProgress([this](int current, int total) {
    updateProgressReport(100.0 * current / total);
});

void VersionHandler::startFirmware(){
    updateStage = FIRMWARE_STAGE;
    std::string url = std::string(config.serverBaseURL) + "version?version=" + versionToDownload;
    printf("[VersionHandler] - Stage 1. Start: %s\n", url.c_str());
    HTTPUpdateResult result = Updater.update(secureClient, url.c_str(), "", [](HTTPClient *http) {
        http->addHeader("identity", config.jsonKey.c_str());
        http->addHeader("Accept", "application/octet-stream");
    });
    if( result != HTTP_UPDATE_OK ){ updateError(result); }else{
        startFileSystem();
    }
}
```

### Debug Message

```plain
[VersionHandler] - Download task started. New version: 1.0.39

[VersionHandler] - Download task started

[VersionHandler] - Stage 1. Start: https://test.hshTest.hu/version?version=1.                                 0.39

[824832][V][HTTPClient.cpp:252] beginInternal(): url: https://test.hshTest.hu/version?version=1.0.39

[824855][D][HTTPClient.cpp:303] beginInternal(): protocol: https, host: test.hshTest.hu port:                                  443 url: /version?version=1.0.39

[826704][D][HTTPClient.cpp:598] sendRequest(): request type: 'GET' redirCount: 0



[826877][V][ssl_client.cpp:62] start_ssl_client(): Free internal heap before TLS 124311

[...]

PR #9824, opened 2024-06-11
fix(client): Implement readBytes in NetworkClient for faster downloads

me-no-dev:
Fixes: https://github.com/espressif/arduino-esp32/issues/9822

## Test downloading 22MB zip and using `readBytes` with ESP32-S3

Test results are relative and actual speeds will vary based on the environment. 

### Download from HTTP server (no SSL):
```cpp
// python3 -m http.server 8080 -d .
const char *fileDownloadUrl = "http://192.168.254.91:8080/esp32-3.0.1.zip";
```

New Method:
```
Download completed!
Total download time: 40.714000 s
Average download speed: 573.149578 Kb/s
```

Old Method:
```
Download completed!
Total download time: 63.890000 s
Average download speed: 365.240443 Kb/s
```

### Download from HTTPS server (with SSL):
```cpp
const char *fileDownloadUrl = "https://github.com/espressif/arduino-esp32/releases/download/3.0.1/esp32-3.0.1.zip";
```

New Method:
```
Download completed!
Total download time: 80.371000 s
Average download speed: 290.343680 Kb/s
```

Old Method:
```
Download completed!
Total download time: 162.368000 s
Average download speed: 143.718047 Kb/s
```

### DIFF
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -483,7 +483,7 @@ size_t NetworkClient::readBytes(char *buffer, size_t length) {
   size_t left = length, sofar = 0;
   int r = 0, to = millis() + getTimeout();
   while (left) {
-    r = read((uint8_t*)buffer+sofar, left);
+    r = read((uint8_t *)buffer + sofar, left);
     if (r < 0) {
       // Error has occurred
       break;
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -481,7 +481,7 @@ int NetworkClient::read(uint8_t *buf, size_t size) {
 
 size_t NetworkClient::readBytes(char *buffer, size_t length) {
   size_t left = length, sofar = 0;
-  int r = 0, to = millis() + _timeout;
+  int r = 0, to = millis() + getTimeout();
   while (left) {
     r = read((uint8_t*)buffer+sofar, left);
     if (r < 0) {
@@ -492,7 +492,7 @@ size_t NetworkClient::readBytes(char *buffer, size_t length) {
       // We got some data
       left -= r;
       sofar += r;
-      to = millis() + _timeout;
+      to = millis() + getTimeout();
     } else {
       // We got no data
       if (millis() >= to) {
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -479,6 +479,14 @@ int NetworkClient::read(uint8_t *buf, size_t size) {
   return res;
 }
 
+size_t NetworkClient::readBytes(char *buffer, size_t length) {
+  int r = read((uint8_t*)buffer, length);
+  if (r < 0) {
+    return 0;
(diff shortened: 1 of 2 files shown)


JAndrassy:
readBytes is a function fo Stream class and should wait Stream's timeout for next byte
https://www.arduino.cc/reference/en/language/functions/communication/stream/streamreadbytes/

me-no-dev:
@JAndrassy then maybe different solution is needed. reading network streams byte by byte is never a good idea

me-no-dev:
@JAndrassy I got it, but readBytes reads byte by byte and that is also not good for Network, is what I meant

me-no-dev:
So you suggest to leave Client to read byte by byte and waste a lot of time doing it?

JAndrassy:
> So you suggest to leave Client to read byte by byte and waste a lot of time doing it?

there is `read)buffer, size)`. that should be used.

me-no-dev:
`Stream.readBytes` says nowhere that we should be waiting for every single byte, therefore we can wait when there are no bytes. Do you get this, or are we just arguing for no reason?

SuGlider:
This may affect OTA and designed time out...

SuGlider:
About `Stream::readBytes()` --> it uses `Stream::timedRead()`. Setting timeout to Zero using `Stream::setTimeout(0)` may help WiFi streams using such function.

This PR should created some other Function Name... Or at least, test timeout for Zero and in case it is not zero, use the Stream Class original function. 

This PR implementation is equivalent to ignoring Timeout setup.

SuGlider:
Please check https://github.com/espressif/arduino-esp32/issues/9823#issuecomment-2160705738
This is why timeout is important for certain Network operations.

me-no-dev:
@SuGlider yes. It should wait the timeout, but should not read a byte by byte with timeout. Nobody said this is final implementation. 

This is network client, it gets many bytes. Currently the code that triggered this was downloading a large file from internet. Now imagine reading this file byte by byte and every time checking if more are available, if it got that single byte before that timeout and so on. Result is that you can not handle anything with speeds above 1Mbps on an S3 (I imagine much less on single core and slow RV chips). If we instead grab all bytes that are available at once and wait up to the timeout for more, then we can achieve speeds 3 times higher. I hope this makes it a bit more clear for you @SuGlider and @JAndrassy

hitecSmartHome:
I'm more than happy to test if it will be compatible with arduino 2.0.17

me-no-dev:
@hitecSmartHome you can try the current changes on top of WiFiClient

JAndrassy:
so the now updated implementation in this PR should use `getTimeout()`, not `_timeout;`.

this implementation of readBytes will be used only if invoked on pointer of reference of type NetworkClient. if the pointer is  ESPLwIPClient, Client or Stream type then the reaBytes from Stream will be executed.

me-no-dev:
>  if the pointer is ESPLwIPClient, Client or Stream type then the reaBytes from Stream will be executed.

That is correct. Stream implementation works fine. This is to fix it for NetworkClient and NetworkClientSecure (and HTTPClient)

PR #9842, opened 2024-06-12
feat: added support for filters in WebServer library

ayushsharma82:
## Description of Change
This PR implements support for `setFilter` in WebServer library, particularly useful for conditional processing of routes based on filter functions set by user.

- `on` methods now return `RequestHandler` which also provides another way for users to remove routes. ( Ref: #9832 )
- Updated `RequestHandler` & `FunctionRequestHandler` class to add `setFilter` function.
- Added 2 default filters: `ON_STA_FILTER` & `ON_AP_FILTER`

## Tests scenarios
I've tested this on arduino-esp32 core v3.0.1 - ESP32 ( Wemos D1 Mini ESP32 board ).

> [!IMPORTANT]
> ~~Potential Bug: @me-no-dev , NetworkClient's `localIP()` method always returns `IPAddress(0,0,0,0)` no matter through which interface the client's request came. This breaks the `ON_STA_FILTER` & `ON_AP_FILTER` filters as we can't detect the right interface correctly.~~
> 
> ~~Eg. In all scenarios (`WIFI_STA` / `WIFI_AP` / `WIFI_AP_STA`), the `client.localIP()` method return `0.0.0.0`).~~ 
> ~~Issue: #9843~~

~~This PR is ready to be merged once the above bug is fixed.~~

Above issue is fixed and PR is ready to be merged.

## Related links
--

### DIFF
--- a/libraries/WebServer/examples/Filters/Filters.ino
+++ b/libraries/WebServer/examples/Filters/Filters.ino
@@ -71,14 +71,14 @@ void setup(void) {
   }
 
   // This route will be accessible by STA clients only
-  server.on("/sta", [&]() {
+  server.on("/", [&]() {
     digitalWrite(led, 1);
     server.send(200, "text/plain", "Hi!, This route is accessible for STA clients only");
     digitalWrite(led, 0);
   }).setFilter(ON_STA_FILTER);
 
   // This route will be accessible by AP clients only
-  server.on("/ap", [&]() {
+  server.on("/", [&]() {
     digitalWrite(led, 1);
     server.send(200, "text/plain", "Hi!, This route is accessible for AP clients only");
     digitalWrite(led, 0);
--- a/libraries/WebServer/examples/Filters/Filters.ino
+++ b/libraries/WebServer/examples/Filters/Filters.ino
@@ -17,12 +17,12 @@ WebServer server(80);
 
 const int led = 13;
 
-// ON_STA_FILTER - Accept requests coming only from the STA interface
+// ON_STA_FILTER - Only accept requests coming from STA interface
 bool ON_STA_FILTER(WebServer &server) {
   return WiFi.STA.hasIP() && WiFi.STA.localIP() == server.client().localIP();
 }
 
-// ON_AP_FILTER - Accept requests coming only from the AP interface
+// ON_AP_FILTER - Only accept requests coming from AP interface
 bool ON_AP_FILTER(WebServer &server) {
   return WiFi.AP.hasIP() && WiFi.AP.localIP() == server.client().localIP();
 }
--- a/libraries/WebServer/examples/Filters/Filters.ino
+++ b/libraries/WebServer/examples/Filters/Filters.ino
@@ -71,18 +71,28 @@ void setup(void) {
   }
 
   // This route will be accessible by STA clients only
-  server.on("/", [&]() {
-    digitalWrite(led, 1);
-    server.send(200, "text/plain", "Hi!, This route is accessible for STA clients only");
-    digitalWrite(led, 0);
-  }).setFilter(ON_STA_FILTER);
+  server
+    .on(
+      "/",
+      [&]() {
+        digitalWrite(led, 1);
+        server.send(200, "text/plain", "Hi!, This route is accessible for STA clients only");
+        digitalWrite(led, 0);
+      }
+    )
+    .setFilter(ON_STA_FILTER);
 
   // This route will be accessible by AP clients only
-  server.on("/", [&]() {
-    digitalWrite(led, 1);
-    server.send(200, "text/plain", "Hi!, This route is accessible for AP clients only");
-    digitalWrite(led, 0);
-  }).setFilter(ON_AP_FILTER);
+  server
+    .on(
+      "/",
+      [&]() {
+        digitalWrite(led, 1);
+        server.send(200, "text/plain", "Hi!, This route is accessible for AP clients only");
+        digitalWrite(led, 0);
+      }
+    )
+    .setFilter(ON_AP_FILTER);
 
   server.on("/inline", []() {
     server.send(200, "text/plain", "this works as well");
--- a/libraries/WebServer/examples/Filters/Filters.ino
+++ b/libraries/WebServer/examples/Filters/Filters.ino
@@ -1,0 +1,100 @@
+#include <WiFi.h>
+#include <NetworkClient.h>
+#include <WebServer.h>
+#include <ESPmDNS.h>
+
+// Your STA WiFi Credentials
+// ( This is the AP your ESP will connect to )
+const char *ssid = "........";
+const char *password = "........";
+
+// Your AP WiFi Credentials
+// ( This is the AP your ESP will broadcast )
+const char *ap_ssid = "ESP32_Demo";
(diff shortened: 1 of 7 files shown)


libraries/WebServer/src/WebServer.cpp
```diff
@@ -796,3 +798,14 @@ String WebServer::_responseCodeToString(int code) {
     default:  return F("");
   }
 }
+
+bool ON_STA_FILTER(WebServer &server) {
+  return (
```
me-no-dev, line 803: Suggest to use the new interfaces `WiFi.STA.localIP()` and `WiFi.AP.localIP()`

libraries/WebServer/src/WebServer.cpp
```diff
@@ -796,3 +798,14 @@ String WebServer::_responseCodeToString(int code) {
...
+    (WiFi.localIP() != IPAddress(0, 0, 0, 0) && server.client().localIP() != IPAddress(0, 0, 0, 0))
+    && WiFi.localIP() == server.client().localIP()
+  );
+}
+
+bool ON_AP_FILTER(WebServer &server) {
+  return WiFi.localIP() == server.client().localIP();
```
me-no-dev, line 810: This will check STA again. See above suggestion

libraries/WebServer/src/WebServer.h
```diff
@@ -26,6 +26,7 @@
 #include <functional>
 #include <memory>
 #include "FS.h"
+#include "WiFi.h"
```
me-no-dev, line 29: This include can go into `WebServer.cpp`. No need to be in the header.

me-no-dev:
to implement ON_STA_FILTER and ON_AP_FILTER you are changing the API much. I suggest to split that in another PR that we could add to 3.1 or whenever makes sense to add such breaking change.

ayushsharma82:
I guess everything's backward compatible with current state of this PR? But, yeah if you want, let me know what I should split.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -796,3 +798,14 @@ String WebServer::_responseCodeToString(int code) {
     default:  return F("");
   }
 }
+
```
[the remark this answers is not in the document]
ayushsharma82: Done

libraries/WebServer/src/WebServer.cpp
ayushsharma82: Done

me-no-dev:
Yes, please split them, because if someone has custom handlers, they will now break. We can add as breaking change when possible

ayushsharma82:
Okay.

Just a thought, should I try to make handlers backwards compatible 🤓? Your team can then deprecate old handlers with v3.1 or later.

me-no-dev:
If they are backward compatible, that would be OK

libraries/WebServer/src/WebServer.cpp
```diff
@@ -796,3 +803,13 @@ String WebServer::_responseCodeToString(int code) {
...
+#if SOC_WIFI_SUPPORTED
+  bool ON_STA_FILTER(WebServer &server) {
+    return WiFi.STA.localIP() == server.client().localIP();
```
me-no-dev, line 809:
```cpp
return WiFi.STA.hasIP() && WiFi.STA.localIP() == server.client().localIP();
```

Else it could return IP 0 which will also match with the client IP being 0

libraries/WebServer/src/WebServer.cpp
```diff
@@ -796,3 +803,13 @@ String WebServer::_responseCodeToString(int code) {
...
+  }
+
+  bool ON_AP_FILTER(WebServer &server) {
+    return WiFi.AP.localIP() == server.client().localIP();
```
me-no-dev, line 813:
```cpp
return WiFi.AP.hasIP() && WiFi.AP.localIP() == server.client().localIP();
```

Else it could return IP 0 which will also match with the client IP being 0

ayushsharma82:
Btw, were you able to check why client IP is always 0? I looked at this log and it showed 0 too. 

https://github.com/espressif/arduino-esp32/blob/master/libraries/WebServer/src/WebServer.cpp#L349

libraries/WebServer/src/WebServer.cpp
ayushsharma82: Done

libraries/WebServer/src/WebServer.cpp
ayushsharma82: Done with [e92f4e6](https://github.com/espressif/arduino-esp32/pull/9842/commits/e92f4e640c4aebb47235f5d290e39ea283957b5b)

me-no-dev:
localIP is fixed in https://github.com/espressif/arduino-esp32/pull/9845

me-no-dev:
here is one thing that I did not think about before. Because of ON_STA_FILTER & ON_AP_FILTER, WiFi will always get dragged, compiled and linked, regardless if it's used or not. How about you remove them from the classes and add them in example? Imagine a board with Ethernet. Basic sketch with ETH will be about half a megabyte, while one with WiFi will probably be more than a megabyte.

ayushsharma82:
I agree.

What about creating separate files for filters which can included by user? Example: `WiFiFilters.h` or `ETHFilters.h`.

me-no-dev:
they are just a few lines each, so better have them inside example.

libraries/WebServer/examples/Filters/Filters.ino
```diff
@@ -0,0 +1,100 @@
...
+  Serial.println(ssid);
+  Serial.print("IP address: ");
+  Serial.println(WiFi.localIP());
+
+  if (MDNS.begin("esp32")) {
+    Serial.println("MDNS responder started");
+  }
+
+  // This route will be accessible by STA clients only
+  server.on("/sta", [&]() {
```
me-no-dev, line 74: can both handlers be for the root path, but open depending on interface?

libraries/WebServer/src/WebServer.h
```diff
@@ -292,4 +293,12 @@ class WebServer {
   String _srealm;  // Store the Auth realm between Calls
 };
 
+/*
+  Request Filters
+*/
+
+bool ON_STA_FILTER(WebServer &server);
```
me-no-dev, line 300: don't forget to remove those

libraries/WebServer/examples/Filters/Filters.ino
ayushsharma82: I purposefully didn't use root path for both interfaces, because that would create an anti-pattern when it comes to recently added `removeRoute` logic.

On a second thought, should we update the `removeRoute` logic to remove all routes for a specific URI or just remove the first one it encounters?

libraries/WebServer/src/WebServer.h
ayushsharma82: Done

me-no-dev (approved):

me-no-dev:
if you have confirmed that localIP works, could you please edit the PR description?

ayushsharma82:
Tested and works well.

lucasssvaz (approved):

Issue #9837, opened 2024-06-12
v3.0.0 detects / defaults to wrong XTAL frequency on ESP32-D0WDQ6-V3 - millis and UART baud are not correct


PaulZC:
### Board

SparkFun ESP32 Thing

### Device Description

The SparkFun ESP32 Thing is a 3rd Party Board based on the ESP32-D0WDQ6-V3 chip with a 26MHz XTAL

https://www.sparkfun.com/products/13907

![image](https://github.com/espressif/arduino-esp32/assets/5690545/8fb0b06b-4b1f-41b6-93be-3795051fc640)

### Hardware Configuration

Only USB

### Version

v3.0.0

### IDE Name

Arduino IDE

### Operating System

Windows

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

A full write-up, diagnosis and fix available at: https://github.com/sparkfun/ESP32_Thing_Plus/issues/7

In cores below 3.0.0, the XTAL frequency was detected correctly. millis and UART baud rates were correct.

From 3.0.0, the XTAL frequency is detected as or defaults to 40MHz. millis and UART baud rates are off by a factor of 1.54.

Using the Serial_All_CPU_Freqs example:

```
For the 2.0.17 version it produces ...
CPU Freq = 240 MHz
XTAL Freq = 26 MHz
APB Freq = 80000000 Hz

For the 3.0.1 version it produces ...
CPU Freq = 240 MHz
XTAL Freq = 40 MHz
APB Freq = 80000000 Hz
```

We have a work-around:

```
#include "soc/rtc.h"

  rtc_clk_xtal_freq_update((rtc_xtal_freq_t)26);
  rtc_clk_cpu_freq_set_xtal();
  setCpuFrequencyMhz(240);
```

But it would be really nice if the core could be corrected / updated... Thanks!

### Sketch

```cpp
https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP32/examples/Serial/Serial_All_CPU_Freqs/Serial_All_CPU_Freqs.ino
```

### Debug Message

```plain
N/A
```

### Other Steps to Reproduce

N/A

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

me-no-dev:
We could not keep the crystal to AUTO, because other things stopped working. I can not recall what was exactly breaking, but it was critical.

[...]

Issue #9783, opened 2024-06-05
ESP32 Dev Module Hardware Serial doesn't work well in 3.0.0 version of boards


hojanico19:
### Board

ESP32 DEV MODULE

### Device Description

Hello, I always use ESP32 DEV MODULE for reading Hardware Serial of a sensor
Thanks

### Hardware Configuration

Hardware Serial ESP32 DEV MODULE WORK FINE IN 2.0.17 VERSION

### Version

v3.0.0

### IDE Name

Arduino IDE

### Operating System

Windows 10 and Ubuntu 24 

### Flash frequency

80 and 40

### PSRAM enabled

yes

### Upload speed

921600

### Description

Hello, I always use ESP32 DEV MODULE for reading Hardware Serial of a sensor, but I've actualizaed to 3.0.0 version of boards, and this doesn't work anymore. When I downgrade to 2.0.17 this work completely. So, for now I will use board version 2.0.17. If yo need that I test that issue ask me, for me there is no problem.

Thanks

### Sketch

```cpp
This is private of my company
```

### Debug Message

```plain
Rean wrong hexadecimal
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

SuGlider:
@hojanico19 - In order to investigate this issue, we need more information about how to reproduce such problem. One good way to allow us to assist you would be by providing a basic sketch that could be used to see the issue and test a possible fix.

SuGlider:
>Hello, I always use ESP32 DEV MODULE for reading Hardware Serial of a sensor

Please specify the sensor. Is there any open Library that we can use to reproduce the issue?

>If yo need that I test that issue ask me, for me there is no problem.

Please provide a minimum Arduino sketch that could be used to reproduce the issue.
We don't need access to your private code, just an simple minimum sketch that can be used to debug the problem.

hojanico19:
I read directly de hexadecimal and then I calculate the response. The ESP32 only read the serial info, when I use 3.0.0 the hexadecimal is wrong. I can't share more info because my company restrict it.

But you can reply it sending hexadecimal info since a device with 9600 bps and receive it by esp32, then verify that is that is correct. The rate of info es ten hexadecimal per 200 millisenconds

SuGlider:
Thanks for the explanation... but it is very hard to be able to reproduce such issue.
[...]

PR #9844, opened 2024-06-12
fix(xtal): Add a way to change the XTAL frequency for SparkFun ESP32 Thing

me-no-dev:
Add support for boards like SparkFun ESP32 Thing that use 26MHz XTAL

Fixes: https://github.com/espressif/arduino-esp32/issues/9837
Fixes: https://github.com/espressif/arduino-esp32/issues/9783

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -252,9 +252,6 @@ extern bool btInUse();
 void initArduino() {
   //init proper ref tick value for PLL (uncomment if REF_TICK is different than 1MHz)
   //ESP_REG(APB_CTRL_PLL_TICK_CONF_REG) = APB_CLK_FREQ / REF_CLK_FREQ - 1;
-#ifdef F_CPU
-  setCpuFrequencyMhz(F_CPU / 1000000);
-#endif
 #if CONFIG_SPIRAM_SUPPORT || CONFIG_SPIRAM
   psramInit();
 #endif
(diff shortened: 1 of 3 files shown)


emilfihlman:
Is this supposed to be working now?
I'm using the latest nightly and esp32 3.0.5 and running into this issue.

emilfihlman:
Initial findings:

So there was a patch, but a) it didn't patch esp32thing**_plus**, just esp32thing, ie esp32thing_plus is missing

`#define F_XTAL_MHZ 26  //SparkFun ESP32 Thing has 26MHz Crystal`
in variants/esp32thing_plus/pins_arduino.h

Secondly it still doesn't work, even though I've closed and opened and cleared /tmp/arduino etc many times for some reason and compiled and upload many times.

emilfihlman:
According to the datasheet (ESP32-WROOM-32D), this device should have 40MHz crystal, yet it definitely outputs at 74880 when Serial.begin(115200) is set.

Even stranger: after upload the device outputs at 74880, but then on plug out and back in or pressing reset button results in the device outputting on 115200 :D

Even even stranger, the device also initially outputs on the correct frequency but then switches to the wrong one after _after_

The only code on the device is Arduino RMT.

emilfihlman:
I'm putting everything in the issue instead of all three places:

https://github.com/espressif/arduino-esp32/issues/9837

emilfihlman:
Issue opened at https://github.com/espressif/arduino-esp32/issues/10388

emilfihlman:
Issue opened at https://github.com/espressif/arduino-esp32/issues/10388

PR #9851, opened 2024-06-13
fix: removeRoute should remove all matching routes (WebServer)

ayushsharma82:
## Description of Change
`removeRoute` now removes all matching routes instead of just first matching route.

## Tests scenarios
Tested on arduino-esp32 core v3.0.1

## Related links
--

### DIFF
--- a/libraries/WebServer/src/WebServer.cpp
+++ b/libraries/WebServer/src/WebServer.cpp
@@ -331,15 +331,28 @@ bool WebServer::removeRoute(const String &uri) {
 }
 
 bool WebServer::removeRoute(const String &uri, HTTPMethod method) {
-  // Loop through all request handlers and see if there is a match
+  bool anyHandlerRemoved = false;
   RequestHandler *handler = _firstHandler;
+  RequestHandler *previousHandler = nullptr;
+
   while (handler) {
     if (handler->canHandle(method, uri)) {
-      return _removeRequestHandler(handler);
+      if (_removeRequestHandler(handler)) {
+        anyHandlerRemoved = true;
+        // Move to the next handler
+        if (previousHandler) {
+          handler = previousHandler->next();
+        } else {
+          handler = _firstHandler;
+        }
+        continue;
+      }
     }
+    previousHandler = handler;
     handler = handler->next();
   }
-  return false;
+
+  return anyHandlerRemoved;
 }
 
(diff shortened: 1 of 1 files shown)


me-no-dev (approved):

libraries/WebServer/src/WebServer.cpp
```diff
@@ -331,15 +331,28 @@ bool WebServer::removeRoute(const String &uri) {
...
     if (handler->canHandle(method, uri)) {
-      return _removeRequestHandler(handler);
+      if (_removeRequestHandler(handler)) {
+        anyHandlerRemoved = true;
+        // Move to the next handler
+        if (previousHandler) {
+          handler = previousHandler->next();
+        } else {
+          handler = _firstHandler;
+        }
```
lucasssvaz, line 347: Isn't this else redundant ? The handler will already start as `_firstHandler` when `previousHandler` is null.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -331,15 +331,28 @@ bool WebServer::removeRoute(const String &uri) {
...
+        continue;
+      }
     }
+    previousHandler = handler;
     handler = handler->next();
   }
-  return false;
+
+  return anyHandlerRemoved;
 }
```
lucasssvaz, line 356: Shouldn't something simpler like this also work ? (I might be misunderstanding something)

```
bool WebServer::removeRoute(const String &uri, HTTPMethod method) {
  bool anyHandlerRemoved = false;
  RequestHandler *handler = _firstHandler;

  while (handler) {
    if (handler->canHandle(method, uri)) {
      if (_removeRequestHandler(handler)) {
        anyHandlerRemoved = true;
      }
    }
    handler = handler->next();
  }

  return anyHandlerRemoved;
}
```

ayushsharma82: This won't work because we need to remove multiple routes that match the same URI. It's now possible to have multiple routes pointing to same URI once filters are implemented.
1. `if (_removeRequestHandler(handler) && !anyHandlerRemoved)` - This will prevent removing more than one matching route
2. `previousHandler` is needed because `_removeRequestHandler` deletes the current handler and essentially 're-chains' linked list of request handlers. Therefore we have to store the previous handler to access the next handler as the current handler will be deleted after calling `_removeRequestHandler`.

image

Issue #9858, opened 2024-06-13
WiFi.setTxPower() always return +19.5dBm


enjoyneering:
### Board

ESP32 Dev Module

### Device Description

plain module

### Hardware Configuration

plain module

### Version

v3.0.1

### IDE Name

Arduino IDE

### Operating System

Windows 10

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

WiFi.setTxPower() always return +19.5dBm in AP or STA mode

I think there is a typo in the [setTxPower()](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiGeneric.cpp#L637) and [getTxPower()](https://github.com/espressif/arduino-esp32/blob/master/libraries/WiFi/src/WiFiGeneric.cpp#L645). 

### Sketch

```cpp
//in setTxPower() and getTxPower() instead of &&

if (!WiFi.STA.started() && !WiFi.AP.started()) {  //&& is wrong
    log_w("Neither AP or STA has been started");

//should be ||
if (!WiFi.STA.started() || !WiFi.AP.started()) { //|| is correct
    log_w("Neither AP or STA has been started");
```

### Debug Message

```plain
return +19.5dBm
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

JAndrassy:
"Neither AP or STA has been started" is 
`!(WiFi.STA.started() || WiFi.AP.started())` 
which is same as
`!WiFi.STA.started() && !WiFi.AP.started()`

me-no-dev:
The following code resulted in `TX Power: 11.0dBm`: 
```cpp
  WiFi.STA.begin();
  if(!WiFi.STA.waitStatusBits(ESP_NETIF_STARTED_BIT, 10000)){
    Serial.println("Failed to start WiFi!");
  }
  WiFi.setTxPower(WIFI_POWER_11dBm);
  int pwr = WiFi.getTxPower();
  Serial.printf("TX Power: %.1fdBm\n", pwr * 0.25);
  WiFi.STA.connect(SSID, PASSWORD);
```

enjoyneering:
Thank you me-no-dev. My bad. It looks like there is already the pull request for this issue. I close mine.

PR #9862, opened 2024-06-14
fix(wifi): Fix WiFi setTxPower and getTxPower

me-no-dev:
We need to wait for the interface to be started in order to be able to set/get TX Power. Code was returning too early, so calling the functions after `begin()` resulted in failure.

Fixes: https://github.com/espressif/arduino-esp32/issues/9858

### DIFF
--- a/libraries/WiFi/src/AP.cpp
+++ b/libraries/WiFi/src/AP.cpp
@@ -187,6 +187,10 @@ bool APClass::begin() {
     log_e("AP enable failed!");
     return false;
   }
+  if(!waitStatusBits(ESP_NETIF_STARTED_BIT, 1000)){
+    log_e("Failed to start AP!");
+    return false;
+  }
   return true;
 }
 
--- a/libraries/WiFi/src/STA.cpp
+++ b/libraries/WiFi/src/STA.cpp
@@ -298,6 +298,10 @@ bool STAClass::begin(bool tryConnect) {
     log_e("STA enable failed!");
     return false;
   }
+  if(!waitStatusBits(ESP_NETIF_STARTED_BIT, 1000)){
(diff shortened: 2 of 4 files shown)


lucasssvaz (approved):

SuGlider (approved):
LGTM

zebraonpc (commented):
The real Tx Power does not changed even if WiFi.getTxPower() shows correct power my esp32-s doesnt change output power it is always the same i musured it

SuGlider:
> The real Tx Power does not changed even if WiFi.getTxPower() shows correct power my esp32-s doesnt change output power it is always the same i measured it

I think that it not measuring the real Power, but just returning the setTxPower() value.

zebraonpc:
> > The real Tx Power does not changed even if WiFi.getTxPower() shows correct power my esp32-s doesnt change output power it is always the same i measured it
> 
> I think that it not measuring the real Power, but just returning the setTxPower() value.

ergo the library doesnt really work correctly with the ESP32-S

me-no-dev:
All that Arduino does is call `esp_wifi_set_max_tx_power(power)`. I suggest you as in the IDF repo and see if they have an idea why in reality power is not changed.

TD-er:
> The real Tx Power does not changed even if WiFi.getTxPower() shows correct power my esp32-s doesnt change output power it is always the same i measured it

What were the TX power values you tried?
Please be aware the ESP cannot change to just any TX power value as there are only some discrete steps.
Also the max. TX power differs based on the type of connection you use.
e.g. 802.11b typically allows for higher TX power compared to 802.11n. (typically 20-ish vs. 13 or 14 dBm respectively, may change per ESP model)

Issue #9230, opened 2024-02-09
"Bad file number" error when server closes the connection


lucasssvaz:
### Board

Any with Wi-Fi

### Device Description

Tested using ESP32 And ESP32-S3 devkits.

### Hardware Configuration

Button with PD on GPIO 35 or change GPIO in the example to the BOOT button.

### Version

latest master (checkout manually)

### IDE Name

Arduino IDE

### Operating System

Linux

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

When the server closes the connection and mbedtls notifies it using `MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY`, many `setSocketOption(): fail on 0, errno: 9, "Bad file number"` errors happen. It appears to be caused when checking connection with `WiFiClientSecure::connected()` (needs confirmation, haven't checked too deeply). It doesn't affect the functionality of the code but expected behavior is for the function to return `false` and not throw errors.

### Sketch

Example from #9103

### Debug Message

```plain
[  1227][V][esp32-hal-uart.c:396] uartBegin(): UART0 baud(115200) Mode(800001c) rxPin(3) txPin(1)
[  1236][V][esp32-hal-uart.c:482] uartBegin(): UART0 not installed. Starting installation
[  1246][V][esp32-hal-uart.c:527] uartBegin(): UART0 initialization done.
[  1362][V][esp32-hal-periman.c:229] perimanSetBusDeinit(): Deinit function for type GPIO (1) successfully set to 0x4016b9d8
[  1374][V][esp32-hal-periman.c:154] perimanSetPinBus(): Pin 0 successfully set to type GPIO (1) with bus 0x1
TOFU pegged to fingerprint: SHA256=d8:d4:da:06:9f:44:48:53:f1:32:0c:8d:80:d0:94:9f:f6:38:f1:28:a4:63:a4:0e:df:ec:4b:3d:10:2b:9f:75
Note: You can check this fingerprint by going to the URL
<https://www.howsmyssl.com> and then click on the lock icon.

Attempting to connect to SSID: Vaz_2.4GHz
[  1421][D][WiFiGeneric.cpp:1055] _eventCallback(): Arduino Event: 0 - WIFI_READY
[  1492][V][WiFiGeneric.cpp:345] _arduino_event_cb(): STA Started
[...]

PR #9869, opened 2024-06-15
ClientSecure.available() fix for connection closed by remote socket

JAndrassy:
for connection closed by remote socket the `NetworkClientSecure.available()` method:
1) logged an error. 
2) returned a negative value (the error code)

The error logging is skipped for error code returned in this case. This error code is skipped in ssl_client.cpp `_handle_error` too.

I tested it with modified WiFiClientInsecure example
```
#include <NetworkClientSecure.h>
#include <WiFi.h>

const char *ssid = "your-ssid";          // your network SSID (name of wifi network)
const char *password = "your-password";  // your network password

const char *server = "arduino.tips";  // Server URL

NetworkClientSecure client;

void setup() {
  Serial.begin(115200);
  delay(500);

  Serial.print("Attempting to connect to SSID: ");
  Serial.println(ssid);
  WiFi.begin(ssid, password);
    Serial.print(".");
    delay(1000);
  }
  Serial.print("Connected to ");
  Serial.println(ssid);

  Serial.println("\nStarting connection to server...");
  client.setInsecure();  //skip verification
  if (!client.connect(server, 443)) {
    Serial.println("Connection failed!");
  } else {
    Serial.println("Connected to server!");
    client.println("GET /asciilogo.txt HTTP/1.1");
    client.print("Host: ");
    client.println(server);
    client.println("Connection: close");
    client.println();

    while (client.connected()) {
      String line = client.readStringUntil('\n');
      if (line == "\r") {
        Serial.println("headers received");
        break;
      }
    }
    while (client.available()) {
      char c = client.read();
      Serial.write(c);
    }
    client.stop();
  }
}

void loop() {
  delay(1);
}
```
without the fix it logged
```
[  2572][E][NetworkClientSecure.cpp:308] available(): Closing connection on failed available check
[  2587][E][NetworkClient.cpp:315] setSocketOption(): fail on 0, errno: 9, "Bad file number"
�
```
the setSocketOption error and the strange character were the result of `read` which was executed because available() didn't return a 0 but a negative value.

### DIFF
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -305,9 +305,11 @@ int NetworkClientSecure::available() {
   res = data_to_read(sslclient.get());
 
   if (res < 0 && !_stillinPlainStart) {
-    log_e("Closing connection on failed available check");
+    if (res != MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY) {
+      log_e("Closing connection on failed available check");
+    }
     stop();
-    return peeked ? peeked : res;
+    return peeked;
   }
   return res + peeked;
 }
--- a/libraries/NetworkClientSecure/src/ssl_client.cpp
+++ b/libraries/NetworkClientSecure/src/ssl_client.cpp
@@ -27,7 +27,7 @@
 const char *pers = "esp32-tls";
 
 static int _handle_error(int err, const char *function, int line) {
-  if (err == -30848) {
+  if (err == MBEDTLS_ERR_SSL_PEER_CLOSE_NOTIFY) {
     return err;
   }
 #ifdef MBEDTLS_ERROR_C


libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
```diff
@@ -305,9 +305,11 @@ int NetworkClientSecure::available() {
   res = data_to_read(sslclient.get());
 
   if (res < 0 && !_stillinPlainStart) {
-    log_e("Closing connection on failed available check");
+    if (res != -30848) {
```
me-no-dev, line 308: what is `-30848`? It should be an error name

me-no-dev:
`NetworkClientSecure::available()` returns `int`, so is it wrong to return negative value?

libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
JAndrassy: it  in ssl_client.cpp _handle_error too. I didn't find that number in any header file

JAndrassy: now I found this https://github.com/Mbed-TLS/mbedtls/issues/2173

JAndrassy: and then this https://github.com/espressif/arduino-esp32/issues/9230

JAndrassy:
https://www.arduino.cc/reference/en/language/functions/communication/stream/streamavailable/
> Returns The number of bytes available to read.

me-no-dev (approved):

lucasssvaz (approved):

everslick:
Will this be backported to 2.0.x also?

me-no-dev:
we have no such plans. IDF 4.4. goes EOL next month. You are welcome to add PR and we will merge it though

PR #9893, opened 2024-06-18
HTTPClient lib - add HTTPCLIENT_NOSECURE build flag

vortigont:
Now it is impossible to use HTTPClient lib without bringing all mbedTLS stuff along, even when no https connections requred.

`HTTPCLIENT_NOSECURE` build flag disables TLS support in HTTPClient library by excluding `NetworkClientSecure.h` header.
This allows linker to strip down mbedTLS lib and certificates bundle, which in turn reduces firmware image for about ~80kib.

Could be handy for the cases when no encrypted http connections required for the sake of saving firmware size.

Tested on all HTTPClient lib examples, firmware image size reduction is about 80kib, RAM reduction is about 600 bytes.

origin:
```
Per-archive contributions to ELF file:
    Archive File     DRAM .data & 0.bss IRAM0 .text & 0.vectors ram_st_total Flash .text & .rodata  & flash_total
    libmbedcrypto.a         144      85          30           0          259       83098     52516         135788
    libmbedtls_2.a           72     488           0           0          560       32160      7339          39571

RAM:   [=         ]  13.3% (used 43712 bytes from 327680 bytes)
Flash: [========  ]  78.5% (used 1029069 bytes from 1310720 bytes)
```

this PR:
```
Per-archive contributions to ELF file:
    Archive File     DRAM .data & 0.bss IRAM0 .text & 0.vectors ram_st_total Flash .text & .rodata    flash_total
    libmbedcrypto.a         144      85          30           0          259       67325     41566         109065

RAM:   [=         ]  13.2% (used 43144 bytes from 327680 bytes)
Flash: [=======   ]  72.0% (used 943213 bytes from 1310720 bytes)
```

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -142,7 +142,9 @@ bool HTTPClient::begin(NetworkClient &client, String url) {
   _secure = (protocol == "https");
 
 #ifdef HTTPCLIENT_NOSECURE
-  if (_secure) return false;
+  if (_secure) {
+    return false;
+  }
 #endif  // HTTPCLIENT_NOSECURE
   return beginInternal(url, protocol.c_str());
 }
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -142,10 +142,9 @@ bool HTTPClient::begin(NetworkClient &client, String url) {
   _secure = (protocol == "https");
 
 #ifdef HTTPCLIENT_NOSECURE
-  return _secure ? false : beginInternal(url, protocol.c_str());
-#else
-  return beginInternal(url, protocol.c_str());
+  if (_secure) return false;
 #endif  // HTTPCLIENT_NOSECURE
+  return beginInternal(url, protocol.c_str());
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vortigont:
@Jason2866 maybe this could be used in Tasmota's Arduino core. As I remember, it excludes mbedTLS completely, and attempt to use `HTTPClient` lib with Tasmota's core results in linker error for `HTTPClient::begin` methods. With this PR it links fine.

mathieucarbou:
I noticed also in my fork of ESPAsyncWebServer (https://github.com/mathieucarbou/ESPAsyncWebServer). I was able to remove the mbedtls lib on Arduino 2 by using other hash implementations, but on Arduino 3, mbedtls comes back from core so cannot be skipped. Would be nice to have a way to more dynamically reduce the firmware size because libraries and code tend to use more and more space always, but our ESP32 devices are for most part still with the same flash memory.

Jason2866:
@vortigont Thx for heads up. Don't see an issue using http client in our fork since the client secure code parts are completly removed

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -214,6 +221,9 @@ bool HTTPClient::begin(String url) {
   clear();
   _port = 80;
   if (!beginInternal(url, "http")) {
+#ifdef HTTPCLIENT_NOSECURE
+    return false;
+#endif  // HTTPCLIENT_NOSECURE
```
me-no-dev, line 226: add `#else` to not have double returns.

vortigont: Done, fixed similar in other places too

PR #9902, opened 2024-06-19
fix(version): Update version to 3.0.2

me-no-dev:
Also enables copy of the jsons to gh-pages


yoursunny:
Version number is incremented, but there's no git tag and the new version is not showing up in the Arduino IDE.

me-no-dev:
yes, it's not released/tagged yet

PR #9948, opened 2024-06-28
feat(net): Enable IPv6 immediately if already connected

me-no-dev:
If the interface is already connected, try to enable IPv6 immediately. Otherwise the interface would need to go through disconnect/connect cycle for IPv6 to be enabled.


SuGlider (approved):
LGTM

Issue #9990, opened 2024-07-05
OTA and upload of larger files fails


TD-er:
### Board

Any with large enough flash

### Device Description

This is about larger uploads of either the sketch, or file to the filesystem.
So any board with large enough flash to have an upload of files > 2 MB

### Hardware Configuration

-

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

40MHz

### PSRAM enabled

yes

### Upload speed

115200

### Description

OTA updates of larger sketches (typically > 2 MB in size) lately tend to fail.
The device then is inaccessible for 10 - 15 minutes.

On the serial console, you'll see the dots per 1436 bytes of upload and they typically stop at 1100 - 1600 dots.
Thus somewhere between 1.4 - 2.2 MB and never seem to make it to the full 2.5 MB needed for my "max" build bin files.

I found 2 issues here.

This function does no longer have a timeout implemented:
https://github.com/espressif/arduino-esp32/blob/4e3523c212375ef65256d09506ee89425dbf5958/libraries/WebServer/src/Parsing.cpp#L346-L358

Related issue: https://github.com/espressif/arduino-esp32/issues/9163
Related PR: https://github.com/espressif/arduino-esp32/pull/9167/files

IMHO the old implementation of this function was just fine:
```c++
int WebServer::_uploadReadByte(WiFiClient& client){
  int res = client.read();
  if(res < 0) {
    // keep trying until you either read a valid byte or timeout
    unsigned long startMillis = millis();
    long timeoutIntervalMillis = client.getTimeout();
    boolean timedOut = false;
    for(;;) {
      if (!client.connected()) return -1;
      // loosely modeled after blinkWithoutDelay pattern
      while(!timedOut && !client.available() && client.connected()){
        delay(2);
        timedOut = millis() - startMillis >= timeoutIntervalMillis;
      }

      res = client.read();
      if(res >= 0) {
        return res; // exit on a valid read
      }
      // NOTE: it is possible to get here and have all of the following
      //       assertions hold true
      //
      //       -- client.available() > 0
      //       -- client.connected == true
      //       -- res == -1
      //
      //       a simple retry strategy overcomes this which is to say the
      //       assertion is not permanent, but the reason that this works
      //       is elusive, and possibly indicative of a more subtle underlying
      //       issue

      timedOut = millis() - startMillis >= timeoutIntervalMillis;
      if(timedOut) {
[...]

PR #9991, opened 2024-07-05
Fix timeout in WebServer::_uploadReadByte and handleClient() (#9990)

TD-er:
Fixes: #9990

-----------
## Description of Change
See #9990 

In short:
- WebServer::_uploadReadByte() does not keep track of some timeout
- WebServer::handleClient() does only set timeout to client after the function call where it is needed.


P-R-O-C-H-Y (approved):
LGTM

libraries/WebServer/src/WebServer.h
```diff
@@ -88,7 +88,7 @@ typedef struct {
   HTTPRawStatus status;
   size_t totalSize;    // content size
   size_t currentSize;  // size of data currently in buf
-  uint8_t buf[HTTP_UPLOAD_BUFLEN];
+  uint8_t buf[HTTP_RAW_BUFLEN];
```
me-no-dev, line 91: why is this change needed?

TD-er: Because somewhere else this buffer is used with this defined size :)
Was a copy/paste error.

me-no-dev (approved):

PR #10002, opened 2024-07-10
FS library position() to return (size_t) -1 on error

### LINKED ISSUE #9992: FS position method return valid position on error, should return -1
### Board

All

### Device Description

ESP32
ESP8266 code base has the same issue

### Hardware Configuration

all 

### Version

### IDE Name

Arduino

### Operating System

Windows 10

### Flash frequency

40Mhz

### PSRAM enabled

yes

### Upload speed

115200

### Description

Current code returns valid file position on error, should return invalid position, i.e. -1

Current code is
```
size_t File::position() const {
  if (!*this) {
    return 0;
  }
  return _p->position();
}
```

should be
```
size_t File::position() const {
  if (!*this) {
    return -1;  // <<<<<<<<<<<<<<<<
  }
  return _p->position();
}
```

### Sketch

```cpp
Not needed
```

### Debug Message

```plain
None
```

### Other Steps to Reproduce

_No response_

drmpf:
Fix for error return from position()
PR Issue #9992

## Description of Change
Currently FS::position() returns 0 if file is invalid,  this change return (size_t) -1 instead, which is a very large number, to indicate an error 

## Tests scenarios
none

## Related links
Closes #9992

### DIFF
--- a/libraries/FS/src/FS.cpp
+++ b/libraries/FS/src/FS.cpp
@@ -105,7 +105,7 @@ bool File::seek(uint32_t pos, SeekMode mode) {
 
 size_t File::position() const {
   if (!*this) {
-    return 0;
+    return (size_t)-1;
   }
 
   return _p->position();
--- a/libraries/FS/src/FS.h
+++ b/libraries/FS/src/FS.h
@@ -64,7 +64,7 @@ public:
   bool seek(uint32_t pos) {
     return seek(pos, SeekSet);
   }
-  size_t position() const; // returns (size_t)-1 on error
+  size_t position() const;  // returns (size_t)-1 on error
   size_t size() const;
   bool setBufferSize(size_t size);
   void close();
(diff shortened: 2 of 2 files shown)


me-no-dev (approved):

P-R-O-C-H-Y (approved):

me-no-dev:
@igrr it is now exactly the same as in official SD lib. While surely we do agree that unsigned returning -1 is wrong, it is what ends up being done upstream.

drmpf:
No what is actually being returned here is NOT -1,
but rather ULONG_MAX
i.e. 4294967295
(size_t)-1  is just a short hand.
You can replace it with ULONG_MAX if you wish and still get the same result.

me-no-dev:
Yes @drmpf , but `ULONG_MAX` is not an invalid size. It just assumes that you will not have files that are that large.

drmpf:
Yes but ULONG_MAX a very file position and far more unlikely a return value for a file position than 0.
The only way to avoid that is to use an out-of-band error return like ferror,
as the design of the method does not allow for a separate error turn.
You could change the method to
bool position(size_t &rtn);
but that is not backward compatible.

PR #10018, opened 2024-07-12
Fix timeout in WebServer::_uploadReadByte and handleClient()

### LINKED ISSUE #9990: OTA and upload of larger files fails

Jason2866:
Backport of https://github.com/espressif/arduino-esp32/pull/9991
Fixes #9990

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -31,8 +31,8 @@
 #endif
 
 #define __STR(a) #a
-#define _STR(a) __STR(a)
-const char * _http_method_str[] = {
+#define _STR(a)  __STR(a)
+static const char *_http_method_str[] = {
 #define XX(num, name, string) _STR(name),
   HTTP_METHOD_MAP(XX)
 #undef XX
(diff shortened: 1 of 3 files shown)


me-no-dev:
@Jason2866 PR is failing. Due to the style changes though, it's hard to track what really changed and what is missing. Will you fix this?

Jason2866:
@me-no-dev to much effort.

Issue #9712, opened 2024-05-28
WIFI can no longer use hostname in HTTP URL


thammr:
### Board

ESP32S2 Development Board

### Device Description

ESP32S2 Development Board

### Hardware Configuration

Two VL53L1Xs

### Version

v3.0.0

### IDE Name

Arduino IDE

### Operating System

Windows 11

### Flash frequency

OTA

### PSRAM enabled

no

### Upload speed

N/A

### Description

My ESP32S2 POSTs a data packet to InfluxDb every 30-40 seconds. After recompiling my code today, now POSTing the data fails after about 5-7 packets. Wireshark shows that the ESP32 starts trying to send packets to 0.0.0.0. However, if I replace hostname with the IP, everything works fine. While the ESP32S2 is failing to POST data to InfluxDb, all the other networking stuff works OK- MQTT, OTA, ping.

The board manager shows version 3.0.0-rc3. I tried downgrading to rc2 - same result, Downgraded to 2.0.16 - works OK now.

I have at least 6 ESP32s sending data to InfluxDb without problem. All use hostname in the URL. 

   uri = String("http://") + host + ":8086/api/v2/write?org=" + org + "&bucket=" + bucket + "&precision=s";

The following code running on a Tiny Pico starts returning -1 after several minutes, Some time later, it will start returning a valid status (405 in my case). This pattern repeats. 

### Sketch

```cpp
#include <WiFi.h>
#include <HTTPClient.h>
#include "arduino_secrets.h"

String postData = "Hello";
String url = "http://myhost:8080/";

void setup() {
  Serial.begin(115200);
  delay(2000);
  WiFi.mode(WIFI_STA);
  WiFi.begin(ssid, pass);
  while (WiFi.status() != WL_CONNECTED) {
    Serial.print('.');
    delay(1000);
  }
}


void loop() {

  WiFiClient client;
  HTTPClient http;

  http.begin(client, url);

  http.addHeader("Content-Type", "text/plain; charset=utf-8");
  http.addHeader("Content-Length", String(postData.length()));
  int statusCode = http.POST(postData);
  
  http.end();
  Serial.println(statusCode);
  delay(1000);
}
```

### Debug Message

```plain
N/A
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

me-no-dev:
[...]

PR #10037, opened 2024-07-16
fix(lwip): Filter out V6 packets if V6 is not enabled

me-no-dev:
the ESP might accept and act on some IPv6 packets, even though IPv6 is not enabled for the interface in Arduino. This change makes the ESP ignore all IPv6 packets if IPv6 address is not available.

Depends on https://github.com/espressif/esp32-arduino-lib-builder/pull/194
Fixes https://github.com/espressif/arduino-esp32/issues/9712


me-no-dev:
Closing as part of https://github.com/espressif/arduino-esp32/pull/10038

PR #10043, opened 2024-07-17
PPP: add "battery status" read stub functions

WebDust21:
## Description of Change

 - Add Arduino stub functions to access the functionality provided by the underlying ESP-MODEM "esp_modem_get_battery_status" call
 - Replaced verbose checks on each Arduino stub function call with a single succint #define for ease of code maintainability.  

## Tests scenarios
Tested on an ESP32-S2 with a SIM7080G modem.  Love the PPP functionality, it just *works*!  Just providing some functional extensibility to functionality that is already present. 

```
    testClient("google.com", 80);
    Serial.print("RSSI: "); Serial.println(PPP.RSSI());
    Serial.print("BtyVolt: "); Serial.println(PPP.getBatteryVoltage());
```
output result
```
[...]
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
Connection Success
RSSI: 18
BtyVolt: 3846
```
showing that the modem VCC power supply is 3.846v.

## Related links
ESP-MODEM "esp_modem_get_battery_status": https://github.com/espressif/esp-protocols/blob/906e447193710c1772305a218a9d0b315a549de4/components/esp_modem/src/esp_modem_c_api.cpp#L280-L293

Initially I thought that "esp_modem_get_battery_status" was leveraging AT+CADC, but it's not: it's using AT+CBC: https://github.com/espressif/esp-protocols/blob/906e447193710c1772305a218a9d0b315a549de4/components/esp_modem/src/esp_modem_command_library.cpp#L207-L234

### DIFF
--- a/libraries/PPP/src/PPP.cpp
+++ b/libraries/PPP/src/PPP.cpp
@@ -8,12 +8,14 @@
 #include "driver/uart.h"
 #include "hal/uart_ll.h"
 
-#define PPP_CMD_MODE_CHECK(x) if (_dce == NULL) {return x;}      \
+#define PPP_CMD_MODE_CHECK(x)                                    \
+  if (_dce == NULL) {                                            \
+    return x;                                                    \
+  }                                                              \
   if (_mode == ESP_MODEM_MODE_DATA) {                            \
     log_e("Wrong modem mode. Should be ESP_MODEM_MODE_COMMAND"); \
     return x;                                                    \
-  }                                                              \
-
+  }
 
 typedef struct {
   void *arg;
@@ -424,7 +426,7 @@ void PPPClass::end(void) {
 
 bool PPPClass::sync() const {
   PPP_CMD_MODE_CHECK(false);
-  
+
   return esp_modem_sync(_dce) == ESP_OK;
 }
 
@@ -495,7 +497,7 @@ bool PPPClass::setPin(const char *pin) {
 
 int PPPClass::RSSI() const {
   PPP_CMD_MODE_CHECK(-1);
-  
+
   int rssi, ber;
   esp_err_t err = esp_modem_get_signal_quality(_dce, rssi, ber);
   if (err != ESP_OK) {
@@ -532,7 +534,7 @@ String PPPClass::IMSI() const {
 
 String PPPClass::IMEI() const {
   PPP_CMD_MODE_CHECK(String());
-  
+
   char imei[32];
   esp_err_t err = esp_modem_get_imei(_dce, (std::string &)imei);
   if (err != ESP_OK) {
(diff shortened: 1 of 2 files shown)


libraries/PPP/src/PPP.h
```diff
@@ -41,6 +41,9 @@ class PPPClass : public NetworkInterface {
   // Modem DCE APIs
   int RSSI() const;
   int BER() const;
+  int getBatteryVoltage() const;
+  int getBatteryLevel() const;
+  int getBatteryStatus() const;
```
me-no-dev, line 46: to conform with the rest of the naming, I will suggest that you drop `get` and have just `batteryVoltage()` etc.

me-no-dev: also maybe put them at the end of the paragraph

libraries/PPP/src/PPP.cpp
```diff
@@ -737,15 +647,44 @@ bool PPPClass::setBaudrate(int baudrate) {
...
-bool PPPClass::sms(const char *num, const char *message) {
-  if (_dce == NULL) {
-    return false;
+int PPPClass::getBatteryVoltage() const {
+  PPP_CMD_MODE_CHECK(-1);
+
+  int volt, bcs, bcl;
+  esp_err_t err = esp_modem_get_battery_status(_dce, volt, bcs, bcl);
+  if (err != ESP_OK) {
+    //log_e("esp_modem_get_battery_status failed with %d %s", err, esp_err_to_name(err));
```
me-no-dev, line 656: why are the log lines commented out?

libraries/PPP/src/PPP.h
WebDust21: OK, makes sense.

libraries/PPP/src/PPP.cpp
WebDust21: Well, I copied the functionality from other functions in PPP.cpp (RSSI + BER), which had them commented out:
https://github.com/espressif/arduino-esp32/blob/690bdb511d9f001e2066da2dda2c631a3eee270f/libraries/PPP/src/PPP.cpp#L515
https://github.com/espressif/arduino-esp32/blob/690bdb511d9f001e2066da2dda2c631a3eee270f/libraries/PPP/src/PPP.cpp#L534

I guess I'm not sure why those two are commented out...when none of the others in the entire file are.

That's literally the only reason I commented them out...

me-no-dev (approved):

libraries/PPP/src/PPP.cpp
WebDust21: LOL, OK...well...I uncommented those two lines referenced above in the latest commit...oops?

I would not expect those commands to be "frequent failures", but maybe they are problematic.  No issues on my end though.

PR #10066, opened 2024-07-23
Enabled the onboard ceramic antenna by default when creating a new project with XIAO_ESP32C6

Cincinnatu:

### Checklist
5. [ ] Please **confirm option to "Allow edits and access to secrets by maintainers"** when opening a Pull Request

-----------
## Description of Change

## Tests scenarios

## Related links


P-R-O-C-H-Y:
Hello @Cincinnatu, can you please provide some description about the change and also add some related links?
I will also recommend to define the antenna pins in `variants/XIAO_ESP32C6/pins_arduino.h` and use the definitions instead of PIN number in variant.cpp.

Jason2866 (changes requested):
the comment is wrong. Correct is `//use onboard antenna`
The code is not needed at all, since using onboard antenna is default (without any setup). Only needed when the extern antenna should be activated.

Cincinnatu:
Due to the hardware design of XIAO_ESP32C6, the WIFI antenna will not be turned on by default during use. Some of our users may not know how to turn on the WIFI by themselves, therefore, so I have submitted this PR.

Cincinnatu:
> 您好，您能否提供一些有关更改的描述并添加一些相关链接？我还建议定义天线引脚，并在 variant.cpp 中使用定义而不是 PIN 号。`variants/XIAO_ESP32C6/pins_arduino.h`
Sure, I'll add it later.

Cincinnatu:
> 您好，您能否提供一些有关更改的描述并添加一些相关链接？我还建议定义天线引脚，并在 variant.cpp 中使用定义而不是 PIN 号。`variants/XIAO_ESP32C6/pins_arduino.h`

Sure, I'll add it later.

variants/XIAO_ESP32C6/variant.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "pins_arduino.h"
+
+extern "C" {
+
+void initVariant(void) {
+  pinMode(WIFI_ENABLE, OUTPUT);
+  digitalWrite(WIFI_ENABLE, LOW);//turn on this function
+
+  pinMode(WIFI_ANT_CONFIG, OUTPUT); 
+  digitalWrite(WIFI_ANT_CONFIG, LOW);//use external antenna
```
P-R-O-C-H-Y, line 19: According to the docs for XIAO_ESP32C6 this line is not true right now:


> GPIO14 is used to select between using the built-in antenna or an external antenna. Before that, you need to set GPIO3 low level to turn on this function. If GPIO14 is set low level, it uses the built-in antenna; if it set to high level, it uses the external antenna. Default is low level. If you want to set it high, you can refer the code below.
```
void setup() {
  pinMode(3, OUTPUT);
  digitalWrite(3, LOW);//turn on this function
  delay(100);
  pinMode(14, OUTPUT); 
  digitalWrite(14, HIGH);//use external antenna
}

```
```suggestion
  digitalWrite(WIFI_ANT_CONFIG, LOW);//use built-in antenna, set HIGH to use external antenna
```

Jason2866:
@Cincinnatu 
> Due to the hardware design of XIAO_ESP32C6, the WIFI antenna will not be turned on by default during use. Some of our users may not know how to turn on the WIFI by themselves, therefore, so I have submitted this PR.

So the behaviour described how it works is wrong at the Web page!
This needs to be corrected.

me-no-dev:
I would suggest adding the following defines to `pins_arduino.h` to help the users:
```cpp
#define useInternalAntenna() digitalWrite(WIFI_ANT_CONFIG, LOW)
#define useExternalAntenna() digitalWrite(WIFI_ANT_CONFIG, HIGH)
```

Then maybe that to be documented in the board web page.

Is there a case where the WiFi should not be enabled?

Jason2866:
> Is there a case where the WiFi should not be enabled?

You never know all uses cases. If it is possible to switch off, there will be probably a use case found for.

P-R-O-C-H-Y (approved):

PR #10076, opened 2024-07-26
fix (library): ESP32-NOW peer removal

craiglink:
in the case were _esp_now_del_peer, the peer could never be added again because `added` would always be true and never reset.  This forces it to be reset

-----------
## Description of Change
in the case where _esp_now_del_peer API calls,, the peer could never be added again by the C++ class because `added` would always be true and never reset.  This forces 'added' to be reset, regardless of the api call.

## Tests scenarios
Fixes a issue in WLED code where a broadcast peer couldn't be added a second time during restart.  Run on an ESP32 dig2go board.


VojtechBartoska:
Thank you for the PR @craiglink, we will review it soon 👍

P-R-O-C-H-Y:
Hello @craiglink, Is there any way on how to reproduce the issue and see if this is needed at all?
From my point of view the `added = false;` is on right place, after successful return from `_esp_now_del_peer();`
If this function fails, it means the peer was not deleted, so it makes sense to keep `added = true`. 
Thank you for more details.

craiglink:
The problem with only setting `added = false` after a successful peer deletion is that the peer list also gets cleared if `esp_now_deinit()` is called.  However this calls exist in a different calls, thus it leaves the peer class in a bad state which can never be reset as its currently written.  A mechanism or codepath is need to reset "added" if  `deinit()` was called.  Or just always set added to false when unregistering a peer

me-no-dev:
@craiglink in that case proper solution would be to seta ll peers to `false` if `esp_now_deinit()` is called (maybe with helper function)

awong1900:
It is break change from 2.0 to 3.0 on esp now. I will watch this.

craiglink:
That code path is actually broken too.  I actually gave up on using this class after seeing all the issues because it seem to brittle.

for example
https://github.com/espressif/arduino-esp32/blob/7fa3afdd0cd4232514c62c52e33168518f977499/libraries/ESP_NOW/src/ESP32_NOW.cpp#L189-L202

Similarly if deinit() fails, the peers list won't be reset even though they no longer exist
And you'd want to create a friend function in Peer that ESP_NOW_Class could call to detach the peer.

P-R-O-C-H-Y:
@craiglink @awong1900 I am closing this in favour of #10102.
Feel free to review or comment the new PR. Thanks

Issue #9974, opened 2024-07-01
Chip Report Does Not Recognize Embedded Flash or Embedded PSRAM


Xylopyrographer:
### Board

seeed studio XIAO ESP32S3, Waveshare ESP32-S3-Matrix

### Device Description

Nothing else attached.

### Hardware Configuration

Nothing else attached.

### Version

v3.0.1

### IDE Name

Arduino IDE v2.3.2

### Operating System

Debian-GNU 12

### Flash frequency

240

### PSRAM enabled

yes

### Upload speed

921600

### Description

When running the sketch below with "Core Debug Level: Debug" on boards using ESP32-S3 chips with in-package (Embedded) flash and/or with in-package (Embedded) PSRAM, the Chip Debug Report shows "No" for Embedded Flash and/or Embedded PSRAM.

Tested using:
seeed studio XIAO ESP32S3 which uses an:
ESP32-S3R8 (8MB Octal SPI In-Package PSRAM)

Waveshare ESP32-S3-Matrix which uses an:
ESP32-S3FH4R2 (4 MB Quad SPI In-Package Flash and 2MB Quad SPI In-Package SPI PSRAM)
Note: As there is no board definition for this device, tested with the following boards selected:
ESP32S3 Dev Module
Lolin S3 Mini

Other boards using the ESP32-S3FH4R2 include:
WeAct Studio ESP32S3-Mini
Waveshare ESP32-S3 Mini
Waveshare ESP32-S3-Zero

### Sketch

```cpp
//
// Q&D sketch to see if embedded flash and/or PSRAM is reported.
//  - see '../esp32-3.0.1/cores/esp32/chip-debug-report.h'

#include "chip-debug-report.h"
#include <esp_psram.h>
#define PS_BUFFER 255

void setup() {

  printBeforeSetupInfo();   // comment out this line if Core Debug Level is set to 'Debug' or 'Verbose'

  Serial.begin(115200);
  Serial.print("\n\nPSRAM Found: ");
  if (psramFound()) {
    Serial.println("Yes");
    psramInit();
    Serial.print("On init, avaialble PSRAM: ");
    Serial.println(ESP.getFreePsram());
    Serial.println("Allocating and filling PSRAM data array.");
    uint8_t* data = (uint8_t*)ps_malloc(PS_BUFFER);
    for (int i = 0; i < PS_BUFFER; i++) {
      data[i] = i;
    }
    Serial.print("After write, avaialble PSRAM: ");
[...]

PR #10080, opened 2024-07-26
feat: Print chip features bitfield

Xylopyrographer:
## Description of Change
Adds printing the value of the chip features bitfield to printChipInfo() to facilitate debugging.

## Tests scenarios
Tested with:

- Arduino IDE v2.3.2
- arduino-esp32 core v3.0.3
- Debian GNU/Linux 12
- Waveshare ESP32-S3-Matrix dev board

## Related links
None


VojtechBartoska:
thanks for the PR @Xylopyrographer, we will review it soon.

me-no-dev:
Question is what does this help if the bits are examined in the following lines? This will just print a meaningless (to the users and us) hex, that we would need to examine anyway. Do you mean that there might be bits not documented?

VojtechBartoska:
@Xylopyrographer any comments? Thanks

Xylopyrographer:
@me-no-dev, @VojtechBartoska 

> Question is what does this help if the bits are examined in the following lines? This will just print a meaningless (to the users and us) hex, that we would need to examine anyway. Do you mean that there might be bits not documented?

So this stems from #9974 where the chip report from a board using an ESP32-S3FH4R2 fails to report that the chip has both on-board flash and PSRAM. To help diagnose further, I had to add the raw chip features dump.

Debugging and diagnostic tools should report accurately, this demonstrates a case where something is amiss. It would also be useful should new silicon be ahead of the chip report as it could be used as a check and yes, show undocumented features.

As for interpretation, the bits are enumerated and can be checked (albeit with a bit more digging) but at least the base information is there to investigate further if needed.

FWIW, haven't yet carried the investigation into the above error further. Not reported in IDF 4.4? Definition of "embedded" means something else? Does it show only foe ESP32 modules?

PR #10089, opened 2024-07-30
esp32-hal-ledc.c: Clear bit in used_channels bitmap

nuclearcat:
When we attach pin to LEDC channel we set bit in used_channels bitmap variable. But used_channels bit was not cleared on detach,
as result simply repeating attach/detach around 16 times will falsely use all available channels and
emit error that channel is not available.

This happen for example in IRSenderESP32.cpp.
This patch fixes this bug.

Sorry for not providing issue, as it require filling big amount of irrelevant form fields for just 2 lines of code.
To reproduce issue enable verbose debug and run in loop:
ledcAttach(13, 12000, 8);
ledcDetach(13);
It will fail, as i recall after 15 or 16th loop cycle.


P-R-O-C-H-Y:
Hello @nuclearcat, thank you very much for providing a PR and reporting the issue.
The fix is actually not on right place, the used_channel bits are cleared in `ledcDetachBus` function, called by Peripheral Manager. I have opened a PR #10094 with a proper fix, so I am closing this PR. Feel free to test with the changes from the PR I linked. Anyway thank you for pointing this out! :)

PR #10128, opened 2024-08-07
Add ability for boards to provide a custom pixel order in neopixelWrite()

sblantipodi:
The hint for the builtin neopixleWrite() function has the wrong information. 
It shows
`void neopixelWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val)`
but then the color order is swapped to GRB.

Supposed that we want to use RGB color order as default one, function must match this order.

The BlinkRGB.ino example also is incorrect in the color order with the current implementation.


SuGlider:
@sblantipodi - Have you tested that?
WS2812B protocol order is GRB.

sblantipodi:
@SuGlider sure... 
WS2812B protocol does not have a specific color order, color orders depends on the LEDs you are using and how it is connected to the board.

Lolin ESP32-C3 Mini for example uses RGB color order,
Lolin ESP32-S3 Mini uses GRB color order.

Programmers should know what kind of color order to use for the board of their interest and call that function accordingly.

To light up the LED red on Lolin C3 I need to call 
  neopixelWrite(LED_BUILTIN, 255, 0, 0); 

To light up the LED red on Lolin S3 I need to call 
  neopixelWrite(LED_BUILTIN, 0, 255, 0); 

Having a double conversion inside the neopixelWrite function is somewhat misleading and not really needed in my opinion.

Please take into consideration that color orders may varies a lot, from my experience there are a lot of leds using those color order:
- GRB
- RGB
- BGR
- BRG
- RBG
- GBR
even though RGB and GRB are the most used.

SuGlider:
@sblantipodi - I see. Thanks for the explanation.
In that case, shall we change this PR to consider that the user may set a color order before using `neoPixelWrite()`?
Maybe a separated API for it?

SuGlider:
@sblantipodi - The way the PR is set, it doesn't work for the ESP DevKits that I have here. The order used by their LEDs ir GRB.

sblantipodi:
> @sblantipodi - I see. Thanks for the explanation. In that case, shall we change this PR to consider that the user may set a color order before using `neoPixelWrite()`? Maybe a separated API for it?

 @SuGlider yes, that will be the best way to do it IMHO.
I can change this PR according to this...

give me some minutes :)

sblantipodi:
@SuGlider 
the old function neopixelWrite() continue to work as before but I changed it from this:
`void neopixelWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val) `
to this
`void neopixelWrite(uint8_t pin, uint8_t green_val, uint8_t red_val, uint8_t blue_val)`

just to make it clear that the function is using GRB color order and not RGB.

I added a
`void neopixelWriteOrdered(uint8_t pin, color_order_t color_order, uint8_t red_val, uint8_t green_val, uint8_t blue_val)`
function to allow the user to select the preferred color order without going crazy with inverting the values.

may it work well for you?

I tested it on S3 from Lolin, Adafruit, UnexpectedMaker
and on a C3 from Lolin.

ESP DevKits will continue to work as before.

me-no-dev:
please leave arguments order to be red green blue as this is the logical order of colors used everywhere

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -11,8 +11,19 @@ extern "C" {
...
 #endif
 
+typedef enum {
+  RGB,
+  BGR,
+  BRG,
+  RBG,
+  GBR,
+  GRB
+} color_order_t;
```
me-no-dev, line 21: naming it so simply is generally not a great idea, because it would clash with any other lib out there  that might have gone the same way. using enum like `rgb_led_color_order_t` and values like `LED_COLOR_RODER_RGB` would lower the possibility of such clash to a minimum.

sblantipodi: you're right, done it.

SuGlider:
There is point here: it shall be as simple as possible and address the onboard LED only.
@sblantipodi said that some other boards use a different RGB order than the one set for Espressif onboard neopixel.

In order to simplify it, a suggestion would be to define the RGB order in the `pins_arduino.h`, which is specific for the board.

The API would continue to be **just** `neopixelWrite(pin, Red, Green, Blue)` and the function would use this board setting, if it is defined, to decide how to send the colors to the WS2812B LED. If nothing is defined, it uses the default which would be R-G-B order.

Example:
`variants/esp32c3/pins_arduino.h` would define
``` cpp 
#define RGB_BUILTIN_ORDER      LED_COLOR_ORDER_RGB
```

`cores/esp32/esp32-hal-rgb-led.h` would just declare the list:
``` cpp
typedef enum {
  LED_COLOR_ORDER_RGB,
  LED_COLOR_ORDER_BGR,
  LED_COLOR_ORDER_BRG,
  LED_COLOR_ORDER_RBG,
  LED_COLOR_ORDER_GBR,
  LED_COLOR_ORDER_GRB
} rgb_led_color_order_t;
```

`cores/esp32/esp32-hal-rgb-led.c` would do something like:
``` cpp
rgb_led_color_order_t rgb_led_color_order = LED_COLOR_ORDER_RGB;
#ifdef RGB_BUILTIN_ORDER
rgb_led_color_order = RGB_BUILTIN_ORDER;
#endif

// + the switch() and so on as proposed in the PR ....
```

----------
For instance, the Lolin ESP32-S3 Mini that uses GRB color order, would declare in its own `pins_arduino.h` file the correct order for its neopixel:

#define RGB_BUILTIN_ORDER      LED_COLOR_ORDER_GRB

---------

Would it be better @robertlipe, @sblantipodi and @me-no-dev?
Let's talk and implement it.

sblantipodi:
> TL;DR: I'm highly skeptical there is an RGB WS2812b. Stick your scope on the data pin and if they're real WS2812b's or even [WS2812c](https://cdn.sparkfun.com/assets/e/1/0/f/b/WS2812C-2020_V1.2_EN_19112716191654.pdf), they're GRB at the packet level.

@robertlipe I double checked some of my ESPs and the Lolin ESP32-S3 is using RGB and not GRB.  

As per WS2812B in general [my firmware](https://github.com/sblantipodi/glow_worm_luciferin) has lots of users using WS2812B RGB strips leave alone the more popular [WLED](https://kno.wled.ge/)

SuGlider:
@sblantipodi @robertlipe - Let's be practical.

@sblantipodi says that he has a couple boards that have an RGB LED that works with a different color order. This can be tested using ESP32 Arduino `neopixelWrite(pin, 255, 0, 0);`. If the LED color is set to RED, the order is GRB. But if the color is  set to GREEN, the color order would be RGB (given that the code sends always GRB order). @sblantipodi - please confirm it.

Considering that it happens, some users may have some trouble with the onboard RGB LED.
@UnexpectedMaker - Can you please give us a hand here? You certainly have the answer in this case.

For this PR, I still suggest the approach described in https://github.com/espressif/arduino-esp32/pull/10128#issuecomment-2276909850

SuGlider:
> You can do with this without introducing any extra CPU overhead

Something like this is not the most elegant way, but it for sure efficient.
No extracode, just define the builtin RGB LED bit order, directly in the code, for any board that needs it.

`variants/esp32c3/pins_arduino.h` would define
``` cpp 
//  neopixelWrite() RMT will use G, R, B bits order
#define RGB_BUILTIN_COLOR_ORDER_STRUCT {green_val, red_val, blue_val} 
```
-----------

`cores/esp32/esp32-hal-rgb-led.c` would do something like:
``` cpp
#ifdef RGB_BUILTIN_COLOR_ORDER_STRUCT
int color[] = RGB_BUILTIN_COLOR_ORDER_STRUCT;
#else
// WS2812B color bit order is G, R, B
int color[] = {green_val, red_val, blue_val};
#endif
```

----------
For instance, the Lolin ESP32-S3 Mini that uses GRB color order, would declare in its own `pins_arduino.h` file the correct order for its neopixel:

``` cpp 
// neopixelWrite() RMT will use R, G, B bits order
#define RGB_BUILTIN_COLOR_ORDER_STRUCT {red_val, green_val, blue_val}
```
---------

@robertlipe, @sblantipodi  @me-no-dev
Let me know.

sblantipodi:
> > You can do with this without introducing any extra CPU overhead
> 
> Something like this is not the most elegant way, but it for sure efficient.
> No extracode, just define the builtin RGB LED bit order, directly in the code, for any board that needs it.
> 
> `variants/esp32c3/pins_arduino.h` would define
> ``` cpp 
> //  neopixelWrite() RMT will use G, R, B bits order
> #define RGB_BUILTIN_COLOR_ORDER_STRUCT {green_val, red_val, blue_val} 
> ```
> -----------
> 
> `cores/esp32/esp32-hal-rgb-led.c` would do something like:
> ``` cpp
> #ifdef RGB_BUILTIN_COLOR_ORDER_STRUCT
> int color[] = RGB_BUILTIN_COLOR_ORDER_STRUCT;
> #else
> // WS2812B color bit order is G, R, B
> int color[] = {green_val, red_val, blue_val};
> #endif
> ```
> 
> ----------
> For instance, the Lolin ESP32-S3 Mini that uses GRB color order, would declare in its own `pins_arduino.h` file the correct order for its neopixel:
> 
> ``` cpp 
> // neopixelWrite() RMT will use R, G, B bits order
> #define RGB_BUILTIN_COLOR_ORDER_STRUCT {red_val, green_val, blue_val}
> ```
> ---------
> 
> @robertlipe, @sblantipodi  @me-no-dev
> Let me know.

I like this approach, I will not have access to my PC for some days, if no one will do it before me, I'll commit this idea as soon as I return home.

sblantipodi:
> @sblantipodi @robertlipe - Let's be practical.
> 
> @sblantipodi says that he has a couple boards that have an RGB LED that works with a different color order. This can be tested using ESP32 Arduino `neopixelWrite(pin, 255, 0, 0);`. If the LED color is set to RED, the order is GRB. But if the color is  set to GREEN, the color order would be RGB (given that the code sends always GRB order). @sblantipodi - please confirm it.
> 
> Considering that it happens, some users may have some trouble with the onboard RGB LED.
> @UnexpectedMaker - Can you please give us a hand here? You certainly have the answer in this case.
> 
> For this PR, I still suggest the approach described in https://github.com/espressif/arduino-esp32/pull/10128#issuecomment-2276909850
> 
> 

I confirm my Lolin S3 Mini has different color order than my Lolin C3 mini. Don't remember who is RGB and who is GRB currently but using the same code, on one board the led is red and on another board is green.

sblantipodi:
> I just checked the schematic https://www.wemos.cc/en/latest/_static/files/sch_s3_v1.0.0.pdf on the cited board. It claims to use WS2812B which, per the manufacturer sheet I referenced before, is GRB. WLED

That schematic uses the WROOM flavour, it's not the mini that uses the "standard" esp32-s3.
I don't know where did you get that schematic but the mini doesn't use the WROOM, the pro does.

In any case don't focus on WS2812, NeoPixels is a family of chips, ws2812 grb is only one of them.

SuGlider:
>Yes, that's very much what I was proposing (or trying to) with the minor
>detail that I'd use the define for the structure contents only and not also
>sometimes the structure initialization, but that's what I'd get behind.
>
>So in a header somewhere Ifndef RGB_BUILTIN_COLOR_ORDER_STRUCT, define it
>and then just always colors = RGB_BUILTIN_COLOR_ORDER_STRUCT with no
>distraction in the source file.

The problem is that we have more than 150 boards in the `boards.txt` file. This file was created by "many hands".
In the way I posted it, it would solve the case that the `pins_arduino.h` has no information about the color order.
Each board Makers/Users would add that piece of code , if necessary, case per case (based on their boards) along the time.


>I suppose making it virtual in the base class and just allowing subclasses
>to override it is out of the question since embedded people like ifdefs,
>right?
>
>Many of us in the industry find that smart data and dumb (or no!) code to
>be quite elegant.

HAL in ESP32 Arduino layer is C code.
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-rgb-led.c
No C++ / Classes / Virtual methods etc.


>Neopixel is a trademark ( https://www.adafruit.com/trademarks of Adafruit
>I've seen them defending it - as they have to - so I try to not use the
>term anywhere, but I want pushing to rename this file, either.

Good point, @robertlipe! Thanks for bringing it. 
@me-no-dev @VojtechBartoska @rtafas -- Should we avoid using "neopixel" name?

@robertlipe - Thank you for your contribution in this discussion.
Let me know if you have more to add to the discussion.

At this time, we would wait for @sblantipodi to modify his PR and then approve it for merging.

SuGlider (changes requested):
@sblantipodi - we will wait for you to change it. Please change also the `pins_arduino.h` of the affected boards that you have tested so far. Thank you for the great contribution!

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -11,8 +11,19 @@ extern "C" {
...
 
+typedef enum {
+  LED_COLOR_ORDER_RGB,
+  LED_COLOR_ORDER_BGR,
+  LED_COLOR_ORDER_BRG,
+  LED_COLOR_ORDER_RBG,
+  LED_COLOR_ORDER_GBR,
+  LED_COLOR_ORDER_GRB
+} rgb_led_color_order_t;
+
```
me-no-dev, line 22: you can add here:
```cpp
#ifndef LED_COLOR_ORDER_DEFAULT
#define LED_COLOR_ORDER_DEFAULT LED_COLOR_ORDER_GRB
#endif
```

Which would allow to define a different color order in `pins_arduino.h`. You would also need to use `LED_COLOR_ORDER_DEFAULT` in the implementation of `neopixelWrite` above

sblantipodi:
I pushed the latest...
Please tell me if this meets your ideas.

SuGlider (approved):
Thanks @sblantipodi for the PR!
Thanks @robertlipe for the support and discussion!
@me-no-dev - it is ready for your final review.

SuGlider:
@sblantipodi - I just used your PR to add the missing license information to the source code. I hope you don't mind.

sblantipodi:
@robertlipe title edited, thanks 👍 
@SuGlider sure, thank you 👍

me-no-dev:
@sblantipodi, @SuGlider and @robertlipe, while the current implementation is the shortest possible solution, I find it a bit unorthodox. I liked it much better before, with the addition I suggested [here](https://github.com/espressif/arduino-esp32/pull/10128#pullrequestreview-2235302947). If that solution is used, it can easily be expanded to support RGBA, DotStar and discrete RGB LEDs in the future. Maybe the function name should change to `rgbLedWriteAdvanced` to spare future renaming, but the approach was good IMHO, even if code was a bit more.

me-no-dev:
There are boards with DotStar and with discrete RGB leds out there. Unexpected Maker and others made boards with DotStar and I have one around with the discrete RGB also. Yes, it is still the target to control one LED and not a strip and I do agree that maybe good idea to add both whites now to the advanced function. I just don't find it intuitive to define the struct as in this PR currently. I've seen it nowhere in Espressif code (or other for that matter). Selecting how it would be is critical to be done now, because whatever is decided will go to be used in the future in board definitions (that we do not want to check and maintain in the future ourselves. see how many boards we have...)

SuGlider:
@sblantipodi - I'll modify it as requested by @me-no-dev, mixing some ideas we have discussed before.
@robertlipe - the proposed code mixes some of the points we discussed and opens space for @me-no-dev request.
@me-no-dev - please review it and let me know.

PR main goals:
- Allow users and board makers to change the RGB LED color order sent using RMT.
- Addresses only 1 LED per pin such as the onboard LED.
- When the board uses standard WS2812b LED, no firmware/memory/flash changes will happen.
- Only affects boards that require such change. It can be defined in "pins_arduino.h" specific board file.

SuGlider:
given that the `switch()` takes a constant value, it is very possible that the compiler will optimize the code and ignore it, by just setting the color order directly as specified.

cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -15,7 +29,40 @@ void neopixelWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue
     return;
   }
 
-  int color[] = {green_val, red_val, blue_val};  // Color coding is in order GREEN, RED, BLUE
+#if defined RGB_BUILTIN_LED_COLOR_ORDER_CHANGE
```
me-no-dev, line 32: maybe remove the `_CHANGE` suffix. `RGB_BUILTIN_LED_COLOR_ORDER` is more readable and makes more sense

SuGlider: Done.

me-no-dev (approved):

P-R-O-C-H-Y (approved):

lucasssvaz (approved):

sblantipodi:
@SuGlider it works as expected

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -11,6 +11,17 @@ extern "C" {
 #define RGB_BRIGHTNESS 64
 #endif
 
+#if defined RGB_BUILTIN_LED_COLOR_ORDER
```
lucasssvaz, line 14:
```suggestion
#ifdef RGB_BUILTIN_LED_COLOR_ORDER
```

cores/esp32/esp32-hal-rgb-led.c
```diff
@@ -15,7 +29,39 @@ void neopixelWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue
     return;
   }
 
-  int color[] = {green_val, red_val, blue_val};  // Color coding is in order GREEN, RED, BLUE
+  // default WS2812B color order is G, R, B
+  const int color[3] = {green_val, red_val, blue_val};
+
+#if defined RGB_BUILTIN_LED_COLOR_ORDER
```
lucasssvaz, line 35:
```suggestion
#ifdef RGB_BUILTIN_LED_COLOR_ORDER
```

me-no-dev:
@sblantipodi @robertlipe @SuGlider PTAL at the latest changes

cores/esp32/esp32-hal-rgb-led.h
```diff
@@ -11,7 +11,26 @@ extern "C" {
...
+  LED_COLOR_ORDER_GRB
+} rgb_led_color_order_t;
+
+void rgbLedWriteOrdered(uint8_t pin, rgb_led_color_order_t order, uint8_t red_val, uint8_t green_val, uint8_t blue_val);
+
+// Will use RGB_BUILTIN_LED_COLOR_ORDER
+void rgbLedWrite(uint8_t pin, uint8_t red_val, uint8_t green_val, uint8_t blue_val);
+
+// Backward compatibility
+#define neopixelWrite(p, r, g, b) rgbLedWrite(p, r, g, b)
```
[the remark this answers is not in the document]
me-no-dev: this is something that we will deal with separately for other reasons.

SuGlider (approved):
Ready for Merging @me-no-dev

PR #10134, opened 2024-08-09
feat(board): Add Alfredo NoU3

BotSpace:
Add board support for Alfredo Systems NoU3.

https://www.alfredosys.com/products/alfredo-nou3/


P-R-O-C-H-Y (changes requested):
Hi @BotSpace, PTAL on my comment.
The `boards.txt` addition looks fine, but the `pins_arduino.h` file is missing a lot of pins declarations.

BotSpace:
I appreciate the swift explanation, I promise I did a lot of research and testing before putting in this PR, but I am in a little over my head here still.

If I'm being honest I would prefer not to define pins this way. to me it seems unnecessary, because the esp32s3 has an internal pin mux, any pin can be any function (except for touch and ADC). If users want to use touch and ADC, they are going to need to look up what those pins are anyway, so A1 or T1 definitions do not help them much.

For RX and TX definitions, I don't understand what those definitions are being used for. My board is intended to be used with JTAG upload mode and Hardware CDC serial debugging on GPIO 19 and 20. I'm afraid if I define RX and TX to those pins, they might get set as UART pins instead of JTAG or OTG pins. There are no other pins intended for serial on the NoU3.

I'm happy to use the SCL and SDA definitions as my board is qwiic compatible on GPIO 34 and 35.

But like serial, there are no pins intended for SPI on my board, I would rather leave that up to the user. I included a random SS pin in my latest commit because that is defiantly needed in CIBoardsTest.ino

If the test fails again I suppose I will just have to add some random serial pins and the rest of the SPI pins.

Any insight on what is strictly necessary would be appreciated!

P-R-O-C-H-Y:
@BotSpace Will you do any action regarding the pins for DS motors, sensors etc? Also please update your branch with master and resolve conflicts. Thanks

BotSpace:
Thanks again for the help. I'm not planning on adding more defs for motors/sensors, I think it makes more sense to have that handled by my [library](https://github.com/AlfredoSystems/Alfredo-NoU3) for the board. I merged my branch with upstream master, hopefully everything is in order now.

P-R-O-C-H-Y (approved):

Issue #10167, opened 2024-08-15
Inconsistant of component's hash files of arduino esp32 component dependencies (DSP, libsodium)


executer-uno:
### Board

ESP32, ESP32-S3 may be others

### Device Description

build system issue

### Hardware Configuration

bare HelloWorld project from espressif__arduino-esp32\examples\hello_world

### Version

v3.0.4

### IDE Name

Visual Studio Code with ESP-IDF extension v1.8.0 (+path https://github.com/espressif/vscode-esp-idf-extension/pull/1252)

### Operating System

Windows 10

### Flash frequency

no matter

### PSRAM enabled

no

### Upload speed

115200

### Description

Got consistent error messages on menuconfig, build or target change on some preconditions (clear project, sdkconfig updated).

Starting point is
Clear copy of espressif__arduino-esp32\examples\hello_world project with main\idf_component.yml one line commented out:
```
dependencies:
  espressif/arduino-esp32:
    version: "*"
    # override_path: "../../../"
    pre_release: false
```

- **:x: menuconfig gives an Error (see "Debug Message") + only on that first step lipsodium hash also appears in error test**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- menuconfig opens fine
- **:x: build - Fails on same error (see "Debug Message")**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- build - Ok (entire build, takes long time)
- menuconfig opens fine, change some parameters, save sdkconfig
- **:x: build - Fails on same error (see "Debug Message")**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- build - Ok (entire rebuild, takes long time)
- main.cpp updated
- build - ok, changes only, fast
- **:x: change target to ESP32-S3 same error (see "Debug Message")**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- change target to ESP32-S3 - OK
- **:x: menuconfig gives an Error (see "Debug Message")**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- menuconfig opens fine
- **:x: build - Fails on same error (see "Debug Message")**
- :heavy_check_mark: espressif__esp-dsp\.component_hash deleted
- build - Ok (entire rebuild, takes long time)

### Sketch

```cpp
main\main.cpp

#include "Arduino.h"

void setup() {
  Serial.begin(115200);
}

void loop() {
  Serial.println("Hello world! Update!");
[...]

PR #10173, opened 2024-08-15
fix components compile issues

Jason2866:
## Description of Change
Fix compile issues when Arduino is used as a component of IDF and files in managed components are changed

## Tests scenarios
Compile fail of example [idf_component_examples/hello_world](https://github.com/espressif/arduino-esp32/tree/master/idf_component_examples/hello_world)

## Related links
fix for https://github.com/espressif/arduino-esp32/issues/10167

@me-no-dev maybe there is a better fix for

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@
 set(min_supported_idf_version "5.1.0")
 set(max_supported_idf_version "5.1.99")
 set(idf_version "${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}.${IDF_VERSION_PATCH}")
+set(ENV{IDF_COMPONENT_OVERWRITE_MANAGED_COMPONENTS} 1)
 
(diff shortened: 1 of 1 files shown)


me-no-dev:
can you try building the libs with this change and [this](https://github.com/espressif/esp32-arduino-lib-builder/blob/master/build.sh#L118-L125) removed?

Jason2866:
Yep, had this already in mind that this should be a solution for the issue.
Will test.

Jason2866:
@me-no-dev No, does not solve the issue in Arduino Lib Builder. Still fails.

me-no-dev:
ok then lets close this and add the fix to the lib builder
add `export IDF_COMPONENT_OVERWRITE_MANAGED_COMPONENTS=1` to the beginning of `build.sh`

me-no-dev:
after all this is not Arduino's fault and does not need to be fixed here.

PR #10177, opened 2024-08-16
Adding CodeCell ESP32C3 new 3rd part board

microbotsio:
Adding CodeCell ESP32C3 new 3rd part board with updating boards.txt and adding the pins_arduino.h in the variant/CodeCell folder.


variants/CodeCell/pins_arduino.h
```diff
@@ -0,0 +1,33 @@
...
+#include "soc/soc_caps.h"
+
+#define EXTERNAL_NUM_INTERRUPTS 22
+#define NUM_DIGITAL_PINS        22
+#define NUM_ANALOG_INPUTS       6
+
+#define analogInputToDigitalPin(p)  (((p)<NUM_ANALOG_INPUTS)?(analogChannelToDigitalPin(p)):-1)
+#define digitalPinToInterrupt(p)    (((p)<NUM_DIGITAL_PINS)?(p):-1)
+#define digitalPinHasPWM(p)         (p < EXTERNAL_NUM_INTERRUPTS)
+
```
P-R-O-C-H-Y, line 14: One small change, this needs to be removed as its defined globally :)
```suggestion
```

P-R-O-C-H-Y (changes requested):
@microbotsio PTAL

P-R-O-C-H-Y (approved):
LGTM

microbotsio:
Can you explain why it the boards test and failed to merge?

boards.txt
```diff
@@ -39802,3 +39802,61 @@ alfredo-nou3.menu.EraseFlash.all=Enabled
...
+
+codecell.upload.speed=921600
+codecell.serial.disableDTR=false
+codecell.serial.disableRTS=false
+
+codecell.build.tarch=riscv32
+codecell.build.target=esp
+codecell.build.mcu=esp32c3
+codecell.build.core=esp32
+codecell.build.variant=codecell
```
P-R-O-C-H-Y, line 39826: The variant name is not matching the variant folder. Please change the variant folder to lowercase.

P-R-O-C-H-Y (changes requested):
@microbotsio

P-R-O-C-H-Y (approved):
LGTM

microbotsio:
Is there anything else we need to do from our side?

P-R-O-C-H-Y:
> Is there anything else we need to do from our side?

@microbotsio All good from your side, the PR ready to be merged (marked as pending merge). We will merge it soon :)

microbotsio:
great thanks!

microbotsio:
Could you please let me know when the next 3.0.5 release is scheduled? :) 

We like our board to show up in Board Manager

guswynn:
@microbotsio why was SCK mapped to pin 4, when that seems to be used for something else (referencing https://github.com/microbotsio/CodeCell/blob/main/hardware/CodeCell%20Schematics.png)? My understanding is that the ESP32 can make peripherals at will, so this can be overridden when constructing the `SPIClass`, but if I am reading https://www.espressif.com/sites/default/files/documentation/esp32-c3_datasheet_en.pdf#cd-pins-io-mux-gpio, correctly, the high throughput using the IO MUX would be:

```
static const uint8_t SS = 5; // this one would need to go through the gpio mux, so it will have a latency hit; the direct pin is 10, as far as i can tell, which is used for something else. It's unclear if this affects the throughput of the spi interface in general.
static const uint8_t MOSI = 2;
static const uint8_t MISO = 7;
static const uint8_t SCK = 6;
```

microbotsio:
Thanks for taking interest in this matter! We're checking if we can
update this in the coming future

PR #10180, opened 2024-08-16
Additional UART_DATA check and pass bits back to callBack

rowingdude:
The proposed change should ensure that the callback is only triggered when appropriate conditions are met and that it passes the number of available bytes to the callback.


Jason2866:
Which use case does generate an issue needing this proposed change?
Do not change file formatting in a PR

rowingdude:
This is a quality of code improvement. I did not mean to include my spaced out file where I was just trying to read better.

me-no-dev:
your quality code improvement breaks compilation

PR #10181, opened 2024-08-16
Prevents the buffer size from changing while the UART is active.

rowingdude:
This change returns the buffer size rather than 0, the goal is to keep the buffer from changing its size while we're using the UART functionality.

-----------
## Description of Change

## Tests scenarios

## Related links


rowingdude:
This is just a quality of code improvement.

me-no-dev:
and how is this a code with higher quality?

PR #10182, opened 2024-08-16
Add _destroyEventTask() to HardwareSerial::Begin ..

rowingdude:
This change assures us that the event task is always re-created when begin() is called.


me-no-dev:
why?

rowingdude:
You guys are right, I was thinking with the mindset of a normal computer and not a microcontroller. I'll add better profiling to my flow and try to make better coding choices.

PR #10183, opened 2024-08-16
Complete an IP address with 0s

rowingdude:
The general idea was to take any partial IP address such as 192. and complete it with 0s, so 192. becomes 192.0.0.0 and so on up to a full address. 

I created a second function, parseIPv4Part to check the validity of each section of the IP address we're assessing, it simply checks the value and returns a true or false, we complete on false.

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -116,48 +116,79 @@ bool IPAddress::fromString(const char *address) {
 }
 
 bool IPAddress::fromString4(const char *address) {
-  // TODO: add support for "a", "a.b", "a.b.c" formats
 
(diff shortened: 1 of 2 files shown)


Jason2866:
Why at all? Heavy code changes in a code part which is mandatory and not trivial.
Don't fix ain't broke. No benefit and code size increases.

rowingdude:
I saw a to do in the comments and made the addition. I’m sorry it’s not
what you're looking for

me-no-dev:
This `ToDo` came from Arduino themselves. I do not think we need it and if it's at all necessary ever, it should be implemented upstream: https://github.com/arduino/ArduinoCore-API/blob/master/api/IPAddress.cpp#L133

Issue #10185, opened 2024-08-16
Add developer contributor guidelines to build and test our PR locally


mathieucarbou:
### Board

N/A

### Device Description

N/A

### Hardware Configuration

N/A

### Version

latest master (checkout manually)

### IDE Name

na

### Operating System

na

### Flash frequency

na

### PSRAM enabled

yes

### Upload speed

na

### Description

Hello,

I am preparing a PR to contribute middleware support (aka Expressif) to WebServer, plus some examples and middleware implementations (cors, security, logging). 

I didn't find any guidelines explaining the process to compile our changes and test the examples we write on a real board before submitting the PR.

Right now, I work around that by adding some platformio.ini files and using PIO to build the examples, but this is quite a hack.

What contributors should do to locally test and run the examples we write against the modifications we did ?

Thanks :-)

### Sketch

```cpp
na
```

### Debug Message

```plain
na
```

### Other Steps to Reproduce

na

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

me-no-dev:
@lucasssvaz @P-R-O-C-H-Y any way this can be tested with wokwi?

Jason2866:
@mathieucarbou Thx, for bringing this up. After the big refactoring of the GH Actions (btw. great work imho) i do not manage to test my changes anymore before opening a PR ;-)

mathieucarbou:
> @mathieucarbou Thx, for bringing this up. After the big refactoring of the GH Actions (btw. great work imho) i do not manage to test my changes anymore before opening a PR ;-)

I've tried running `github/scripts/on-push.sh` locally, it succeeded, which proves it didn't consider my changes... 🤣

PR #10186, opened 2024-08-17
feat(webserver): Middleware with default middleware for cors, authc, curl-like logging

mathieucarbou:
This PR improves WebServer with the following additions:

- Add Middleware support (aka Expressif) with some built-in middlewares like cors, authc and logging
- Add ability to collect all incoming request headers
- Added response code and info on server

As discussed in https://github.com/espressif/arduino-esp32/issues/10185, I tested the example thanks to a pio config.

Ideally it would be nice if @me-no-dev and @ayushsharma82 could review this PR, both were involved in these parts.

### DIFF
--- a/libraries/WebServer/examples/Middleware/Middleware.ino
+++ b/libraries/WebServer/examples/Middleware/Middleware.ino
@@ -1,0 +1,186 @@
+/**
+ * Basic example of using Middlewares with WebServer
+ * 
+ * Middleware are common request/response processing functions that can be applied globally to all incoming requests or to specific handlers.
+ * They allow for a common processing thus saving memory and space to avoid duplicating code or states on multiple handlers.
+ * 
+ * Once the example is flashed (with the correct WiFi credentials), you can test the following scenarios with the listed curl commands:
+ * - CORS Middleware: answers to OPTIONS requests with the specified CORS headers and also add CORS headers to the response when the request has the Origin header
+ * - Logging Middleware: logs the request and response to an output in a curl-like format
+ * - Authentication Middleware: test the authentication with Digest Auth 
+ * 
+ * You can also add your own Middleware by extending the Middleware class and implementing the run method.
+ * When implementing a Middleware, you can decide when to call the next Middleware in the chain by calling next().
+ * 
+ * Middleware are execute in order of addition, the ones attached to the server will be executed first.
+ */
+#include <WiFi.h>
+#include <WebServer.h>
+#include <Middlewares.h>
+
+// Your AP WiFi Credentials
+// ( This is the AP your ESP will broadcast )
+const char *ap_ssid = "ESP32_Demo";
+const char *ap_password = "";
+
+WebServer server(80);
+
+LoggingMiddleware logger;
+CorsMiddleware cors;
+AuthenticationMiddleware auth;
+
+void setup(void) {
+  Serial.begin(115200);
+  WiFi.softAP(ap_ssid, ap_password);
+
+  Serial.print("IP address: ");
+  Serial.println(WiFi.AP.localIP());
+
+  // curl-like output example:
+  //
+  // > curl -v -X OPTIONS -H "origin: http://192.168.4.1" http://192.168.4.1/
+  //
+  //  Connection from 192.168.4.2:51683
+  // > OPTIONS / HTTP/1.1
+  // > Host: 192.168.4.1
+  // > User-Agent: curl/8.10.0
+  // > Accept: */*
+  // > origin: http://192.168.4.1
+  // >
+  // * Processed in 5 ms
+  // < HTTP/1.HTTP/1.1 200 OK
+  // < Content-Type: text/html
+  // < Access-Control-Allow-Origin: http://192.168.4.1
+  // < Access-Control-Allow-Methods: POST,GET,OPTIONS,DELETE
+  // < Access-Control-Allow-Headers: X-Custom-Header
+  // < Access-Control-Allow-Credentials: false
+  // < Access-Control-Max-Age: 600
+  // < Content-Length: 0
+  // < Connection: close
+  // <
+  logger.setOutput(Serial);
+
+  cors.setOrigin("http://192.168.4.1");
+  cors.setMethods("POST,GET,OPTIONS,DELETE");
+  cors.setHeaders("X-Custom-Header");
+  cors.setAllowCredentials(false);
+  cors.setMaxAge(600);
+
+  auth.setUsername("admin");
+  auth.setPassword("admin");
+  auth.setRealm("My Super App");
+  auth.setAuthMethod(DIGEST_AUTH);
+  auth.setAuthFailureMessage("Authentication Failed");
+
+  server.addMiddleware(&logger);
+  server.addMiddleware(&cors);
+
+  // Not authenticated
+  //
+  // Test CORS preflight request with:
+  // > curl -v -X OPTIONS -H "origin: http://192.168.4.1" http://192.168.4.1/
+  //
+  // Test cross-domain request with:
+  // > curl -v -X GET -H "origin: http://192.168.4.1" http://192.168.4.1/
+  //
+  server.on("/", []() {
+    server.send(200, "text/plain", "Home");
+  });
+
+  // Authenticated
+  //
+  // > curl -v -X GET -H "origin: http://192.168.4.1"  http://192.168.4.1/protected
+  //
+  // Outputs:
+  //
+  // * Connection from 192.168.4.2:51750
+  // > GET /protected HTTP/1.1
+  // > Host: 192.168.4.1
+  // > User-Agent: curl/8.10.0
+  // > Accept: */*
+  // > origin: http://192.168.4.1
+  // >
+  // * Processed in 7 ms
+  // < HTTP/1.HTTP/1.1 401 Unauthorized
+  // < Content-Type: text/html
+  // < Access-Control-Allow-Origin: http://192.168.4.1
+  // < Access-Control-Allow-Methods: POST,GET,OPTIONS,DELETE
+  // < Access-Control-Allow-Headers: X-Custom-Header
+  // < Access-Control-Allow-Credentials: false
+  // < Access-Control-Max-Age: 600
+  // < WWW-Authenticate: Digest realm="My Super App", qop="auth", nonce="ac388a64184e3e102aae6fff1c9e8d76", opaque="e7d158f2b54d25328142d118ff0f932d"
+  // < Content-Length: 21
+  // < Connection: close
+  // <
+  //
+  // > curl -v -X GET -H "origin: http://192.168.4.1" --digest -u admin:admin  http://192.168.4.1/protected
+  //
+  // Outputs:
+  //
+  // * Connection from 192.168.4.2:53662
+  // > GET /protected HTTP/1.1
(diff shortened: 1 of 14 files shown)


Jason2866:
@mathieucarbou Do you see the possibility to make this enhancement optional?
The size increase is not just a few bytes.

mathieucarbou:
> @mathieucarbou Do you see the possibility to make this enhancement optional? The size increase is not just a few bytes.

yeas I saw that and I was surprises by it... I don't feel to have written 2k of code... What can cause that ?

mathieucarbou:
> One suggestion could be to move all the 3 middlewares (authc, cors, logging) in examples....

I've added a second commit that is moving the "Middlewares.h" file (containing implementation classes for Authentication, CORS and Logging) to the example folder. Users will have to copy these classes if they want to use it.

Let's see the effect now...

mathieucarbou:
A

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+  explicit LoggingMiddleware(Stream &out) : _out(&out) {}
+
+  bool run(WebServer &server, Middleware::Callback next) override {
+    _out->print(F("* Connection from "));
+    _out->print(server.client().remoteIP().toString());
+    _out->print(F(":"));
+    _out->println(server.client().remotePort());
+
+    _out->print(F("< "));
+    HTTPMethod method = server.method();
```
safocl, line 24: from
```HTTPMethod method```
to
```const HTTPMethod method```
or
```const auto method```
?

File: libraries/WebServer/src/Middlewares.h [hunk 2]
```diff
@@ -0,0 +1,208 @@
...
+    if (method == HTTP_ANY) {
+      _out->print(F("HTTP_ANY"));
+    } else {
+      _out->print(http_method_str((http_method)method));
```
safocl, line 28: from
```(http_method)method```
to
```method```
?

```typedef enum http_method HTTPMethod;```
https://github.com/espressif/arduino-esp32/blob/ae052f414ae6eb0f927b1acfb25fa23fa4f91524/libraries/WebServer/src/HTTP_Method.h#L6C1-L6C37

File: libraries/WebServer/src/Middlewares.h [hunk 3]
```diff
@@ -0,0 +1,208 @@
...
+        _out->print(F("< "));
+        _out->print(server.headerName(i));
+        _out->print(F(": "));
+        _out->println(server.header(i));
+      }
+    }
+
+    _out->println(F("<"));
+
+    bool ret = next();
```
safocl, line 49: similar to that case https://github.com/espressif/arduino-esp32/pull/10186/files#r1774121401

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+      _out->println(F("* Not processed!"));
+    }
+
+    return ret;
+  }
+};
+
+class AuthenticationMiddleware : public Middleware {
+private:
+  // authenticate state
```
safocl, line 82: replace with enum class instead of comments?
from
```
// authenticate state
// 0: not authenticated
// 1: callback
// 2: username/password
// 3: sha1
```
to
```
enum class AuthState : std::int8_t {
             NOT_AUTHENTICATED, // 0: not authenticated
             CALLBACK, // 1: callback
             USERNAME_PASSWD, // 2: username/password
             SHA1// 3: sha1     
};
```       
?

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+  // authenticate request
+  HTTPAuthMethod _mode = BASIC_AUTH;
+  String _realm;
+  String _authFailMsg;
+
+public:
+  AuthenticationMiddleware &authenticate(WebServer::THandlerFunctionAuthCheck fn) {
+    assert(fn);
+    _fn = fn;
+    _auth = 1;
```
safocl, line 102: considering https://github.com/espressif/arduino-esp32/pull/10186/files#r1774149721

from
```_auth = 1;```
to ```_auth = AuthState::CALLBACK```
?

File: libraries/WebServer/src/Middlewares.h [hunk 6]
```diff
@@ -0,0 +1,208 @@
...
+    return *this;
+  }
+
+  AuthenticationMiddleware &authenticate(const char *username, const char *password) {
+    if (strlen(username) == 0 || strlen(password) == 0) {
+      _auth = 0;
```
safocl, line 108: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774158001

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+      return *this;
+    } else {
+      _username = username;
+      _password = password;
+      _auth = 2;
```
safocl, line 113: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774158001

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+      return *this;
+    }
+  }
+
+  AuthenticationMiddleware &authenticateBasicSHA1(const char *username, const char *sha1AsBase64orHex) {
+    if (strlen(username) == 0 || strlen(sha1AsBase64orHex) == 0) {
+      _auth = 0;
```
safocl, line 120: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774158001

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+      return *this;
+    }
+    _username = username;
+    _sha1 = sha1AsBase64orHex;
+    _auth = 3;
```
safocl, line 125: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774158001

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+    return *this;
+  }
+
+  bool run(WebServer &server, Middleware::Callback next) override {
+    switch (_auth) {
+      case 1:
```
safocl, line 131: from 
```case 1:```
to
```case AuthState::CALLBACK:```

File: libraries/WebServer/src/Middlewares.h [hunk 11]
```diff
@@ -0,0 +1,208 @@
...
+        if (server.authenticate(_fn)) {
+          return next();
+        } else {
+          server.requestAuthentication(_mode, _realm.c_str(), _authFailMsg);
+          return true;
+        }
+
+      case 2:
```
safocl, line 139: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774159969

libraries/WebServer/src/Middlewares.h
```diff
@@ -0,0 +1,208 @@
...
+        if (server.authenticate(_username.c_str(), _password.c_str())) {
+          return next();
+        } else {
+          server.requestAuthentication(_mode, _realm.c_str(), _authFailMsg);
+          return true;
+        }
+
+      case 3:
```
safocl, line 147: similar with https://github.com/espressif/arduino-esp32/pull/10186/files#r1774159969

mathieucarbou:
@safocl : I would be glad to revisit this PR but only if there is a real motivation coming from the Espressif team to merge it. 

Otherwise I won't spend my time updating a PR for nothing.

I have already implemented middleware support in the fork of https://github.com/mathieucarbou/ESPAsyncWebServer we maintain (https://github.com/mathieucarbou/ESPAsyncWebServer) and also in v2 branch of PsychicHttp (https://github.com/hoeken/PsychicHttp) which is the future version, with a lot of supported middlewares for CORS, Authc, authz, curl like logging, etc.

me-no-dev:
>  I would be glad to revisit this PR but only if there is a real motivation coming from the Espressif team to merge it.

There is interest. we are trying to finish up some big tasks and have a good look at this PR

mathieucarbou:
@safocl : thank your for your review

@me-no-dev : thank you for your confirmation

I have reworked the PR to clean it - as it was it was more a PoC waiting for some interest. So I correctly did it and cleaned up the things saw bu @safocl and much more.

I saw several things that I really don't like in WebServer:

1. Returning `String` instead of `const String&`: I saw a bunch of placed where a String copy would be done in the return instead of returning a ref. Sadly, fixing that would break backward compatibility

2. Use of chained pointers: this just makes the code error prone and harde to read with all these boilerplate loops. I would rather use `std::list` instead, but since I didn't see any usage of that (and only a few about vector), I don't know if there's a reason why the Arduino core code would not use it ? Especially that in the case of WebServer, the number of elements in such list would not be high enough to trigger any downsides of using them.

3. Last point is more about organisation: the file structure inside `src` folder is quite fuzzy I find... Also `WebServer.h` would need a bit of re-ordering / api doc : there are so many public methods, and a mix of them for request and response.

libraries/WebServer/examples/Middleware/Middleware.ino
```diff
@@ -0,0 +1,170 @@
+#include <WiFi.h>
```
lucasssvaz, line 1: Please add a couple of lines in the beginning of the example explaining how to use the example.

lucasssvaz (approved):
LGTM. Tested and it works as expected. Just a couple nitpicks.

mathieucarbou:
@lucasssvaz : thanks! I will update the PR soon.

lucasssvaz:
@mathieucarbou Also, as there are some changes to the WebServer. Maybe it would be better to target branch `release/v3.1.x` to make sure we don't introduce any unwanted breaking change by mistake.

mathieucarbou:
> @mathieucarbou Also, as there are some changes to the WebServer. Maybe it would be better to target branch `release/v3.1.x` to make sure we don't introduce any unwanted breaking change by mistake. Probably you just need to rebase this PR and the commits on top of it.

Yes that would be better.

What do you think about the 3 bullet points above ? Could I also switch to std::list ?

lucasssvaz:
> > @mathieucarbou Also, as there are some changes to the WebServer. Maybe it would be better to target branch `release/v3.1.x` to make sure we don't introduce any unwanted breaking change by mistake. Probably you just need to rebase this PR and the commits on top of it.
> 
> Yes that would be better.
> 
> What do you think about the 3 bullet points above ? Could I also switch to std::list ?

I think it should be fine now that we are targeting v3.1.x as long as the flash usage increase is not very significant. There are too many changes from IDF recently that increased flash usage a lot and we are trying to optimize things for space.

lucasssvaz:
@mathieucarbou Could you please reopen this targeting `master` ?

mathieucarbou:
> @mathieucarbou Could you please reopen this targeting `master` ?

FYI, for next time, before deleting branches, just know that there is a github option to edit PRs and change the target branch. This allows keeping the history of PRs instead of having to recreate them. Here since the target branch was deleted, github closed the PR without a way to re-open it and change the target branch.

me-no-dev:
@mathieucarbou I did not expect the PR to be closed. The one for the libs was not closed and I switched the target branch after I deleted the 3.1.x one

me-no-dev:
Interesting what made this PR different. Maybe the fact that we were not given rights to edit it?

mathieucarbou:
Oh, that's right: _"Allow edits and access to secrets by maintainers"_ was not checked, I just checked it now.

Why ? I do not know. This is always an option I usually keep on (I hate to block people or being blocked by this option). 
I don't even know if there is a github option to keep it off by default.

Definitley, that must be this thing.

Do you know if this could be unchecked by default with the required CLA agreement ?

me-no-dev:
trying to reopen

me-no-dev:
It's not allowing me to reopen it 😦

Jason2866:
"Edit" is off when the PR is opened from an ORG repository.

PR #10203, opened 2024-08-19
add required components

Jason2866:
compile fails if other components are removed

@me-no-dev thx for helping to find the reason for failing of my Arduino Lib Builder compiles


me-no-dev:
you will probably need to redo this PR, so I suggest to hold for a bit, or I will add them myself

Jason2866:
Whatever you prefer :-)

Issue #6749, opened 2022-05-15
FFAT exists throws error when file does not exist


frankcohen:
### Board

Adafruit Feather ESP32-S3

### Device Description

I added these lines to the examples->FFat->FFat_Test at line 182:

    if ( FFat.exists( "/hello.txt" ) )
    {
      Serial.println( "hello.txt exists");
    }
    else
    {
      Serial.println( "hello.txt does not exist");
    }
      
    if ( FFat.exists( "/foo.txt" ) )
    {
      Serial.println( "foo.txt exists");
    }
    else
    {
      Serial.println( "foo.txt does not exist");
    }

I see this in the Arduino IDE Serial Monitor:

...
- read from file:
Hello World!
[  6785][E][vfs_api.cpp:104] open(): /ffat/hello.txt does not exist, no permits for creation
foo.txt does not exist
foo.txt exists
Listing directory: /
  FILE: foo.txt	SIZE: 14
...

I expected file.exists() would return an error.

-Frank

### Hardware Configuration

SD NAND card on the SPI bus with chip select on GPIO 10

### Version

latest development Release Candidate (RC-X)

### IDE Name

Arduino IDE 1.8.10

### Operating System

MacOS 12.3.1

### Flash frequency

40

### PSRAM enabled

no

### Upload speed

115200

### Description

I initialize the Serial object with:
    Serial.begin(115200);
    Serial.setDebugOutput(true);
    delay(5000);

exists() does not seem to return a boolean value when the file does not exist.

### Sketch

```cpp
#include "FS.h"
#include "FFat.h"

// This file should be compiled with 'Partition Scheme' (in Tools menu)
// set to 'Default with ffat' if you have a 4MB ESP32 dev module or
// set to '16M Fat' if you have a 16MB ESP32 dev module.

// You only need to format FFat the first time you run a test
#define FORMAT_FFAT true

void listDir(fs::FS &fs, const char * dirname, uint8_t levels){
    Serial.printf("Listing directory: %s\r\n", dirname);

    File root = fs.open(dirname);
    if(!root){
        Serial.println("- failed to open directory");
        return;
    }
    if(!root.isDirectory()){
        Serial.println(" - not a directory");
        return;
    }

    File file = root.openNextFile();
    while(file){
        if(file.isDirectory()){
            Serial.print("  DIR : ");
            Serial.println(file.name());
            if(levels){
                listDir(fs, file.path(), levels -1);
            }
        } else {
[...]

PR #10217, opened 2024-08-21
fix(littlefs): Use VFSImpl::exists() to avoid false error log

### LINKED ISSUE #7615: LittleFS.exists() is giving an error
### Board

ESP32-S3

### Device Description

ESP32-S3 16MB Dev-Kit

### Hardware Configuration

Nothing

### Version

v2.0.5

### IDE Name

Arduino IDE 2.0.4

### Operating System

Windows 10

### Flash frequency

240MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

I am using the LittleFS library.
When I try to check if a file exists, this function is executed:
https://github.com/espressif/arduino-esp32/blob/bb8c8559b62a0d2e888d51bf12bd50cc070f0b4b/libraries/LittleFS/src/LittleFS.cpp#L46
```
bool LittleFSImpl::exists(const char* path)
{
    File f = open(path, "r",false);
    return (f == true);
}
```

As it try to open a file, I get this error:
[E][vfs_api.cpp:104] open(): /littlefs/g00000/c001 does not exist, no permits for creation

From this line:
https://github.com/espressif/arduino-esp32/blob/bb8c8559b62a0d2e888d51bf12bd50cc070f0b4b/libraries/FS/src/vfs_api.cpp#L105

I think this error should not be logged, if we just check if a file exists.

### Sketch

```cpp
char path[20];
createPath(path);  
if(!LittleFS.exists(path))
{
  LittleFS.mkdir(path);
}
```

### Debug Message

```plain
[E][vfs_api.cpp:104] open(): /littlefs/g00000/c001 does not exist, no permits for creation
```

### Other Steps to Reproduce

_No response_

BlueAndi:
## Description of Change
Use exists() from VFSImpl base class and avoid using open() to check whether the file exists, because open() will log an error in case a file doesn't exists which is confusing, but might be relevant in case a developer uses open() directly and expects that the file is there.

## Related links
Closes #7615


me-no-dev:
I think the proper action here would be to make the error a warning or even debug.

BlueAndi:
> I think the proper action here would be to make the error a warning or even debug.

This could be done additional. But I would not expext any log by using exists.

me-no-dev:
If you remove the method, then `exists` will stop functioning correctly. The log is there to let you know that the file was not found and will not be created. If the log is the issue, then that is where it should be fixed.

me-no-dev:
I do not recall, because LittleFS was not implemented by us, but I see it being implemented much the same for the other file systems. Issue here seems the same as https://github.com/espressif/arduino-esp32/issues/6749 though for FFat. @lbernstone do you recall why FFat needs to open the file to check if it exists? If the issue is the error log, I do agree that it's a bit to much to issue an error, where a warning or even debug will suffice for the cases where one will need to get more information.

BlueAndi:
> If you remove the method, then `exists` will stop functioning correctly.

Didn't realize this in my tests. Whats exactly is the problem by using base class exists method?

me-no-dev:
@lucasssvaz can you please test this same solution with LittleFS, FFat and SPIFFS?

ioviedodev:
I experiment the same issue in arduino esp32 version 3.0.4 as component idf with version 5.1.4

aeljandro2:
Same here. Been struggling with the and then realized is just a log thing "Seems". Anyway I think supporting littleFS is kinda important.

lucasssvaz:
@me-no-dev Tested for LittleFS, SPIFFS and FFat. This solution works for LittleFS and FFat. For SPIFFS this error does not appear and when trying this solution it causes `exists()` to always return true.

lucasssvaz (approved):

me-no-dev (approved):

GeorgeIoak:
I have a situation where I'm not able to update the library yet so I was trying to implement a code fix to directly use `VFSImpl::exists()` in place of `fs->exists()` which currently throws the error. I've made several attempt at implementing it but it's always returning false even if the file exists. This is what I had to do to get it to compile so I'm thinking I've botched this up and perhaps someone can help fix it who knows better.

```
bool IHPFirmware::doesFirmwareFileExist() {
    VFSImpl* vfsImpl = reinterpret_cast<VFSImpl*>(this->deps.fs);
    return vfsImpl->VFSImpl::exists(this->firmwarePath);
}
```

I confirmed that I'm passing the same path as I was using with `fs->exists()`

Sorry for asking here but I wasn't sure where else would be appropriate to ask

PR #10224, opened 2024-08-22
Change "neopixel" (Adafruit TM) references to use RGB LED names - compliance

SuGlider:
## Description of Change

Internal request to remove all references to Trade Mark terms (neopixel).
Documentation, commentaries, symbols and functions were modified.

@ladyada - Please Take a Look and comment. Thanks.

## Tests scenarios
CI Only

## Related links
None - Internal.


ladyada:
hi, this was closed? we would not want to remove the name NEOPIXEL from our board definitions and we can "grant use" for any other purposes.

me-no-dev:
Hi @ladyada :) There will be another PR. We are removing it from other boards and have renamed the method in our API. NEOPIXEL definitions will stay for Adafruit boards. Seems improper to call leds "Neopixel" when they are not coming from you, so we decided to clean up a bit. There is a backward compatible define for `neopixelWrite()` so old code and your boards can continue to use it. I hope this is OK?

PR #10225, opened 2024-08-22
Change "neopixel" references to use RGB LED naming

SuGlider:
## Description of Change

Internal request to remove all references to Trade Mark terms (neopixel).
Documentation, commentaries, symbols and functions were modified.

## Tests scenarios

CI Only.

## Related links

None - Internal.


SuGlider:
@me-no-dev | @rftafas | @VojtechBartoska 

PTAL.

me-no-dev:
@SuGlider my only "comment" is: Should we maybe keep the `neopixelWrite` in pins_arduino for adafruit boards?

@ladyada this is the new PR. Please have a look

tonhuisman:
Hmm, how much confusion will this cause for people trying to connect/use 4-pin RGB leds with their boards (and as mounted on some boards instead of a NeoPixel), only to find they are supposed to use ~NeoPixel~ 'Serial Controlled RGB(W)' leds... 🤔

me-no-dev:
@tonhuisman these functions are generally to be used with board LEDs and we do plan to add support for 4 pin RGB and DotStar LEDs

ladyada:
reviewed - we're OK with any variant using the word 'NEOPIXEL' in the board def .... you may want to contact those board makers to give them the alert that a built in pin name has changed :)

fyi DotStar is also a trademark

me-no-dev:
> fyi DotStar is also a trademark

It's a household name at this point, just like NeoPixel :) (props to you)

SuGlider:
> @SuGlider my only "comment" is: Should we maybe keep the `neopixelWrite` in pins_arduino for adafruit boards?
> 
> @ladyada this is the new PR. Please have a look

Based on recent note from @ladyada:

Given that the `neopixelWrite()` is now called `rgbLedWrite()`, all commentaries have changed for all boards, including Adafruit.
I think that we should keep the PR because those commentaries were added by myself, not by Adafruit.

I also think that, for Arduino Core 3.0.x,  we should remove the `#define neopixelWrite rgbLedWrite` declaration and use a `neopixelWrite()` function, marking it as deprecated, instead.
Latter in Arduino Core 3.1, we should just definitively remove it.

Please let me know your considerations.

docs/en/api/rmt.rst
```diff
@@ -17,7 +17,7 @@ To get started with RMT, you can try:
 RMT Write Neo Pixel
```
lucasssvaz, line 17:
```suggestion
RMT Write RGB LED
```

SuGlider:
@me-no-dev @P-R-O-C-H-Y @lucasssvaz - Ready for review.

P-R-O-C-H-Y (approved):

me-no-dev (approved):

lucasssvaz (approved):
Works fine

PR #10236, opened 2024-08-24
fix(comment): Fix comment in `tests.yml`

DemoYeti:


P-R-O-C-H-Y (approved):

lucasssvaz (commented):
I don't think this is correct.

The files going to the right (`tests_results.yml` and `tests_wokwi.yml`) run on `workflow_run`, meaning that they will only be run after the previous one finishes.

The files listed below, on the other hand, run on `workflow_call`. Meaning it will be called by `tests.yml` and run as it is a job inside this file/workflow.

me-no-dev:
Closing per comment from @lucasssvaz

PR #10239, opened 2024-08-25
NetworkUDP - in parsePacket handle previous parsed packet

JAndrassy:
`parsePacket` now returns 0 if  previous packet was not read. what is the logic of that?

this PR proposes to changes it to return the existing packet if it was not read yet and discard the packet if the user read some part of it.

### DIFF
--- a/libraries/Network/src/NetworkUdp.cpp
+++ b/libraries/Network/src/NetworkUdp.cpp
@@ -292,7 +292,10 @@ void NetworkUDP::flush() {}
 
 int NetworkUDP::parsePacket() {
   if (rx_buffer) {
-    return 0;
+    if (rx_buffer->full()) { // packet was not read yet
+      return rx_buffer->available();
+    }
+    clear(); // discard the rest of the packet
   }
   struct sockaddr_storage si_other_storage;  // enough storage for v4 and v6
   socklen_t slen = sizeof(sockaddr_storage);


libraries/Network/src/NetworkUdp.cpp
```diff
@@ -292,18 +292,14 @@ void NetworkUDP::flush() {}
...
-    return 0;
+    if (rx_buffer->full()) { // packet was not read yet
+      return rx_buffer->available();
+    }
+    clear(); // discard the rest of the packet
   }
   struct sockaddr_storage si_other_storage;  // enough storage for v4 and v6
   socklen_t slen = sizeof(sockaddr_storage);
-  int len = 0;
-  if (ioctl(udp_server, FIONREAD, &len) == -1) {
```
TD-er, line 300: How come this check for the length can be removed? 
I do see it is later assigned via `recvfrom`, but it does allow for an early escape so no buffer needs to allocated if the length cannot be determined or is 0 (is length of 0 possible?)

JAndrassy: sorry. it was a merge error

libraries/Network/src/NetworkUdp.cpp
```diff
@@ -292,7 +292,10 @@ void NetworkUDP::flush() {}
 
 int NetworkUDP::parsePacket() {
   if (rx_buffer) {
-    return 0;
+    if (rx_buffer->full()) { // packet was not read yet
```
TD-er, line 295: See both implementations of `room()` and `full()`:
```c++
size_t cbuf::room() const {
  size_t _room = 0;
  if (_buf != NULL) {
    _room = xRingbufferGetCurFreeSize(_buf);
  }
  return _room;
}

bool cbuf::full() const {
  return room() == 0;
}
```

As you can see, when checking for `full()`, you only check if there is no more room left.
However I do not know how these buffers are allocated.
For example if I process a packet of 7 bytes, what guarantee do we have the buffer is full then and there isn't any extra byte of free space?

See also the note in the comments in the header file where the buffer is created:
```c++
/**
 * @brief       Create a ring buffer
 *
 * @param[in]   xBufferSize Size of the buffer in bytes. Note that items require
 *              space for a header in no-split/allow-split buffers
 * @param[in]   xBufferType Type of ring buffer, see documentation.
 *
 * @note    xBufferSize of no-split/allow-split buffers will be rounded up to the nearest 32-bit aligned size.
 *
 * @return  A handle to the created ring buffer, or NULL in case of error.
 */
RingbufHandle_t xRingbufferCreate(size_t xBufferSize, RingbufferType_t xBufferType);
```

TD-er: There is however another obvious bug in the NetworkUDP.cpp which may need some fixing...

```c++
  if (len > 0) {
    rx_buffer = new (std::nothrow) cbuf(len);
    rx_buffer->write(buf, len);
  }
```

JAndrassy: I tested the cbuf class to check if it behaves as I expected.

TD-er: But the note at the `xRingbufferCreate` function clearly states it will round up `xBufferSize` to the nearest 4-byte.
If this happens, then you can't use the `full()` function to detect whether it is still untouched.

JAndrassy: RINGBUF_TYPE_BYTEBUF is used here. the note is for RINGBUF_TYPE_NOSPLIT and RINGBUF_TYPE_ALLOWSPLIT

TD-er: Ah check!

libraries/Network/src/NetworkUdp.cpp
```diff
@@ -292,7 +292,10 @@ void NetworkUDP::flush() {}
...
+      return rx_buffer->available();
```
me-no-dev, line 296: can you show a benefit from returning the previous packet len more than once? If code did not care about it the first time, why should it care after that?

libraries/Network/src/NetworkUdp.cpp
```diff
@@ -292,7 +292,10 @@ void NetworkUDP::flush() {}
...
+    }
+    clear(); // discard the rest of the packet
```
me-no-dev, line 298: What will happen in you call `parsePacket` from one thread, but you read it's data from another?

libraries/Network/src/NetworkUdp.cpp
JAndrassy: for the library it is better than destroying it then allocate the next then destroy it ...

me-no-dev: library will not destroy it. It will just return 0, because old packet is still there and new one was not received. The point of returning positive length is to signal that a new packet has been received. This happens only once per packet.

JAndrassy: but the PR is about not returning a zero if the packet was not read until the end

PR #10242, opened 2024-08-26
fix(client): Make flush() call clear() and deprecate it

me-no-dev:
This is a compromise for issues caused by https://github.com/espressif/arduino-esp32/pull/9453

Example output:
```
/var/folders/74/n3f3k71s7lv_snk832kfgjfw0000gn/T/arduino_modified_sketch_265854/WiFiIPv6.ino: In function 'void wifiConnectedLoop()':
/var/folders/74/n3f3k71s7lv_snk832kfgjfw0000gn/T/arduino_modified_sketch_265854/WiFiIPv6.ino:53:20: warning: 'virtual void NetworkUDP::flush()' is deprecated: Use clear() instead. [-Wdeprecated-declarations]
   53 |     ntpClient.flush();
      |     ~~~~~~~~~~~~~~~^~
In file included from /Users/user/Documents/Arduino/hardware/espressif/esp32/libraries/Network/src/Network.h:14,
                 from /Users/user/Documents/Arduino/hardware/espressif/esp32/libraries/WiFi/src/WiFiGeneric.h:39,
                 from /Users/user/Documents/Arduino/hardware/espressif/esp32/libraries/WiFi/src/WiFiSTA.h:29,
                 from /Users/user/Documents/Arduino/hardware/espressif/esp32/libraries/WiFi/src/WiFi.h:33,
                 from /var/folders/74/n3f3k71s7lv_snk832kfgjfw0000gn/T/arduino_modified_sketch_265854/WiFiIPv6.ino:1:
/Users/user/Documents/Arduino/hardware/espressif/esp32/libraries/Network/src/NetworkUdp.h:68:8: note: declared here
   68 |   void flush();  // Print::flush tx
      |        ^~~~~
```

### DIFF
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -369,7 +369,9 @@ int NetworkClient::read() {
   return data;
 }
 
-void NetworkClient::flush() {}
+void NetworkClient::flush() {
+  clear();
+}
 
 size_t NetworkClient::write(const uint8_t *buf, size_t size) {
   int res = 0;
--- a/libraries/Network/src/NetworkClient.h
+++ b/libraries/Network/src/NetworkClient.h
@@ -56,6 +56,7 @@ public:
   size_t write(const uint8_t *buf, size_t size);
   size_t write_P(PGM_P buf, size_t size);
   size_t write(Stream &stream);
+  [[deprecated("Use clear() instead.")]]
   void flush();  // Print::flush tx
   int available();
   int read();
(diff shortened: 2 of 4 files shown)


JAndrassy:
but there are multiplatform libraries which invoke `flush()` for TX and they will get the deprecation warning too

Jason2866:
@JAndrassy Ten times better than your code breaking change!!

JAndrassy:
@Jason2866 loosing RX data is not good.

me-no-dev:
> but there are multiplatform libraries which invoke flush() for TX and they will get the deprecation warning too

if those libs worked on ESP32, then they would expect to really `clear()`, so getting the deprecation warning is OK.

JAndrassy:
@me-no-dev multiplatform libraries which don't care about the specific platform and just are at risk not working correctly on the esp32 plaform

me-no-dev:
@JAndrassy you need to be able to accept some compromise. Breaking ESP specific libs is also not a good thing. This is such compromise that will allow libs to move to clear and in the future have your way.

Issue #10238, opened 2024-08-25
ESP.getChipModel() returns incorrect chip name: "ESP32-D0WD" instead of "ESP32-D0WD-V3" 


mscott370:
### Board

ESP32-DEVKITC-VIE

### Device Description

ESP32 WROVER-IE chip on DevKitC V4 board. 
- Using Arduino IDE board definition "ESP32 Wrover Kit (all versions)"

### Hardware Configuration

No, only serial USB connection to Arduino IDE host.

### Version

v3.0.4

### IDE Name

Arduino IDE v2.3.2

### Operating System

Both on Windows 11 and on Ubuntu Linux 24.04

### Flash frequency

80Mhz

### PSRAM enabled

yes

### Upload speed

115200

### Description

When uploading code to ESP32, ESPtool.py identifies the chip model correctly as ESP32-D0WD-V3, but ESP.getChipModel() returns a value of "ESP32-D0WD"  which indicates chip model V1 not V3.
This error seems to be caused by the following code in ESP.cpp 
 270   case EFUSE_RD_CHIP_VER_PKG_ESP32D0WDQ5:
 271       if (getChipRevision() == 3) {
 272         return "ESP32-D0WD-V3";
 273       } else {
 274          return "ESP32-D0WD";
 275       }
The above code was valid in IDF V4.4 when the value returned by getChipRevision() was simply the major wafer revision number (3) (uint8_t). This value was changed in IDF v5 to uint16_t, consisting of [major version]*100 + [minor version]. In my case, the value returned is 303, so as 303 is not == 3, we fall through into the else clause and return Chip Model = "ESP32-D0WD".
  
A quick+dirty change of the if-statement to  "if (getChipRevision() **/100**   == 3) {" produces the correct result, so it looks like this is an IDF v5 / Arduino-ESP32 V3 migration issue. 

### Sketch

```cpp
String chip_name =  ESP.getChipModel();
  Serial.println("Chip Model = " + chip_name);   // Display initial result from GetChipModel()
  esp_chip_info(&chipinfo);                      // Fill chipinfo structure
  bool rev3 = (chipinfo.revision/100) == 3;
  bool single_core = chipinfo.cores == 1;
  if (chip_name == "ESP32-D0WD" && rev3)  {      // If result invalid (Rev3 not valid with D0WD-V1),    
     chip_name += "-V3";                         // modify chip name
     Serial.println("Real Chip Model = " + chip_name);
     }
```

### Debug Message

```plain
None
[...]

PR #10243, opened 2024-08-26
fix(chip): Return correct ESP32 chip model

me-no-dev:
Fixes: https://github.com/espressif/arduino-esp32/issues/10238


P-R-O-C-H-Y (approved):
Also tested by the issue reporter @mscott370.

Issue #10237, opened 2024-08-24
PPP.begin() fails when modem is in CMUX mode


szerwi:
### Board

ESP32 S3

### Device Description

ESP32 S3 DevKitC

### Hardware Configuration

ESP32 connected to SIM7600E (BK-SIM7600E board) using only RX and TX, without RST pin.

### Version

v3.0.4

### IDE Name

PlatformIO (Arduino as ESP IDF component)

### Operating System

Windows 10

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

When RST pin in GSM modem is not used, it is not possible to reconnect to the module (eg. call `PPP.begin()` after ESP32 restart) when the modem was already switched to CMUX mode.

`PPP::begin()` tries to switch from data mode to command mode when it does not get the proper response for `AT` command, but it does not consider the case when the modem is in CMUX mode.

I have managed to temporary fix it by making changes in `PPP::begin()`, starting from line 312: https://github.com/espressif/arduino-esp32/blob/master/libraries/PPP/src/PPP.cpp#L312
```cpp
/* Wait for Modem to respond */
  if (_pin_rst >= 0) {
    // wait to be able to talk to the modem
    log_v("Waiting for response from the modem");
    while (esp_modem_sync(_dce) != ESP_OK && trys < 100) {
      trys++;
      delay(500);
    }
    if (trys >= 100) {
      log_e("Failed to wait for communication");
      goto err;
    }
  } else {
    // try to communicate with the modem
    if (esp_modem_sync(_dce) != ESP_OK) {
      log_v("Modem does not respond to AT, maybe in DATA mode? ...exiting network mode");

    // 
    // FIX START
    //
      log_v("Force switching mode to CMUX");
      _mode = ESP_MODEM_MODE_CMUX;

      log_v("Switching mode to command");
      esp_err_t cmd_err = esp_modem_set_mode(_dce, ESP_MODEM_MODE_COMMAND);
      if(cmd_err != ESP_OK) {
        log_w("Failed to switch mode to COMMAND! %d", cmd_err);
      }
      else {
        log_v("Mode switched to COMMAND");
      }

      log_v("Switching mode to CMUX");
      esp_err_t cmux_err = esp_modem_set_mode(_dce, ESP_MODEM_MODE_CMUX);
      if(cmux_err != ESP_OK) {
        log_w("Failed to switch mode to CMUX! %d", cmux_err);
      }
      else {
[...]

PR #10264, opened 2024-08-30
fix(ppp): Fix PPP not working after reboot if no RST

me-no-dev:
Modem will try to switch to CMUX and then back to COMMAND in order to allow communication with the modem.

This does not work on SIM800, so RST pin should always be connected to ensure proper reset is possible on boot.

Fixes: https://github.com/espressif/arduino-esp32/issues/10237


szerwi:
I can confirm that the fix works for SIM7600E modem, so it resolves my issue.

lucasssvaz (approved):

Issue #10281, opened 2024-09-03
WiFi scanNetworks may not find all APs


TD-er:
### Board

ESP32-C3/C6

### Device Description

-

### Hardware Configuration

-

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

Any

### PSRAM enabled

yes

### Upload speed

115200

### Description

The example WiFi scan code may not show all available APs.
Usually only a very small subset.

In ESP-IDF5.3 the struct `wifi_scan_config_t` does have a few new members:

```c++
    uint8_t home_chan_dwell_time;                      /**< time spent at home channel between scanning consecutive channels. */
    wifi_scan_channel_bitmap_t channel_bitmap;         /**< Channel bitmap for setting specific channels to be scanned. For 2.4ghz channels set ghz_2_channels from BIT(1) to BIT(14) from LSB to MSB order to indicate channels to be scanned. Currently scanning in 5ghz channels is not supported. Please note that the 'channel' parameter above needs to be set to 0 to allow scanning by bitmap. */
```

[link](https://github.com/espressif/esp-idf/blob/2c46030bbfd923d98a8fc28f24860e8ac935a151/components/esp_wifi/include/esp_wifi_types_generic.h#L193-L195)

According to the comments and docs it looks like whenever these members are set to 0, some predefined default is used. See: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_wifi.html#_CPPv428esp_wifi_set_scan_parametersPK26wifi_scan_default_params_t
However for the channel bitmap no default is mentioned, so I guess we should just set those bits ourselves.

Suggested code change for `WiFiScanClass::scanNetworks`

```c++
  if (channel == 0) {
    config.channel_bitmap.ghz_2_channels = 0xFFFF;
  } else {
    config.channel_bitmap.ghz_2_channels = BIT(channel);
  }
  config.home_chan_dwell_time = 30; // The home_chan_dwell_time needs to be a minimum of 30ms and a maximum of 150ms.
```

N.B. The esp-idf example code does do a `calloc` to sneakily set all to 0 and thus hiding the fact they do not properly initialize their members.
See: https://github.com/espressif/esp-idf/blob/6673376297b921d438790d195014b860eaf8bb17/examples/wifi/scan/main/scan.c#L174-L182

[...]

PR #10282, opened 2024-09-03
fix(scan): Make sure scan struct is cleared before set

me-no-dev:
new parameters could cause issues if scan is not cleared

fixes: https://github.com/espressif/arduino-esp32/issues/10281

### DIFF
--- a/libraries/WiFi/src/WiFiScan.cpp
+++ b/libraries/WiFi/src/WiFiScan.cpp
@@ -77,6 +77,7 @@ int16_t
   scanDelete();
 
   wifi_scan_config_t config;
+  memset(&config, 0, sizeof(wifi_scan_config_t));
   config.ssid = (uint8_t *)ssid;
   config.bssid = (uint8_t *)bssid;
   config.channel = channel;


Jason2866:
Tested with https://github.com/TD-er/MCVE_ESPxx and a C6. Before only two APs where found. With the fix all APs around are found.

TD-er:
Also tried here using the same MCVE code ;) on a C3 which was failing very well before and now working just fine.

TD-er:
This PR should also address this macro:
```c++
#define WIFI_SCAN_PARAMS_DEFAULT_CONFIG() { \
    .scan_time.active.min = WIFI_ACTIVE_SCAN_MIN_DEFAULT_TIME, \
    .scan_time.active.max = WIFI_ACTIVE_SCAN_MAX_DEFAULT_TIME, \
    .scan_time.passive = WIFI_PASSIVE_SCAN_DEFAULT_TIME, \
    .home_chan_dwell_time = WIFI_SCAN_HOME_CHANNEL_DWELL_DEFAULT_TIME\
}
```

This macro is suggested to be used in the comments for `esp_wifi_set_scan_parameters` or maybe it is used inside this function too???
So it should at least also set the channel bitmap to some default or 0.

P-R-O-C-H-Y (approved):

brentru:
@P-R-O-C-H-Y @me-no-dev @TD-er  As that macro is important and part of the ESP-IDF code, would it need to be addressed before a new version arduino-esp32 is released next or does this PR work without it?

me-no-dev:
@brentru we do not use that macro in Arduino

PR #10293, opened 2024-09-04
change(esp_now_serial): No teardown on retry limit

ClockeNessMnstr:
## Description of Change

After max retries is met once the ESP_NOW_Serial_Class performs "end()" which removes the peer from ESP_NOW. 
Further messages to and from ESP_NOW_Serial are not received or sent. 
Peer should stay in ESP_NOW to re-establish connection even with data loss. 
This change will "retry and drop" the data piece by piece instead of aborting the connection.

ESP_NOW_Serial "Peer" will not remove itself from ESP_NOW_Peer list when data sent is not received by a peer. ESP_NOW_Serial::end() is still called when ESP_NOW::send() fails.

## Tests scenarios

Tested on a pair of M5Stack Stamp ESP32-C3.
Confirmed that transmission is able to continue after:  disconnecting the Receiving module from power.
Attempting to send data.
Restoring power to Receiving module.

Previously, transmission halts as the cleanup/teardown in ESP_NOW_Serial::end() is triggered. 

## Related links

I did not create an issue for this.
Issue was NowSerial had to be re-initialized essentially for every byte without knowing if the peer was present/restored.

Workaround was to use broadcast address for TX and a separate ESP_NOW_Serial peer with MAC address for RX.

### DIFF
--- a/libraries/ESP_NOW/src/ESP32_NOW_Serial.cpp
+++ b/libraries/ESP_NOW/src/ESP32_NOW_Serial.cpp
@@ -264,9 +264,10 @@ void ESP_NOW_Serial_Class::onSent(bool success) {
       //the data is lost in this case
       vRingbufferReturnItem(tx_ring_buf, queued_buff);
       queued_buff = NULL;
-      xSemaphoreGive(tx_sem);
-      end();
       log_e(MACSTR " : RE-SEND_MAX[%u]", MAC2STR(addr()), resend_count);
+      //send next packet?
+      //log_d(MACSTR ": NEXT", MAC2STR(addr()));
+      checkForTxData();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ClockeNessMnstr:
This seems to stop ESP_NOW_Serial::write from returning 0 although it returns "1" for the first attempt anyway. 
Going to make a couple issues instead.

ClockeNessMnstr:
Actually it returns "1" until it does the "teardown" even though it did not succeed in sending the first byte(s). 

This is not due to my change.

me-no-dev:
The idea makes sense, @P-R-O-C-H-Y please make sure it's implemented in the best way.

P-R-O-C-H-Y:
@ClockeNessMnstr Hi, I did a small change to your PR so we are not bringing any breaking change and have this option as optional. Can you please test if it's ok for you? Thanks

me-no-dev (approved):

PR #10310, opened 2024-09-09
Fixed Stream find hang.

NathanSweet:
## Description of Change
Check `timedRead` is < 0, then use `char` (rather than `int`) for the rest of `findMulti`. This avoids sign extension and incorrect comparisons which result in a hang.

## Tests scenarios
This code hangs forever, even when Serial repeatedly receives 0xCA, tested on Arduino Mega 2560 R3:
```
while (!Serial1.find(0xCA)) ;
```

The reason is in Stream `findMulti`:
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/Stream.cpp#L132

Consider this comparison:
```
if (c == t->str[t->index]) {
```
`c` is `int` with an unsigned value 0xCA.
`t->str` is `const char *` with (on Arduino Mega 2560 R3) a signed value -54 decimal which as an unsigned value is 0xCA.
In the comparison, the char is sign extended to an int, resulting in 0xFFFFFFCA giving:
```
if (0xCA == 0xFFFFFFCA) {
```
This is never true and not the intention. There appear to be other comparisons in Stream.cpp with this problem.

### DIFF
--- a/cores/esp32/Stream.cpp
+++ b/cores/esp32/Stream.cpp
@@ -122,10 +122,11 @@ int Stream::findMulti(struct Stream::MultiTarget *targets, int tCount) {
   }
 
   while (1) {
-    int c = timedRead();
-    if (c < 0) {
+    int cc = timedRead();
+    if (cc < 0) {
       return -1;
     }
+    char c = cc;
 
     for (struct MultiTarget *t = targets; t < targets + tCount; ++t) {
       // the simple case is if we match, deal with that first.


me-no-dev:
Hi @NathanSweet ! Thank you for the PR. Please note that Stream and other Arduino classes come from upstream. I suggest you propose your fix first [in the official Arduino core API](https://github.com/arduino/ArduinoCore-API/blob/master/api/Stream.cpp#L267) and when accepted, let us know so we can update it on our end.

I will close this PR for now

NathanSweet:
@me-no-dev I looked at Stream.cpp over there and they have fixed the problem in 2018:
https://github.com/arduino/ArduinoCore-API/blob/master/api/Stream.cpp#L273

Why is the Arduino code in your repo so out of date?
https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/Stream.cpp#L132

me-no-dev:
> Why is the Arduino code in your repo so out of date?

There are quite a few reasons why that is so. I'll update the class. Can you please post an issue, so we can track it's progress?

me-no-dev:
Here is the update https://github.com/espressif/arduino-esp32/pull/10328

NathanSweet:
Cool, thank you!

I didn't mean to be rude, it's just surprising to see a dependency forked and missing fixes from 6 years ago. I guess the Arduino codebase has done things that aren't agreeable with ESP32?

PR #10319, opened 2024-09-10
Fix(NetworkEvents): Don't skip event callbacks in NetworkEvents::removeEvent

LeeLeahy2:
## Description of Change
Modified the four (4) NetworkEvents::removeEvent routines in libraries/Network/src/NetworkEvents.cpp to decrement "i" after the event callback is erased.

## Tests scenarios
Requires Fix [10317] to prevent crashes.

Used the submitted sketch for issue [10318](https://github.com/espressif/arduino-esp32/issues/10318).  Without this change each actual event generates 4 callbacks initially and 2 callbacks after calling removeEvent.  With this change all event handlers are removed and no further callbacks are called.

## Related links
Closes issue [10318](https://github.com/espressif/arduino-esp32/issues/10318)

### DIFF
--- a/libraries/Network/src/NetworkEvents.cpp
+++ b/libraries/Network/src/NetworkEvents.cpp
@@ -221,6 +221,7 @@ void NetworkEvents::removeEvent(NetworkEventCb cbEvent, arduino_event_id_t event
     NetworkEventCbList_t entry = cbEventList[i];
     if (entry.cb == cbEvent && entry.event == event) {
       cbEventList.erase(cbEventList.begin() + i);
+      i--; // Don't skip over any entries in the list
     }
   }
 }
@@ -240,6 +241,7 @@ void NetworkEvents::removeEvent(NetworkEventFuncCb cbEvent, arduino_event_id_t e
     NetworkEventCbList_t entry = cbEventList[i];
     if (getStdFunctionAddress(entry.fcb) == getStdFunctionAddress(cbEvent) && entry.event == event) {
       cbEventList.erase(cbEventList.begin() + i);
+      i--; // Don't skip over any entries in the list
     }
   }
 }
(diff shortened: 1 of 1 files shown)


me-no-dev:
Hi @LeeLeahy2 ! I do disagree with the solution a bit. IMHO the fix in this case should be in `onEvent`, where the code should check if this same callback is already assigned and not allow to assign it again. You are attaching the same function 4 times, which is not a valid use case. Only exception is to attach the same function only for specific events, which the solution I propose covers as well. Do you agree?

VojtechBartoska:
Hello @LeeLeahy2, thanks for the proposal.

Please sign CLA too, it is needed to unblock this PR. One of the requirements, we will see how the discussion will be shaped.

me-no-dev:
> instead of `i--` I would add `break`

Actually return and then also add log error at the end that callback was not found

LeeLeahy2:
I made the choice based upon the original implementation which continued scanning the list instead of having the break after the entry was found.  As such, it looked like multiple entries were to be allowed, so I provided an example sketch to support this behavior.  This also was the minimal change and did change the implementations behavior.

The alternative was a much larger change and was also possibly a change in desired behavior, although this is the behavior that I would have expected:
* Add the breaks in the removeEvent routines
* Possibly issue an error when the entry was not found in removeEvent routines
* Modify the onEvent routines to verify that the entry was not in the list before adding the entry to the list
* Modify the onEvent routines to issue an error when the entry is found on the list

I will submit another patch that implements the alternative behavior above.

me-no-dev:
Thanks @LeeLeahy2 looking forward the new patch

me-no-dev:
@LeeLeahy2 we  can close this one now, correct?

PR #10335, opened 2024-09-12
[WiFiScan] Allow allocation in _scanDone() to fail and prevent memory leak

TD-er:
## Description of Change
When there are many AP's seen during a scan, the allocation of `_scanResult` may fail. 
Thus add `(std::nothrow)` to the `new` call.  Without it, the next check for a valid pointer was not even done as the application would already have crashed.

Also it is possible the array was still present before allocating a new one, so we must make sure it is deleted first.

## Related links

Follow-up of this PR: https://github.com/espressif/arduino-esp32/pull/10312

## Possible (known) issue

There is 1 possible known issue left which is not being dealt with by this PR.
Since the `_scanDone()` function is called from a callback from an async scan, in theory it is possible someone is still processing a previous scan result via `_getScanInfoByIndex()`.
However if this does happen, then without this fix we would have had a memory leak for sure.

To work around such corner cases, quite a lot of code changes may be needed, which is out of scope of this PR.

### DIFF
--- a/libraries/WiFi/src/WiFiScan.cpp
+++ b/libraries/WiFi/src/WiFiScan.cpp
@@ -117,8 +117,13 @@ int16_t
  */
 void WiFiScanClass::_scanDone() {
   esp_wifi_scan_get_ap_num(&(WiFiScanClass::_scanCount));
+  if (WiFiScanClass::_scanResult) {
+    delete[] reinterpret_cast<wifi_ap_record_t *>(WiFiScanClass::_scanResult);
+    WiFiScanClass::_scanResult = 0;
+  }
+
   if (WiFiScanClass::_scanCount) {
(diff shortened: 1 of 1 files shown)


libraries/WiFi/src/WiFiScan.cpp
```diff
@@ -117,8 +117,13 @@ int16_t
  */
 void WiFiScanClass::_scanDone() {
   esp_wifi_scan_get_ap_num(&(WiFiScanClass::_scanCount));
+  if (WiFiScanClass::_scanResult) {
+    delete[] reinterpret_cast<wifi_ap_record_t *>(WiFiScanClass::_scanResult);
+    WiFiScanClass::_scanResult = 0;
```
me-no-dev, line 122: using `NULL` or `nullptr` here is preferable (even though it all resolves to `0`)

me-no-dev (approved):

P-R-O-C-H-Y (approved):

PR #10339, opened 2024-09-12
[WebServer] Mark functions as const + reduce copy of strings

TD-er:
## Description of Change
A few functions in WebServer does have `String` arguments, which does lead to copying of the strings when calling those functions.
For those where applicable, I changed the function arguments to `const String&`.

Also some functions which do not modify members, were not marked as `const`. Added those too.


me-no-dev (approved):

P-R-O-C-H-Y (approved):

TD-er:
Wow, quick merges :)

PR #10345, opened 2024-09-13
Reduce copy of Strings in WebServer RequestHandler

TD-er:
## Description of Change
Nearly each function of WebServer RequestHandler does copy the String arguments instead of using `const String&`.

N.B. This might break some classes which inherit from `RequestHandler`

### DIFF
--- a/libraries/WebServer/src/detail/RequestHandlersImpl.h
+++ b/libraries/WebServer/src/detail/RequestHandlersImpl.h
@@ -21,7 +21,7 @@ public:
     delete _uri;
   }
 
-  bool canHandle(HTTPMethod requestMethod, String requestUri) override {
+  bool canHandle(HTTPMethod requestMethod, const String& requestUri) override {
     if (_method != HTTP_ANY && _method != requestMethod) {
       return false;
     }
@@ -29,7 +29,7 @@ public:
     return _uri->canHandle(requestUri, pathArgs);
   }
 
-  bool canUpload(String requestUri) override {
+  bool canUpload(const String& requestUri) override {
     if (!_ufn || !canHandle(HTTP_POST, requestUri)) {
       return false;
     }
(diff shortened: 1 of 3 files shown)


libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -146,21 +146,20 @@ class StaticRequestHandler : public RequestHandler {
...
 
     String path(_path);
-    String eTagCode;
 
     if (!_isFile) {
       // Base URI doesn't point to a file.
       // If a directory is requested, look for index file.
       if (requestUri.endsWith("/")) {
-        requestUri += "index.htm";
+        return handle(server, requestMethod, String(requestUri + "index.htm"));
```
me-no-dev, line 162: why?

libraries/WebServer/src/detail/RequestHandlersImpl.h
```diff
@@ -146,21 +146,20 @@ class StaticRequestHandler : public RequestHandler {
...
-  bool handle(WebServer &server, HTTPMethod requestMethod, String requestUri) override {
+  bool handle(WebServer &server, HTTPMethod requestMethod, const String& requestUri) override {
     if (!canHandle(server, requestMethod, requestUri)) {
       return false;
     }
 
     log_v("StaticRequestHandler::handle: request=%s _uri=%s\r\n", requestUri.c_str(), _uri.c_str());
 
     String path(_path);
-    String eTagCode;
```
me-no-dev, line 157: moving it changes nothing

libraries/WebServer/src/detail/RequestHandlersImpl.h
TD-er: So we can keep the `requestUri` const

libraries/WebServer/src/detail/RequestHandlersImpl.h
TD-er: Less allocated on the stack when it returns early (e.g. due to the added single recursive call you asked about in the previous comment.

me-no-dev (approved):

P-R-O-C-H-Y (approved):

PR #10348, opened 2024-09-14
Fix missing virtual declarations in Stream.h

TD-er:
## Description of Change
Fixes some changes made in PR #10328

- Virtual destructor was removed
- Some functions had `virtual` removed

These changes do cause issues when operating on a `Stream` pointer instead of the derived class.

### DIFF
--- a/cores/esp32/Stream.h
+++ b/cores/esp32/Stream.h
@@ -61,6 +61,7 @@ public:
   Stream() {
     _timeout = 1000;
   }
+  virtual ~Stream() {}
 
   // parsing methods
 
@@ -105,8 +106,8 @@ public:
   float parseFloat(LookaheadMode lookahead = SKIP_ALL, char ignore = NO_IGNORE_CHAR);
   // float version of parseInt
 
-  size_t readBytes(char *buffer, size_t length);  // read chars from stream into buffer
-  size_t readBytes(uint8_t *buffer, size_t length) {
+  virtual size_t readBytes(char *buffer, size_t length);  // read chars from stream into buffer
+  virtual size_t readBytes(uint8_t *buffer, size_t length) {
     return readBytes((char *)buffer, length);
   }
   // terminates if length characters have been read or timeout (see setTimeout)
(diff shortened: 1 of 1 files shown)


JAndrassy:
the destructor should be in the Print class

TD-er:
> the virtual destructor should be in the Print class

OK, but the virtual declaration of the other functions should be there.

N.B. the Print class does have a virtual destructor, so I agree it is not needed in the `Stream` class.

me-no-dev (approved):

me-no-dev:
@TD-er  have you thought of proposing this upstream?

P-R-O-C-H-Y (approved):

TD-er:
> @TD-er have you thought of proposing this upstream?

Nope, as I don't know why it was already different on this repo to begin with.

To me it makes perfect sense why those functions should be virtual, but I can't think of any reason why they are not made virtual in the upstream code. 
Nor can I oversee what effect it may have on small microcontrollers as marking it virtual may add some to the build size and I guess `Stream.h` may be used in quite a lot of code.

me-no-dev:
it also adds up to 3K of flash here too. Should you maybe rethink your code and not overwrite those?

TD-er:
It does make both ESP32 and ESP8266 implementations different.
Also I don't know which other libraries may break because of this.

Not sure what will happen when calling these functions when you have a pointer to the `Stream` base class, I guess then the implementation of the derived class will not be called right?

me-no-dev:
You can give it a shot. I think the derived class will be called.

TD-er:
> You can give it a shot. I think the derived class will be called.

Nope, it will not.
See: https://www.programiz.com/cpp-programming/virtual-functions

However, according to this, we could declare these functions with `override` in the `Stream` class? (see linked page roughly half way)
Then it isn't declared virtual. 
This should reduce the bin file a bit, though I don't think it will be the same size as it is now.

JAndrassy:
> override specifier (since C++11) Specifies that a virtual function overrides another virtual function.

compiler will check if a virtual function with the specified name and parameter exists in the base class

TD-er:
> > override specifier (since C++11) Specifies that a virtual function overrides another virtual function.
> 
> compiler will check if a virtual function with the specified name and parameters exists in the base class

Yep, that's how I found this as the compiler complained.

However the question is if we could do without `virtual` here?
If not, then it is for sure also a bug in the upstream Arduino code.

PR #10385, opened 2024-09-27
ci(json): Add configuration requirements to ci.json files

lucasssvaz:
## Description of Change

Add the `requires` field to the CI JSON. This field will only compile the sketches for the targets that have the requirements in the `sdkconfig` file.

## Tests scenarios

Locally and CI

### DIFF
--- a/docs/en/contributing.rst
+++ b/docs/en/contributing.rst
@@ -109,17 +109,39 @@ Also:
 Testing
 *******
 
-Be sure you have tested the example in all the supported targets. If the example works only with specific targets,
-edit/add the ``ci.json`` in the same folder as the sketch to specify the supported targets. By default,
-all targets are assumed to be supported.
+Be sure you have tested the example in all the supported targets. If the example some specific hardware requirements,
+edit/add the ``ci.json`` in the same folder as the sketch to specify the required configurations from ``sdkconfig``.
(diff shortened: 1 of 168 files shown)


docs/en/contributing.rst
```diff
@@ -130,17 +157,17 @@ For example, in the sketch:
...
 
 .. code-block:: markdown
 
-    Currently, this example supports the following targets.
+    Currently, this example requires Wi-Fi and supports the following targets.
 
-    | Supported Targets | ESP32 | ESP32-S2 | ESP32-C3 | ESP32-S3 |
-    | ----------------- | ----- | -------- | -------- | -------- |
+    | Supported Targets | ESP32 | ESP32-H2 | ESP32-S3 | ESP32-C3 | ESP32-C6 |
+    | ----------------- | ----- | -------- | -------- | -------- | -------- |
```
P-R-O-C-H-Y, line 170: Just a comment here. I know its an example of the configuration, but H2 does not have wifi so idk if its good example here :D

P-R-O-C-H-Y (approved):

me-no-dev (approved):

docs/en/contributing.rst
lucasssvaz: Ooops, missed that. Thanks!

PR #10392, opened 2024-10-01
ci(compilation): Use default partition and add append to FQBN option

lucasssvaz:
## Description of Change

- Change the partition used by the CI to the same as Arduino by default
- Fix issues with requirements check
- Add option to append to all FQBNs

## Tests scenarios

CI and Locally

### DIFF
--- a/libraries/RainMaker/examples/RMakerCustomAirCooler/ci.json
+++ b/libraries/RainMaker/examples/RMakerCustomAirCooler/ci.json
@@ -1,5 +1,5 @@
 {
-  "fqbn_append": "PartitionScheme=huge_app",
+  "fqbn_append": "PartitionScheme=rainmaker_4MB",
   "requires": [
     "CONFIG_SOC_WIFI_SUPPORTED=y",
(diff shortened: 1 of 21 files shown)


me-no-dev (approved):

me-no-dev:
@P-R-O-C-H-Y PTAL

@lucasssvaz we will merge this later today when you are at work, because I expect there to be issues with the 3.1 branch, plus will need to make it also work with P4

libraries/RainMaker/examples/RMakerCustomAirCooler/ci.json
```diff
@@ -1,6 +1,7 @@
 {
+  "fqbn_append": "PartitionScheme=huge_app",
```
P-R-O-C-H-Y, line 2: For RainMaker I suggest to use "RainMaker 4MB no OTA partition".

lucasssvaz: I just used the same one that was being used by the examples. But I can change that

Issue #10296, opened 2024-09-05
ESP32-S3 - Version 3.0 breaks USBHIDKeyboard LEDs events (Caps Lock, Num Lock, etc.)


TheCrypt0:
### Board

ESP32-S3

### Device Description

Custom ESP32-S3 keyboard but it should work on any ESP32-S3 devkit with USB connection.

### Hardware Configuration

No.

### Version

v3.0.4

### IDE Name

Arduino IDE

### Operating System

macOS 14.6.1

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

I'm currently building a USB keyboard using the ESP32-S3 and when upgrading to the latest Arduino Core, the reporting of LEDs (Caps Lock, Num Lock, etc.) did not happen anymore, even though the typing works. Downgrading to 2.0.13 makes the code work again. 

I attached a minimum reproducible example with the expected output when the devkit is connected to a PC.

### Sketch

```cpp
#include <USB.h>
#include <USBCDC.h>
#include <USBHIDKeyboard.h>

USBHIDKeyboard m_keyboard;
USBCDC         m_serial;
ESPUSB*        m_usb;

static void
usbEventCallback(void*            arg,
                 esp_event_base_t event_base,
                 int32_t          event_id,
                 void*            event_data)
{
  if (event_base == ARDUINO_USB_EVENTS) {
    arduino_usb_event_data_t* data = (arduino_usb_event_data_t*)event_data;
    switch (event_id) {
      case ARDUINO_USB_STARTED_EVENT:
        m_serial.println("USB PLUGGED");
        break;
      case ARDUINO_USB_STOPPED_EVENT:
        m_serial.println("USB UNPLUGGED");
        break;
      case ARDUINO_USB_SUSPEND_EVENT:
        m_serial.printf("USB SUSPENDED: remote_wakeup_en: %u\n",
                        data->suspend.remote_wakeup_en);
        break;
      case ARDUINO_USB_RESUME_EVENT:
        m_serial.println("USB RESUMED");
        break;

      default:
        break;
    }
  } else if (event_base == ARDUINO_USB_HID_EVENTS) {
    arduino_usb_hid_event_data_t* data =
      (arduino_usb_hid_event_data_t*)event_data;
    switch (event_id) {
      case ARDUINO_USB_HID_SET_PROTOCOL_EVENT:
        m_serial.printf("HID SET PROTOCOL: %s\n",
[...]

PR #10398, opened 2024-10-02
fix(hid): Fix OUTPUT report not received

me-no-dev:
Fixes an issue where the keyboard LEDs will not trigger the proper event

Fixes: https://github.com/espressif/arduino-esp32/issues/10296


niarbrnd:
Thanks, this fix is good

PR #10409, opened 2024-10-02
test(psram): Add PSRAM test

lucasssvaz:
## Description of Change
Add validation test for PSRAM

## Tests scenarios

CI and Locally

### DIFF
--- a/tests/validation/psram/ci.json
+++ b/tests/validation/psram/ci.json
@@ -1,0 +1,26 @@
+{
+  "fqbn": {
+    "esp32": [
+      "espressif:esp32:esp32:PSRAM=enabled"
+    ],
+    "esp32s2": [
+      "espressif:esp32:esp32s2:PSRAM=enabled"
+    ],
+    "esp32s3": [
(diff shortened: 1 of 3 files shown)


me-no-dev (approved):

me-no-dev:
@P-R-O-C-H-Y PTAL

P-R-O-C-H-Y (approved):
LGTM. @lucasssvaz I am worried about the S3 as there are 2 PSRAM options (OPI and quad). Do we have both runners available and also how the test recognize which to chose?

lucasssvaz:
> LGTM. @lucasssvaz I am worried about the S3 as there are 2 PSRAM options (OPI and quad). Do we have both runners available and also how the test recognize which to chose?

Yeah, forgot to consider this. I need to check which PSRAM type our hardware runners have. If they were running properly that would be easier XD.

I also forgot to add the diagram files for wokwi with the PSRAM enabled

P-R-O-C-H-Y (approved):

Issue #10391, opened 2024-09-30
Arduino WiFi AP API coexistence with esp-matter ConnectivityManager Implementation


HamzaHajeir:
### Board

ESP32-S3-DevKitC-1-N8R2

### Device Description

.

### Hardware Configuration

..

### Version

v3.0.4

### IDE Name

PlatformIO

### Operating System

Windows 11

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

.

### Description

I'm building a custom project which coexists esp-matter and Arduino in one project. 

Such that the WiFi management is went through Arduino library (WiFi), and to inform CHIP's core that the application controls the WiFi. through:
```cpp
auto &connMgr = chip::DeviceLayer::ConnectivityMgrImpl();
connMgr.SetWiFiStationMode(chip::DeviceLayer::ConnectivityManager::WiFiStationMode::kWiFiStationMode_ApplicationControlled);
connMgr.SetWiFiAPMode(chip::DeviceLayer::ConnectivityManager::WiFiAPMode::kWiFiAPMode_ApplicationControlled);
```

All up to here is okay, but when engaging with `WiFi.enableAP(true)` API an error happens, which is a result of incompatibility with esp-matter's code regarding WiFi. 

Both stacks initializes the netif`'s ([Arduino](https://github.com/espressif/arduino-esp32/blob/e403f0b4815643ccd2d98f62b7accf5a8108f2cd/libraries/WiFi/src/WiFiGeneric.cpp#L261-L266))(esp-matter[[1](https://github.com/espressif/esp-matter/blob/9350d9d5f948d3b7c61c8659c4d6990d0ff00ea4/components/esp_matter/esp_matter_core.cpp#L996-L1001)][[2](https://github.com/espressif/connectedhomeip/blob/17b1a38e909e7874593bcb87c31be03a5866f1d4/src/platform/ESP32/ESP32Utils.cpp#L323-L343)]). While there should be exactly one netif initialized, the program crashes in the second call of.

A suggestion of using the name keys, and the same pattern of code in esp-matter's `start()`, such that to ask for the netif by name, and if not existed to create it:

```cpp
      if ((esp_netifs[ESP_IF_WIFI_AP] = esp_netif_get_handle_from_ifkey("WIFI_AP_DEF")) == NULL) {
        esp_netifs[ESP_IF_WIFI_AP] = esp_netif_create_default_wifi_ap();
      }
[...]

PR #10415, opened 2024-10-03
fix(asyncudp): Fixes and implements tcpip thread locking

HamzaHajeir:
## Description of Change
Implements and fixes TCPIP Core locking in AsyncUDP when enabled.

AsyncUDP calls Raw LwIP APIs that are not safe to call from other threads without locking. While some raw APIs are being [called by using `tcpip_api_call()`](https://github.com/espressif/arduino-esp32/blob/733373a049d8a06fab00628bcae585550fac6d07/libraries/AsyncUDP/src/AsyncUDP.cpp#L48-L135), other calls are not considering this method, as [[1]](https://github.com/espressif/arduino-esp32/blob/733373a049d8a06fab00628bcae585550fac6d07/libraries/AsyncUDP/src/AsyncUDP.cpp#L31), [[2]](https://github.com/espressif/arduino-esp32/blob/733373a049d8a06fab00628bcae585550fac6d07/libraries/AsyncUDP/src/AsyncUDP.cpp#L476-L481), [[3]](https://github.com/espressif/arduino-esp32/blob/733373a049d8a06fab00628bcae585550fac6d07/libraries/AsyncUDP/src/AsyncUDP.cpp#L495), and [[4]](https://github.com/espressif/arduino-esp32/blob/master/libraries/AsyncUDP/src/AsyncUDP.cpp#L640).

When `CONFIG_LWIP_CHECK_THREAD_SAFETY` is enabled, the system gets aborted when the lwip core is not locked; Therefore exposing a source of bugs.

Any call to `tcpip_api_call()` in tcpip core locked results in a deadlock, thus removed the corresponding calls.

I've prefered to check against the state before locking and unlocking, helpful when any concerned AsyncUDP API being called from within LwIP thread, highly through callbacks, if any.

This will not affect the direct Arduino users but until [esp32-arduino-libs-builder PR](https://github.com/espressif/esp32-arduino-lib-builder/pull/186) gets merged it will fix a source of bugs.

Currently: Arduino-as-ESPIDF-component users can easily have this fix by enabling `CONFIG_LWIP_TCPIP_CORE_LOCKING` option.

## Related links

https://www.nongnu.org/lwip/2_1_x/multithreading.html

https://github.com/espressif/arduino-esp32/issues/10391#issuecomment-2385443646

### DIFF
--- a/libraries/AsyncUDP/src/AsyncUDP.cpp
+++ b/libraries/AsyncUDP/src/AsyncUDP.cpp
@@ -488,6 +488,7 @@ bool AsyncUDP::_init() {
   UDP_MUTEX_LOCK();
   _pcb = udp_new();
   if (!_pcb) {
+    UDP_MUTEX_UNLOCK();
     return false;
   }
   udp_recv(_pcb, &_udp_recv, (void *)this);
--- a/libraries/AsyncUDP/src/AsyncUDP.cpp
+++ b/libraries/AsyncUDP/src/AsyncUDP.cpp
@@ -16,17 +16,19 @@ extern "C" {
 #include "lwip/priv/tcpip_priv.h"
 
 #ifdef CONFIG_LWIP_TCPIP_CORE_LOCKING
-#define UDP_MUTEX_LOCK()      if (!sys_thread_tcpip(LWIP_CORE_LOCK_QUERY_HOLDER)) { \
-                                LOCK_TCPIP_CORE();                                  \
-                              }
+#define UDP_MUTEX_LOCK()                                \
+  if (!sys_thread_tcpip(LWIP_CORE_LOCK_QUERY_HOLDER)) { \
+    LOCK_TCPIP_CORE();                                  \
+  }
 
-#define UDP_MUTEX_UNLOCK()    if (sys_thread_tcpip(LWIP_CORE_LOCK_QUERY_HOLDER)) {  \
-                                UNLOCK_TCPIP_CORE();                                \
-                              }
-#else // CONFIG_LWIP_TCPIP_CORE_LOCKING
+#define UDP_MUTEX_UNLOCK()                             \
+  if (sys_thread_tcpip(LWIP_CORE_LOCK_QUERY_HOLDER)) { \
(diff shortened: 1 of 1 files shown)


me-no-dev (approved):

libraries/AsyncUDP/src/AsyncUDP.cpp
```diff
@@ -473,12 +485,13 @@ bool AsyncUDP::_init() {
   if (_pcb) {
     return true;
   }
+  UDP_MUTEX_LOCK();
   _pcb = udp_new();
   if (!_pcb) {
     return false;
```
TD-er, line 491: Shouldn't there be an unlock before the return?

HamzaHajeir:
Well spotted, missed to consider that path. Now should I add another commit
or amend the commit?

me-no-dev:
> Now should I add another commit or amend the commit?

just add another commit to this PR

HamzaHajeir:
> just add another commit to this PR

Done

VojtechBartoska:
Hello @HamzaHajeir, please sign CLA without that we are not able to merge this PR. Thanks!

HamzaHajeir:
> Hello @HamzaHajeir, please sign CLA without that we are not able to merge this PR. Thanks!

That bureaucratic :) 

Couldn't find that document, can you guide how to?

TD-er:
> > Hello @HamzaHajeir, please sign CLA without that we are not able to merge this PR. Thanks!
> 
> That bureaucratic :)
> 
> Couldn't find that document, can you guide how to?

See the 2nd post in this PR: https://github.com/espressif/arduino-esp32/pull/10415#issuecomment-2391191122

HamzaHajeir:
> > > Hello @HamzaHajeir, please sign CLA without that we are not able to merge this PR. Thanks!
> > 
> > 
> > That bureaucratic :)
> > Couldn't find that document, can you guide how to?
> 
> See the 2nd post in this PR: [#10415 (comment)](https://github.com/espressif/arduino-esp32/pull/10415#issuecomment-2391191122)

Thanks, done!

PR #10428, opened 2024-10-07
feat(uart) 3.1.x Core: fixes serialEventRun() to avoid calling available() if serialEvent() is not declared

### LINKED ISSUE #10397: In ESP32 core version V3.0.5 speed is 30-50% slower than V2.0.17 with Arduino Nano ESP32.
### Board

Arduino Nano ESP32

### Device Description

Arduino Nano ESP32

### Hardware Configuration

Nothing attached

### Version

v3.0.4

### IDE Name

Arduino IDE 1.0 and 2.0

### Operating System

Windows 11

### Flash frequency

NA

### PSRAM enabled

no

### Upload speed

NB Arduino settings in board

### Description

I noticed while transferring a large Arduino sketch on an Arduino Nano ESP32 from ESP32 core version 2.0.17 to ESP32 core V3..0.5 the software loop speeds dropped from 220,000 to 169,000 loops per second. 
The speed reduction is also with versions 3.0.1 - 3.0.4

With the sketch below the speed reduction is even more and almost 50%. 
I did not test the ESP32 Zero and C3-12F board with the ESP32 V2 core but I expect the same results in speed differences

Below the 'Loops per second' when adding one to the loop counter for one second.
No difference noticed using uint64_t or uint32_t

*** Arduino Nano ESP32 board on Arduino ESP32 core V2.0.17
Loops per second: 737744
Loops per second: 734583
Loops per second: 734451
Loops per second: 734451
Loops per second: 734451
Loops per second: 734451

*** Processors on V3.0.5

Arduino Nano ESP32
Loops per second: 380670
Loops per second: 380709
Loops per second: 380726
Loops per second: 380724
Loops per second: 380727

ESP32 S3 Zero 
Loops per second: 382485
Loops per second: 382488
Loops per second: 382487
Loops per second: 382487
Loops per second: 382487

ESP32-C3-12F
Loops per second: 125696
Loops per second: 126240
Loops per second: 125695
Loops per second: 126241
Loops per second: 125695
Loops per second: 126240
Loops per second: 125696

### Sketch

```cpp
uint32_t  Loopcounter = 0;
static uint32_t msTick;      

void setup() 
{
 Serial.begin(115200);                     // Setup the serial port to 115200 baud //
 msTick = millis(); 
}

void loop() 
{
 Loopcounter++;
 EverySecondCheck();
}

//--------------------------------------------//
// COMMON Update routine done every second
//--------------------------------------------
void EverySecondCheck(void)
{
 uint32_t msLeap = millis() - msTick;     // 
 if (msLeap >999)                         // Every second enter the loop
 {
  msTick = millis();
 Serial.printf("Loops per second: %u\n",Loopcounter);
 Loopcounter = 0;
 }  
}
```

### Debug Message

```plain
NA
```

### Other Steps to Reproduce

NA

SuGlider:
## Description of Change
loop() calls Serial Events functions when those are declared. 
The way it was declared, was forcing to always call avalable() to then call an empty function.
This commit fixes it.

Now when the sketch has no serialEvent() function declared, it won't be executed neither will available().
There are is a performance degradation when calling available() every single time when loop() was executed, whenever Serial.begin() is executed.

It shall be merged to 3.0.x and 3.1.x

## Tests scenarios

``` cpp
uint32_t  Loopcounter = 0;
static uint32_t msTick;      

void setup() 
{
 Serial.begin(115200);                     // Setup the serial port to 115200 baud //
 msTick = millis(); 

 Serial0.begin(115200); // THIS Line starts UART Arduino Driver and also forces to executing serialEventRun() that "eats" CPU time...
}

void loop() 
{
 Loopcounter++;
 EverySecondCheck();
}

//--------------------------------------------//
// COMMON Update routine done every second
//--------------------------------------------
void EverySecondCheck(void)
{
 uint32_t msLeap = millis() - msTick;     // 
 if (msLeap >999)                         // Every second enter the loop
 {
  msTick = millis();
 Serial.printf("Loops per second: %u\n",Loopcounter);
 Loopcounter = 0;
 }  
}
```

## Related links
Closes #10397

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -24,16 +24,13 @@
 #endif
 
 void serialEvent(void) __attribute__((weak));
-void serialEvent(void) {}
 
 #if SOC_UART_NUM > 1
 void serialEvent1(void) __attribute__((weak));
-void serialEvent1(void) {}
 #endif /* SOC_UART_NUM > 1 */
 
 #if SOC_UART_NUM > 2
 void serialEvent2(void) __attribute__((weak));
-void serialEvent2(void) {}
 #endif /* SOC_UART_NUM > 2 */
 
 #if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_SERIAL)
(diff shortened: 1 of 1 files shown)


TD-er:
Still why does it need to be called every loop() call?
Isn't every like msec or 100 usec enough?   (at 115200 you can only get a new character every 100 usec)
And shouldn't the available() check only be run when `begin()` was called? 

The best location for both checks is probably in the `available()` function of each serial device.

me-no-dev (approved):

me-no-dev:
> Still why does it need to be called every loop() call?

because there are higher speeds used by some and they do expect to get the event as soon as possible, else they come here and submit issue reports :)

P-R-O-C-H-Y (approved):

lucasssvaz (approved):

PR #10447, opened 2024-10-10
fix(uart): Add missing HP UARTs for ESP32-P4

lucasssvaz:
## Description of Change

Add missing HP UART 3 and 4 for ESP32-P4


SuGlider (approved):
LGTM.

PR #10470, opened 2024-10-16
fix(esp-rainmaker): Bump esp-rainmaker version

Asc91:
## Description of Change
Update esp-rainmaker version to 1.5.0 as asked in #10456


me-no-dev:
please create a new PR that targets the `release/v3.0.x` branch. Because of matter, versions in the `release/v3.1.x` branch are fixed to the older versions and updated ones will be available for now only on the 3.0.x releases

Issue #10500, opened 2024-10-22
PSRAM stops working when importing `arduino-esp32` into ESP-IDF. Flag `CONFIG_SPIRAM` becomes undefined when `BOARD_HAS_PSRAM` is not set.


cjdell:
### Board

Xiao ESP32-S3 Sense

### Device Description

Xiao ESP32-S3 Sense connected to PC with USB-C

### Hardware Configuration

Just USB-C to PC

### Version

latest master (checkout manually)

### IDE Name

VSCode

### Operating System

NixOS

### Flash frequency

40MHz

### PSRAM enabled

yes

### Upload speed

460800

### Description

Unfortunately my PSRAM stopped working as soon as I started using the `arduino-esp32` component. This code literally undefines the `CONFIG_SPIRAM` and `CONFIG_SPIRAM_SUPPORT` flag out of existence:

https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-psram.h#L29

There is no way to override them back. I even tried defining `BOARD_HAS_PSRAM` as a build flag but it hasn't worked. I haven't found docs on how to set global builds flags nor is there any information on the `BOARD_HAS_PSRAM` flag and any explanation as to why it exists when we already have flags in `sdkconfig` for those.

Manually removing lines 24-31 in `esp32-hal-psram.h` will get PSRAM working again but this is an uncomfortable workaround. 

Thanks

### Sketch

```cpp
#include <Arduino.h>

void setup()
{
  delay(2000);

  // DOES NOT WORK WITHOUT "BOARD_HAS_PSRAM" WHICH IS NOT DEFINED !!!
#if CONFIG_SPIRAM
  ESP_LOGI("main", "CONFIG_SPIRAM DEFINED");
#else
  ESP_LOGI("main", "CONFIG_SPIRAM ** NOT ** DEFINED");
#endif

  ESP_LOGI("main", "PSRAM Size: %i", (int)ESP.getPsramSize());
}

void loop()
{
}
```

### Debug Message

```plain
No CONFIG_SPIRAM defined
PSRAM reports 0 bytes (incorrect)
```

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [X] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

cjdell:
Better workaround for now, add the following line to `CMakeLists.txt` in the root of your project:

```cmake
add_compile_options(-DBOARD_HAS_PSRAM=1)
```

[...]

PR #10502, opened 2024-10-22
fix(component): Do not undefine PSRAM in ESP-IDF

me-no-dev:
Currently if Arduino is used as ESP-IDF component, `CONFIG_PSRAM` will always get undefined due to the fact that `BOARD_HAS_PSRAM` is not defined. This PR attempts to mitigate this problem by always defining (to 0 or 1) `BOARD_HAS_PSRAM` when in Arduino and when in ESP-IDF, to define `BOARD_HAS_PSRAM` based on `CONFIG_PSRAM `

Fixes: https://github.com/espressif/arduino-esp32/issues/10500

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -79,7 +79,7 @@ esp32c2.build.flash_freq=60m
 esp32c2.build.flash_mode=qio
 esp32c2.build.boot=qio
 esp32c2.build.partitions=minimal
-esp32c2.build.defines=
+esp32c2.build.defines=-DBOARD_HAS_PSRAM=0
 
 esp32c2.menu.PartitionScheme.minimal=Minimal (1.3MB APP/700KB SPIFFS)
 esp32c2.menu.PartitionScheme.minimal.build.partitions=minimal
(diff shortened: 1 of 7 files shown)


me-no-dev:
@P-R-O-C-H-Y you will need to keep in mind this new configuration in boards.txt

Jason2866:
@me-no-dev Will this may break existing Platformio boards.json manifests?
The are like this:
```
{
  "build": {
    "arduino":{
      "memory_type": "qio_opi"
    },
    "core": "esp32",
    "extra_flags": [
      "-DARDUINO_ESP32S3",
      "-DARDUINO_USB_CDC_ON_BOOT=1",
      "-DARDUINO_RUNNING_CORE=1",
      "-DARDUINO_EVENT_RUNNING_CORE=1",
      "-DBOARD_HAS_PSRAM"
    ],
    "f_cpu": "240000000L",
    "f_flash": "80000000L",
    "flash_mode": "qio",
.
.
.
```

me-no-dev:
@Jason2866 it will definitely break things. Will brainstorm some more

PR #10524, opened 2024-10-25
feat(esptool): Update esptool to v4.8.1

lucasssvaz:
## Description of Change

Use version 4.8.1 of esptool


Jason2866:
Why not esptool.py v4.8.1

lucasssvaz:
> Why not esptool.py v4.8.1

I updated it to 4.8.0 a while back but forgot to change this entry. The next esptool version will contain fixes need for arduino as well, so we will update again soon.

me-no-dev:
@Jason2866 4.8.1 contains the fix for usb reboot?

Jason2866:
@me-no-dev v4.8.1 does not contain the fix for USB reboot.
There is yet no release done with the commit containing the fix.
But using v4.8.1 is better anyways. It includes SHA update when the image header is changed

PR #10530, opened 2024-10-27
feat(ESP32): Add "-DARDUINO_ESP32_MCU_{build.mcu}" macro to platform.txt file

kiwi64ajs:
## Description of Change
Adding this macro definition allows MCU-specific preprocessor conditional expressions, enabling libraries to handle MCU-specific differences between the ESP32 MCUs and IDF data structures.

The specific issue I could not find a workaround for was handling the difference in UART State Structures which this macro allowed me to write code like the sample below, where the UART State register is located in different location/structure on some ESP32 MCUs.

```
			uart_dev_t *hw = UART_LL_GET_HW(_uart_nr);		 	

 		#if defined(ARDUINO_ESP32_MCU_esp32c3) || defined(ARDUINO_ESP32_MCU_esp32s3)    
		  	return hw->fsm_status.st_urx_out != 0;
			
 		#elif defined(ARDUINO_ESP32_MCU_esp32)  
 			return hw->status.st_urx_out != 0;

 		#endif
```

### DIFF
--- a/platform.txt
+++ b/platform.txt
@@ -148,13 +148,13 @@ recipe.hooks.prebuild.8.pattern=/usr/bin/env bash -c "cp -f "{compiler.sdk.path}
 recipe.hooks.prebuild.8.pattern.windows=cmd /c COPY /y "{compiler.sdk.path}\sdkconfig" "{build.path}\sdkconfig"
 
 ## Compile c files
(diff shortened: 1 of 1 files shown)


lucasssvaz:
There is no need to create new macros, you can use the ones already provided by ESP-IDF. Here are some examples:

https://github.com/espressif/arduino-esp32/blob/master/cores/esp32/esp32-hal-uart.c#L1023C5-L1023C28
https://github.com/espressif/arduino-esp32/blob/master/libraries/RainMaker/examples/RMakerSwitch/RMakerSwitch.ino#L12

PR #10533, opened 2024-10-28
fix(bm): Board Manager install will fail on Windows

me-no-dev:
On some Windows installations, Board Manager will fail to install the RC2 core, because it could not find esptool for the given OS.

### DIFF
--- a/package/package_esp32_index.template.json
+++ b/package/package_esp32_index.template.json
@@ -491,6 +491,13 @@
               "size": "29828542"
             },
             {
+              "host": "i686-mingw32",
+              "url": "https://github.com/espressif/arduino-esp32/releases/download/3.1.0-RC1/esptool-v4.8.1-win64.zip",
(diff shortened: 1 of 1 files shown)


me-no-dev:
@lucasssvaz if you want to prevent future issues like this, you can test in get.py for Windows if both OS targets are present and fail if not. It will cause the Windows CI to fail.

lucasssvaz:
@me-no-dev I think we should do the same with `i686-pc-linux-gnu` as the same issue might occur

me-no-dev:
@lucasssvaz can it be proven/tested?

lucasssvaz:
@me-no-dev Just tested and it will only happen if the 32bit version of the arduino IDE is installed (it will also fail to find other tools like openocd)

lucasssvaz (approved):

SuGlider (approved):

P-R-O-C-H-Y (approved):

Issue #7432, opened 2022-11-06
Add a Arduino Matter library to the road map?


Humancell:
### Related area

Arduino Library

### Hardware specification

All Boards

### Is your feature request related to a problem?

The Matter specification has been released, and there has been IDF support for quite some time. At one level is would be good to see examples of using these APIs from the Arduino environment. At another level it would be great if there was a higher level Arduino library created that leveraged these APIs and provided examples for both implementing and controlling a Matter device.

### Describe the solution you'd like

It would be great if there was a higher level Arduino library created that leveraged the IDF Matter APIs and provided examples for both implementing and controlling a Matter device.

### Describe alternatives you've considered

There are some other people outside of this project starting to experiment with this, but they are not fully open source nor sponsored by Espressif.

### Additional context

For the Espressif platform to dominate in the market for Matter devices it could be engaging the Arduino developers better.

### I have checked existing list of Feature requests and the Contribution Guide

- [X] I confirm I have checked existing list of Feature requests and Contribution Guide.

lbernstone:
Why does this need to be part of the core?  It can be handled in a 3rd party library.

VojtechBartoska:
Hello @Humancell,

thanks for opening this feature Request.

I'm adding this to next major milestone 3.0.0 and we will evaluate it.

jakubdybczak:
You can try library I have created few days ago https://github.com/jakubdybczak/esp32-arduino-matter

Humancell:
Hello,

> You can try library I have created few days ago https://github.com/jakubdybczak/esp32-arduino-matter

I did see this library, but from my understanding you have not made this 100% open source, which is a problem for us.  We can not use any closed source code in our projects.

If I am incorrect, and this really is 100% open source then we could take a look.

jakubdybczak:
@Humancell 
Why do you think that this one is not open-source?
[arduino-esp32](https://github.com/espressif/arduino-esp32) also contains precompiled libraries and scripts that do this are in [esp32-arduino-lib-builder](https://github.com/espressif/esp32-arduino-lib-builder) repo.
Same with [esp32-arduino-matter](https://github.com/jakubdybczak/esp32-arduino-matter), you can find scripts used to build it there [esp32-arduino-matter-builder](https://github.com/jakubdybczak/esp32-arduino-matter-builder).

SuGlider:
We will consider this request for next releases in 2023... please be patient.

SuGlider:
[...]

PR #10543, opened 2024-10-29
feat(Matter): New Matter Endpoint - Dimmable Light

SuGlider:
## Description of Change

Organizes the source code folder creating a folder called `src/MatterEndpoints/` 
Adds a new Matter Endpoint: Dimmable Light, with example.
Ajusts the log level of main Matter Attribute processing callback.
Adds license information to the code.

## Tests scenarios
ESP32C6 and ESP32S2 using the provided example.

## Related links
Related to #7432


CMakeLists.txt
```diff
@@ -169,7 +169,8 @@ set(ARDUINO_LIBRARY_OpenThread_SRCS
   libraries/OpenThread/src/OThreadCLI_Util.cpp)
 
 set(ARDUINO_LIBRARY_Matter_SRCS
-  libraries/Matter/src/MatterOnOffLight.cpp
+libraries/Matter/src/MatterEndpoints/MatterOnOffLight.cpp
+libraries/Matter/src/MatterEndpoints/MatterDimmableLight.cpp
```
lucasssvaz, line 173:
```suggestion
  libraries/Matter/src/MatterEndpoints/MatterOnOffLight.cpp
  libraries/Matter/src/MatterEndpoints/MatterDimmableLight.cpp
```

libraries/Matter/examples/MatterDimmableLight/MatterDimmableLight.ino
```diff
@@ -0,0 +1,157 @@
+// Matter Manager
```
lucasssvaz, line 1: Could you please add a brief description in the sketch of how the example works ?

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.cpp
```diff
@@ -0,0 +1,158 @@
+#include <sdkconfig.h>
```
lucasssvaz, line 1: Please add the Espressif's license header (IDK if the other Matter files might be missing it)

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.h
```diff
@@ -0,0 +1,48 @@
+#pragma once
```
lucasssvaz, line 1: Please add the Espressif's license header (IDK if the other Matter files might be missing it)

libraries/Matter/examples/MatterDimmableLight/MatterDimmableLight.ino
SuGlider: I'll prepare a whole documentation latter when I add a few more endpoints.

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.cpp
SuGlider: I'll do that latter for all the files.

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.h
SuGlider: I'm do that latter for all the files.

CMakeLists.txt
SuGlider: done.

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.cpp
SuGlider: done to all source code files.

libraries/Matter/src/MatterEndpoints/MatterDimmableLight.h
SuGlider: done to all source code files.

lucasssvaz (approved):

P-R-O-C-H-Y (approved):
Tested with HomeAssistant running on VM.

libraries/Matter/examples/MatterDimmableLight/MatterDimmableLight.ino
```diff
@@ -0,0 +1,171 @@
...
+
+// Matter Protocol Endpoint Brightness Change Callback
+bool setLightBrightness(uint8_t brightness) {
+  Serial.printf("User Callback :: New Light Brigthness = %.2f%% [Val = %d]\r\n", ((float) brightness * 100) / MatterDimmableLight::MAX_BRIGHTNESS, brightness);
+  setRGBLight(DimmableLight.getOnOff(), brightness);
+  // store last Brightness for when the Light is restarted / power goes off
+  lastStatePref.putUChar("lastBrightness", brightness);
+  // This callback must return the success state to Matter core
+  return true;
+}
```
me-no-dev, line 71: can't those 3 be one callback?

SuGlider: I was thinking about something related yesterday.
There are other 3 light modes to be implemented (temperature, color, all together).

I was considering to use C++ inheritance. Some light modes have the same cluster as the base light and upper lights.

Therefore, keeping the callbacks separated allows me to call parent's methods, reducing the code for the implementation of the children's classes.

Let me implement the other light classes. We can discuss it together looking into the final code.

SuGlider: @me-no-dev - I have changed the example code. It makes it simplier. Let me know.

SuGlider: @me-no-dev 
It is better to keep the callbacks separated for a simple reason:
Matter protocol allows the controller to change differnt attributes to create effects, like lighting ramp up and down.

Example: A light is ON and level = 50. The user pressed the APP button to turn it OFF. Matter Controller will:
1- Change Level to 1 (keeping it ON) for a ramp down
2- Change on-off to 0 (OFF).
3- Change the Level back to 50, but still keeping it OFF to allow the Light Device to keep/know last brightness (before turning it off).

Therefore, a simple "turn it off" creates  3 different events.
A single callback would be called 3 times but in order to know what has changed (and take action), it would need to compare each attribute with its previous value.

Making it with a callback to each attribute change allows the application to better understand possible individual actions.

SuGlider: Last change: used lambda funtions to allow the code to have a single callback function.

me-no-dev:
Please include all endpoint headers into `Matter.h` so that only that header needs to be included by the users. Otherwise LGTM

SuGlider:
> Please include all endpoint headers into `Matter.h` so that only that header needs to be included by the users. Otherwise LGTM

@me-no-dev - Done.

me-no-dev (approved):

PR #10553, opened 2024-11-01
feat(Example): Remove o call to a no existing method from Serial example

thalesdemond:
## Description of Change
On the exemple *Serial_ALL_CPU_Freqs* there were a call to a method that does not exit o the library,
** Serial.updateBaudRate(115200) **.

## Tests scenarios
The used hardware was a ESP32-S3-BOX
I was only trying to run the exemple code.


lucasssvaz:
@thalesdemond The function does indeed exist. You can find it here:

https://github.com/espressif/arduino-esp32/blob/release/v3.0.x/cores/esp32/HardwareSerial.cpp#L392

PR #10582, opened 2024-11-07
fix(net): Allow to compile without IPv6 enabled

me-no-dev:
cc: @Jason2866

### DIFF
--- a/cores/esp32/IPAddress.cpp
+++ b/cores/esp32/IPAddress.cpp
@@ -406,7 +406,7 @@ void IPAddress::to_ip_addr_t(ip_addr_t *addr) const {
     addr->u_addr.ip4.addr = _address.dword[IPADDRESS_V4_DWORD_INDEX];
   }
 #else
-    addr->addr = _address.dword[IPADDRESS_V4_DWORD_INDEX];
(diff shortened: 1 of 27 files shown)


Jason2866:
@me-no-dev Great! Thx. Will test ASAP

Jason2866:
Tested and build a framework without IPv6 support which worked to build an Arduino sketch with IPv4 only.

Jason2866:
When IPv6 is disabled in IDF there is something broken. The example sketch https://github.com/espressif/arduino-esp32/blob/master/libraries/WebServer/examples/HelloServer/HelloServer.ino 
does compile connects to the AP but the WebServer is not reachable. Ping gets replys from the esp IP address.

Jason2866:
@me-no-dev Webserver is now reachable with last commit. Build a small Tasmota Variant without IPv6 support. Looks like it is fully working.

SuGlider (approved):
All Good. Tested WiFi Server, Client and Scan.
No IPv6 saves about 28KB flash space.

P-R-O-C-H-Y (approved):
Tested WebServer + Client, both working.

PR #10593, opened 2024-11-11
refactor(uart): Refactor UART test to work with any number of UARTs

lucasssvaz:
## Description of Change

Refactors UART test to handle any number of UART peripherals.

Also add a diagram file for esp32 to set the simulation CPU frequency to 40 MHz. Sometimes Wokwi simulation can fail with the default value.

Note that the Wokwi UART test for ESP32 will fail until this PR is merged.

## Tests scenarios

Tested locally and CI.

### DIFF
--- a/tests/validation/uart/diagram.esp32.json
+++ b/tests/validation/uart/diagram.esp32.json
@@ -1,0 +1,24 @@
+{
+  "version": 1,
+  "author": "lucasssvaz",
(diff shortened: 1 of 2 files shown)


SuGlider (approved):
LGTM

P-R-O-C-H-Y (approved):

PR #10595, opened 2024-11-11
fix(wifi): fixes WiFi.isconnected() to return true when it is connected and it has an IP Addr

### LINKED ISSUE #10580: WiFi.localIP() reports 0.0.0.0 after succesfull connect - Arduino Core 3.0.7
### Board

esp32-s3

### Device Description

Just a plain old esp32-s3

### Hardware Configuration

Nothing of consequence

### Version

v3.0.7

### IDE Name

PlatformIO

### Operating System

Mint Linux

### Flash frequency

80

### PSRAM enabled

yes

### Upload speed

921600

### Description

`WiFi.localIP()` should report the IP address. Instead it displays `0.0.0.0` after a succesfull connection. After a while it does show the correct IP.

### Fix

A quick fix here was to change [`isConnected()` in `WiFiSTA.cpp`](https://github.com/espressif/arduino-esp32/blob/3.0.7/libraries/WiFi/src/WiFiSTA.cpp#L218) to the snippet below.

This change gives the expected behaviour.

```c++
/**
 * is STA interface connected?
 * @return true if STA is connected and has a IP
 */
bool WiFiSTAClass::isConnected() {
  return STA.connected() && STA.hasIP();
}
```

### Sketch

```cpp
#include <Arduino.h>
#include <WiFi.h>

const char *SSID = "xxx";
const char *PSK = "xxx";

void setup()
{
    Serial.begin(115200);
    while (!Serial)
        delay(10);

    Serial.println("connecting..");

    WiFi.begin(SSID, PSK);

    while (!WiFi.isConnected())
        delay(10);

}

void loop()
{
    delay(1000);
}
```

### Debug Message

```plain
connecting..
0.0.0.0
192.168.0.253
192.168.0.253
192.168.0.253
192.168.0.253
192.168.0.253
192.168.0.253
192.168.0.253
192.168.0.253
```
### Other Steps to Reproduce

I did not try any other hw.

SuGlider:
## Description of Change
After 3.0.x (including 3.1.x) the function `WiFiSTAClass::isConnected()` returns when it is connected to the WiFi AP but not necessarily when it has an IP address.
This is a breaking change because 2.0.x only returns true when both, it is connected and it has an IP.

This PR fixes it.

## Tests scenarios
``` cpp
#include <Arduino.h>
#include <WiFi.h>

const char *SSID = "xxx";
const char *PSK = "xxx";

void setup()
{
    Serial.begin(115200);
    while (!Serial)
        delay(10);

    Serial.println("connecting..");

    WiFi.begin(SSID, PSK);

    while (!WiFi.isConnected())
        delay(10);

}

void loop()
{
    delay(1000);
}
```

## Related links
Fix #10580

### DIFF
--- a/libraries/WiFi/src/WiFiSTA.cpp
+++ b/libraries/WiFi/src/WiFiSTA.cpp
@@ -215,7 +215,7 @@ bool WiFiSTAClass::bandwidth(wifi_bandwidth_t bandwidth) {
  * @return true if STA is connected to an AP
  */
 bool WiFiSTAClass::isConnected() {
-  return STA.connected();
+  return STA.connected() && STA.hasIP();
 }
 
 /**


lucasssvaz (approved):

libraries/WiFi/src/WiFiSTA.cpp
```diff
@@ -215,7 +215,7 @@ bool WiFiSTAClass::bandwidth(wifi_bandwidth_t bandwidth) {
  * @return true if STA is connected to an AP
  */
 bool WiFiSTAClass::isConnected() {
-  return STA.connected();
+  return STA.connected() && STA.hasIP();
```
me-no-dev, line 218: just `STA.hasIP()` is enough here. You can not have IP if you are not connected.

SuGlider: Even if it is comfigured as static IP?

me-no-dev: you are correct. `hasIP()` checks if either IP has been assigned or that it has static. Since even when static `GOT_IP` event is received, maybe worth changing there too. Not sure what is more correct for `gotIP()`, but we can discuss that later.

me-no-dev (approved):

beta-tester:
i have a question.
how is or was the logic for `WiFi.status()` ?

```C
if (WiFi.status() == WL_CONNECTED) {
}
```

does `WL_CONNECTED` mean connected and got an IP as well?

PR #10597, opened 2024-11-12
fix(build): Allow Arduino to compile in IDF without components

me-no-dev:
This adds guards for the rest of the extra components


Jason2866:
@me-no-dev  There are some `#include "sdkconfig.h"` missing. Without the check(s) will fail

me-no-dev:
@Jason2866 where?

PR #10609, opened 2024-11-15
Guard ethernet parts to make it possible to use without

Jason2866:
Currently it is not possible to switch off Ethernet support in IDF and compile Arduino as an component. This PR adds guards to make it possible to compile only with WiFi and no Ethernet.
Follow up from #10597

@me-no-dev

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -19,13 +19,10 @@
  */
 
 // Disable the automatic pin remapping of the API calls in this file
-
-#include "sdkconfig.h"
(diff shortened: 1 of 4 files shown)


libraries/Ethernet/src/ETH.cpp
```diff
@@ -19,6 +19,10 @@
  */
 
 // Disable the automatic pin remapping of the API calls in this file
+
+#include "sdkconfig.h"
+#if CONFIG_ETH_ENABLED
```
me-no-dev, line 24: this check can be after `#include "ETH.h" and then you do not need to include sdkconfig again (already included by ETH.h.

PR #10630, opened 2024-11-20
Guard crypto parts in "Update" to save resources 

Jason2866:
This PR guards the crypto parts in library Update.
By doing this flash space is saved, since most users don't use crypto updates. To enable the possibility of crypto updates again just adding #define UPDATE_CRYPT to the sketch to enable the functionality again.
(similar to https://github.com/espressif/arduino-esp32/pull/9893)

### DIFF
--- a/libraries/Update/src/Update.h
+++ b/libraries/Update/src/Update.h
@@ -63,6 +63,7 @@ public:
     */
   bool begin(size_t size = UPDATE_SIZE_UNKNOWN, int command = U_FLASH, int ledPin = -1, uint8_t ledOn = LOW, const char *label = NULL);
 
+#ifdef UPDATE_CRYPT
   /*
      Setup decryption configuration
      Crypt Key is 32bytes(256bits) block of data, use the same key as used to encrypt image file
@@ -71,6 +72,7 @@ public:
      Crypt Mode,    used to select if image files should be decrypted or not
     */
   bool setupCrypt(const uint8_t *cryptKey = 0, size_t cryptAddress = 0, uint8_t cryptConfig = 0xf, int cryptMode = U_AES_DECRYPT_AUTO);
+#endif /* UPDATE_CRYPT */
 
   /*
       Writes a buffer to the flash and increments the address
@@ -99,6 +101,7 @@ public:
     */
   bool end(bool evenIfRemaining = false);
 
+#ifdef UPDATE_CRYPT
   /*
       sets AES256 key(32 bytes) used for decrypting image file
     */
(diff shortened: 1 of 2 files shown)


libraries/Update/src/Update.h
```diff
@@ -63,6 +63,7 @@ class UpdateClass {
     */
   bool begin(size_t size = UPDATE_SIZE_UNKNOWN, int command = U_FLASH, int ledPin = -1, uint8_t ledOn = LOW, const char *label = NULL);
 
+#ifdef UPDATE_CRYPT
```
me-no-dev, line 66: where would this `UPDATE_CRYPT` come from? Something is wrong here

me-no-dev:
We can not accept this as-is.

1. Defining `UPDATE_CRYPT` in the sketch will not propagate the definition to `Updater.cpp`
2. Your changes should not turn things OFF by default

Jason2866:
1.) Why in #9893 it is done the same way?
2.) Logic changed to let crypto feature on, which is imho a bad default setting since nearly no one does crypto updates with Arduino. Waste of resources.

me-no-dev:
1. It's not done the same way. It's the opposite, if something is NOT defined, then code stays the same. In your case something MUST be defined for the feature to stay the same
2. There are people using the feature and we do get issue reports from time to time

me-no-dev:
If you make it so it turns the feature off if `UPDATER_NO_CRYPT` is defined, then it would be acceptable

me-no-dev (approved):

Jason2866:
The last commit change the behaviour to not change anything until the user sets the define.
Adding every possible feature and enabling by default is questionable. The feature is added not long ago. Core 3.0.x has more relevant breaking changes....

me-no-dev:
@Jason2866 can you please take care of the changes outlined by the precomit hook? Seems it can not automatically apply them: https://github.com/espressif/arduino-esp32/actions/runs/12007402991/job/33467892197#step:9:39

me-no-dev:
There are a few more to have a look at. basically all red/green things https://github.com/espressif/arduino-esp32/actions/runs/12008009265/job/33469780093?pr=10630#step:9:45

Jason2866:
@me-no-dev Changed the red stuff to the suggested green format

Jason2866:
@me-no-dev  idk what to change here?!
https://github.com/espressif/arduino-esp32/actions/runs/12008156960/job/33470223392

Jason2866:
trailing space.... check is green now

PR #10647, opened 2024-11-24
feat(matter): adds new Matter Endpoint: Color Light

SuGlider:
## Description of Change
This PR adds a new Matter Endpoint: MatterColorLight.
It has also an example.

## Tests scenarios
ESP32-S3 with the provided example.

## Related links
Related to #7432


me-no-dev:
@P-R-O-C-H-Y  will the color formats be usable within Zigbee also?

P-R-O-C-H-Y:
> @P-R-O-C-H-Y will the color formats be usable within Zigbee also?

@SuGlider @me-no-dev Yes, I am using the XYtoRGB and RGBtoXY also. So that part of code is duplicated :)
Maybe we can create some common lib or place for that?

me-no-dev:
I'd say to put it in the core. It can be used for our own color led too. Only suggestion is to maybe prefix the function names, so that they do not clash with libraries that possibly have them as well

SuGlider:
> I'd say to put it in the core. It can be used for our own color led too. Only suggestion is to maybe prefix the function names, so that they do not clash with libraries that possibly have them as well

We can add a namespace to it. `ArduinoColorSpace::`

me-no-dev:
No namespace please: https://stackoverflow.com/questions/46626494/does-c-have-namespaces-similar-to-c

PR #10666, opened 2024-11-29
Remove 3rd party references in code and documentation

SuGlider:
## Description of Change
This PR changes documentation and examples to leave only those platforms that are officially supported by Espressif.

Pending changes to CI and scripts (e.g. renaming files) will be done in a separate PR to minimize breaking changes with existing projects.

## Tests scenarios
CI Only

## Related links
None


SuGlider:
@Lucasczm - Please review the CI Tools changes.
@me-no-dev - I'll also add a PR to Lib Builder.

@pedrominatel | @rftafas | @P-R-O-C-H-Y 
These are all places that I found a referece to PIO in this repository.

Jason2866:
@SuGlider @me-no-dev  This PR removes support for Platformio / pioarduino
Why doing this?

mathieucarbou:
Is there an equivalent alternative or backward compatibility "path" somewhere, or the motivation from expressif is really to break the daily dev / build of everyone using pio / way of working?

SuGlider:
This PR is under review. Please fell free to comment with a change request.

We want remove Trade Marks, Brand Names and other reserved name in order to avoid issues.
It has been already done with NeoPixel (Adafruit Brand Name).

tools/platformio-build.py
Jason2866, line 1: Keep this file and rename it to `pioarduino-build.py` any appearance of Platformio can be replaced with `pioarduino` if needed

.github/scripts/on-release.sh
Jason2866, line 1: The deleted file is needed. Rename to `pioarduino-build.py`

SuGlider: We shall change the name of this file. What is he name of the Community Project?

SuGlider: I've found PioArduino (https://github.com/pioarduino/platform-espressif32) 
Is that the right repository, @Jason2866?

mathieucarbou: That's the right one. There is a Discord also if you need.

Jason2866: @SuGlider Well, if only trademarks and brands name appearing is the reason, please do the suggested changes without removing. Remove does kill the Open Source Project pioarduino.

rftafas: We are also removing third party software. Exception is Arduino IDE  because Arduino Core for ESP32 is somewhat a component for that IDE.

We may add a "third party tools" section.

Jason2866: What 3rd party software is there?

mathieucarbou: > We are also removing third party software. 

> Arduino Core for ESP32 is somewhat a component for that IDE.

Since when ? Arduino Core has always been the Arduino API on ESP32 chips, that's what has always been stated in the doc so far.

IDE compatibility has nothing to do with a framework and obviously a framework is not a component of an IDE: to build Arduino apps on ESP32 chips, one should not have to be forced to use Arduino IDE, which IMHO, is far from being the best available IDE out there for complex projects and it is about time Espressif recognise this fact. There are reasons why so many people are not using Arduino or ESP-IDF CLI. If it was so  good, people would not bother using alternatives.

Some of your doc

- https://docs.espressif.com/projects/arduino-esp32/en/latest/getting_started.html#about-arduino-esp32
- https://docs.espressif.com/projects/arduino-esp32/en/latest/esp-idf_component.html
- https://docs.espressif.com/projects/arduino-esp32/en/latest/installing.html#installing-using-platformio
- https://docs.espressif.com/projects/arduino-esp32/en/latest/guides/core_compatibility.html

So... Should the community understand that Espressif is actually changing the direction of the Arduino Core project completely so that it only becomes available from Arduino IDE ? Which, consequently would break all projects out there based on platformio or pioarduino and would have far more implications than just dropping support for PlatformIO ^

rftafas: Vojtech is no longer part of Espressif team. He decided to venture for other endeavors. I've been the Product Manager since 2022.

There are 3 products here:

1. **The Arduino Core for ESP32**, that is an 'add-on' to Arduino IDE, reason why even being a 3rd Party Software, Arduino IDE will be kept.
1. **The Arduino Component for ESP-IDF**, which most people use along normal/usual ESP-IDF tools.
1. **The source code.** People will use it however they please.

Third party software (i.e. not owned or 'managed' by Espressif) will be moved to a specific place in documentation, if/when using updated/supported Espressif software. Changes are on documentation for now, and we are only aiming the software (i.e. the source code) when it is related to trademarks (SuGlider examples). We are making it very clear about what is an Espressif Product and what is not.

Why?

Well... This very thread is an example in itself. The very statement that "Espressif will no longer support PlatformIO" is problematic in its core and it does break one PIO Labs trademark rules, making it appear that we own PlatformIO. We don't. PlatformIO is a commercial product, which is owned, supported and maintained by PIO Labs. We can't be the 'maintainers' or even the 'supporters' of it. It is NOT shared property. PIO Labs is the sole owner.

Simple warnings do not make a difference on defining this (they are there, yet, confusion remains). So, in order to be compliant to any third party projects and their trademarks, what is our product and so on, we are reviewing all content. This is the scope of this PR.

SuGlider:
@Jason2866 @mathieucarbou - We support community initiatives. 
No danger or bad intentions to your initiative.
As I said, we shall not use PIO or PlatformIO terms anymore.

Let's focus on that. Please let's work together on it.

.github/scripts/on-release.sh
rftafas: As for renaming to `pioarduino-build.py`, for example, AFAIK, PIOArduino isn't a trademark (yet). We might think this later, but it doesn't need to be now. Goal for now is documentation update.

Jason2866:
@rftafas Thx for the explanation and clarification. Good to hear to keep support for community pioarduino :-)

rftafas (approved):

TD-er:
I don't understand it... There were numerous changes requested by @Jason2866, like not to delete files.
There are no new commits since and still the PR is approved???

What am I missing here?

lucasssvaz:
@SuGlider @me-no-dev 
Updated description to make intentions more clear and reverted changes to scripts as discussed in our internal meeting.

lucasssvaz (approved):

SuGlider:
> I don't understand it... There were numerous changes requested by @Jason2866, like not to delete files. There are no new commits since and still the PR is approved???
> 
> What am I missing here?

No worries. @Jason2866 request is fine and executed.

Jason2866 (approved):

pedrominatel (approved):

P-R-O-C-H-Y (approved):

PR #10673, opened 2024-12-03
WiFiProv: Split provisioning into two parts for better synchronization.

kvp1703:
- Split the beginProvision into initProvision and beginProvision.
- The user node association (cloud_user_assoc) endpoint failed to register in the provisioning workflow because provisioning began before the endpoint creation.

### DIFF
--- a/libraries/WiFiProv/src/WiFiProv.cpp
+++ b/libraries/WiFiProv/src/WiFiProv.cpp
@@ -1,4 +1,4 @@
-/*
+  /*
     WiFiProv.cpp - WiFiProv class for provisioning
     All rights reserved.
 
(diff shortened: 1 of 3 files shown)


libraries/WiFiProv/src/WiFiProv.cpp
```diff
@@ -72,13 +72,14 @@ static void get_device_service_name(prov_scheme_t prov_scheme, char *service_nam
...
 }
 
-void WiFiProvClass ::beginProvision(
-  prov_scheme_t prov_scheme, scheme_handler_t scheme_handler, network_prov_security_t security, const char *pop, const char *service_name,
-  const char *service_key, uint8_t *uuid, bool reset_provisioned
-) {
-  bool provisioned = false;
-  static char service_name_temp[32];
+bool provInitDone = false;
+bool provisioned = false;
```
me-no-dev, line 76: please make those static

PR #10677, opened 2024-12-04
fix: Split provisioning into two parts for better synchronization

kvp1703:
## Description of Change
Split provisioning into two parts for better synchronization.
The user node association (cloud_user_assoc) endpoint failed to register in the provisioning workflow because provisioning began before the endpoint creation.

### DIFF
--- a/libraries/RainMaker/examples/RMakerSwitch/RMakerSwitch.ino
+++ b/libraries/RainMaker/examples/RMakerSwitch/RMakerSwitch.ino
@@ -94,6 +94,12 @@ void setup() {
 
   RMaker.enableSystemService(SYSTEM_SERV_FLAGS_ALL, 2, 2, 2);
 
+#if CONFIG_IDF_TARGET_ESP32S2
+  WiFiProv.initProvision(NETWORK_PROV_SCHEME_SOFTAP, NETWORK_PROV_SCHEME_HANDLER_NONE);
+#else
+  WiFiProv.initProvision(NETWORK_PROV_SCHEME_BLE, NETWORK_PROV_SCHEME_HANDLER_FREE_BTDM);
+#endif
+
   RMaker.start();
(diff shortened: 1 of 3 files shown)


libraries/RainMaker/examples/RMakerSwitch/RMakerSwitch.ino
```diff
@@ -94,6 +94,8 @@ void setup() {
 
   RMaker.enableSystemService(SYSTEM_SERV_FLAGS_ALL, 2, 2, 2);
 
+  WiFiProv.initProvision(NETWORK_PROV_SCHEME_BLE, NETWORK_PROV_SCHEME_HANDLER_FREE_BTDM);
```
P-R-O-C-H-Y, line 97: There should be a #if CONFIG_IDF_TARGET_ESP32S2 to use a SOFTAP scheme like is done in `WiFiProv.beginProvision`, instead of forcing the BLE.
```
#if CONFIG_IDF_TARGET_ESP32S2
  WiFiProv.beginProvision(NETWORK_PROV_SCHEME_SOFTAP, NETWORK_PROV_SCHEME_HANDLER_NONE, NETWORK_PROV_SECURITY_1, pop, service_name);
#else
  WiFiProv.beginProvision(NETWORK_PROV_SCHEME_BLE, NETWORK_PROV_SCHEME_HANDLER_FREE_BTDM, NETWORK_PROV_SECURITY_1, pop, service_name);
#endif
```

kvp1703: yes, pushed the chnages.

me-no-dev (approved):

SuGlider (approved):

P-R-O-C-H-Y (approved):

PR #10700, opened 2024-12-09
feat(chip): Add definition for BOOT_PIN for all chips

me-no-dev:
For use in sketches as default button


cores/esp32/esp32-hal.h
```diff
@@ -61,6 +61,20 @@ extern "C" {
...
+const uint8_t BOOT_PIN = 9;
+#define BOOT_PIN BOOT_PIN
+#elif CONFIG_IDF_TARGET_ESP32P4
+const uint8_t BOOT_PIN = 35;
+#define BOOT_PIN BOOT_PIN
+#elif CONFIG_IDF_TARGET_ESP32C5
+const uint8_t BOOT_PIN = 28;
+#define BOOT_PIN BOOT_PIN
+#endif
+
```
lucasssvaz, line 77: Maybe a little simpler like this ?

```suggestion
#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
const uint8_t BOOT_PIN = 0;
#elif CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32H2 || CONFIG_IDF_TARGET_ESP32C61
const uint8_t BOOT_PIN = 9;
#elif CONFIG_IDF_TARGET_ESP32P4
const uint8_t BOOT_PIN = 35;
#elif CONFIG_IDF_TARGET_ESP32C5
const uint8_t BOOT_PIN = 28;
#endif
#define BOOT_PIN BOOT_PIN
```

me-no-dev: no. because you want to be able to test in BOOT_PIN is defined

lucasssvaz: But isnt it always defined unless we are adding a new chip ?

me-no-dev: yup. that is the case. will change

SuGlider (approved):
LGTM!

lucasssvaz (approved):

PR #10702, opened 2024-12-10
feat(matter_examples): apply boot button change to all examples

SuGlider:
## Description of Change
Applies PR#10700 to all Matter examples.
Adds a way to decommission the Matter devices by pressing the Button for longer than 10 seconds to all examples, but `MatterCommisionTest.ino` that decommission itself as part of the sketch.

It also changes `MatterComposedLights.ino` to drive 3 different Light endpoints (onoff/dimmable/color).
It change the name of a couple example to keep the same naming style.

## Tests scenarios
All examples were tested using ESP32-S3 and ESP32-C6.

## Related links
Related to #10700


libraries/Matter/examples/MatterColorLight/MatterColorLight.ino
```diff
@@ -35,7 +35,12 @@ const uint8_t ledPin = 2;  // Set your pin here if your board has not defined LE
 #endif
 
 // set your board USER BUTTON pin here
-const uint8_t buttonPin = 0;  // Set your pin here. Using BOOT Button. C6/C3 use GPIO9.
+#ifdef BOOT_PIN
+const uint8_t buttonPin = BOOT_PIN;  // Set your pin here. Using BOOT Button.
```
me-no-dev, line 39: BOOT_PIN is always defined. Leave just this line and it will be enough for someone to change the pin to something else. Same for the other examples.

P-R-O-C-H-Y (approved):

lucasssvaz (approved):

libraries/Matter/examples/MatterColorLight/MatterColorLight.ino
SuGlider: done.

me-no-dev (approved):

PR #10712, opened 2024-12-11
feat(matter): creates a Matter Contact Sensor Endpoint

SuGlider:
## Description of Change
Adds the Matter Contact Sensor Endpoint with an example.

## Tests scenarios
Using ESP32-C6 and the provided example

## Related links
Related to #7432


libraries/Matter/src/MatterEndpoints/MatterContactSensor.cpp
```diff
@@ -0,0 +1,98 @@
...
+  started = false;
+}
+
+bool MatterContactSensor::setContact(bool _contactState) {
+  if (!started) {
+    log_e("Matter Contact Sensor device has not begun.");
+    return false;
+  }
+
+  // avoid processing the a "no-change"
```
lucasssvaz, line 72:
```suggestion
  // Skip if there is no change
```

SuGlider: Thanks.

lucasssvaz (approved):

me-no-dev (approved):

PR #10730, opened 2024-12-15
Use INPUT_PULLUP instead of INPUT

leochen4891:
## Description of Change
The digitalRead of an INPUT pin is LOW, which stops the timer right after start. 
Setting it to INPUT_PULLUP will fix it.

It aims to make repeat timer example work out of the box on some of the popular S3 boards.

## Tests scenarios
Tested on ESP32-S3 with Arduino-esp32 core v3.0.7 using Arduino 2.3.4

## Related links
https://www.amazon.com/dp/B0C9H7Y66W


P-R-O-C-H-Y (approved):

SuGlider (approved):
LGTM - Thank you @leochen4891 for the contribution!

PR #10750, opened 2024-12-18
feat(webserver): Middleware with default middleware for cors, authc, curl-like logging 

mathieucarbou:
For history, please see: https://github.com/espressif/arduino-esp32/pull/10186

-------

This PR improves WebServer with the following additions:

- Add Middleware support (aka Expressif) with some built-in middlewares like cors, authc and logging (curl-like)
- Add ability to collect all incoming request headers
- Added response code and info on server

### DIFF
--- a/libraries/WebServer/src/Parsing.cpp
+++ b/libraries/WebServer/src/Parsing.cpp
@@ -78,8 +78,14 @@ bool WebServer::_parseRequest(NetworkClient &client) {
   String req = client.readStringUntil('\r');
   client.readStringUntil('\n');
   //reset header value
-  for (int i = 0; i < _headerKeysCount; ++i) {
-    _currentHeaders[i].value = String();
+  if (_collectAllHeaders) {
+    // clear previous headers
+    collectAllHeaders();
+  } else {
+    // clear previous headers
+    for (RequestArgument *header = _currentHeaders; header; header = header->next) {
+      header->value = String();
+    }
   }
 
   // First line of HTTP request looks like "GET /path HTTP/1.1"
@@ -153,9 +159,6 @@ bool WebServer::_parseRequest(NetworkClient &client) {
       headerValue = req.substring(headerDiv + 1);
       headerValue.trim();
       _collectHeader(headerName.c_str(), headerValue.c_str());
-
-      log_v("headerName: %s", headerName.c_str());
-      log_v("headerValue: %s", headerValue.c_str());
 
       if (headerName.equalsIgnoreCase(FPSTR(Content_Type))) {
         using namespace mime;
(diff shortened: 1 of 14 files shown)


mathieucarbou:
Note: 

I kept using chained pointers, I don't know why STL lists are not used across the code, if it is because Arduino Core has to support AVR-like boards where C++ STL is not available ?

Because if it is the case, then why not creating a sort of `ArduinoList` (like it was done with `String`) that could be used everywhere, instead of repeating the CRUD logic to manage a list everywhere ? Maybe these structures exist already ?

me-no-dev:
It's been on my backlog to create something that is thread-safe, but so is a nice Task class... maybe we will find time in near future to do it. In order to have real impact and because the core is partially in C, it would be nice to have a C version that would be implemented in the CPP as well. Then we can go around and replace all lists with that

lucasssvaz (approved):

me-no-dev (approved):

SuGlider (approved):

libraries/WebServer/src/Parsing.cpp
```diff
@@ -272,12 +272,29 @@ bool WebServer::_parseRequest(NetworkClient &client) {
...
+    }
+    if (header->key.equalsIgnoreCase(headerName)) {
+      header->value = headerValue;
+      log_v("header collected: %s: %s", headerName, headerValue);
       return true;
     }
   }
+  assert(last);
+  if (_collectAllHeaders) {
+    last->next = new RequestArgument();
```
safocl, line 288: maybe replace default construction with member initialization?
`last->next = new RequestArgument{.key = headerName, .value = headerValue};`

safocl: https://godbolt.org/z/4cjahhdcE

mathieucarbou: we can use consteval in arduino code ? removing it to your sample and switching to c++11 or 17 leads to a completely different output. So is it worth doing it ?

safocl: > we can use consteval in arduino code ?

`consteval` introduce in **c++20** -- if this standard is available, `consteval` will be available.
but I gave that code as a demonstration that everything will be fine (even in constexpr context)

safocl: https://godbolt.org/z/bxb99Weoh for c++11(14)

safocl: I wanted to say in general that instead of several operations, you can do all of them at once with one.
```cpp
last->next = new RequestArgument();
last->next->key = headerName;
last->next->value = headerValue;
```
->
`last->next = new RequestArgument{ headerName, headerValue };`

mathieucarbou: yes, completely agree!

PR #10776, opened 2024-12-23
Restore Arduino Client API

pennam:
Should fix https://github.com/espressif/arduino-esp32/issues/2755#issuecomment-2547632462


JAndrassy:
ESP32/examples/Camera/CameraWebServer memory consumption increased extremely, but that example doesn't use classes modified in this PR

me-no-dev (approved):

PR #10788, opened 2024-12-28
Update the I2S Simple_tone.ino

greenyleaf:
Fixed problems in the I2S tone example.

-----------
## Description of Change
Fixed the problem so that the tone example could be played normally on ESP32 board.

## Tests scenarios

I have tested my Pull Request on Arduino-esp32 core v3.1.0 with ESP32 and ESP32-S3 Board with this scenario.


me-no-dev:
Closing because CLA has not been signed for over a month

PR #10805, opened 2025-01-04
NetworkEvents allow stack size to be changed.

thekurtovic:
## Description of Change
Replaced hardcoded stack size value for "arduino_events" task. Defaults to the original value. Application may override the value with a build flag i.e. `-D ARDUINO_NETWORK_EVENT_TASK_STACK_SIZE=8192`

## Tests scenarios
N/A

## Related links
N/A


vortigont:
I'm wondering if we'd better reuse default event loop's task for this also?

me-no-dev:
@vortigont not sure I understand what you mean?

vortigont:
Hi @me-no-dev,
what I wanted to highlight is that currently `NetworkEvents` implementation uses it's own RTOS queue and task to generate/propagate and run callbacks for Network events. While IDF has exactly same thing implemented and used for similar network/Bluetooth event handling via [Event Loop Library](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/esp_event.html). An [example](https://github.com/espressif/esp-idf/tree/master/examples/system/esp_event/default_event_loop) from IDF. Under the hood it has same set of objects - a queue and a handling task, the default instance is created [here](https://github.com/espressif/esp-idf/blob/b5ac4fbdf9e9fb320bb0a98ee4fbaa18f8566f37/components/esp_event/default_event_loop.c#L92). As soon as WiFi or BT code is used in the project this default event loop is started anyway.
So maybe reusing this queue could a feasible approach - at least we could save resources for same and reuse the IDF's event API.

I can reimplement `NetworkEvents` class over event loop's API if there is an interest in this.

me-no-dev (approved):

thekurtovic:
@vortigont I think this sounds like a cool idea and would be worth opening a PR

PR #10884, opened 2025-01-21
Arduino ESP32 v3.3 based on ESP-IDF v5.5 (master)

me-no-dev:


cores/esp32/Esp.cpp
```diff
@@ -63,6 +64,9 @@ extern "C" {
 #elif CONFIG_IDF_TARGET_ESP32P4
 #include "esp32p4/rom/spi_flash.h"
 #define ESP_FLASH_IMAGE_BASE 0x2000  // Esp32p4 is located at 0x2000
+#elif CONFIG_IDF_TARGET_ESP32C5
+#include "esp32c5/rom/spi_flash.h"
+#define ESP_FLASH_IMAGE_BASE 0x0000  // Esp32c5 is located at 0x0000
```
Jason2866, line 69: @me-no-dev c5 flash image base 0x0? Shouldn't it be 0x2000?

docs/conf_common.py
pedrominatel, line 1: LGTM.

pedrominatel (approved):

SuGlider (approved):

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 #   idf.py build
 
 set(min_supported_idf_version "5.3.0")
```
SuGlider, line 8: Should it be 5.5.0 because of the C5 addition?
```suggestion
set(min_supported_idf_version "5.5.0")
```

SuGlider: It may be better testing building some empty Arduno as IDF Component sketch using IDF 5.4|5.3

me-no-dev: sure, but the branch does support 5.3+, just like 3.2 supports 5.3 and 5.4. C5 addition is enabled only when available. Everything is guarded

me-no-dev: 5.3 is now breaking due to uart: https://github.com/espressif/arduino-esp32/actions/runs/16365909770/job/46242981911?pr=10884#step:5:1988

Please fix :)

PR #10885, opened 2025-01-21
fix version in package.json

Jason2866:
core version is 3.2.0


me-no-dev:
I'll close this, because there is another way to update the package versions (not just pioardiuno). Still this is not something you should use yet :)

Jason2866:
Yes, i do it automatically with Github in a project.

PR #10886, opened 2025-01-21
Update core version to 3.3.0

Jason2866:
use correct core version for this branch


me-no-dev:
same as the other PR. I will update versions once everything is OK. This is a very test branch that should not be used for anything outside of the core team

PR #10897, opened 2025-01-23
Changed setTimeout() to setConnectionTimeout() in WebServer

gonzabrusco:
## Description of Change
In WebServer setTimeout() was being used instead of the new method of NetworkClient setConnectionTimeout()


me-no-dev:
`setConnectionTimeout` is used to set the timeout for a client to connect to a server. In this case, this is a server client that is already connected, so `setConnectionTimeout` will do nothing. It is correct to have `setTimeout` as it is in the current code.

I'm closing this PR. If you are having some sort of issue that get's somehow fixed by this change, please open an issue and we can discuss it there.

Issue #10930, opened 2025-02-04
Lack of proper cleanup in ArduinoOTA class (missing end() in ~ArduinoOTA )


cziter15:
### Board

Not strictly related to hardware, so I'm skipping this section

### Description

I'm the author of ksIotFrameworkLib (iot library for arduino esp), working extensively with various ESP devices. Recently, I encountered a frustrating issue that led to crashes under specific conditions. The problem was challenging to pinpoint due to its random nature.

During the transition between the provisioning app and the regular operational app, I faced random exceptions:

`Guru Meditation Error: Core 0 panic'ed (LoadProhibited). Exception was unhandled.`

My code is written in modern C++, utilizing `unique_ptr` as the standard practice. After investigating, I discovered that the root cause of the crash was linked to the mDNS service (an IDF component). Since I use ArduinoOTA without any global instances enabled, this was another area I needed to explore.

And - bingo!

It turns out that the ArduinoOTAClass does not correctly clean up mDNS in its destructor. It should be calling the end() method, but it doesn't. This issue also impacts the ESP8266 framework.

https://github.com/espressif/arduino-esp32/blob/0302b4db4741ae51ff45a3b1b533175ee0c0d92c/libraries/ArduinoOTA/src/ArduinoOTA.cpp#L32

### Workaround
Explicitly invoke the `end()` method in the code that owns ArduinoOTA code before the object is destroyed.

me-no-dev:
Nice catch! I guess we never noticed it, because globals are enabled in Arduino by default. Fix is coming

cziter15:
It was really tricky (and my luck) to find the cause. Actually stacktrace was not pointing to the OTA at all.

Rather it was randomly crashing somewhere after device connected the network. The only thing I've noticed was the logline like "mdns service is already running" or something like this, but this unfortunatelly didn't get my attention earlier.

PR #10932, opened 2025-02-04
fix(ota): Make sure that ArduinoOTA.end() is called in the destructor

me-no-dev:
fixes: https://github.com/espressif/arduino-esp32/issues/10930


cziter15:
Is it safe to get `_udp_ota.stop();` called multiple times?

me-no-dev:
yes it is

Issue #10938, opened 2025-02-06
Crash when using String::move on empty string


TD-er:
### Board

ESP32-classic

### Device Description

Not tried on several boards, but got it reproducible on ESP32-classic.

### Hardware Configuration

-

### Version

latest master (checkout manually)

### IDE Name

PlatformIO

### Operating System

Windows 11 / WSL2

### Flash frequency

-

### PSRAM enabled

yes

### Upload speed

-

### Description

Assigning an empty string using `std::move` may cause a crash.

So also assigning an empty string as result from a function will also eventually end up in crashes.

The fix is extremely simple; the memmove should only be called when `rhs.len() > 0`

Maybe the check should even be extended to `if (rhs.len() && rhs.buffer())`

```c++
void String::move(String &rhs) {
  if (buffer()) {
    if (capacity() >= rhs.len()) {
      if (rhs.len()) {  // <<== This check added
        memmove(wbuffer(), rhs.buffer(), rhs.length() + 1);
      }
      setLen(rhs.len());
      rhs.invalidate();
      return;
    } else {
      if (!isSSO()) {
        free(wbuffer());
        setBuffer(nullptr);
      }
    }
  }
  if (rhs.isSSO()) {
    setSSO(true);
    memmove(sso.buff, rhs.sso.buff, sizeof(sso.buff));
  } else {
    setSSO(false);
    setBuffer(rhs.wbuffer());
  }
  setCapacity(rhs.capacity());
  setLen(rhs.len());
  rhs.setSSO(false);
  rhs.setCapacity(0);
  rhs.setBuffer(nullptr);
  rhs.setLen(0);
}
#endif
```

N.B. it is a bit confusing that `buffer()` can return both a nullptr as well as a pointer when it is an empty string.
This might cause other issues as well as the rest of the coding-style in the class suggests it is more like checking for a non-empty string.

N.B.2. This bug is likely present for quite a long time, so maybe also needs backporting?

### Sketch

```cpp
N.B. I did not exactly test this sketch below, but it is a very stripped down gist of the code which is causing crashes in my project.

++
String foo() {
  return String();
}

// Important to note that the strings have to be constructed first 
// and then assigned a new value.
String bar, working, crashing, crashing_too;

bar = foo();
working = bar;
crashing = std::move(bar);
crashing_too = foo();
```

### Debug Message

```plain
Part of crash log, pointing clearly to `String::move`


[...]

PR #10940, opened 2025-02-06
Fix crash when using String::move on empty string (#10938)

TD-er:
## Description of Change
`memmove` should not be called on a `nullptr` or size of `0`.
See: https://en.cppreference.com/w/cpp/string/byte/memmove

When an empty string was handed over to move, the application would crash.
The change is just adding an extra check to see if there is something to move.

## Related links
Fixes: #10938


me-no-dev (approved):

TD-er:
Included in single PR: #10945

PR #10942, opened 2025-02-06
Fix potential out of range in String::concat

TD-er:
## Description of Change

There is no prerequisite the given array has to be a 0-terminated char array. So we should only copy the length that has been given.

The `setLen()` function will make sure the internal string is 0-terminated. 
Thus there is no need to dangerously assume there will be 1 more byte to copy.

Example of code which may fail:
```c++
String str;
char buf[] = {'a', 'b'};
str.concat(buf, sizeof(buf));
```


TD-er:
Included in single PR: #10945

PR #10954, opened 2025-02-10
Fix(I2S example): make fix to the ESP32 I2S simple tone example

greenyleaf:
Fixed a problem in the I2S tone example.

-----------
## Description of Change
Did some fix so that the tone example could play normally on ESP32 boards with an I2S decoder attached.

## Tests scenarios

Tested the Pull Request on Arduino-esp32 core v3.1.0 with ESP32 and ESP32-S3 Boards with this scenario.


me-no-dev:
@greenyleaf we will test it and comment. No worries

me-no-dev (approved):

PR #10779, opened 2024-12-24
feat(board): Add custom board Huidu HD-WF2/HD-WF4 - esp32-s3 HUB75 driver boards

vortigont:
Huidu HD-WF2 - esp32-s3 HUB75 driver board
Chip: esp32-s3
Flash: 8MiB
RTC: BM8563
Vendor: https://www.hdwell.com/Product/index46.html

IOs:
 - 2xHUB75 ports
 - button
 - status LED
 - USB-A

Huidu HD-WF4 - esp32-s3 HUB75 driver board
Chip: esp32-s3
Flash: 8MiB
RTC: BM8563
Vendor: https://www.hdwell.com/Product/index46.html

IOs:
 - 4xHUB75 ports
 - button
 - status LED
 - USB-A
 - P/S IO ports

PR #10957, opened 2025-02-11
feat(board): Update 3rd party board Huidu HD-WF2/HD-WF4

vortigont:
Followup #10779

 - fixed flash mode to 'dio'
 - removed psram flags
 - corrected gpio mapping defines

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -48923,20 +48923,6 @@ huidu_hd_wf2.menu.CPUFreq.80.build.f_cpu=80000000L
 huidu_hd_wf2.menu.CPUFreq.40=40MHz
 huidu_hd_wf2.menu.CPUFreq.40.build.f_cpu=40000000L
 
-huidu_hd_wf2.menu.PSRAM.enabled=QSPI Flash fix
-huidu_hd_wf2.menu.PSRAM.enabled.build.defines=-DBOARD_HAS_PSRAM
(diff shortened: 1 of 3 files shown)


boards.txt
```diff
@@ -48860,8 +48860,8 @@ huidu_hd_wf2.build.partitions=default
 huidu_hd_wf2.build.defines=
 huidu_hd_wf2.build.loop_core=
 huidu_hd_wf2.build.event_core=
-huidu_hd_wf2.build.flash_type=qio
-huidu_hd_wf2.build.psram_type=qspi
+huidu_hd_wf2.build.flash_type=dio
+huidu_hd_wf2.build.psram_type=
```
me-no-dev, line 48864: Please do not edit those lines or board will never compile. flash and psram types are required and not meaning that there is psram or that the flash mode could not be changed to DIO. Same for the other board

vortigont: oh, that is unexpected... then I assume that I could just leave it as-as and it would be fine 
tnx!

me-no-dev: those two lines control which libs to be used and were added to accommodate OPI flash and psram on S3

me-no-dev:
looks like something went wrong with your rebase? there are commits that should not be here.

vortigont:
ugh ( I've rebased to a clean commit

P-R-O-C-H-Y (approved):

PR #11012, opened 2025-02-25
ci(zigbee): Check if Zigbee is enabled for CI tests

lucasssvaz:
## Description of Change

To match recent changes to the libs, check if Zigbee is enabled rather than if it is supported.

## Tests scenarios

CI

### DIFF
--- a/libraries/Zigbee/examples/Zigbee_Analog_Input_Output/ci.json
+++ b/libraries/Zigbee/examples/Zigbee_Analog_Input_Output/ci.json
@@ -1,6 +1,6 @@
 {
   "fqbn_append": "PartitionScheme=zigbee,ZigbeeMode=ed",
   "requires": [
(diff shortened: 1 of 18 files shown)


P-R-O-C-H-Y:
The ED mode (enddevice) is supported only on native Zigbee radio supported chips.

lucasssvaz:
> The ED mode (enddevice) is supported only on native Zigbee radio supported chips.

So I should keep `"CONFIG_SOC_IEEE802154_SUPPORTED=y"` for these ?

P-R-O-C-H-Y:
Let's see if the others will pass in the way the examples are configured. Because the Native Zigbee is used there instead the config for the UART RCP, but it may be ok :) If it passes  it would be nice

P-R-O-C-H-Y (approved):
LGTM

PR #11016, opened 2025-02-26
feat(gpio): allows mixing digital and analog read/write operations

SuGlider:
## Description of Change
This PR will improve the message when the sketch wants to set a pin with `digitalWrite()` that has a different function than GPIO. It also allow `digitalRead()` to read the io level for any IO function. A warning message will be sent to the console in case that the pin is not set as GPIO.

Example:

``` cpp
// half birghtness LED blink
uint8_t ledPin = 2;

void setup() {
  pinMode(ledPin, OUTPUT);  // sets the pin as output
}

void loop() {
  analogWrite(ledPin,  64);
  delay(500);
  pinMode(ledPin, OUTPUT);  // neecssary otherwise digitalWrite() will fail
  digitalWrite(ledPin, LOW);
  delay(500); 
}
```

``` cpp
uint8_t pin = 2;

void setup() {
  pinMode(pin, INPUT);  // sets the pin as input
  Serial.begin(115200);
}

void loop() {
  if (digitalRead(pin)) {
    Serial.print("Reading analog value: ");
    Serial.println(analogRead(pin));
  } else {
   Serial.println("No signal");
  }
  delay(1000);
}
``` 
## Tests scenarios
Using the sketches above.

## Related links
None.


cores/esp32/esp32-hal-gpio.c
```diff
@@ -184,12 +184,11 @@ extern int ARDUINO_ISR_ATTR __digitalRead(uint8_t pin) {
...
 #endif
 
-  if (perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) != NULL) {
-    return gpio_get_level((gpio_num_t)pin);
-  } else {
-    log_e("IO %i is not set as GPIO.", pin);
-    return 0;
+  // if the pin is not in GPIO mode, make it happen
+  if (perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) == NULL) {
+    __pinMode(uint8_t pin, INPUT);
```
SuGlider, line 189:
```suggestion
    __pinMode(pin, INPUT);
```

me-no-dev:
I am firmly against allowing `digitalWrite` to pins not in GPIO mode. For `digitalRead` we can just remove the periman check and do `gpio_get_level`. That would be safe, though results will not be ensured if pin is in LP or Analog mode.

SuGlider:
@me-no-dev - it is updated as requested. PTAL.

cores/esp32/esp32-hal-gpio.c
```diff
@@ -182,14 +182,12 @@ extern int ARDUINO_ISR_ATTR __digitalRead(uint8_t pin) {
   if (pin == RGB_BUILTIN) {
     return RGB_BUILTIN_storage;
   }
-#endif
-
+#endif  // RGB_BUILTIN
+  // This work when the pin is set as GPIO and in INPUT mode. For all other pin functions, it may return inconsistent response
   if (perimanGetPinBus(pin, ESP32_BUS_TYPE_GPIO) != NULL) {
```
me-no-dev, line 187: This should check if getPinBus is equal to NULL, else you are sending warning when the pin is actually set as GPIO

SuGlider: Correct.

me-no-dev (approved):

lucasssvaz (approved):

Issue #11024, opened 2025-03-01
UART speeds above 460800 on ESP32 no longer functional since v3.0.


ewpa:
### Board

ESP32 Dev Module

### Device Description

ESP32-WROOM-32E on simple board.

### Hardware Configuration

Using hardware serial 2 UART at 3.3V, pins 16 and 17.

### Version

v3.0.5

### IDE Name

arduino-cli

### Operating System

GNU/Linux

### Flash frequency

80MHz

### PSRAM enabled

no

### Upload speed

115200

### Description

Hardware serial 2 BAUD rates of 921600 and above were usable on ESP32 Arduino 2.0.17.  On 3.0 and 3.1 they are no longer usable, garbage is seen.

### Sketch

```cpp
HardwareSerial hs0(2);
hs0.begin(intendedBaud, SERIAL_8N1, RX2, TX2);
```

### Debug Message

```plain
Garbage.
```

### Other Steps to Reproduce

Appears related to espressif/esp-idf PR#8572 reported by b1ackviking.  I have created a PR with this fix and it resolves this regression.

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [x] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

ewpa:
Hi @SuGlider, based on your PR comments I traced the calls where the clocks were determined.  In my code, this was being executed just once, hence the wrong clock was selected for higher BAUD rates, e.g.
```
  Serial.begin(9600);
  // starts UART2 at 921,600 on RX2(18) and TX2(19)
  hs0.begin(921600, SERIAL_8N1, 18, 19);
```

This _is_ a change in behavior from 2.0, and my PR did inadvertently fix it, however the fix/workaround was simple adding `Serial.end` before subsequent faster `Serial.begin`s:
```
  Serial.begin(9600);
  // starts UART2 at 921,600 on RX2(18) and TX2(19)
  hs0.end();
  hs0.begin(921600, SERIAL_8N1, 18, 19);
```

So I am happy to withdraw PR#11025 and close this issue.  I am not certain on the intended behavior of multiple `begin`s in the Arduino platform.

SuGlider:
@ewpa - Using Arduino Core 3.1.3, all these use cases shall work correctly, for any ESP32 UART port (Serial, Serial1, Serial2) :
Examples using Serial2 (UART2):
``` cpp
[...]

PR #11025, opened 2025-03-01
fix(uart): Permit UART speeds over 460800 on ESP32 Arduino 3.0

ewpa:
## Description of Change
Regression fix: Permit UART speeds over 460800 on ESP32 Arduino 3.0 and later.  UART speeds above 460800 were functional on ESP32 Arduino 2.0.

## Tests scenarios
Tested on ESP32 Arduino 3.0.5.

## Related links
  - b1ackviking: "Console sets UART clock source to REF_TICK on ESP32 and ESP32S2 by default, however, APB clock frequency does not change when power management is disabled. Using APB clock source allows higher baud rates for UART console."
  - Merge espressif/esp-idf PR#8572: "fix(console): use apb clock source for uart when power management is disabled (IDFGH-6952)"
  - Closes Issue #11024.


SuGlider:
@ewpa - 
ESP32 Arduino has all SoC with `CONFIG_PM_ENABLE` disabled:
**sdkconfig**
```
#
# Power Management
#
# CONFIG_PM_ENABLE is not set
# CONFIG_PM_SLP_IRAM_OPT is not set
# end of Power Management
```

The suggested change (testing `#elif defined(CONFIG_PM_ENABLE) && SOC_UART_SUPPORT_REF_TICK`) would simply remove part of the code forever:

```
  // there is an issue when returning from light sleep with the C6 and H2: the uart baud rate is not restored
  // therefore, uart clock source will set to XTAL for all SoC that support it. This fix solves the C6|H2 issue.
#if SOC_UART_SUPPORT_XTAL_CLK
  uart_config.source_clk = UART_SCLK_XTAL;  // valid for C2, S3, C3, C6, H2 and P4
#elif defined(CONFIG_PM_ENABLE) && SOC_UART_SUPPORT_REF_TICK
  if (baudrate <= REF_TICK_BAUDRATE_LIMIT) {
    uart_config.source_clk = UART_SCLK_REF_TICK;  // valid for ESP32, S2 - MAX supported baud rate is 250 Kbps
  } else {
    uart_config.source_clk = UART_SCLK_APB;  // if power management is enabled baudrate may change with the APB Frequency!
  }
#else
  // Default CLK Source: CLK_APB for ESP32|S2|S3|C3 -- CLK_PLL_F40M for C2 -- CLK_PLL_F48M for H2 -- CLK_PLL_F80M for C6
  uart_config.source_clk = UART_SCLK_DEFAULT;  // baudrate may change with the APB Frequency!
#endif
```

If the baudrate is set to `460,800`, it will be over `REF_TICK_BAUDRATE_LIMIT` (250,000) and the UART Source Clock will be set as `UART_SCLK_APB`, which is the same as `UART_SCLK_DEFAULT`.

### Please explain how this PR would be effective for the issue you have presented.

ewpa:
Hi @SuGlider, thank you for taking a look.  What you are saying makes sense so I will need to review this PR further to see why it appears to make a positive impact in _my_ case (i.e. UART rate of 921600 and above).

SuGlider (commented):
This change is not recommended.

ewpa:
Agreed, I withdraw this PR.  See Issue #11024 for alternatives.

PR #11048, opened 2025-03-07
fix(eth): Fix RMII Ethernet not being able to be restarted

me-no-dev:
This pull request includes changes to the `ETHClass::begin` method in the `libraries/Ethernet/src/ETH.cpp` file. The main purpose of these changes is to improve the initialization process of the Ethernet MAC and PHY by storing the created instances in class member variables instead of local variables.

Initialization improvements:

* Changed the `mac` variable to a class member variable `_mac` and updated the error handling to use `_mac`.
* Changed the `phy` variable to a class member variable `_phy` and updated the error handling to use `_phy`.
* Updated the switch cases for different PHY types to assign the new PHY instances to `_phy` instead of `phy`.
* Modified the `eth_config` initialization to use the class member variables `_mac` and `_phy` instead of the local variables `mac` and `phy`.

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -271,8 +271,8 @@ bool ETHClass::begin(eth_phy_type_t type, int32_t phy_addr, int mdc, int mdio, i
   eth_mac_config.sw_reset_timeout_ms = 1000;
   eth_mac_config.rx_task_stack_size = _task_stack_size;
 
-  esp_eth_mac_t *mac = esp_eth_mac_new_esp32(&mac_config, &eth_mac_config);
-  if (mac == NULL) {
+  _mac = esp_eth_mac_new_esp32(&mac_config, &eth_mac_config);
+  if (_mac == NULL) {
     log_e("esp_eth_mac_new_esp32 failed");
     return false;
   }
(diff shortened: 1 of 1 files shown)


libraries/Ethernet/src/ETH.cpp
```diff
@@ -281,26 +281,25 @@ bool ETHClass::begin(eth_phy_type_t type, int32_t phy_addr, int mdc, int mdio, i
...
-    case ETH_PHY_RTL8201: phy = esp_eth_phy_new_rtl8201(&phy_config); break;
-    case ETH_PHY_DP83848: phy = esp_eth_phy_new_dp83848(&phy_config); break;
-    case ETH_PHY_KSZ8041: phy = esp_eth_phy_new_ksz80xx(&phy_config); break;
-    case ETH_PHY_KSZ8081: phy = esp_eth_phy_new_ksz80xx(&phy_config); break;
+    case ETH_PHY_LAN8720: _phy = esp_eth_phy_new_lan87xx(&phy_config); break;
+    case ETH_PHY_TLK110:  _phy = esp_eth_phy_new_ip101(&phy_config); break;
+    case ETH_PHY_RTL8201: _phy = esp_eth_phy_new_rtl8201(&phy_config); break;
+    case ETH_PHY_DP83848: _phy = esp_eth_phy_new_dp83848(&phy_config); break;
+    case ETH_PHY_KSZ8041: _phy = esp_eth_phy_new_ksz80xx(&phy_config); break;
+    case ETH_PHY_KSZ8081: _phy = esp_eth_phy_new_ksz80xx(&phy_config); break;
```
TD-er, line 293: These functions suggest there is memory allocated.
Shouldn't there be a `delete _phy;` and `delete _mac;` in the destructor and also in the `begin()` and `end()` functions?

TD-er: I did miss this piece of code where those are deleted:
https://github.com/espressif/arduino-esp32/blob/5a710bede70de4c9d34230d44d98040889a5ce78/libraries/Ethernet/src/ETH.cpp#L914-L922

So it will be deleted when calling `end()` or destruct the object.

P-R-O-C-H-Y (approved):

SuGlider (approved):
LGTM

Issue #10923, opened 2025-01-31
Support setting the Link Speed for Ethernet


Ing-Dom:
### Related area

(Wired) Ethernet driver

### Hardware specification

especially for LAN8720, but relevant for all PHYs with the capability of setting the LINK speed by software

### Is your feature request related to a problem?

While a lot of drivers for the LAN8720 PHY support setting the link speed manually, it is not possible within arduino-esp32.
It IS indeed possible when you are using plain IDF, but the Arduino Ethernet driver (ETH.h) hides that from the user.
One could argue why this is neccessary - in 10Base-T Ethernet, which is often sufficient for embedded devices like the ESP32, the PHY needs much less power (50mA @3,3V). While you can set you PHY to 10MBIT in general by using the boot strap pins on the phy, this can cause serious issues with newer switches that do not support 10Base-T anymore.
So the option should be in software.

### Describe the solution you'd like

I would like to have an interface in the ethernet driver to be usable like

```
ETH.setLinkSpeed(eth_speed_t ::ETH_SPEED_MAX);
ETH.setLinkSpeed(eth_speed_t ::ETH_SPEED_10M);
ETH.setLinkSpeed(eth_speed_t ::ETH_SPEED_100M);
```

and also a `ETH.getLinkSpeed();`

### Describe alternatives you've considered

I tried to bypass the Ethernetdriver like this:

 eth_speed_t speed = ETH_SPEED_10M;
    esp_eth_ioctl(ETH.handle(), ETH_CMD_S_SPEED, &speed);

but his does not work.
It is not possible when you do it before ETH.begin(), and als not afterwards.
It has to be done somewhere in the middle of ETH.begin()..

### I have checked existing list of Feature requests and the Contribution Guide

- [x] I confirm I have checked existing list of Feature requests and Contribution Guide.

Avenitos:
I will support the addition of this functionality, I also encountered the fact that the standard autonegotiation mechanism does not work, it was necessary to disable Full Duplex so that the network would work normally, I added the code to the library manually and it worked, but I want a normal way of management.

I use rtl8201 by the way.

Need Add:
1. Auto-negotiation on/off
2. Speed setting
3. Duplex Mode Settings

```      // Set auto-negotiation status
      bool auto_nego_en = false; // Disable auto-negotiation      
      esp_eth_ioctl(ETH.handle(), ETH_CMD_S_AUTONEGO, &auto_nego_en);

      // Set speed
[...]

PR #11053, opened 2025-03-07
feat(eth): Add setters for negotiation, speed and duplex modes

me-no-dev:
This pull request introduces several new methods and modifications to the `ETHClass` in the Ethernet library to enhance its functionality. The primary changes include adding methods to set duplex mode, auto-negotiation, and link speed, as well as updating the return type of the `linkSpeed` method.

### Enhancements to `ETHClass`:

* Added `setFullDuplex` method to allow setting the duplex mode of the Ethernet connection.
* Added `setAutoNegotiation` method to enable or disable auto-negotiation on the Ethernet connection.
* Added `setLinkSpeed` method to set the link speed of the Ethernet connection.
* Updated `linkSpeed` method to return `uint16_t` instead of `uint8_t` for better compatibility with link speed values.

### Header file updates:

* Updated `ETH.h` to declare the new methods and modify the return type of `linkSpeed`.

fixes: https://github.com/espressif/arduino-esp32/issues/10923

### DIFF
--- a/libraries/Ethernet/src/ETH.cpp
+++ b/libraries/Ethernet/src/ETH.cpp
@@ -1014,7 +1014,7 @@ bool ETHClass::setFullDuplex(bool on) {
   if (_eth_handle == NULL) {
     return false;
   }
-  eth_duplex_t link_duplex = on?ETH_DUPLEX_FULL:ETH_DUPLEX_HALF;
+  eth_duplex_t link_duplex = on ? ETH_DUPLEX_FULL : ETH_DUPLEX_HALF;
   esp_err_t err = esp_eth_ioctl(_eth_handle, ETH_CMD_S_DUPLEX_MODE, &link_duplex);
   if (err != ESP_OK) {
     log_e("Failed to set duplex mode: 0x%x: %s", err, esp_err_to_name(err));
@@ -1064,7 +1064,7 @@ bool ETHClass::setLinkSpeed(uint16_t speed) {
   if (_eth_handle == NULL) {
     return false;
   }
-  eth_speed_t link_speed = (speed == 10)?ETH_SPEED_10M:ETH_SPEED_100M;
+  eth_speed_t link_speed = (speed == 10) ? ETH_SPEED_10M : ETH_SPEED_100M;
   esp_err_t err = esp_eth_ioctl(_eth_handle, ETH_CMD_S_SPEED, &link_speed);
   if (err != ESP_OK) {
     log_e("Failed to set link speed: 0x%x: %s", err, esp_err_to_name(err));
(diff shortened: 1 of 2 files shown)


libraries/Ethernet/src/ETH.cpp
```diff
@@ -1037,6 +1060,18 @@ uint8_t ETHClass::linkSpeed() const {
...
 
+bool ETHClass::setLinkSpeed(uint16_t speed) {
+  if (_eth_handle == NULL) {
+    return false;
+  }
+  eth_speed_t link_speed = (speed == 10)?ETH_SPEED_10M:ETH_SPEED_100M;
+  esp_err_t err = esp_eth_ioctl(_eth_handle, ETH_CMD_S_SPEED, &link_speed);
+  if (err != ESP_OK) {
+    log_e("Failed to set link speed: 0x%x: %s", err, esp_err_to_name(err));
+  }
```
P-R-O-C-H-Y, line 1071: Adding a log_v or log_i might be good, if someone select other speed then 10, to see what link speed is used (100).

P-R-O-C-H-Y (approved):
Overall LGTM

libraries/Ethernet/src/ETH.cpp
me-no-dev: I left it as-is because I do not expect this to ever be used and also because I expect 1Gbps support to come soon

P-R-O-C-H-Y: ok, so later on maybe a check if its 10,100 or 1000 might be good.

WebDust21:
Note that as pointed out in the initial post in #10923, these added functions are effectively unusable.  The ioctl call stipulations require the driver to be created, yet not started--which is not a possible state from Arduino-ESP32 Eth, as "eth.begin" both creates and starts the driver before returning.

ETH_CMD_S_SPEED -> https://github.com/espressif/esp-idf/blob/1f46216a7214e27c1e351b4310acf7fcd77029d4/components/esp_eth/src/esp_eth.c#L474
ETH_CMD_S_AUTONEGO -> https://github.com/espressif/esp-idf/blob/1f46216a7214e27c1e351b4310acf7fcd77029d4/components/esp_eth/src/esp_eth.c#L460
ETH_CMD_S_DUPLEX_MODE -> https://github.com/espressif/esp-idf/blob/1f46216a7214e27c1e351b4310acf7fcd77029d4/components/esp_eth/src/esp_eth.c#L495

About the only way these can effectively be implemented with the current ecosystem, is if they are added as arguments to "eth.begin()".  That way, if the user wanted to change the rate, they could stop it with "eth.end()", then restart with different args to "eth.begin()".

me-no-dev:
@WebDust21 will be possible to do with the same API. I'll rework how things are done. Basically you will call the methods before `begin()` and in `begin()` they will get executed with the selected settings

PR #11070, opened 2025-03-10
fix(matter): examples must set pin to Digital Mode after analogWrite() and before digitalWrite()

SuGlider:
## Description of Change
When no RGB LED is used, a regular LED is used, in the examples, to demonstrate Light Brightness or the Fan Motor Power. 

The regular LED is driven with `analogWrite()` and set OFF with `digitalWrite(pin, LOW)`. In order to set it OFF, it is necessary to set the pin to Digital Mode OUTPUT, otherwise an error will be displayed in the UART console. 

## Tests scenarios
Using ESP32.

None

### DIFF
--- a/libraries/Matter/examples/MatterFan/MatterFan.ino
+++ b/libraries/Matter/examples/MatterFan/MatterFan.ino
@@ -49,11 +49,11 @@ void fanDCMotorDrive(bool fanState, uint8_t speedPercent) {
   // drive the Fan DC motor
   if (fanState == false) {
     // turn off the Fan
- #ifndef RGB_BUILTIN
+#ifndef RGB_BUILTIN
     // after analogWrite(), it is necessary to set the GPIO to digital mode first
     pinMode(dcMotorPin, OUTPUT);
 #endif
-   digitalWrite(dcMotorPin, LOW);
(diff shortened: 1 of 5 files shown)


P-R-O-C-H-Y (approved):

SuGlider:
Wow, @P-R-O-C-H-Y, --- That was really FAST!!!! 🥇

libraries/Matter/examples/MatterFan/MatterFan.ino
```diff
@@ -49,7 +49,11 @@ void fanDCMotorDrive(bool fanState, uint8_t speedPercent) {
   // drive the Fan DC motor
   if (fanState == false) {
     // turn off the Fan
-    digitalWrite(dcMotorPin, LOW);
+ #ifndef RGB_BUILTIN
+    // after analogWrite(), it is necessary to set the GPIO to digital mode first
+    pinMode(ledPin, OUTPUT);
```
me-no-dev, line 54: I think you meant `dcMotorPin` instead of `ledPin`

SuGlider: Thanks!

me-no-dev (approved):

PR #11075, opened 2025-03-11
[WiFiScan] Fix clear 'scanning bit' and clear scanCount

TD-er:
`WIFI_SCANNING_BIT` may not always be cleared on timeout. Also `_scanCount` will only be cleared if there is a `_scanResult`.

See: https://github.com/espressif/arduino-esp32/issues/8952

N.B. this is a summary of discussion with @me-no-dev 
Still a draft


me-no-dev:
Closing in favor of https://github.com/espressif/arduino-esp32/pull/11188

PR #11077, opened 2025-03-12
fix(uart): Fixes UART CI script to work with Arduino Core 3.2.x

SuGlider:
## Description of Change
Fixes the UART CI script to work correctly with Arduino Core 3.1.2+ / 3.2.x

## Tests scenarios
ESP32 | ESP32-S2 | ESP32-S3 | ESP32-C3 | ESP32-C6 | ESP32-P4 using the CI script

## Related links
None

### DIFF
--- a/tests/validation/uart/uart.ino
+++ b/tests/validation/uart/uart.ino
@@ -53,7 +53,7 @@ public:
     : uart_num(num), serial(serial_ref), peeked_char(-1), default_rx_pin(rx_pin), default_tx_pin(tx_pin), recv_msg("") {}
 
   void begin(unsigned long baudrate) {
-    // pinMode will force enabing the internal pullup resistor (IDF 5.3.2 Change)
+    // pinMode will force enabling the internal pullup resistor (IDF 5.3.2 Change)
     pinMode(default_rx_pin, INPUT_PULLUP);
     serial.begin(baudrate, SERIAL_8N1, default_rx_pin, default_tx_pin);
     while (!serial) {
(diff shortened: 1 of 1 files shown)


me-no-dev (approved):

lucasssvaz:
Is ESP32 test also supposed to pass ?

SuGlider:
> Is ESP32 test also supposed to pass ?

Yes.  It works with the real device. Not sure why would it fail in Wokwi.

SuGlider:
@lucasssvaz -  I have found out that Wokwi runs Arduino Core 3.1.1. The target of this patch is for Core 3.1.2 onwards. That may explain the issue with ESP32 and Wokwi.

lucasssvaz:
> @lucasssvaz - I have found out that Wokwi runs Arduino Core 3.1.1. The target of this patch is for Core 3.1.2 onwards. That may explain the issue with ESP32 and Wokwi.
> 
> Note: using Arduino Core 3.1.1 + ESP32, it won't fail in the Real Device... only Wokwi fails.

Wokwi runs the binaries we compile, in this case 3.2.0. Wokwi online editor has fixed versions like this but its not our case.

Anyway, if it works in the actual hardware. That's what matters. I will report to Uri.

lucasssvaz:
@SuGlider Can you test wokwi locally with the changes I pushed to see if it works ?

SuGlider:
> @SuGlider Can you test wokwi locally with the changes I pushed to see if it works ?

No, it also fails locally.

```
/tmp/hexi-esp32:esp32:esp32/sketch/sketch.ino:453:periman_test:FAIL: Expected '' Was '\xFF'
Change pins test successful
/tmp/hexi-esp32:esp32:esp32/sketch/sketch.ino:557:change_pins_test:PASS
End when stopped test successful
/tmp/hexi-esp32:esp32:esp32/sketch/sketch.ino:558:end_when_stopped_test:PASS

-----------------------
11 Tests 1 Failures 0 Ignored 
```

P-R-O-C-H-Y (dismissed):

lucasssvaz (approved):

PR #11115, opened 2025-03-13
feat(NetworkEvent lib) refactor network events handling

vortigont:
the main idea here is to reuse ESP IDF's event bus to propagate notifications for Arduino network events too.
Currently `NetworkEvents` lib uses it's own instances of queue and Task to deliver event among componects, but ESP IDF already has a very similar implementation and what is currently happening is we resend messages from one queue to another.
Reusing IDF's event loop for Arduino networking events could give some benefits:
     - a unified event exchange bus between arduino and IDF components
     - save resources for extra Task stack and Queue to handle Arduino's own event bus

I've come further and refactored other code related to NetworkEvents to close some gaps and unify messaging.
Since code changes are quite large, would appreciate any feedbacks if this worth the efforts to continue.

@me-no-dev @P-R-O-C-H-Y


me-no-dev:
> - a unified event exchange bus between arduino and IDF components

Arduino events are in reality copies of already existing IDF events that span multiple components and that are just collected in on place, plus some internal handling to update states and such, because most do not listen to events and expect to get the states otherwise

vortigont:
yeah, users are tend to misbehave :) you point is reasonable. Although nothing prevents user to attach to the same default event bus in his code.
But if the question is to just have a separate event handling task IDF's API allows to spawn user queues with dedicated task but using same event bus API. Does that makes sense? 

As for mutexes, I must have skipped it, sorry, will catch up.
But if I'm not mistaken in esp's toolchain cpp std mutexes are just wraps around same rtos semaphores under the hood, need to double check.

Jason2866:
There is always a bell ringing when i see the chance to save some RAM :-)
Do you have an rough idea how much RAM would be "saved"?
There are cases where 1KB more makes the difference between working or not...

vortigont:
@Jason2866 nah... I was not for the RAM saving here at all, it's just the thing everybody loves in MCU world, we do have that `Memory usage test` here for a reason I guess :) But straight answer is - it would save 8k for task's stack if reusing default event bus, which is not mandatory, we could have our own dedicated event task if that makes more sense.
My intention was to try to unify event messaging with it's origin - IDF's API, actually it is the source of network messages. Currently arduino's network event lib has 3 different types of callbacks - static by id, static by struct reference and functional by id + struct by value and none of those is matching the origin
```c++
using NetworkEventCb = void (*)(arduino_event_id_t event);
using NetworkEventFuncCb = std::function<void(arduino_event_id_t event, arduino_event_info_t info)>;
using NetworkEventSysCb = void (*)(arduino_event_t *event); 
```
and kind of not-so-nice scheme to register/unregister by function pointers while IDF is using instance handlers and  keep by-ptr as legacy. OK, ok, `NetworkEvent` lib returns handlers too but under the hood most of the components are still not using it as I checked. This all definitely has some historic reasons and compatibility obligations. I was trying to move it all into some distinctive direction, if esp32 arduino's platform tends to reuse IDF's drivers and other components where possible. At least it makes a bit clear where to more forward if at all.

Well, anyway it was a nice exercise for me to understand the relation and exchange between Arduino's network events and IDF's origin.

PR #11122, opened 2025-03-13
fix(UART): sets the correct uart clock source when using begin(baudrate)

### LINKED ISSUE #11121: Don't use begin() method to just change baudrate
> Ok, in that case, we need a warning for the `begin()` method to say that if you just want to change the Baudrate you must use `updateBaudRate() ` method even if `uartBegin()` seems to allow this change. 

 _Originally posted by @GitHubLionel in [#10314](https://github.com/espressif/arduino-esp32/issues/10314#issuecomment-2721850568)_

SuGlider:
## Description of Change
Arduino HardwareSerial was changed to use XTAL source clock whenever possible. This assures that the UART port doesn't chage its baudrate when APB frequency changes. The exception is related to ESP32 and ESP32-S2 that have no such clock source option.

In that case, ESP32|ESP32-S2 will use REF_TICK as clock source when the baudrate is lower than 250,000.
For a higher baudrate, it will use APB as clock source, which of course, may change the baudrate if its frequency is changed.

This PR fixes an issue using ESP32 | ESP32-S2 when it calls first `begin(baud_rate_lower_than_250K)` followed by `begin(baud_rate_higher_than_250K)`. In the issue, the UART clock source isn't  changed and the baud rate is not set correctly.

## Tests scenarios
Using ESP32 | ESP32-S2

``` cpp
Serial.begin(9600);
... configuration
Serial.begin(256000);
```

## Related links
Fixes #11121


lucasssvaz (approved):
LGTM

me-no-dev (approved):

PR #11153, opened 2025-03-20
feat(zigbee): Add check, boolean returns, fix Analog, add optional reset on factoryReset

### LINKED ISSUE #11178: Make Reboot after Factory Reset on an EP optional
### Related area

Zigbee 

### Hardware specification

C6, H2

When an EP executes a factory reset, it also executes a device reboot. When I execute a factory reset, I  may want to shut the device down, not reboot. Currently, I need to set a shutdown flag in flash, and upon reboot, if the flag is set, clear it and shut the device down. That works, but it's a workaround. 

Add a boolean to factoryReset(bool reboot) 
If true, reboot.
If false, return to the caller 

Flash-based flag to signal shutdown after the reboot from a factoryReset() call.

### Additional context

_No response_

P-R-O-C-H-Y:
## Description of Change

- Fix memory leak
- Updated almost all methods to return a boolean value and print error message with code and string of the error.
- `factoryReset` method now have an optional parameter `bool restart`, by default its set to TRUE, but if false is passed, the Zigbee will just erase the nvram and the reset or reboot will need to be managed in user sketch. Example of how it can be used is the Temperature Humidity Sleepy example.
- Simplified the `AnalogSensor` Endpoint -> removed AnalogValue, as its not commonly used in any device
- Fixed `AnalogSensor` compilation warning message

## Related links

Closes #11178


me-no-dev:
Please format errors like this:
```cpp
  esp_err_t err = esp_wifi_set_max_tx_power(power);
  if (err != ESP_OK) {
    log_e("Failed to set TX Power: 0x%x: %s", err, esp_err_to_name(err));
  }
```

SuGlider:
@P-R-O-C-H-Y  - please review the code in ZigbeeEP.cpp :: `void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute)`. It has 2 `malloc()` with no correspondent `free()`. This will certainly cause memory leak.

Better think and review the whole zigbee code when using `new` or `malloc`.

SuGlider (changes requested):
From a fast code read, I understand that ZigbeeEP.cpp can cause memory leaks.
It may happen in other code files as well.

I recommend reviewing all places where `new` or `malloc` is used.

P-R-O-C-H-Y:
Thanks @SuGlider, will take a look.

lucasssvaz (approved):
LGTM

P-R-O-C-H-Y:
All done now.

me-no-dev (approved):

PR #11157, opened 2025-03-21
Remove superfluous warnings for board Arduino nano

Jason2866:
Align with other user settings choosen from user. There is no warning for the selections made  anywhere else.

To do remapping or not is a setting as every other selection done in config.
There is no reason to generate warnings.

### DIFF
--- a/variants/arduino_nano_nora/io_pin_remap.cpp
+++ b/variants/arduino_nano_nora/io_pin_remap.cpp
@@ -6,14 +6,6 @@
 #error This build system is not supported. Please rebuild without BOARD_HAS_PIN_REMAP.
 #endif
 
-#if !defined(BOARD_HAS_PIN_REMAP)
-// This board uses pin mapping but the build system has disabled it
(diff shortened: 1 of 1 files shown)


me-no-dev:
@pillo79 PTAL!

pillo79:
Thanks for the ping! I am OK with deleting the second part, it is an user choice after all, so why complain :slightly_smiling_face: But please **keep the first part**:
```C
#if !defined(BOARD_HAS_PIN_REMAP)
// This board uses pin mapping but the build system has disabled it
#warning The build system forces the Arduino API to use GPIO numbers on a board that has custom pin mapping.
```

We had lots of users that complained about using third party build systems that did not properly export the macros related to pin mapping, and that helped expose those issues. It definitely does never appear unless there is a problem with the build system (and if so, it needs to be clear!).

Jason2866:
Closing since there will still be unnecessary warnings. So it can be left as it is.

PR #11159, opened 2025-03-21
feat(logging): Arduino log redirection

mathieucarbou:
`ets_sys.h` supports log redirection when `ets_printf()` is used and some functions are set with `ets_install_putc1()` or `ets_install_putc1()`.

```c++

/**
  * @brief  Printf the strings to uart or other devices, similar with printf, simple than printf.
  *         Can not print float point data format, or longlong data format.
  *         So we maybe only use this in ROM.
  *
  * @param  const char *fmt : See printf.
  *
  * @param  ... : See printf.
  *
  * @return int : the length printed to the output device.
  */
int ets_printf(const char *fmt, ...);

/**
  * @brief  Output a char to uart, which uart to output(which is in uart module in ROM) is not in scope of the function.
  *         Can not print float point data format, or longlong data format
  *
  * @param  char c : char to output.
  *
  * @return None
  */
void ets_write_char_uart(char c);

/**
  * @brief  Ets_printf have two output functions： putc1 and putc2, both of which will be called if need ouput.
  *         To install putc1, which is defaulted installed as ets_write_char_uart in none silent boot mode, as NULL in silent mode.
  *
  * @param  void (*)(char) p: Output function to install.
  *
  * @return None
  */
void ets_install_putc1(void (*p)(char c));

/**
  * @brief  Ets_printf have two output functions： putc1 and putc2, both of which will be called if need ouput.
  *         To install putc2, which is defaulted installed as NULL.
  *
  * @param  void (*)(char) p: Output function to install.
  *
  * @return None
  */
void ets_install_putc2(void (*p)(char c));
```

I need to add a custom `ets_install_putc1()` handler to redirect arduino logs, but sadly the implementation in `esp32-hal-uart.c` is bypassing the `ets_printf()` internal mechanism and is directly calling `ets_write_char_uart()` instead . 

https://github.com/espressif/arduino-esp32/blob/6c04a9315309b530120de98321f8d900c151a9b9/cores/esp32/esp32-hal-uart.c#L1123-L1132

Why ? I really was not able to figure out...

But this piece of code prevents everybody from redirecting Arduino logs, except with the boards in the macro conditions listed above...

Since I do not know the logic why `ets_printf()` is not called in all cases, I am proposing to add a macro: `ARDUINO_LOG_FORCE_ETS_PRINTF=1` which, when set, will make sure all the logs go through `ets_printf()`.

The user setting this macro will then be responsible to:

1. Either call `ets_install_putc1(ets_write_char_uart)`
2. Or call `ets_install_putc1(my_custom_character_log_callback)` with his custom function to redirect the logs.

### Use case example:

```c++
static StreamString* _arduinoLogBuffer = nullptr;
static Mycila::Logger* _arduinoLogDestination = nullptr;

static void log_char(char c) {
  if (!_arduinoLogDestination)
    return;
  _arduinoLogBuffer->write(c);
  if (c == '\n') {
    for (auto& output : _arduinoLogDestination->_outputs) {
      output->print("SYSTEM: "); // TODO: REMOVE THIS - PROVES ARDUINO LOGS WERE REDIRECTD
      output->print(*_arduinoLogBuffer);
    }
    _arduinoLogBuffer->clear();
  }
}

static void redirectArduinoLogs(Logger& logger) {
  _arduinoLogBuffer = new StreamString();
  _arduinoLogBuffer->reserve(MYCILA_LOGGER_BUFFER_SIZE);
  _arduinoLogDestination = &logger;

  // will override default arduino installed functions which sends logs to UART
  ets_install_putc1(log_char);
}
```

With the macro def and the code above Arduino logs with be printed:

```c++
  redirectArduinoLogs(logger);

  logger.setLevel(ARDUHAL_LOG_LEVEL_DEBUG);
  logger.forwardTo(Serial);

  log_d("Arduino debug message");
  log_i("Arduino info message");
  log_w("Arduino warning message");
  log_e("Arduino error message");
```

outputs:

```
SYSTEM: [   819][D][Logger.ino:17] setup(): Arduino debug message
SYSTEM: [   826][I][Logger.ino:18] setup(): Arduino info message
SYSTEM: [   833][W][Logger.ino:19] setup(): Arduino warning message
SYSTEM: [   840][E][Logger.ino:20] setup(): Arduino error message
```


SuGlider:
@mathieucarbou - I remember that the reason for that is because ESP32|ESP32-S2|ESP32-S3 (Xtensa SoC) had a problem with `ets_printf()`. This code works for UART and also USB CDC logging, when `Serial` is actually a HWSerial or USBCDC port.

We can review if this is still valid.

SuGlider:
@mathieucarbou - What SoC do you use with your projects?

mathieucarbou:
> @mathieucarbou - What SoC do you use with your projects?

I am currently testing with an esp32dev board, but the project has to also work for some other boards such as s2, s3, wt32_eth01, etc. This is a feature that will only be activated in debug mode to grab the boot logs from the ESP since users cannot connect easily a USB computer.

I did the modification locally in the c file for the esp32dev board and it works but I don't know the possible impact of this on other boards.

I was also intrigued by this comment in the header:

> To install putc1, which is defaulted installed as ets_write_char_uart

So this is my understanding that putc1 (so a call to ets_printf) should already use ets_write_char_uart behind ?

SuGlider (commented):
I'll retest log output within all SoC using just `ets_printf()` to verify if it works for all possible Serial ports (UART/CDC).

Let set this PR on hold until the results are out.

mathieucarbou:
> `ets_install_putc1()` and `ets_install_putc2()` should just define a pointer to a function that shall output anything that uses `ets_printf()`. In other words, `ets_printf()` calls the output function defined as `putc1()` or `putc2()`.

That's exactly my understanding also.

> Therefore, if your software needs to redirect log output, it could just change the function pointed by `ets_install_putc1()` for UART and `ets_install_putc2()` for USB CDC and use `ets_printf()` whenever logging.

That's what does not work.

If you mean: 

_if your software needs to redirect log output when using `ets_printf()`_

then I agree with you.

But this is not the use case explained above.
What I want and need is to redirect all calls to log_e, log_w, log_i, etc Basically, all Arduino logs, to a custom function. And for that, Arduino would need to call **every time** `ets_printf()`, which is not the case now because it directly calls instead the uart callback.

SuGlider:
@mathieucarbou - We will change `esp32-hal-uart.c` to only use `ets_printf()`.

SuGlider (approved):

mathieucarbou:
@SuGlider : thanks a lot !

lucasssvaz (approved):

me-no-dev (approved):

P-R-O-C-H-Y (approved):

mathieucarbou:
Hello,

Just wanted to to drop a note that I was able to verify the feature in my application logger implementation (https://github.com/mathieucarbou/MycilaLogger)

```c++
#include <Arduino.h>
#include <MycilaLogger.h>
#include <StreamString.h>

Mycila::Logger logger;
StreamString logBuffer;

void setup() {
  Serial.begin(115200);
  while (!Serial)
    continue;

  Mycila::Logger::redirectArduinoLogs(logger);

  logger.setLevel(ARDUHAL_LOG_LEVEL_DEBUG);

  logger.forwardTo(&logBuffer);

  log_d("Arduino debug message");
  log_i("Arduino info message");
  log_w("Arduino warning message");
  log_e("Arduino error message");

  Serial.println("Captured Arduino logs:");
  Serial.println(logBuffer);
  logBuffer.clear();

  logger.debug("APP", "A debug message");
  logger.info("APP", "An info message");
  logger.warn("APP", "A warning message");
  logger.error("APP", "An error message");

  Serial.println("Captured Mycila logger logs:");
  Serial.println(logBuffer);
  logBuffer.clear();
}

void loop() {
  delay(500);
}
```

outputs:

image

That's a really super cool feature for apps with a non-dev user base in order to save the startup logs (up to a max amount) on disk when debug mode is activated in one's app.

I am using it like that:

```c++
Mycila::Logger logger;

class LogStream : public Print {
  public:
    LogStream(const char* path, size_t limit) : _path(path), _limit(limit) {
      File file = LittleFS.open(_path, "r");
      _logSize = file ? file.size() : 0;
      file.close();
      _accepting = _logSize < _limit;
    }
    size_t write(const uint8_t* buffer, size_t size) override {
      if (!_accepting)
        return 0;
      if (_logSize + size > _limit) {
        _accepting = false;
        return 0;
      }
      File file = LittleFS.open(_path, "a");
      if (!file)
        return 0;
      size_t written = file.write(buffer, size);
      file.close();
      _logSize += written;
      return written;
    }
    size_t write(uint8_t c) override {
      assert(false);
      return 0;
    }

  private:
    const char* _path;
    const size_t _limit;
    size_t _logSize = 0;
    bool _accepting = false;
};

[...]
```

```c++
logger.redirectArduinoLogs();
logger.forwardTo(&Serial); // Serial output
logger.forwardTo(webSerial); // WebSerial console at http://1.2.3.4/console
logger.forwardTo(new LogStream(YASOLR_LOG_FILE, 32 * 1024)); // first 32k of startup logs go on disk
```

Then I allow the user to download the startup logs for troubleshooting.

Issue #8176, opened 2023-05-09
BLE provisioning sample doesn't work on release/v2.x branch


sidwarkd:
### Board

ESP32 Dev Module

### Device Description

DevKitC plain

### Hardware Configuration

No additional hardware connected.

### Version

latest master (checkout manually)

### IDE Name

VSCode

### Operating System

Linux

### Flash frequency

40MHz

### PSRAM enabled

no

### Upload speed

2000000

### Description

The use of `esp_bt_controller_mem_release` does not seem correct in esp32-hal-misc.c. In the function `initArduino()` if you have set BT_ENABLED in sdkconfig, it will call `btInUse` to determine whether to release the BT memory. There are two problems.

First, in using the sample sketch below (a slightly modified version of the provisioning sample included with this library) it does not appear to link correctly. The sample will fail to run properly as the BT memory is released in `initArduino` so trying to enable BT later for provisioning fails. The weak linking is not working as expected.

Secondly, this pattern seems wrong as well. The code in `initArduino()` would imply that you could have CONFIG_BT_ENABLED but have `btInUse()` return false which, per the code, would never be possible. If CONFIG_BT_ENABLED is yes then esp32-hal-bt.c declares `btInUse()` to return true always. So there is no scenario where if the config is enabled `btInUse()` _should_ return false (although it does per the first problem mentioned above).

This makes the following block of code in esp32-hal-misc.c puzzling:
```
#ifdef CONFIG_BT_ENABLED
    if(!btInUse()){
        esp_bt_controller_mem_release(ESP_BT_MODE_BTDM);
    }
#endif
```

To get the provisioning example code to work with BLE the call to `esp_bt_controller_mem_release` must NOT be called in `initArduino`.

### Sketch

```cpp
#include "WiFiProv.h"
#include "WiFi.h"
#include "nvs_flash.h"
#include "nvs.h"

bool is_provisioned = false;
void SysProvEvent(arduino_event_t *sys_event)
{
    switch (sys_event->event_id)
    {
    case ARDUINO_EVENT_WIFI_STA_GOT_IP:
        Serial.print("\nConnected IP address : ");
        Serial.println(IPAddress(sys_event->event_info.got_ip.ip_info.ip.addr));
        is_provisioned = true;
        break;
    case ARDUINO_EVENT_WIFI_STA_DISCONNECTED:
        Serial.println("\nDisconnected. Connecting to the AP again... ");
        break;
    case ARDUINO_EVENT_PROV_START:
[...]

PR #11167, opened 2025-03-23
Fix error #8176: simple_ble_start enable controller failed 259

Jason2866:
which happens when example https://github.com/espressif/arduino-esp32/tree/master/libraries/WiFiProv/examples/WiFiProv is compiled with pioarduino. The full pioarduino example is here https://github.com/pioarduino/test-rainmaker/tree/wifiprov

This PR fixes this bug by comment out the call to `esp_bt_controller_mem_release` since, once you release that memory, it cannot be reclaimed per the docs.

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -305,11 +305,11 @@ void initArduino() {
   if (err) {
     log_e("Failed to initialize NVS! Error: %u", err);
   }
-#if defined(CONFIG_BT_ENABLED) && SOC_BT_SUPPORTED
-  if (!btInUse()) {
-    esp_bt_controller_mem_release(ESP_BT_MODE_BTDM);
-  }
-#endif
+//#if defined(CONFIG_BT_ENABLED) && SOC_BT_SUPPORTED
+//  if (!btInUse()) {
+//    esp_bt_controller_mem_release(ESP_BT_MODE_BTDM);
+//  }
+//#endif
   init();
   initVariant();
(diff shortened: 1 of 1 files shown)


me-no-dev:
This is the classic BT memory being released, not BLE, so I wonder why you see the issue that you do. Maybe we can rethink this so that provisioning will cause it to not be released?

Jason2866:
@me-no-dev Isnt the provided PR just doing this? Keep the allocated memory for BT / BLE needs. Imho the weak if [here](https://github.com/espressif/arduino-esp32/blob/1145aec2b86d7ee08060caad86f72bd370f19e69/cores/esp32/esp32-hal-misc.c#L246-L257) does not work always as it should

me-no-dev:
@Jason2866 it is keeping it always, regardless if BT is used or not. The point of the commented code is to free that memory if BT is not in use. You in effect never free that memory with this PR

Jason2866:
Removed the in pioarduino non working weak stuff defines.

Admsher:
Nope it still gives the same error, if I am correct only the duplicate call from esp-hal-bt.c is removed?

Jason2866:
The PR is **not** needed for pioarduino. The use of weak definitions in Arduino core can be forced in Platformio / pioarduino by adding `lib_archive = no` in platformio.ini
The [example](https://github.com/pioarduino/test-rainmaker/tree/wifiprov) has this setting added and is working now.

Jason2866:
closing in favour of #12192

PR #11196, opened 2025-03-28
fix(zigbee): memory leak issue with malloc

SuGlider:
## Description of Change
In ` libraries/Zigbee/src/ZigbeeEP.cpp` file, `void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute)` has a memory leak due to a `malloc()` with no `free()`.

The same is applied to `bool ZigbeeEP::setManufacturerAndModel(const char *name, const char *model)` with necessary NULL test of `basic_cluster`.

Changed code to use `std::vector<char>` whenever possible. 

## Tests scenarios
CI only.

## Related links
None

### DIFF
--- a/libraries/Zigbee/src/ZigbeeEP.cpp
+++ b/libraries/Zigbee/src/ZigbeeEP.cpp
@@ -40,8 +40,8 @@ bool ZigbeeEP::setManufacturerAndModel(const char *name, const char *model) {
   size_t name_length = strlen(name);
   size_t model_length = strlen(model);
   if (name_length > ZB_MAX_NAME_LENGTH || model_length > ZB_MAX_NAME_LENGTH) {
-      log_e("Manufacturer or model name is too long");
-      return false;
+    log_e("Manufacturer or model name is too long");
+    return false;
   }
   // Allocate a new array of size length + 2 (1 for the length, 1 for null terminator)
   std::vector<char> zb_name(name_length + 2);  // +2 for length byte and null terminator
--- a/libraries/Zigbee/src/ZigbeeEP.cpp
+++ b/libraries/Zigbee/src/ZigbeeEP.cpp
@@ -19,6 +19,9 @@ ZigbeeEP::ZigbeeEP(uint8_t endpoint) {
   _ep_config.endpoint = 0;
   _cluster_list = nullptr;
   _on_identify = nullptr;
+  _read_model = nullptr;          // Explicitly initialize here
+  _read_manufacturer = nullptr;   // Explicitly initialize here
+
   _time_status = 0;
   if (!lock) {
     lock = xSemaphoreCreateBinary();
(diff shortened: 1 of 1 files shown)


lucasssvaz (changes requested):
VLAs like `char zb_name[name_length + 2]` are considered Undefined Behavior in C++ (although GCC supports it as an extension).

Better just to add the missing `free` to avoid the memory leak.

SuGlider:
> VLAs like `char zb_name[name_length + 2]` are considered Undefined Behavior in C++

In such case, using `std::vector<char>` declaration is even better in that case. No need to consider `free()`.

SuGlider:
GCC supporting it is important. I would prefer allocating memory for a few bytes in Stack instead of Heap. Reduces memory fragmentation and it is faster.

libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -245,7 +249,7 @@ void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute) {
   /* Basic cluster attributes */
   if (attribute->id == ESP_ZB_ZCL_ATTR_BASIC_MANUFACTURER_NAME_ID && attribute->data.type == ESP_ZB_ZCL_ATTR_TYPE_CHAR_STRING && attribute->data.value) {
     zbstring_t *zbstr = (zbstring_t *)attribute->data.value;
-    char *string = (char *)malloc(zbstr->len + 1);
+    char string[zbstr->len + 1];
```
me-no-dev: TBH I do not like using the reserved word "string" here. Maybe use "str" or something else

SuGlider:
@P-R-O-C-H-Y - I can't test it here to make sure that it works. Please test it using HA and the Zigbee dongle. Thanks!

SuGlider:
Reviewing it over again... code change is wrong.

SuGlider:
Sorry guys. Current code is 100% fine. No issues. My bad.

libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -163,9 +168,7 @@ char *ZigbeeEP::readManufacturer(uint8_t endpoint, uint16_t short_addr, esp_zb_i
   read_req.attr_number = ZB_ARRAY_LENTH(attributes);
   read_req.attr_field = attributes;
 
-  if (_read_manufacturer != nullptr) {
-    free(_read_manufacturer);
-  }
+  free(_read_manufacturer);  // CPP tests it for nullptr
```
me-no-dev, line 171: what does the comment mean?

libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -33,38 +35,41 @@ void ZigbeeEP::setVersion(uint8_t version) {
...
-  if (name_length > 32 || model_length > 32) {
+  if (name_length > ZB_MAX_NAME_LENGTH || model_length > ZB_MAX_NAME_LENGTH) {
     log_e("Manufacturer or model name is too long");
     return false;
   }
   // Allocate a new array of size length + 2 (1 for the length, 1 for null terminator)
-  char *zb_name = new char[name_length + 2];
-  char *zb_model = new char[model_length + 2];
+  std::vector<char> zb_name(name_length + 2);
+  std::vector<char> zb_model(model_length + 2);
```
me-no-dev, line 48: what will happen if there was not enough memory?

libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -245,20 +246,22 @@ void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute) {
...
-    memcpy(string, zbstr->data, zbstr->len);
-    string[zbstr->len] = '\0';
-    log_i("Peer Manufacturer is \"%s\"", string);
-    _read_manufacturer = string;
+    std::vector<char> zb_manufacturer(zbstr->len + 1);
+    memcpy(zb_manufacturer.data(), zbstr->data, zbstr->len);
+    zb_manufacturer[zbstr->len] = '\0';
+    log_i("Peer Manufacturer is \"%s\"", zb_manufacturer.data());
+    free(_read_manufacturer);  // Free any previously allocated memory
+    _read_manufacturer = strdup(zb_manufacturer.data());  // Duplicate the information for persistent storage
```
me-no-dev, line 254: `strdup` will return NULL if there was not enough memory

PR #11199, opened 2025-03-28
fix(zigbeeEP): review of names and memory allocation

SuGlider:
## Description of Change
Just a general review of memory allocation routines and changing `string` variable name to avoid using reserved words.

## Tests scenarios
CI only

## Related links
https://github.com/espressif/arduino-esp32/pull/11196

### DIFF
--- a/libraries/Zigbee/src/ZigbeeEP.cpp
+++ b/libraries/Zigbee/src/ZigbeeEP.cpp
@@ -251,7 +251,7 @@ void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute) {
   if (attribute->id == ESP_ZB_ZCL_ATTR_BASIC_MANUFACTURER_NAME_ID && attribute->data.type == ESP_ZB_ZCL_ATTR_TYPE_CHAR_STRING && attribute->data.value) {
     zbstring_t *zbstr = (zbstring_t *)attribute->data.value;
     char *_read_manufacturer = (char *) malloc(zbstr->len + 1);
-    if (_read_manufacturer == nullptr) {
+    if (_read_manufacturer == NULL) {
       log_e("Failed to allocate memory for manufacturer data");
       xSemaphoreGive(lock);
       return;
@@ -264,7 +264,7 @@ void ZigbeeEP::zbReadBasicCluster(const esp_zb_zcl_attribute_t *attribute) {
   if (attribute->id == ESP_ZB_ZCL_ATTR_BASIC_MODEL_IDENTIFIER_ID && attribute->data.type == ESP_ZB_ZCL_ATTR_TYPE_CHAR_STRING && attribute->data.value) {
     zbstring_t *zbstr = (zbstring_t *)attribute->data.value;
     char *_read_model = (char *) malloc(zbstr->len + 1);
-    if (_read_model == nullptr) {
+    if (_read_model == NULL) {
       log_e("Failed to allocate memory for model data");
       xSemaphoreGive(lock);
       return;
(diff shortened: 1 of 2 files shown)


libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -40,9 +42,22 @@ bool ZigbeeEP::setManufacturerAndModel(const char *name, const char *model) {
...
+  if (basic_cluster == nullptr) {
+    log_e("Failed to get basic cluster");
+    return false;
+  }  
   // Allocate a new array of size length + 2 (1 for the length, 1 for null terminator)
-  char *zb_name = new char[name_length + 2];
-  char *zb_model = new char[model_length + 2];
+  char *zb_name = (char *) malloc(name_length + 2);
+  char *zb_model = (char *) malloc(model_length + 2);
+  if (zb_name == nullptr || zb_model == nullptr) {
```
me-no-dev, line 54: nitpick: You are using "C" AP (`malloc`)I. You should check against C value (`NULL`) not "C++" (`nullptr`). Same for the rest of the code

SuGlider: I'll change it back to new/delete.

me-no-dev: no. please leave it as malloc/free, just check against NULL for posterity

SuGlider: Changed it to use stack alocated memory instead. It is 68 bytes on stack.
It is faster and there is no heap memory fragmentation side effect.

libraries/Zigbee/src/ZigbeeEP.cpp
```diff
@@ -33,16 +35,24 @@ void ZigbeeEP::setVersion(uint8_t version) {
 }
 
 bool ZigbeeEP::setManufacturerAndModel(const char *name, const char *model) {
+  // Allocate a new array of size length + 2 (1 for the length, 1 for null terminator)
+  // This memory space can't be deleted or overwritten, therefore using static
```
SuGlider, line 39: @P-R-O-C-H-Y  - Could you please check if `esp_zb_basic_cluster_add_attr(...)` copies (duplicates) or just stores the pointer to both name strings?

Please let me know. Thanks!

libraries/Zigbee/src/ZigbeeEP.h
```diff
@@ -138,8 +142,8 @@ class ZigbeeEP {
   }
 
 private:
-  char *_read_manufacturer;
-  char *_read_model;
+  char _read_manufacturer[ZB_MAX_NAME_LENGTH + 1];  // + 1 for the extra '\0'
+  char _read_model[ZB_MAX_NAME_LENGTH + 1];  // + 1 for the extra '\0'
```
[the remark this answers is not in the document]
SuGlider: What I see here is that this code is being used for both, Server Clusters (statefull EPs that hold a attribute database) and also for Client Clusters (stateless EPs that query and set Server Clusters).

+1 for being printable, but it shall not be necessary...
+2 because the ATTR data model demands first byte to contain the size the the remaining to be the data (not necessary a '\0' at the end, actually.

libraries/Zigbee/src/ZigbeeEP.cpp
SuGlider: data is copied into a separated memory space.

libraries/Zigbee/src/ZigbeeEP.h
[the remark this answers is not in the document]
SuGlider: In my opinion. This should be +1 (size) for the ATTR format:
[size][data]
Keep it as the internal database and set names into it, read values from it.

lucasssvaz (approved):

me-no-dev (approved):

PR #11206, opened 2025-03-31
fix(board): Alfredo NoU3 pins_arduino.h uses uint8_t but it causes error with esp32-hal-gpio.h

SaintSampo:
## Description of Change
Starting in core version 3.2.0, this pins_arduino.h file fails to compile without stdint.h

## Tests scenarios
Tested with 3.1.3 and 3.2.0. File compiles in the former but not the latter (uint8_t not defined). I'm curious what caused this change but just hoping this fixes in. I would be surprised if other pins files don't have this issue.


lucasssvaz (approved):

variants/alfredo-nou3/pins_arduino.h
```diff
@@ -1,6 +1,8 @@
 #ifndef Pins_Arduino_h
 #define Pins_Arduino_h
 
+#include <stdint.h>
```
SuGlider, line 4: It is included in `<Arduino.h>` along with many other:

``` cpp
#include <stdbool.h>
#include <stdint.h>    // <<<<<<< Here. Line 24 of Arduino.h
#include <stdarg.h>
#include <stddef.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <inttypes.h>
```

joshua-8:
Hi @SuGlider, thank you for taking a look at this issue for us.

I found that I could not compile for an Alfredo NoU3 board with esp32 v3.2.0.
The Alfredo NoU3 `pins_arduino.h` uses the `uint8_t` type for some pins https://github.com/SaintSampo/arduino-esp32/blob/1b79810df333c46517a634c3e8a666d47ad55ced/variants/alfredo-nou3/pins_arduino.h#L21-L27

While Arduino.h includes `stdint.h`, `pins_arduino.h` does not include `Arduino.h`, and `pins_arduino.h` appears to compile before another file can include `Arduino.h`.

Here's an example that failed to compile. https://github.com/AlfredoSystems/Alfredo-NoU3/tree/main/examples/MotorParty

I have attached the output from the Arduino IDE showing that with esp32 version 3.2.0 the example only compiles when I edited `pins_arduino.h` on my computer to add `stdint.h`.

[failed compile log with esp32 version 3.2.0.txt](https://github.com/user-attachments/files/19622919/failed.compile.log.with.esp32.version.3.2.0.txt)
[successful compiling with old esp32 version 3.0.7.txt](https://github.com/user-attachments/files/19622920/successful.compiling.with.old.esp32.version.3.0.7.txt)
[successful compiling with edited esp32 version 3.2.0.txt](https://github.com/user-attachments/files/19622921/successful.compiling.with.edited.esp32.version.3.2.0.txt)

Is there any harm in including `stdint.h` in pins_arduino.h? One other potential solution would be to change the variable type used for the pins in `pins_arduino.h` to `byte`.

Thank you, espressif community, for your time.
-- joshua-8

SuGlider:
@joshua-8 - Thanks for the explanation and log files.
The issue is not in **Alfredo NoU3** `pins_arduino.h` file.
It is due to wrong include order in `cores\esp32\esp32-hal-gpio.h`.

The way to fix it for this board and all other possible boards that may use `stdint.h` types is by fixing `cores\esp32\esp32-hal-gpio.h`.

I'll push the commits necessary to this PR.

Thanks.

SuGlider:
@joshua-8 - Please test this PR with the target board and sketches. It shall be solves now.
The proposed changes will fix not just  *Alfredo NoU3* `pins_arduino.h` file, but all other boards that may want to use stdint or stdbool types.

SuGlider:
@SaintSampo @joshua-8 - My bad.... actually all other `pins_arduino.h` files include <stdint>... This boards should do the same.
The initial PR was correct. Reverting it to the original PR.

SuGlider (approved):
LGTM

Issue #11200, opened 2025-03-28
RMT: RMTLoopback example hangs and gives core error


fanfanlatulipe26:
### Board

ESP32 Dev Kit v1

### Device Description

ESP32 Dev Kit v1

### Hardware Configuration

Jumper between pin 18 and 21

### Version

v3.1.2

### IDE Name

Arduino IDE 1.8.19

### Operating System

Windows 10

### Flash frequency

80Mhz

### PSRAM enabled

no

### Upload speed

921600

### Description

In fact I use the very last version of ESP32: 3.2.0  (release not yet available in the dropdown list here above)

The standard RMTLoopback  example fails with an error and hangs
`E (1190) rmt: rmt_receive(386): partial receive not supported
`   
If the example is compiled with debug level None we just get:  
`real tick set to: 100ns   Please connect GPIO 18 to GPIO 21, now.  E (11) rmt: rmt_receive(386): partial receive not supported
`
If we compile with debug level Error, we get more output from the testcase as can be seen in the "Debug Message" section

As can be seen in the output of the test, the results are not exactly the one expexted, but not really garbage. Some shift ??

More: the frequency used in the example is 10 000 000
`#define RMT_FREQ               10000000  // tick time is 100ns
`
In fact the generated waveform shows that it is not 100ns, and the wave form is not exactly the one given by the data array.
I found little documentation and the tick time is not at all always as described in the comment in esp32-hal-rmt.h for rmtInit

`New Parameters in Arduino Core 3: RMT tick is set in the rmtInit() function by the
    frequency of the RMT channel. Example: 100ns tick => 10MHz, thus frequency will be 10,000,000 Hz
`
Seems OK for 1Mhz, 2Mhz,8Mhz, but bad for 5Mhz, 10Mhz.

### Sketch

```cpp
The example given in for ESP32 RMT: RMTLoopback
No change.
```

### Debug Message

```plain
ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:1
load:0x3fff0030,len:4888
load:0x40078000,len:16516
load:0x40080400,len:4
load:0x40080404,len:3476
entry 0x400805b4
[...]

PR #11221, opened 2025-04-06
refactor(rmt): refactored RMT loopback example

SuGlider:
## Description of Change
The way RMT loopback example was done, it had a sync problem that caused the result to do not match.
Low level signal was sent first and then the high level, therefore, it would only detect the second signal (high) after a delay in nanoseconds. This delay would cause the received signal to be shifted by one RMT symbol.

Output of the original example:
```
Got 30 RMT symbols
000e8001=8001000e 000d8001=8002000d 000b8002=8003000c 000a8004=8004000b 
00098005=8005000a 00088006=80060009 00078007=80070008 00068008=80080007 
```

With this refactoring/fix, RMT symbols start with High Level and the output matches the received RMT symbols.
This PR makes the example easier for understanding as well as the RMT symbol will better match.

```
Got 32 RMT symbols
0001800e=0001800e 0002800d=0002800d 0003800c=0003800c 0004800b=0004800b 
0005800a=0005800a 00068009=00068009 00078008=00078008 00088007=00088007 
```

## Tests scenarios
Tested with ESP32, ESP32-C3 and ESP32-S3 using the provided example.

## Related links
Related to #11200

### DIFF
--- a/libraries/ESP32/examples/RMT/RMTLoopback/RMTLoopback.ino
+++ b/libraries/ESP32/examples/RMT/RMTLoopback/RMTLoopback.ino
@@ -57,10 +57,10 @@ void setup() {
 
   // create multiple pulses with different width to be sent
   for (int i = 0; i < 255; i++) {
-    data[i].level0 = 1; // HIGH
-    data[i].duration0 = 1 + 13 - (i % 13); // number of Tick on High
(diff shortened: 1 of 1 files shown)


lucasssvaz (approved):
LGTM

me-no-dev (approved):

PR #11232, opened 2025-04-08
fix(core): StreamString performance improvements

mathieucarbou:


SuGlider:
Question about this PR: if there are 2 tasks (producer/consumer), one writes data to the StreamString and the other reads data from the same StreamString, wouldn't its Strinng Buffer grow for ever?

SuGlider:
For the issue with buffers, I'd suggest using some other structure, such as this one:
https://github.com/arduino/ArduinoCore-samd/blob/master/libraries/I2S/src/utility/I2SDoubleBuffer.cpp
It implements all necessary functions, but it has a fixed size buffer. 
An extension to this implementation could be a linked list of "buffers" (String pointers) that could grow as necessary. 
It can use a positional index within the String that is being read at that moment.  
The linked list can grow or shrink as necessary by creating/destroying Strings and adding/removing it pointers in a linked list.
As data is read, whenever a memory block is fully read, memory can be deleted, returning to the heap.

Issue #11254, opened 2025-04-15
esp32-c5 fails to compile with branch `release/v3.3.x`


Jason2866:
### Description

Forked Arduino Lib Builder https://github.com/Jason2866/esp32-arduino-lib-builder/tree/v5.5_Arduino
fails with https://github.com/Jason2866/esp32-arduino-lib-builder/actions/runs/14479931341
using https://github.com/tasmota/arduino-esp32/tree/release/v3.3.x and https://github.com/tasmota/esp-idf

```plain
/Users/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/components/arduino/cores/esp32/esp32-hal-uart.c:1160:58: error: 'LP_UART_SCLK_LP_FAST' undeclared (first use in this function); did you mean 'LP_UART_SCLK_RC_FAST'?
 1160 |       case UART_SCLK_RTC:     uart->_uart_clock_source = LP_UART_SCLK_LP_FAST; break;
      |                                                          ^~~~~~~~~~~~~~~~~~~~
      |                                                          LP_UART_SCLK_RC_FAST
/Users/runner/work/esp32-arduino-lib-builder/esp32-arduino-lib-builder/components/arduino/cores/esp32/esp32-hal-uart.c:1160:58: note: each undeclared identifier is reported only once for each function it appears in
```

Reverting change in file `esp32-hal-uart.c` from #11191 solves the compile fail

Jason2866:
@SuGlider can you have a look?

PR #11255, opened 2025-04-17
fix(esp32c5): Add proper RTC clock source to uartSetClockSource

Jason2866:
C5 fails to compile when trying to build a firmware. Can't test if the fix is correct since not having a C5. If correct fix for #11254

@SuGlider


lucasssvaz (approved):
LGTM

SuGlider:
@Jason2866 - Yes, this is the proper fix for the ESP32-C5 LP UART clock source.

SuGlider (approved):

PR #11309, opened 2025-04-28
feat(uart): simplifies UART example based on MODBUS standard

SuGlider:
## Description of Change
Improves and simplifies the Serial OnReceive example.
The example uses Rx Timeout based on MODBUS standard.

This change will demonstrate how to receive any number of bytes and then execute a callback as soon as defined timeout happens.

## Tests scenarios
Using this example with ESP32, ESP32-S2, ESP32-S3, ESP32-C3 and ESP32-C6.

## Related links
None.


libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -57,67 +62,52 @@
...
-      delay(2000);
-    }
-  }
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
+  // UART_CLK_SRC_APB will allow higher values of RX Timeout
+  // default for ESP32 and ESP32-S2 is REF_TICK which limits the RX Timeout to 1
+  // setClockSource() must be called before begin()
+  UART0.setClockSource(UART_CLK_SRC_APB);
+#endif
+  // the amount of data received or waiting to be proessed shall not exceed this limit of 1024 bytes
```
SuGlider, line 88:
```suggestion
  // the amount of data received or waiting to be processed shall not exceed this limit (1024 bytes)
```

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -57,67 +62,52 @@
...
-    uint32_t now = millis();  // tracks timeout
-    while ((millis() - now) < communicationTimeout_ms) {
-      if (UART0.available()) {
-        uart_buffer += (char)UART0.read();
-        now = millis();  // reset the timer
-      }
-    }
-    // releases the mutex for data processing
-    xSemaphoreGive(uart_buffer_Mutex);
+  while (UART0.available()) {
```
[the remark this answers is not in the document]
SuGlider: At the same time, it may lead to a dead lock due to the callback running on a higher priority task. If the semaphore is taken by loop() task and data is received with timeout before it can give the semaphore, the higher priotity task that runs onReceive() will be dead locked.

Therefore, for this example, we just let the callback to add more data to the String if loop() data processing is interrupted.

lucasssvaz (approved):
LGTM

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -57,67 +62,52 @@
```
me-no-dev, line 62: Given that `Serial0` always exists, why not use that instead of mapping to UART0?

SuGlider: Ok, let me just make sure it won't fail and then I'll change it.

me-no-dev (approved):

libraries/ESP32/examples/Serial/onReceiveExample/onReceiveExample.ino
```diff
@@ -34,90 +34,73 @@
...
-
-// Create a mutex for the access to uart_buffer
-// only one task can read/write it at a certain time
-SemaphoreHandle_t uart_buffer_Mutex = NULL;
-
-// UART_RX_IRQ will be executed as soon as data is received by the UART
-// This is a callback function executed from a high priority
-// task created when onReceive() is used
+// The Modbus RTU standard prescribes a silent period corresponding to 3.5 characters between each
+// message, to be able fo figure out where one message ends and the next one starts.
```
SuGlider, line 59:
```suggestion
// message, to be able to figure out where one message ends and the next one starts.
```

PR #11315, opened 2025-04-28
Restore proper pin remapping functionality after 3.2.0

pillo79:
Issue reported at the Arduino Forums in [this thread](https://forum.arduino.cc/t/esp32-core-3-2-0-gives-many-compile-errors-with-by-arduino-pin-setting/1374758).

Commit 0773dd7619b1117bd59a068b69a76cdb22484f7c from PR #10841 broke pin remapping by moving its application too early in the definition process. This commit restores the original order of includes, ensuring that pin remapping is applied correctly.


me-no-dev (approved):

SuGlider:
> @pillo79 - thanks for the PR. Just let me check if it work with the #10841 use case.

All good. LGTM for merging.

SuGlider (approved):

PR #11341, opened 2025-05-07
fix compile fail esp_hosted

Jason2866:
branch master has updates which needs esp_hosted >= 2.0.0
@me-no-dev


me-no-dev:
waiting on CI and will merge, then re-trigger the lib-builder. C5 will still fail there, but I can fix that

PR #11350, opened 2025-05-11
Added new Unexpected Maker SQUiXL and EDGES3[D] boards.

UnexpectedMaker:
Added support for 2 of my new boards: SQUiXL and EdgeS3[D]

Both ESP32-S3 boards.

Thanks :)

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -3672,7 +3672,7 @@ um_edges3_d.build.target=esp32s3
 um_edges3_d.build.mcu=esp32s3
 um_edges3_d.build.core=esp32
 um_edges3_d.build.variant=um_edges3_d
-um_edges3_d.build.board=EDGES3[D]
+um_edges3_d.build.board=EDGES3D
 
 um_edges3_d.build.usb_mode=1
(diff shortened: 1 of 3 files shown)


UnexpectedMaker:
It seems SQUiXL is failing your CI compile tests because it's trying to do tests with SPI - There is no SPI available to users on this board - so no SPI pins have been defined in arduino_pins.h - this is intentional, so I am not sure how to get past this issue.

Is this a real blocker @me-no-dev ?

Also, I've signed and confirmed the CLA - but it never changes to says I've signed.

Thanks :)

P-R-O-C-H-Y:
@UnexpectedMaker If you don't want to specify the SPI pins, you can set them as -1, but the SPI pins definition must exist :)

UnexpectedMaker:
> @UnexpectedMaker If you don't want to specify the SPI pins, you can set them as -1, but the SPI pins definition must exist :) CLA seems to be fine.

Oh, I didn't realize I could set them to -1 - I ended up adding IO assigned to them anyway :)

PR #11360, opened 2025-05-13
feat(board): add Satellite 1 CORE (rev 5.1)

ericlewis:
## Description of Change
Add the [Satellite 1 CORE (rev 5.1)](https://futureproofhomes.net/products/satellite1-core-board) from FutureProofHomes.

## Tests scenarios
I have tested my Pull Request on the board itself using:

```cpp
#include <Arduino.h>

void setup() {
  Serial.begin(115200);
  while (!Serial && millis() < 3000) ;   // wait for the host (max 3 s)

  pinMode(LED_BUILTIN, OUTPUT);

  if (!psramFound()) {
    Serial.println("PSRAM not detected!");
    return;
  }

  uint32_t bytes = ESP.getPsramSize();   // 0, 2 048 000, 4 194 304, 8 388 608
  Serial.printf("PSRAM size : %u bytes\n", bytes);
  Serial.printf("PSRAM mode : %s (inferred)\n",
                bytes >= 8 * 1024 * 1024 ? "Octal" : "Quad");
}

void loop() {
  digitalWrite(LED_BUILTIN, HIGH);
  delay(500);
  digitalWrite(LED_BUILTIN, LOW);
  delay(500);
}
```

## Links
The relevant spec: https://github.com/FutureProofHomes/Satellite1-Hardware/blob/main/core/rev5.1coreSCH.pdf

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -49849,3 +49849,157 @@ cyobot_v2_esp32s3.menu.ZigbeeMode.zczr.build.zigbee_mode=-DZIGBEE_MODE_ZCZR
 cyobot_v2_esp32s3.menu.ZigbeeMode.zczr.build.zigbee_libs=-lesp_zb_api.zczr -lzboss_stack.zczr -lzboss_port.remote
 
 ##############################################################
+
+# Satellite 1 CORE (rev 5.1) 
+
+sat1core_rev5_1.name=Satellite 1 CORE (rev 5.1) 
+
+sat1core_rev5_1.bootloader.tool=esptool_py
+sat1core_rev5_1.bootloader.tool.default=esptool_py
+
+sat1core_rev5_1.upload.tool=esptool_py
+sat1core_rev5_1.upload.tool.default=esptool_py
+sat1core_rev5_1.upload.tool.network=esp_ota
+
+sat1core_rev5_1.upload.maximum_size=1310720
+sat1core_rev5_1.upload.maximum_data_size=327680
+sat1core_rev5_1.upload.flags=
+sat1core_rev5_1.upload.extra_flags=
+sat1core_rev5_1.upload.use_1200bps_touch=false
+sat1core_rev5_1.upload.wait_for_upload_port=false
+
+sat1core_rev5_1.serial.disableDTR=false
+sat1core_rev5_1.serial.disableRTS=false
+
+sat1core_rev5_1.build.tarch=xtensa
+sat1core_rev5_1.build.bootloader_addr=0x0
+sat1core_rev5_1.build.target=esp32s3
+sat1core_rev5_1.build.mcu=esp32s3
+sat1core_rev5_1.build.core=esp32
+sat1core_rev5_1.build.variant=satellite1_core_rev5_1
+sat1core_rev5_1.build.board=SATELLITE1_CORE_REV5_1
+
+sat1core_rev5_1.build.usb_mode=1
+sat1core_rev5_1.build.cdc_on_boot=1
+sat1core_rev5_1.build.msc_on_boot=0
+sat1core_rev5_1.build.dfu_on_boot=0
+sat1core_rev5_1.build.f_cpu=240000000L
+sat1core_rev5_1.build.flash_size=16MB (128Mb)
+sat1core_rev5_1.build.flash_freq=80m
+sat1core_rev5_1.build.flash_mode=dio
+sat1core_rev5_1.build.boot=qio
+sat1core_rev5_1.build.boot_freq=80m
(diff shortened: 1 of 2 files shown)


boards.txt
```diff
@@ -50155,3 +50155,157 @@ cyobot_v2_esp32s3.menu.ZigbeeMode.zczr.build.zigbee_mode=-DZIGBEE_MODE_ZCZR
...
+
+sat1core_rev5_1.menu.PSRAM.opi=Enabled
+sat1core_rev5_1.menu.PSRAM.opi.build.defines=-DBOARD_HAS_PSRAM
+sat1core_rev5_1.menu.PSRAM.opi.build.psram_type=opi
+
+sat1core_rev5_1.menu.FlashMode.qio=QIO 80MHz
+sat1core_rev5_1.menu.FlashMode.qio.build.flash_mode=dio
+sat1core_rev5_1.menu.FlashMode.qio.build.boot=qio
+sat1core_rev5_1.menu.FlashMode.qio.build.boot_freq=80m
+sat1core_rev5_1.menu.FlashMode.qio.build.flash_freq=80m
```
P-R-O-C-H-Y, line 50213: I think you can run a OPI Flash mode as your module has OPI PSRAM, if you don't have something custom. Can you check?

boards.txt
```diff
@@ -50155,3 +50155,157 @@ cyobot_v2_esp32s3.menu.ZigbeeMode.zczr.build.zigbee_mode=-DZIGBEE_MODE_ZCZR
...
+
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme1=Small App w/ OTA + Huge FS (2MB APP/2MB OTA/12MB SPIFFS)
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme1.build.custom_partitions=gen4esp32_2MBapp_2MBota_12MBspiffs
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme1.upload.maximum_size=2097152
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme2=Medium App w/ OTA + Large FS (4MB APP/4MB OTA/7MB SPIFFS)
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme2.build.custom_partitions=gen4esp32_4MBapp_4MBota_7MBspiffs
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme2.upload.maximum_size=4718592
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme3=Large App w/ OTA (8MB APP/8MB OTA)
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme3.build.custom_partitions=gen4esp32_8MBapp_8MBota
+sat1core_rev5_1.menu.PartitionScheme.gen4esp32scheme3.upload.maximum_size=8323072
```
P-R-O-C-H-Y, line 50263: This won't work, as those partitions does not exist in the core. You have to use the ones we have and provide a "Custom" option there also, where users can you your custom ones you can provide in your docs or some repo :)

variants/satellite1_core_rev5_1/pins_arduino.h
```diff
@@ -0,0 +1,60 @@
...
+#define BOARD_HAS_PSRAM   1
+#define DEFAULT_PSRAM_OCT 1         // 8-bit OPI @ 120 MHz
+
+/* ───────── ADC & touch aliases ───────── */
+static const uint8_t A0 = 1;   static const uint8_t T4  = 4;
+static const uint8_t A1 = 2;   static const uint8_t T5  = 5;
+static const uint8_t A2 = 3;   static const uint8_t T6  = 6;
+static const uint8_t A3 = 4;   static const uint8_t T7  = 7;
+static const uint8_t A4 = 6;   static const uint8_t T8  = 8;
+static const uint8_t A5 = 7;
```
P-R-O-C-H-Y, line 58: Just a formatting here.
```suggestion
static const uint8_t A0 = 1;   
static const uint8_t A1 = 2;   
static const uint8_t A2 = 3;  
static const uint8_t A3 = 4;   
static const uint8_t A4 = 6;   
static const uint8_t A5 = 7;

static const uint8_t T4  = 4;
static const uint8_t T5  = 5;
static const uint8_t T6  = 6;
static const uint8_t T7  = 7;
static const uint8_t T8  = 8;
```

P-R-O-C-H-Y (changes requested):
@ericlewis PTAL on my comments

ericlewis:
yeah im gonna go ahead and close this bc there is a lot of issues with this board.

PR #11403, opened 2025-05-26
fix(board): Alfredo-NoU3 changed default upload mode from Hardware CDC and JTAG to USB-OTG

SaintSampo:
Hardware CDC and JTAG with the ESP32S3 sometimes gets into a state where the only way to upload code is to first put the MCU into BOOT mode: https://github.com/espressif/arduino-esp32/discussions/11327

To fix this, and now that the tiny-USB implementation on ESP32-S3 is in decent shape, upload over USB-OTG is the way to go.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -43777,8 +43777,8 @@ alfredo-nou3.upload.maximum_size=3342336
 alfredo-nou3.upload.maximum_data_size=327680
 alfredo-nou3.upload.flags=
 alfredo-nou3.upload.extra_flags=
-alfredo-nou3.upload.use_1200bps_touch=false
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y:
Hi @SaintSampo, you added a new boards in last commit.
PLease revert or add all necessary changes like the variants/pins_arduino.h file for the new board and change title and description of the PR :)

P-R-O-C-H-Y:
@SaintSampo The best you can do is to create a branch on your fork and commit changes to the branch. Then you can open a PR that will be coming from the branch. This way you can do multiple separate changes/PRs :)

PR #11430, opened 2025-06-04
fix: Add PSRAM Option for Geekble nano

syong0921:
fix: Add PSRAM Option for Geekble nano

## Description of Change
we noticed that Geekble nano users need PSRAM option.
So enabled Geekble nano's PSRAM option in board.txt files

## Tests scenarios
Tested under Arduino IDE v2.3.6, Arduino-esp32 core v3.2.0 with Geekble nano

## Related links


lucasssvaz:
Your base branch is behind our master and is reverting many changes. Please rebase first.

PR #11437, opened 2025-06-04
fix: Updated the tools options for Geekble Mini

syong0921:
fix: Updated the tools options for Geekble Mini

Description of Change

Geekble Mini is a Module for Arduino - ESP beginners 
We thought Geekble Mini still have too much options, which can confuse beginners 
So deleted rarely used options

Tests scenarios

Tested under Arduino IDE v2.3.6, Arduino-esp32 core v3.2.0 with Geekble Mini

Related links


P-R-O-C-H-Y (approved):
LGTM

PR #11466, opened 2025-06-14
fix(example): led flash not working if not using default model in camera example.

HighDoping:
## Description of Change
LED flash do not work on some boards with esp32-cam example because the pin is always set to 22. This PR changes the pin definition to be set dynamically by setupLedFlash().

## Tests scenarios
I have tested my Pull Request on Arduino-esp32 core v3.2.0 with AI Thinker ESP32-CAM

### DIFF
--- a/libraries/ESP32/examples/Camera/CameraWebServer/board_config.h
+++ b/libraries/ESP32/examples/Camera/CameraWebServer/board_config.h
@@ -1,8 +1,13 @@
+#ifndef BOARD_CONFIG_H
+#define BOARD_CONFIG_H
+
+//
 // WARNING!!! PSRAM IC required for UXGA resolution and high JPEG quality
 //            Ensure ESP32 Wrover Module or other board with PSRAM is selected
 //            Partial images will be transmitted if image exceeds buffer size
 //
 //            You must select partition scheme from the board menu that has at least 3MB APP space.
+
 // ===================
 // Select camera model
 // ===================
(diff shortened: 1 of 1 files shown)


libraries/ESP32/examples/Camera/CameraWebServer/app_httpd.cpp
```diff
@@ -30,9 +30,9 @@
 // LED FLASH setup
 #if CONFIG_LED_ILLUMINATOR_ENABLED
 
-#define LED_LEDC_GPIO            22  //configure LED pin
 #define CONFIG_LED_MAX_INTENSITY 255
 
+int led_pin = 22;  // Store the actual LED pin used
```
me-no-dev: 22 is invalid GPIO on some chips. Code should be reworked so that `CONFIG_LED_ILLUMINATOR_ENABLED` and `LED_LEDC_GPIO` are defined in the camera board config and not just be enabled by default

HighDoping: Thanks for review, I'm working on a version that use ```#if defined(LED_GPIO_NUM)``` to check if LED is available.
I'll test on real hardware this weekend and upload the new fix.

libraries/ESP32/examples/Camera/CameraWebServer/board_config.h
lucasssvaz, line 1: Please add the proper guards for header files

lucasssvaz (approved):

me-no-dev (approved):

PR #11492, opened 2025-06-21
feat(uart): fixes loopback function after IDF changes

SuGlider:
## Description of Change
IDF 5.4.1 has added a new function called uart_release_pin() that is called whenever new pins are set or when uart driver is deleted.

This has a side effect that causes RX pin to never work again with the loopback function. Other changes also have removed some GPIO setup that was necessary for the GPIO loopback mode work.

This PR forces a full RX Pin setup in order to make it work in GPIO Matrix with Loopback TX Signal

## Tests scenarios
CI Only

## Related links
CI error

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -1399,10 +1399,10 @@ void uart_internal_loopback(uint8_t uartNum, int8_t rxPin) {
     return;
   }
   // forces rxPin to use GPIO Matrix and setup the pin to receive UART TX Signal - IDF 5.4.1 Change with uart_release_pin()
-  gpio_func_sel((gpio_num_t) rxPin, PIN_FUNC_GPIO);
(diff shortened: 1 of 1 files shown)


SuGlider:
@me-no-dev : This PR shall fix #11416 after the IDF Change done by the commit https://github.com/espressif/esp-idf/commit/60f5828a4a790d4481aaf7c69eac72e3af1079fe

me-no-dev (approved):

Jason2866:
The PR does **not** fix the issue @SuGlider @me-no-dev

SuGlider:
> The PR does **not** fix the issue @SuGlider @me-no-dev 

This is for the CI.
Investigation continues.

SuGlider:
> The PR does **not** fix the issue @SuGlider @me-no-dev

@Jason2866 - If you have time, please open an issue with, as many as possible, test cases related to the issue that the IDF change created in Arduino Serial pins. Thanks!

PR #11523, opened 2025-06-27
Fix appending to Strings longer than 64k

MattiasTF:
If oldLen is truncated to uint16_t, appending to a String that is longer than 65535 bytes will create a broken String.

### DIFF
--- a/cores/esp32/WString.cpp
+++ b/cores/esp32/WString.cpp
@@ -180,7 +180,7 @@ bool String::changeBuffer(unsigned int maxStrLen) {
   if (maxStrLen < sizeof(sso.buff) - 1) {
     if (isSSO() || !buffer()) {
       // Already using SSO, nothing to do
-      uint16_t oldLen = len();
+      size_t oldLen = len();
       setSSO(true);
       setLen(oldLen);
     } else {  // if bufptr && !isSSO()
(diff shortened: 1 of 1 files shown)


cores/esp32/WString.cpp
```diff
@@ -180,15 +180,15 @@ bool String::changeBuffer(unsigned int maxStrLen) {
   if (maxStrLen < sizeof(sso.buff) - 1) {
     if (isSSO() || !buffer()) {
       // Already using SSO, nothing to do
-      uint16_t oldLen = len();
+      size_t oldLen = len();
       setSSO(true);
       setLen(oldLen);
```
lucasssvaz, line 185: `setLen` receives an `int` as argument. `size_t` might overflow it as it is bigger than that

MattiasTF: `len` returns `ptr.len`, which is an `uint32_t`, as an `unsigned int`. `setLen` assigns its `int` argument to `ptr.len`, so back to where it came from. Therefore, no overflow is possible in this case and the only issue is `setLen` pretending to accept an `int` in the first place.

`newSize` was already a `size_t` before, and it gets passed to `setCapacity`, which also takes an `int` argument and then assigns it to an `uint32_t`.

The whole String class is a mess that uses integer types inconsistently all the time. Rewriting the whole String class in an integer type-safe way is beyond the scope of this bugfix.

me-no-dev (approved):

Issue #11261, opened 2025-04-19
Support ESP-Now Version 2 packet size


carstenkoester:
### Related area

WiFi, ESP-Now

### Hardware specification

No specific hardware

### Is your feature request related to a problem?

As per the [ESP-NOW User Guide](https://github.com/espressif/esp-now/blob/master/User_Guide.md) and [Programming guide](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_now.html), ESP-NOW supports Version v2.0 with an increased maximum packet length.

However, the current Arduino ESP-NOW library still truncates the packet at 250 bytes - ref. https://github.com/espressif/arduino-esp32/blob/master/libraries/ESP_NOW/src/ESP32_NOW.cpp#L389-L391 and other places in that file.

### Describe the solution you'd like

It would be useful to expose the increased packet size through the library, to allow users of the ESP-Now library to leverage the functionality of sending large packets.

Based on the documentation, it looks like ESP-NOW v1.0 vs v2.0 is merely a function of ESP-IDF versions. ie. older versions of ESP-IDF support version 1 packet size and newer versions of ESP-IDF support version 2 packet size, with no change in API and no hardware dependency (as in, one doesn't need to run a specific board to benefit from the ESP-NOW v2.0)? If that is the case, then it *should* be safe to simply replace references to `ESP_NOW_MAX_DATA_LEN` with `ESP_NOW_MAX_DATA_LEN_V2` in these files.

I've tested this with the fork at https://github.com/espressif/arduino-esp32/compare/master...carstenkoester:arduino-esp32:espnow-v2-packet-size?expand=1 and at least in a basic test, was able to to transmit a 1000-byte ESP-NOW packet from an ESP32S3 dev board to an (older) Adafruit Huzzah32, with the change strictly required *on the transmitter side only*. I haven't been able to test other variations eg. ESP_NOW_Serial though. 

### I have checked existing list of Feature requests and the Contribution Guide

- [x] I confirm I have checked existing list of Feature requests and Contribution Guide.

PR #11524, opened 2025-06-27
feat(esp_now): Add support for ESP NOW V2

lucasssvaz:
## Description of Change

Add support for ESP-NOW v2.0 data sizes.

## Tests scenarios

Tested locally

## Related links

Closes [#11261](https://github.com/espressif/arduino-esp32/issues/11261)

### DIFF
--- a/libraries/ESP_NOW/src/ESP32_NOW.cpp
+++ b/libraries/ESP_NOW/src/ESP32_NOW.cpp
@@ -140,7 +140,10 @@ static void _esp_now_tx_cb(const uint8_t *mac_addr, esp_now_send_status_t status
   }
 }
 
-ESP_NOW_Class::ESP_NOW_Class() {}
+ESP_NOW_Class::ESP_NOW_Class() {
+  max_data_len = 0;
+  version = 0;
+}
 
 ESP_NOW_Class::~ESP_NOW_Class() {}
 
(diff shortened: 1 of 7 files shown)


P-R-O-C-H-Y (approved):
LGTM

libraries/ESP_NOW/src/ESP32_NOW.cpp
```diff
@@ -238,16 +258,38 @@ int ESP_NOW_Class::getEncryptedPeerCount() {
   return num.encrypt_num;
 }
 
+int ESP_NOW_Class::getMaxDataLen() const {
+  if (max_data_len == 0) {
+    log_e("ESP-NOW not initialized. Please call begin() first to get the max data length.");
+    return -ESP_ERR_ESPNOW_NOT_INIT;
```
lucasssvaz:
```suggestion
    return -1;
```

libraries/ESP_NOW/src/ESP32_NOW.cpp
```diff
@@ -238,16 +258,38 @@ int ESP_NOW_Class::getEncryptedPeerCount() {
...
+  }
+
+  return max_data_len;
+}
+
+int ESP_NOW_Class::getVersion() const {
+  if (version == 0) {
+    log_e("ESP-NOW not initialized. Please call begin() first to get the version.");
+    return -ESP_ERR_ESPNOW_NOT_INIT;
```
lucasssvaz:
```suggestion
    return -1;
```

libraries/ESP_NOW/src/ESP32_NOW_Serial.cpp
```diff
@@ -70,8 +70,25 @@ bool ESP_NOW_Serial_Class::begin(unsigned long baud) {
...
   }
-  setRxBufferSize(1024);  //default if not preset
-  setTxBufferSize(1024);  //default if not preset
+
+  size_t buf_size = 0;
+  if (ESP_NOW.getVersion() == 2) {
+    // ESP-NOW v2.0 has a larger maximum data length, so we need to increase the buffer sizes
+    // to hold around 3-4 packets
+    buf_size = setRxBufferSize(4096);
+    buf_size &= setTxBufferSize(4096);
```
lucasssvaz, line 79: Maybe we should keep the 1024 buffer size but print a warning that the buffers might be too small.

lucasssvaz:
Also I couldn't find any information if the version depends only on the ESP-IDF version. If yes we can just replace everything with the new size.

libraries/ESP_NOW/src/ESP32_NOW_Serial.cpp
me-no-dev: It's OK to have a bigger buffer. Let's not bother users with unnecessary warnings

PR #11528, opened 2025-06-29
Update ZigbeeColorDimmableLight to clamp color hue and saturation to 0-254 (Fixes #11527)

### LINKED ISSUE #11527: Zigbee Color Dimmable Light - Invalid Saturation
### Board

XIAO ESP32-C6

### Device Description

Raw board, disconnected from everything

### Hardware Configuration

Raw board, disconnected from everything

### Version

latest stable Release (if not listed below)

### IDE Name

VSCode + Platformio

### Operating System

NA

### Flash frequency

NA

### PSRAM enabled

yes

### Upload speed

115200

### Description

When I initialize a Zigbee color dimmable light in code to have a color value of RGB(4, 255, 0), I get an error message when the ZigbeeColorDimmableLight code tries to set the color:

```
Bus 1: Light change - State: OFF, RGB: (4,255,0), Level: 99
[  6303][E][ZigbeeColorDimmableLight.cpp:180] setLight(): Failed to set light saturation: 0x87: Invalid value
```

Doing some digging, what it appears is happening is that the function to [convert RGB color to HSV color](https://github.com/espressif/arduino-esp32/blob/21640ac82a1bb5efa8cf0b3841be1ac80add6785/cores/esp32/ColorFormat.c#L92-L119) is called as part of `setLight`, and the results are then passed - unadjusted - to `esp_zb_zcl_set_attribute_val`. The issue is that `espRgbColorToHsvColor` can return a saturation value from 0-255, but the valid range for Zigbee is 0-254.

### Sketch

```cpp
Although I have not tested this as I don't have the Arduino IDE, I believe this example should cause the error due to the addition of Line 118 to the stock example:

https://gist.github.com/thorrak/c3e957ed79313b80655e2ab666d662c5

Line 118 is the only line that differs from the stock example.
```

### Debug Message

```plain
[  6303][E][ZigbeeColorDimmableLight.cpp:180] setLight(): Failed to set light saturation: 0x87: Invalid value
```

### Other Steps to Reproduce

_No response_

thorrak:
## Description of Change
The Zigbee standard requires color saturation in the range of 0-254, but `espRgbColorToHsvColor` returns saturations in the range of 0-255. As referenced in #11527 when a full saturation color is set, the following error is received:

```
[  6303][E][ZigbeeColorDimmableLight.cpp:180] setLight(): Failed to set light saturation: 0x87: Invalid value
```

This PR clamps the range of the saturation used in `setLight()` to 0-254 (reducing it to 254 from 255 when full saturation is set), eliminating the "Invalid value" error.

## Tests scenarios
I have tested this change on my XIAO ESP32-C6 

## Related links
Fixes #11527

### DIFF
--- a/libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp
+++ b/libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp
@@ -127,7 +127,8 @@ bool ZigbeeColorDimmableLight::setLight(bool state, uint8_t level, uint8_t red, 
 
   espXyColor_t xy_color = espRgbColorToXYColor(_current_color);
   espHsvColor_t hsv_color = espRgbColorToHsvColor(_current_color);
-  uint8_t hue = (uint8_t)hsv_color.h;
+  uint8_t hue = (uint8_t)hsv_color.h; // Recast from uint16 to uint8
+  hue = (hue > 254) ? 254 : hue;  // Clamp to 0-254 (per the Zigbee standard)
   uint8_t saturation = (hsv_color.s > 254) ? 254 : (uint8_t)hsv_color.s;  // Clamp to 0-254 (per the Zigbee standard)
 
   log_v("Updating light state: %d, level: %d, color: %d, %d, %d", state, level, red, green, blue);
--- a/libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp
+++ b/libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp
@@ -128,6 +128,7 @@ bool ZigbeeColorDimmableLight::setLight(bool state, uint8_t level, uint8_t red, 
   espXyColor_t xy_color = espRgbColorToXYColor(_current_color);
   espHsvColor_t hsv_color = espRgbColorToHsvColor(_current_color);
   uint8_t hue = (uint8_t)hsv_color.h;
+  uint8_t saturation = (hsv_color.s > 254) ? 254 : (uint8_t)hsv_color.s;  // Clamp to 0-254 (per the Zigbee standard)
 
   log_v("Updating light state: %d, level: %d, color: %d, %d, %d", state, level, red, green, blue);
   /* Update light clusters */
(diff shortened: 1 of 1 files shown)


thorrak:
One downside to this fix (and with the Zigbee standard in general) is that there is necessarily a slight loss of information when mapping from a 0-255 saturation range to a 0-254 saturation range. This means that the color value known to Zigbee will always differ from any cached value set in code for colors with 255 saturation.

**For example** - if I set a color of #FF0000 in code (0H, 255S, 255V), it will get converted to #FF0101 (0H, 254S, 255V).

In _practice_, this may not be as much of an issue, as any color received via Zigbee will already have a maximum saturation value of 254 per the standard, but any color set from code will not be similarly limited.

I have no idea how to handle this other than potentially to log a warning to the console, but in my PR this error will just silently occur.

thorrak:
On a hunch I tested what happens when the hue was set to something that mapped to 255, and - sure enough - saw similar behavior:

```
[ 57201][E][ZBCDL.cpp:178] setLight(): Failed to set light hue: 0x87: Invalid value
```

I went ahead and applied the same fix for hue as I did for saturation. 

I expect the same issue here as for saturation, unfortunately - any colors that would map to a hue of 255 (if set directly within code) will end up being clamped to a hue of 254. As this is a limitation of the Zigbee standard, I don't know if there is any way to correct for this, however.

libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp
```diff
@@ -127,7 +127,9 @@ bool ZigbeeColorDimmableLight::setLight(bool state, uint8_t level, uint8_t red,
 
   espXyColor_t xy_color = espRgbColorToXYColor(_current_color);
   espHsvColor_t hsv_color = espRgbColorToHsvColor(_current_color);
-  uint8_t hue = (uint8_t)hsv_color.h;
+  uint8_t hue = (uint8_t)hsv_color.h; // Recast from uint16 to uint8
+  hue = (hue > 254) ? 254 : hue;  // Clamp to 0-254 (per the Zigbee standard)
+  uint8_t saturation = (hsv_color.s > 254) ? 254 : (uint8_t)hsv_color.s;  // Clamp to 0-254 (per the Zigbee standard)
```
lucasssvaz: IMO it would be better to use `std::min` to be more intuitive when reading the code

thorrak: Works for me. Updated.

P-R-O-C-H-Y:
@thorrak This is what cluster specification says: 
Screenshot 2025-06-30 at 12 03 31

thorrak:
> @thorrak This is what cluster specification says

Is your preference then to scale rather than to truncate? `Hue254 = (uint8/t) (hue255 / 255 * 254)`?

Given we’re dropping one value of 256, I would expect this would just have the effect of doing the following (but with floating point math):

`Hue254 = (Hue255>=128) ? (Hue255-1) : Hue255`

thorrak:
If either of those are the way you would want to go (compression rather than truncation) I’m guessing we would probably need to expand the value received from the Zigbee network as well. My guess - having not reviewed the code for values received - is that the 0-254 values received from the network aredirectly mapped to the 0-255 space used by the ColorFormat library on the ESP rather than being expanded. 

I have not reviewed any of the code dealing with received values however.

P-R-O-C-H-Y:
I will check and let you know what solution would be the best.

lucasssvaz (approved):

PR #11530, opened 2025-06-30
fix(csrf): Fix SCRF vulnerability in OTA examples and libraries

me-no-dev:
Closes: https://github.com/espressif/arduino-esp32/security/advisories/GHSA-9vfw-wx65-c872

### DIFF
--- a/libraries/Update/examples/OTAWebUpdater/OTAWebUpdater.ino
+++ b/libraries/Update/examples/OTAWebUpdater/OTAWebUpdater.ino
@@ -7,10 +7,15 @@
 
 #define SSID_FORMAT "ESP32-%06lX"  // 12 chars total
 //#define PASSWORD "test123456"    // generate if remarked
+const char * authUser = "admin";
+const char * authPass = "admin";
 
 WebServer server(80);
 Ticker tkSecond;
 uint8_t otaDone = 0;
+
+const char * csrfHeaders[2] = {"Origin", "Host"};
+static bool authenticated = false;
 
 const char *alphanum = "0123456789!@#$%^&*abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 String generatePass(uint8_t str_len) {
@@ -38,6 +43,9 @@ void apMode() {
 }
 
 void handleUpdateEnd() {
+  if (!authenticated) {
+    return server.requestAuthentication();
+  }
   server.sendHeader("Connection", "close");
   if (Update.hasError()) {
     server.send(502, "text/plain", Update.errorString());
@@ -45,6 +53,7 @@ void handleUpdateEnd() {
     server.sendHeader("Refresh", "10");
     server.sendHeader("Location", "/");
     server.send(307);
+    delay(500);
     ESP.restart();
   }
 }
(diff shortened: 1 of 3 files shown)


P-R-O-C-H-Y (approved):

me-no-dev:
@JLLeitschuh PTAL

libraries/Update/examples/OTAWebUpdater/OTAWebUpdater.ino
```diff
@@ -7,11 +7,16 @@
 
 #define SSID_FORMAT "ESP32-%06lX"  // 12 chars total
 //#define PASSWORD "test123456"    // generate if remarked
+const char * authUser = "admin";
+const char * authPass = "admin";
```
JLLeitschuh: I'd suggest generating a random password that you print to the console to let the user know what it is over using a hard-coded password.

libraries/WebServer/examples/WebUpdate/WebUpdate.ino
```diff
@@ -11,11 +11,16 @@
 const char *host = "esp32-webupdate";
 const char *ssid = "........";
 const char *password = "........";
+const char * authUser = "admin";
+const char * authPass = "admin";
```
JLLeitschuh: Same feedback here. Generate a random password, and serial print it. I would suggest avoiding having this hard-coded in this example

JLLeitschuh (commented):
This will work, but I'd suggest implementing this as a standard middleware implementation over this solution. That way the logic is reusable. But that's my ten cents. One concern around the use of the `Host` and `Origin` header is that this won't work behind middleware proxies.

That concern is raised here:

> Use the Host header value: If you want your application to find its own target so it doesn't have to be configured for each deployed instance, we recommend using the Host family of headers. The Host header is meant to contain the target origin of the request. **_But, if your app server is sitting behind a proxy, the Host header value is most likely changed by the proxy to the target origin of the URL behind the proxy, which is different than the original URL. This modified Host header origin won't match the source origin in the original Origin or Referer headers._**

https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html?utm_source=chatgpt.com#identifying-the-target-origin

I've more frequently seen CSRF protection implemented via a CSRF token in the form post body, but this will likely also work.

libraries/WebServer/examples/WebUpdate/WebUpdate.ino
me-no-dev: I think you are missing the point that this is an Arduino basic example, whose purpose is not to secure every possible scenario, but instead to showcase a particular use (the Updater class). No sane person will use `admin:admin` to secure anything. I can put comment to change the default values or define the user/pass as dots (like the SSID and Pass for WiFi), but anything beyond that is over-complication of otherwise basic example that needs to be readable and easy to understand by novice users. Same goes for uses behind proxies, sessions, etc things that are beyond the scope of the examples and whose implementation is not straight-forward on memory constrained devices. Surely feel free to to propose changes in form of Pull Request that will satisfy your requirements :)

Using `Host` and `Origin` was stated to be enough of a protection by the links in your initial report. Session tokens are not an option in our case at all.

PR #11552, opened 2025-07-04
fix(board): Update PSRAM configuration for RAK3112 to fix PSRAM error

DanielCao0:
fix : quad_psram: psram id read error: 0x00ffffff, psram chip not found or not supported, or wrong psram line mode arduino

Configuration updates for `rakwireless_rak3112`:
* Added a new `psram_type` property with the value `opi` to specify the type of PSRAM used.
* Introduced a `memory_type` property that dynamically combines `build.boot` and `build.psram_type` for better memory type management.
* Updated the `menu.PSRAM.enabled` description to "OPI PSRAM" for clarity.
* Add wisblock pin define

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -50809,8 +50809,10 @@ rakwireless_rak3112.build.flash_mode=dio
 rakwireless_rak3112.build.boot=dio
 rakwireless_rak3112.build.partitions=default
 rakwireless_rak3112.build.defines=
+rakwireless_rak3112.build.psram_type=opi
+rakwireless_rak3112.build.memory_type={build.boot}_{build.psram_type}
 
-rakwireless_rak3112.menu.PSRAM.enabled=Enabled
+rakwireless_rak3112.menu.PSRAM.enabled=OPI PSRAM
 rakwireless_rak3112.menu.PSRAM.enabled.build.defines=-DBOARD_HAS_PSRAM
 rakwireless_rak3112.menu.PSRAM.enabled.build.psram_type=opi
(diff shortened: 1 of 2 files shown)


boards.txt
```diff
@@ -50809,8 +50809,10 @@ rakwireless_rak3112.build.flash_mode=dio
 rakwireless_rak3112.build.boot=dio
```
me-no-dev, line 50809: boot can probably be `qio` here for faster flash access

boards.txt
```diff
@@ -50809,8 +50809,10 @@ rakwireless_rak3112.build.flash_mode=dio
...
 rakwireless_rak3112.build.partitions=default
 rakwireless_rak3112.build.defines=
+rakwireless_rak3112.build.psram_type=opi
+rakwireless_rak3112.build.memory_type={build.boot}_{build.psram_type}
 
-rakwireless_rak3112.menu.PSRAM.enabled=Enabled
+rakwireless_rak3112.menu.PSRAM.enabled=OPI PSRAM
```
me-no-dev, line 50815: I think that `Enabled` here makes more sense, given that it does not have QSPI option

DanielCao0: Thank you for your suggestion. It has been updated "Enabled"

boards.txt
DanielCao0: rakwireless_rak3112.build.flash_mode=qio
rakwireless_rak3112.build.boot=qio   
Has been modified

me-no-dev: only `boot` should be `qio`. Flash mode should stay `dio`

PR #11581, opened 2025-07-10
feat(matter): removing wifi requirement for H2 and C5

SuGlider:
## Description of Change
Arduino Matter 3.3.x Library runs over WiFi or Thread depending on the target SoC.
CI JSON shall be adjusted to test Thread Matter by removing the WiFi requirement.

Removes WiFi Provisioning examples given that it doesn't make any sense because Matter uses BLE for network provisioning.
Updates Library Key Words.

## Tests scenarios
CI Only

## Related links
None

### DIFF
--- a/libraries/Matter/examples/MatterLambdaSingleCallbackManyEPs/MatterLambdaSingleCallbackManyEPs.ino
+++ b/libraries/Matter/examples/MatterLambdaSingleCallbackManyEPs/MatterLambdaSingleCallbackManyEPs.ino
@@ -46,16 +46,11 @@ const uint8_t MAX_LIGHT_NUMBER = 6;
 MatterOnOffLight OnOffLight[MAX_LIGHT_NUMBER];
 
 // all pins, one for each on-off light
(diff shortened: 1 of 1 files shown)


libraries/Matter/examples/MatterLambdaSingleCallbackManyEPs/MatterLambdaSingleCallbackManyEPs.ino
```diff
@@ -0,0 +1,128 @@
...
+    // waits for Matter Plugin Commissioning.
+    uint32_t timeCount = 0;
+    while (!Matter.isDeviceCommissioned()) {
+      delay(100);
+      if ((timeCount++ % 50) == 0) {  // 50*100ms = 5 sec
+        Serial.println("Matter Node not commissioned yet. Waiting for commissioning.");
+      }
+    }
+  } else {
+    if (!Matter.isDeviceConnected()) {
```
SuGlider:
```suggestion
    if (Matter.isDeviceConnected()) {
```

PR #11608, opened 2025-07-18
Add senseBox Eye board

PaulaScharf:
## Description of Change
Adding the senseBox Eye board.

## Tests scenarios
I have tested this pull request on the senseBox Eye board.

## Related links
- [schematics](https://github.com/sensebox/senseBox_Eye/blob/main/hardware/schematics/v1.2/senseBox-Eye_v1.2.pdf)
- [sensebox.de](https://sensebox.de/)


boards.txt
lucasssvaz, line 1: You are deleting other users' boards. Please rebase your branch

P-R-O-C-H-Y: @PaulaScharf please fix this

PaulaScharf: sorry about that. I hope its good now

variants/sensebox_eye/variant.cpp
```diff
@@ -0,0 +1,50 @@
+#include "esp32-hal-gpio.h"
+#include "pins_arduino.h"
+#include "driver/rmt_tx.h"
+
+extern "C" {
+
+void blinkLED(uint8_t color[3], rmt_channel_handle_t led_chan, rmt_encoder_handle_t ws2812_encoder, rmt_transmit_config_t tx_config) {
```
me-no-dev: Please use Arduino APIs for RMT and do not use `ESP_ERROR_CHECK` because it will `abort` in case of error and bootloop the device. We have ready-to-use APIs for controlling ws2812 (even as LED_BUILTIN). Have a look at those

PaulaScharf:
thank you, that makes the variant.cpp much slimmer :)

PR #11624, opened 2025-07-22
fix(zigbee): Fix RGB color calculation

### LINKED ISSUE #11619: [Bug] Color conversion in espXYToRgbColor() too dark – XYZ values scaled down by /100
The current implementation of espXYToRgbColor() divides the already-normalized XYZ values by 100.0f before the XYZ→sRGB matrix multiplication.
This scales X, Y, Z down by a factor 100 and therefore makes all RGB output ≈100× too dark.

```
// current code (snippet)
X = X / 100.0f;   // <-- should be removed
Y = Y / 100.0f;   // <-- should be removed
Z = Z / 100.0f;   // <-- should be removed
r = (X * 3.2406f) - (Y * 1.5372f) - (Z * 0.4986f);
...
```

https://github.com/espressif/arduino-esp32/blob/6015fd73e0d9d065c80fc79ef0f35d3273a217de/libraries/Zigbee/src/ep/ZigbeeColorDimmableLight.cpp#L86C24-L86C39

P-R-O-C-H-Y:
## Description of Change
This pull request introduces changes to improve color conversion functionality and refactors related code for clarity and flexibility. The most significant updates include adding an optional scaling parameter to the `espXYToRgbColor` function, modifying its usage across the codebase, and updating the header file to reflect these changes.

Before:
RED color on max level (100%) - RGB(68,0,0)

After fix:
RED color on max level (100%) - RGB(255,0,0)

## Tests scenarios
Tested using ColorDimmableLight example

## Related links
Closes #11619

### DIFF
--- a/cores/esp32/ColorFormat.h
+++ b/cores/esp32/ColorFormat.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <stdint.h>
+#include <stdbool.h>
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -49,7 +50,7 @@ typedef struct HsvColor_t espHsvColor_t;
 typedef struct XyColor_t espXyColor_t;
 typedef struct CtColor_t espCtColor_t;
 
-espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y);
+espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y, bool addXYZScaling);
 espRgbColor_t espXYColorToRgb(uint8_t Level, espXyColor_t xy);
 espXyColor_t espRgbColorToXYColor(espRgbColor_t rgb);
 espXyColor_t espRgbToXYColor(uint8_t r, uint8_t g, uint8_t b);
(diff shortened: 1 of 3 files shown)


cores/esp32/ColorFormat.h
```diff
@@ -49,7 +50,7 @@ typedef struct HsvColor_t espHsvColor_t;
 typedef struct XyColor_t espXyColor_t;
 typedef struct CtColor_t espCtColor_t;
 
-espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y);
+espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y, bool addXYZScaling);
```
lucasssvaz, line 53: Just to avoid breaking changes:

```suggestion
espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y, bool addXYZScaling = false);
```

P-R-O-C-H-Y: You can't do this in C, its a C++ feature.

lucasssvaz: You are right, my bad. Then we can do something like this:

```suggestion
espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y, bool addXYZScaling);
espRgbColor_t espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y);
```

And then do:

``` c
espXYToRgbColor(uint8_t Level, uint16_t current_X, uint16_t current_Y) {
    espXYToRgbColor(Level, current_X, current_Y, false);
}
```

P-R-O-C-H-Y: Also a C++ feature :D. Lets keep it like this

me-no-dev (approved):

PR #11631, opened 2025-07-23
Update tinyuf2 0.35

tyeth:
## Description of Change
Updates the bootloader and tinyuf2 app bin files for the Adafruit boards.

Also fixes the csv name for the XIAO S3 tinyuf2 partition. It's got a partitions-8mb.csv but instead refers to a partitions-8mb-tinuf2.csv which is one of the tinyuf2 partition files that were renamed ages ago.
As the board appeared to be pointing to a non existent file, rather than point at the one in board folder this switches to the communal CSV (same content) and adds the 8MB No OTA partition scheme. Did the tinyuf2 bins for those xiao s3's too, did both after verifying same content between them.

## Tests scenarios
Grabbed latest release zips from tinyuf2, as our CI has been, built with the bootloader and tinyuf2 files updated (plus boot_app0.bin), tested normal and tinyuf2 functionality.

Tested both tinyuf2 partitions for XIAO S3 in arduino ide blink example, then checked partition scheme correctly written, and tinyuf2 bootloader still works as expected.


tyeth:
Hey, if we want this partition scheme update in the next release plus in older version releases like 3.1.x and 3.2.x and 3.3.x then what's the correct way to go about it?

tyeth:
DangerJS failure: [Checks are being rate limited at GH](https://github.com/espressif/arduino-esp32/actions/runs/16478227559/job/46585424104?pr=11631#step:4:212), so CI needs kicking

tyeth:
~~Drafting this to also add some verbiage updates for the partition schemes.~~
EDIT: Will do partition label updates in another PR if needed

me-no-dev:
@P-R-O-C-H-Y PTAL

PR #11640, opened 2025-07-25
feat(github): Update template type

Parsaabasi:
## Description of Change
To include and make the "Type" of the issue mandatory.
This way the users would need to select the type of the issue. This will allow the bot to perform more accurately in managing the backlog as well as the engineers. 

This change was discussed with the team and agreed.


pedrominatel (approved):
LGTM.

PR #11687, opened 2025-08-02
feat(modem): execution of AT command in raw mode

fobos777:
## Description of Change

Regular `cmd` function does not return the entire reply of the given AT command that is problematic for some use cases, e.g., quering GNSS data in CMUX mode. So, lets introduce `cmd_raw` call that, assuming correct CMUX behaviour, returns entire reply of the given AT command.

## Tests scenarios

Tested with `Waveshare ESP32-S3 SIM7670G 4G Development Board` using B05 modem firmware.

### DIFF
--- a/libraries/PPP/src/PPP.cpp
+++ b/libraries/PPP/src/PPP.cpp
@@ -747,6 +747,18 @@ String PPPClass::cmd(const char *at_command, int timeout) {
   return String(out);
 }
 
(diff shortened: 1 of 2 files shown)


me-no-dev:
Ho @fobos777 please sign the CLA

me-no-dev:
@fobos777 closing for lack of response. Feel free to sign CLA and reopen the pull request

PR #11768, opened 2025-08-28
feat(HTTPClient): add support for collecting all HTTP headers

### LINKED ISSUE #10802: HTTPClient code does not collect header names
### Related area

WiFi

### Hardware specification

ESP32

I finally got header fetch working.  I find a problem with it.

Firstly, this feature should have an example.  I can provide one since I finally got it working on my fourth attempt.

Second:
1. One has to know the name of the header(s) and list them in the C code to be able to get them.
2. The http.headers() call does not report the actual number of headers in the returned data, it simply reports the number of names you entered into the headerKeys array.  I already know that number, I had to enter them.  It should return the number of headers that are available so they can be iterated through.
3. The author uses size_t and the current Arduino compiler release has a problem with that definition.  I forgot the error message but it is repeatable, for the most part.  Sometimes it care, sometimes it does not care.  It would be easier on us programmers if you used a long integer or even just integer.  There will not be over 2 billion headers in one file.  size_t is surely overkill and an obscure variable type that confuses the compiler.
4.  headerKeysCount already has the same number that http.headers() returns.  Redundant and, therefore, useless.  This needs correction to return the actual count of found headers.  In the data I receive, there are 4 headers but .headers returns 1 because I listed only 1 in the array.
5. Let Firefox be your guide.  It shows all four returned headers.  That's how I verified they are there and verified the name of the header I needed.

Here is the code I ended up with (after 3 wrong/incomplete AI responses and then more in-depth digging through web pages):

  ```
const char *headerKeys[] = {"x-ratelimit-remaining-month"};  // The only header I care about.
  const size_t headerKeysCount = sizeof(headerKeys) / sizeof(headerKeys[0]);  // Returns 1
  http.collectHeaders(headerKeys, headerKeysCount);

  if (iHttpResponseCode == 200) {  // 200 is goodness!

      // Print all headers
      int headerCount = http.headers();  // Simply a count of elements of the headerKeys array.  Returns 1
      //      Serial.printf("\r\n%i HTTP header(s) sought:\r\n", headerCount);
      for (int i = 0; i < headerCount; i++) {
        //        Serial.println(http.header((size_t) 0));  // size_t is required to be coded!
        sTemp = http.header((size_t) 0);  // size_t is required to be coded!
        sTemp = sTemp.substring(0, sTemp.lastIndexOf(":"));
        //        Serial.printf("%s: %s gave %s\r\n",
        //                      http.headerName(i).c_str(),
        //                      http.header(i).c_str(), sTemp);
        Serial.printf("There are %s /*%i*/ XRate calls left for this key"
                      " this month.\r\n", sTemp/*, sTemp.toInt()*/);
  }
```

See above.  

Return all headers you find.
Return the count of all headers you find for iteration through them.
Don't require the user to specify the header name in advance.  Collect all of them.
Create a working example. (I have one if you want to start from there)

None.  There are no alternatives.  I use HTTPClient.

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by JakubAndrysek**
## Description of Change
This pull request refactors the HTTP header collection logic in the `HTTPClient` library to use a `std::vector` for managing response headers, adds support for collecting all headers, and updates the example and CI configuration accordingly. The changes simplify memory management, improve flexibility, and make it easier to work with dynamic sets of headers.

## Test Scenarios
Code was tested on ESP32 and ESP32S3

## Related links
Closes issue  #10802 … example, close #10802

### CODE REVIEW

File: libraries/HTTPClient/examples/CustomHeaders/CustomHeaders.ino [hunk 1]
```diff
@@ -0,0 +1,86 @@
+#include <Arduino.h>
+
+#include <WiFi.h>
+#include <assert.h>
+#include <HTTPClient.h>
+
+#define USE_SERIAL Serial
```
me-no-dev: why is this needed? Please remove

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -90,9 +90,6 @@ HTTPClient::~HTTPClient() {
   if (_client) {
     _client->stop();
   }
-  if (_currentHeaders) {
-    delete[] _currentHeaders;
-  }
```
me-no-dev, line 95: this is the destructor, so how will you ensure that all memory is cleared if you remove the lines?

libraries/HTTPClient/examples/CustomHeaders/CustomHeaders.ino
JakubAndrysek: I have created this example based on this [BasicHttpClient](https://github.com/espressif/arduino-esp32/blob/master/libraries/HTTPClient/examples/BasicHttpClient/BasicHttpClient.ino#L15), which uses the same macro. Now I have looked at the implementation of [next HTTPClient examples](https://github.com/search?q=repo%3Aespressif%2Farduino-esp32%20USE_SERIAL&type=code) and some of them use it as well.
Do you want me to remove it in the next PR?

libraries/HTTPClient/src/HTTPClient.cpp
JakubAndrysek: I think the _currentHeaders.clear() call in the destructor is unnecessary. Since [_currentHeaders](https://github.com/espressif/arduino-esp32/pull/11768/files#diff-39772ac72f68278f77f2b5e82dd9c3fcd3b188f532d4fe70ba0f7ece4f7cd163R309) is a std::vector containing objects with automatic destructors, C++'s RAII principle guarantees that all memory will be automatically freed when the HTTPClient object is destroyed, making manual cleanup redundant.
Or do I have to convert it to a unique_ptr with a vector inside like [_tcpDeprecated](https://github.com/espressif/arduino-esp32/blob/master/libraries/HTTPClient/src/HTTPClient.h#L284)?

libraries/HTTPClient/examples/CustomHeaders/CustomHeaders.ino
me-no-dev: yes please

libraries/HTTPClient/src/HTTPClient.h
```diff
@@ -304,8 +306,7 @@ class HTTPClient {
   String _acceptEncoding = "identity;q=1,chunked;q=0.1,*;q=0";
 
   /// Response handling
-  RequestArgument *_currentHeaders = nullptr;
-  size_t _headerKeysCount = 0;
+  std::vector<RequestArgument> _currentHeaders;
```
Jason2866, line 309: Why, vector probably needs more bytes.

JakubAndrysek: You are right, the vector probably needs more bytes, but since I do not know the exact size and I am reading until the `_client` has some data, it is a lot easier to just use a vector and let it grow as needed
The previous implementation used only a fixed-size array, but after the update, I need to account for dynamic resizing to allow storing multiple headers. It is possible to do it with a resize operation of a static vector, but I think it is more convenient to use a dynamic vector for this purpose.

Jason2866: Fully understand, over time every part of the core does need more resources. One day we will have a core which can not be used for anything more than just "blink".

SuGlider: Blame says that `#include <vector>` was added 3 years ago to HTTPClient.h for supporting Coockies.
I see no issue here.

P-R-O-C-H-Y (approved):

me-no-dev (approved):

PR #11830, opened 2025-09-14
Update esp32-hal-misc.c

arersen:
# Fix millis() returning nanoseconds instead of milliseconds

## Description of Change
Fixed the `millis()` implementation.  
Previously, it incorrectly returned values in nanoseconds due to dividing `esp_timer_get_time()` (microseconds) by `1000ULL`.  
Now the function multiplies by `1000ULL`, so it properly returns elapsed time in milliseconds.  

## Test Scenarios
- Tested on **Arduino-esp32 core v3.0.0**  
- Hardware: ** LILYGO ESP32** 
- Verified that `millis()` increments correctly every 1 ms compared to `micros()`.

### DIFF
--- a/cores/esp32/esp32-hal-misc.c
+++ b/cores/esp32/esp32-hal-misc.c
@@ -206,7 +206,7 @@ unsigned long ARDUINO_ISR_ATTR micros() {
 }
 
 unsigned long ARDUINO_ISR_ATTR millis() {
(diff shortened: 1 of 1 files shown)


Jason2866:
One tick is a micro second, to get the time for one tick in milliseconds it needs to be divided by 1000.
The current implementation is correct.

PR #11848, opened 2025-09-16
feat(board): add support for Silicognition ManT1S

xorbit:
## Description of Change

Add new board Silicognition ManT1S: https://www.crowdsupply.com/silicognition/mant1s
- Add board configuration to `boards.txt`.
- Add pin definitions and T1S Ethernet config  to `variants/mant1s`.

mant1s-angle-01

## Test Scenarios

This requires LAN867x support so depends on #11843, which is not in any release.  So it was tested on `master` on a prototype ManT1S rev 4.  Works as expected with unmodified ETH_LAN8720 example, since `pins_arduino.h` defines `ETH_PHY_MDC` correctly for the ManT1S the config block on top of the example is ignored and uses the correct Ethernet definitions from `pins_arduino.h` instead.

## Related links

Depends on #11843 and library built with the LAN867x supports this adds.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -29578,6 +29578,96 @@ wesp32.menu.EraseFlash.all.upload.erase_cmd=-e
 
 ##############################################################
 
+mant1s.name=Silicognition ManT1S
+
+mant1s.bootloader.tool=esptool_py
+mant1s.bootloader.tool.default=esptool_py
+
+mant1s.upload.tool=esptool_py
+mant1s.upload.tool.default=esptool_py
+mant1s.upload.tool.network=esp_ota
+
+mant1s.upload.maximum_size=1310720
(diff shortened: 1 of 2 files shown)


P-R-O-C-H-Y:
Hi @xorbitm, if this board is unrelated to existing ones in boards.txt file (which I guess) please place the board at the end of boards.txt file.

xorbit:
I received a comment about the placement in `boards.txt` in my email but I can't see it here.  This is a Silicognition board, so I put it right below the other Silicognition board, the wESP32.

P-R-O-C-H-Y:
@xorbit All good, I have removed the comment, as the board above is from Silicognition. Missed that first time. Its ready to be merged ;)

PR #11854, opened 2025-09-18
fix(hash): add selective build option for Hash, WebServer dependency

fill1890:
The Hash library introduced in version v3.3.1 doesn't have an option to be included in a selective build, and in a selective build it defaults to not included. As the WebServer library depends on the Hash library, selective builds that include WebServer break as the dependency is not satisfied.

This PR adds an appropriate option to Kconfig, and adds the WebServer → Hash dependency.

Tested on a selective build using the Arduino core as an ESP-IDF component on an ESP32-S3. As far as I know the aren't any open issues for this.

### DIFF
--- a/Kconfig.projbuild
+++ b/Kconfig.projbuild
@@ -341,6 +341,11 @@ config ARDUINO_SELECTIVE_PPP
     depends on ARDUINO_SELECTIVE_COMPILATION
     default y
 
+config ARDUINO_SELECTIVE_Hash
+    bool "Enable Hash"
+    depends on ARDUINO_SELECTIVE_COMPILATION
+    default y
+
 config ARDUINO_SELECTIVE_ArduinoOTA
     bool "Enable ArduinoOTA"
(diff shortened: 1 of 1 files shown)


me-no-dev:
@lucasssvaz PTAL

lucasssvaz (approved):

Kconfig.projbuild
```diff
@@ -341,6 +341,11 @@ config ARDUINO_SELECTIVE_PPP
...
     default y
 
+config ARDUINO_SELECTIVE_Hash
+    bool "Enable Hash"
+    depends on ARDUINO_SELECTIVE_COMPILATION
+    default y
+
 config ARDUINO_SELECTIVE_ArduinoOTA
     bool "Enable ArduinoOTA"
     depends on ARDUINO_SELECTIVE_COMPILATION && ARDUINO_SELECTIVE_Network
```
lucasssvaz, line 351:
```suggestion
    depends on ARDUINO_SELECTIVE_COMPILATION && ARDUINO_SELECTIVE_Network
    select ARDUINO_SELECTIVE_Hash
```

PR #11875, opened 2025-09-25
feat(release): Update version on release and add XZ archives

lucasssvaz:
## Description of Change

This pull request adds a few features to the release script:

- Automatically update version based on the release tag
  - If release is successful, push a commit with the version update and move the release tag to it.
- Package and publish the core as `.tar.xz` (in addition to `.zip`)
- Download, repack and publish libs as both `.zip` and `.tar.xz`

## Test Scenarios

Tested in fork:

https://github.com/lucasssvaz/arduino-esp32/actions/runs/18012905331/job/51250139963#step:6:1
https://github.com/lucasssvaz/arduino-esp32/releases/tag/9.9.9-RC1
https://github.com/lucasssvaz/arduino-esp32/commit/9c78e1bf723c30ac0b457c7a346b46513d30031e

### DIFF
--- a/.github/scripts/on-release.sh
+++ b/.github/scripts/on-release.sh
@@ -54,6 +54,30 @@ if [ -n "${VENDOR}" ]; then
     echo "Setting packager: $VENDOR"
 fi
 
+function update_version {
+    set -e
+    set -o pipefail
+
+    local tag=$1
+    local major
+    local minor
+    local patch
+
+    # Extract major, minor, and patch from the tag
+    # We need to make sure to remove the "v" prefix from the tag and any characters after the patch version
+    tag=$(echo "$tag" | sed 's/^v//g' | sed 's/-.*//g')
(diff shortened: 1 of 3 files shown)


.github/scripts/on-release.sh
```diff
@@ -293,22 +340,100 @@ if [ $? -ne 0 ]; then
...
+# Extract ZIP
+
+echo "Repacking libs to XZ ..."
+unzip -q "$OUTPUT_DIR/$LIBS_ZIP" -d "$OUTPUT_DIR"
+pushd "$OUTPUT_DIR" >/dev/null
+tar -cJf "$LIBS_XZ" "esp32-arduino-libs"
+popd >/dev/null
+
+# Upload ZIP and XZ libs to release page
+
```
me-no-dev, line 419: you should also update the json to use the copied zip libs

lucasssvaz: Done

PR #11879, opened 2025-09-27
fix(board): Fix wESP32 board config with new options

xorbit:
### Checklist

## Description of Change
- Provide flash size option to choose 4MB flash (rev 5 and below) or 16MB flash (default, rev 7 and higher).
- Provide appropriate partition scheme options to effectively use either 4MB or 16MB flash sizes.
- Change default Ethernet PHY from LAN8720 to RTL8201 (rev 7 and higher default).

## Test Scenarios
Tested on Arduino 2.3.6 locally with wESP32 rev 8 board.

### DIFF
--- a/variants/wesp32/pins_arduino.h
+++ b/variants/wesp32/pins_arduino.h
@@ -35,7 +35,7 @@ static const uint8_t T9 = 32;
 #define ETH_PHY_POWER -1
 #define ETH_PHY_MDC   16
 #define ETH_PHY_MDIO  17
-#define ETH_PHY_TYPE  ETH_PHY_LAN8720
+#define ETH_PHY_TYPE  ETH_PHY_RTL8201
 #define ETH_CLK_MODE  ETH_CLOCK_GPIO0_IN
 
 #endif /* Pins_Arduino_h */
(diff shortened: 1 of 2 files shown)


variants/wesp32/pins_arduino.h
```diff
@@ -35,7 +35,7 @@ static const uint8_t T9 = 32;
 #define ETH_PHY_POWER -1
 #define ETH_PHY_MDC   16
 #define ETH_PHY_MDIO  17
```
P-R-O-C-H-Y, line 37: What about having also `#define ETH_PHY_TYPE_REV4 ETH_PHY_LAN8720` in case someone uses the old revision to keep backward compatibility.

xorbit: In my examples, I have usually called out just setting all parameters:
https://wesp32.com/software/#arduino-esp32

With the change in this PR, users now could just use `ETH.begin()` for any wESP32 sold in the last 4 years.

The only revisions ever sold are rev 3, 4 and 5 with LAN8720, sold for about 3 years (2018 ~ 2021), and rev 7 and 8 with RTL8201, sold the last 4 years (2021 ~ present).  I'm not sure how to clearly communicate what another define would be for exactly.  It's not just revision 4, it's 3, 4 and 5 you'd use LAN8720 for.  I don't see how any new `#define` we'd come up with would be clearer than to just say "this board needs LAN8720" and use the `ETH_PHY_LAN8720` name directly.

P-R-O-C-H-Y (approved):

PR #11896, opened 2025-10-03
feat(uart): refactor PeriMan detaching function init

SuGlider:
## Description of Change
Removed uart_init_PeriMan function and related comments.
Peripehral Manager detaching initialization was creating multiple DGB messages everytime that a UART Serial port was created.
It will now log once and only for the pins that are really used.

## Test Scenarios
ESP32 - Basic UART Initilization - CI 

## Related links
None.

### DIFF
--- a/cores/esp32/esp32-hal-uart.c
+++ b/cores/esp32/esp32-hal-uart.c
@@ -479,6 +479,10 @@ static bool _uartAttachPins(uint8_t uart_num, int8_t rxPin, int8_t txPin, int8_t
       ret &= perimanSetPinBus(rxPin, ESP32_BUS_TYPE_UART_RX, (void *)uart, uart_num, -1);
       if (ret) {
         uart->_rxPin = rxPin;
+        // set Peripheral Manager deInit Callback for this UART pin
+        if (perimanGetBusDeinit(ESP32_BUS_TYPE_UART_RX) == NULL) {
+          perimanSetBusDeinit(ESP32_BUS_TYPE_UART_RX, _uartDetachBus_RX);
+        }
       }
     }
     if (!ret) {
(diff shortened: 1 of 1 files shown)


lucasssvaz (approved):

cores/esp32/HWCDC.cpp
```diff
@@ -324,14 +322,19 @@ void HWCDC::begin(unsigned long baud) {
 
   // Peripheral Manager setting for USB D+ D- pins
   uint8_t pin = USB_INT_PHY0_DM_GPIO_NUM;
+  if (perimanGetBusDeinit(ESP32_BUS_TYPE_USB_DM) == NULL) {
```
me-no-dev, line 325: is this check necessary? could the deinit callback be set to something else? If not, there is nothing wrong in just assigning it to the same thing

cores/esp32/esp32-hal-uart.c
```diff
@@ -479,6 +479,10 @@ static bool _uartAttachPins(uint8_t uart_num, int8_t rxPin, int8_t txPin, int8_t
       ret &= perimanSetPinBus(rxPin, ESP32_BUS_TYPE_UART_RX, (void *)uart, uart_num, -1);
       if (ret) {
         uart->_rxPin = rxPin;
+        // set Peripheral Manager deInit Callback for this UART pin
+        if (perimanGetBusDeinit(ESP32_BUS_TYPE_UART_RX) == NULL) {
```
me-no-dev, line 483: same question as in USB

cores/esp32/HWCDC.cpp
SuGlider: Given that this call is inside `begin()`, depending of the sketch it may be called more than one time along execution. This test prevents issuing not necessary VERBOSES log messages.

This is not about having or not already set the `deinit` function, but about preventing excessive log messages in the esp32-hal-periman.c | function `perimanSetBusDeinit()` | Line 235
`  log_v("Deinit function for type %s (%u) successfully set to %p", perimanGetTypeName(type), (unsigned int)type, cb);` 

When debugging using log messages, in my opinion, less messages is better. 
Let me know.

cores/esp32/esp32-hal-uart.c
SuGlider: The explanation is identical to what was stated earlier.

cores/esp32/HWCDC.cpp
me-no-dev: then I am OK with it

PR #11897, opened 2025-10-05
feat(board): expose logging subsystem arduino IDE

guppy0130:
## Description of Change

* introduce `Logging Subsystem` option in arduino IDE
  * introduces `arduino` and `esp_idf` options
* `esp_idf` option sets the `build.log_subsystem` variable to `-DUSE_ESP_IDF_LOG`, which is passed into `build.extra_flags`
* `arduino` option is an empty str and therefore no new flags are added to `build.extra_flags`
* cannot use `build.defines` as that conflicts with the PSRAM `-D` flag?

note: this only adds it for XIAO ESP32S3; I can add it to the rest of the boards if that's desirable, but hoping y'all might have some automation for that.

## Test Scenarios

Hardware: `FQBN: espressif:esp32:XIAO_ESP32S3:DebugLevel=debug,PSRAM=opi,LoggingSubsystem=ESP_IDF`
Software: whatever is pulled in by current `main`

Example sketch:

```c++
// define some custom logger fx
int syslogger(const char *format, va_list args) {
  char buffer[256];
  vsnprintf(buffer, sizeof(buffer), format, args);

  // some custom write (e.g., for me I wrote to a remote syslog endpoint)
  Serial.print(buffer);

  va_end(args);
  return strlen(buffer);
}

void setup() {
  esp_log_set_vprintf(syslogger);
}

void loop() {
  ESP_LOGD("my_app", "hello world");
}
```

Without this change, `ESP_LOGD` only goes to serial? with this change, `syslogger` is used correctly.

## Related links

#4845 introduced `USE_ESP_IDF_LOG`

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -19,6 +19,7 @@ menu.EraseFlash=Erase All Flash Before Sketch Upload
 menu.JTAGAdapter=JTAG Adapter
 menu.ZigbeeMode=Zigbee Mode
 menu.PinNumbers=Pin Numbering
+menu.LoggingSubsystem=Logging Subsystem
 
 # Custom options
 menu.Revision=Board Revision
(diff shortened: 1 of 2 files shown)


me-no-dev:
Anyone can do this at any time by adding `-DUSE_ESP_IDF_LOG` to `build_opt.h` in their sketch folder.  You can also overwrite the default log output by setting `ets_install_putc1` or `ets_install_putc2`. Therefore I am not in favor of this PR

guppy0130:
> Anyone can do this at any time by adding `-DUSE_ESP_IDF_LOG` to `build_opt.h` in their sketch folder.

Oh, didn't realize this; tried this out and works as expected. Don't really mind if this is closed then.

Will note that the change introduced in the PR may make it easier/more obvious for end users though? Happy to amend this to a docs update if that's desirable.

What makes the current set of options valid and this one less so? Or is it the overhead of tracking/maintaining another option (totally understandable)?

me-no-dev:
The `USE_ESP_IDF_LOG` option was added to help users who compile Arduino as component in their ESP-IDF projects. It is not meant to be used in Arduino IDE, because we have no control over IDF logs there. We even translate IDF logs back to Arduino logs if some IDF source is compiled (from the sketch or library). In Arduino IDE we use only Arduino logging, because that is what we can control (enabled, level, output and so on). All `Serial` outputs (UART, USB, etc.) already have a way to be used as log output and it's very easy to add that functionality to other ones too (as I have proposed above).

PR #11898, opened 2025-10-05
getFlashFrequencyMHz

Jason2866:
Currently the flash frequency is read from magic header. The value is not in every case the real used clock frequency.
The PR changes this and read the config from the hardware registers which sets the clock frequency.


Jason2866:
Closing, trying to find higher level function which can be used

Issue #11787, opened 2025-09-02
I2C.Master logging


wrlaw:
### Board

ESP32

### Device Description

ESP32 Do IT V1

### Hardware Configuration

Issue potentially lots of devices but the example below is for an I2C based LCD

### Version

v3.3.0

### Type

Bug

### IDE Name

Arduino IDE 1.8.19

### Operating System

Debian Linux

### Flash frequency

80 MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Arduino Core release 3.2.0 based on ESP-IDF Rel 5.4.1 introduced increased logging for i2c.master and this change was subsequently included in ESP-IDF 5.5.0 for Arduino Core 3.3.0.

Problem is, that increased logging produces messages which would be best classified as DEBUG but instead are reproduced on the serial port even if the Arduino IDE log level is configured as NONE.   This action fills the serial port with messages from i2c.master which the user may not be interested in.

This increased logging was not generated in ESP-IDF 5.3 used by Arduino Core release 3.1.3 so you either have to keep yourself at Arduino core release 3.1.3 or put up with 100s of messages coming out of i2c.master burying any messages from your application.

### Sketch

```cpp
/*
 * I2C based LCD as an example
 * 
 * On a dev board where for unrelated software development
 * the LCD may not be connected.
 *
 * The serial port is then filled with messages
 * if using Arduino Core releases 3.2.0 and 3.3.0
 * 
 * Release 3.1.3 has no such issue
 *
 * It should be possible to disable these messages
 * by setting the Arduino debug level to NONE
 * 
 */

#include <LiquidCrystal_I2C.h>  

LiquidCrystal_I2C lcd(0x27,16,2);

void setup() {
  Serial.begin(115200);

  lcd.init();            
  lcd.clear();          
  lcd.backlight();      
  lcd.setCursor(0,0);
  lcd.print("Hello");
}

void loop() {

}
```

### Debug Message

```plain
Debug level is the problem, you can't turn these messages off

13:33:15.776 -> E (1071) i2c.master: I2C hardware NACK detected
13:33:15.776 -> E (1071) i2c.master: I2C transaction unexpected nack detected
13:33:15.776 -> E (1071) i2c.master: s_i2c_synchronous_transaction(945): I2C transaction failed
13:33:15.776 -> E (1077) i2c.master: i2c_master_multi_buffer_transmit(1214): I2C transaction failed
[...]

PR #11931, opened 2025-10-18
fix(i2c): Set I2C master log level to none

me-no-dev:
Silence log messages from the I2C master driver.

Closes: https://github.com/espressif/arduino-esp32/issues/11787


P-R-O-C-H-Y (approved):
Me may document this, so when needed users can turn it on again.

me-no-dev:
@P-R-O-C-H-Y not sure that is needed, since users will get our logs. They just will not get spammed by the IDF driver

lucasssvaz (approved):

PR #11945, opened 2025-10-24
WebServer: Use client() instead of _currentClient

### LINKED ISSUE #11944: WebServer: Make it possible to create and use derived classes
### Related area

WebServer

### Hardware specification

Doesn't matter

In the WebServer class, `void WebServer::sendContent_P(PGM_P content, size_t size)` is not virtual, i.e.. not overridable. However, its implementation uses `_currentClient`. That makes it impossible to implement a derived class, e.g. to implement a TLS-protected SecureWebServer. For that, I'd have to use a `SecureClient` member.

At least that applies when I want to do "proper polymorphism", i.e. access my potential SecureWebServer via a WebServer pointer.

- Make all methods using `_currentClient` virtual
- Alternatively, make use of a newly added `getClient()` virtual method?

_No response_

### Additional context

_No response_

hurricanefrog:
This makes it possible to create derived classes properly.

## Description of Change
This change replaces all uses of `_currentClient` in the `WebServer` class by calling the virtual method `client()`.
(Except in `handleClient`, which makes to sense to touch, I think).

This makes it possible to create derived classes, e.g to create TLS-protected web servers.
This was not possible previously due to the hardcoded "plain" `_currentClient`.

Derived classes can now simply override `client()`, e.g. to return the TLS protected socket, and continue using the rest of the class (i.e. all of the HTTP parsing).

## Test Scenarios
I haven't done any tests. I expect no impact on the function of the code. I haven't been able to create a HTTPS server with the changes yet, to confirm it leads to my goal, this takes some additional time.

## Related links

closes #11944

### DIFF
--- a/libraries/WebServer/src/WebServer.cpp
+++ b/libraries/WebServer/src/WebServer.cpp
@@ -565,7 +565,7 @@ void WebServer::chunkResponseBegin(const char *contentType) {
   }
 
   _chunkedResponseActive = true;
-  _chunkedClient = _currentClient;
+  _chunkedClient = client();
 
   _contentLength = CONTENT_LENGTH_UNKNOWN;
 
@@ -574,7 +574,7 @@ void WebServer::chunkResponseBegin(const char *contentType) {
   _currentClientWrite(header.c_str(), header.length());
 
   _chunkedResponseActive = true;
-  _chunkedClient = _currentClient;
+  _chunkedClient = client();
 }
 
 void WebServer::chunkWrite(const char *data, size_t length) {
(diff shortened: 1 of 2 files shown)


hurricanefrog:
Should I really lowercase the WebServer component to `fix(webserver)` in the commit message?

me-no-dev:
> Should I really lowercase the WebServer component to fix(webserver) in the commit message?

It's not a big deal, but it's nice to have consistency. Will not influence wether this gets merged or or not :)

me-no-dev:
> I haven't done any tests. I expect no impact on the function of the code. I haven't been able to create a HTTPS server with the changes yet, to confirm it leads to my goal, this takes some additional time.

Lets keep this open then (I'll convert it to Draft) until you can test it, in case other changes also need to be added. Please let us know when you have done that :)

PR #11954, opened 2025-10-27
Update of all variant files to add BAT_VOLT_PIN for all variants that define a battery voltage pin

### LINKED ISSUE #11953: Battery pin definition non standard and untestable
### Board

Multiple - e.g. Lilygo T3S3 

### Device Description

None (attached to laptop, but not relevant)

### Hardware Configuration

No

### Version

latest stable Release (if not listed below)

### Type

Task

### IDE Name

PlatformIO

### Operating System

N/A

### Flash frequency

N/A

### PSRAM enabled

yes

### Upload speed

N/A

### Description

The variant files contain at least 15 different ways of telling you on which pin the battery voltage can be measured. 

Try `grep BAT /Users/mitra/.platformio/packages/framework-arduinoespressif32/variants/*/pins_arduino.h`
VBAT_VOLTAGE BAT_VOLTS BAT_VOLT BAT_MEASURE VBAT_SENSE BATT_ADC_PIN BAT_ADC_PIN BAT_ADC BATT_MONITOR BAT_MEASURE BAT_LV WS_BAT_ADC PIN_VBAT PIN_BATTERY BATTERY_PIN

What is worse is that these are mostly (though not all) `static const uint8_t` so it is next to impossible to write board independent (library) code without testing by board name

I am sure there is code that depends on most of these different definitions, so I would suggest we pick one and progressively add #defines to all variant files that define one common value. 

I don't particularly care which is used but suggest it indicates its an ADC as for example `VBAT_VOLTAGE` could imply its the voltage level, so if we chose `VBAT_BUILTIN`

then e.g. on `lilygo_t3_s3_sx1262/pins_arduino.h` we could add `#define VBAT_BUILTIN BAT_VOLT`

I'm sure someone could write a quick python script to make all the changes, but I don't want to submit this as a PR if there is likely to be (reasonable) pushback. 

### Sketch

```cpp
#ifdef ARDUINO_LILYGO_T3_S3_V1_X
  analogRead(BAT_VOLT); // Specific to a few boards
#elif defined(xxxxx)
  analogRead(VBAT_SENSE)
#elif ... repeat for every board that has built in battery measurement
#endif
```

### Debug Message

```plain
n/a
```

### Other Steps to Reproduce

_No response_

mitra42:
Closes #11953 

## Description of Change
Variant files define the battery voltage pin in at least 15 different ways, this makes it impossible to write library code that works with the battery across multiple boards without a VERY long ifdef which (since existing variant files define `const uint8_t`) can only test against board names. 

Adding this #define allows for board and variant independent battery code for all boards that define a battery pin. 

Note - name of pin BAT_VOLT_PIN as suggested by @me-no-dev 

## Test Scenarios

I have tested on lilygo_t3_s3 using the variant lilygo_t3_s3_sx127x, changing a battery sensor to use this new #define. 

## Related links
Closes #11953

### DIFF
--- a/variants/Pcbcupid_GLYPH_C3/pins_arduino.h
+++ b/variants/Pcbcupid_GLYPH_C3/pins_arduino.h
@@ -11,7 +11,7 @@ static const uint8_t LED_BUILTIN = 1;
 //MSR Used in on-board battery measurement
 static const uint8_t BAT_MEASURE = 0;
 #define BAT_VOLT_PIN BAT_MEASURE
-#define MSR BAT_MEASURE
+#define MSR          BAT_MEASURE
 
 static const uint8_t TX = 21;
 static const uint8_t RX = 20;
--- a/variants/Pcbcupid_GLYPH_C6/pins_arduino.h
+++ b/variants/Pcbcupid_GLYPH_C6/pins_arduino.h
@@ -11,7 +11,7 @@ static const uint8_t LED_BUILTIN = 14;
 //MSR Used in on-board battery measurement
 static const uint8_t BAT_MEASURE = 0;
 #define BAT_VOLT_PIN BAT_MEASURE
-#define MSR BAT_MEASURE
+#define MSR          BAT_MEASURE
 
 static const uint8_t TX = 16;
 static const uint8_t RX = 17;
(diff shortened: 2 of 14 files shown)


me-no-dev:
@P-R-O-C-H-Y one more thing to keep track of

mitra42:
Thanks

PR #11967, opened 2025-10-29
fix(esp32): issues on starting touch sensor and attaching interrupt

AlexAls:
This PR fixes the following issues:
 - error on setting a callback;
 - error on setting a custom threshold;
 - wrong condition for starting touch sensor.

The order of arguments is mixed up when calling a function `__touchConfigInterrupt`.
Setting the threshold resets all other configuration parameters.
In function `touchStart` using "enable" instead of "not enable" in "if" condition.

### DIFF
--- a/cores/esp32/esp32-hal-touch-ng.c
+++ b/cores/esp32/esp32-hal-touch-ng.c
@@ -117,8 +117,12 @@ bool touchDisable() {
   if (!enabled) {  // Already disabled
     return true;
   }
-  if (!running && (touch_sensor_disable(touch_sensor_handle) != ESP_OK)) {
-    log_e("Touch sensor still running or disable failed!");
+  if (running) {
+    log_e("Touch sensor still running!");
+    return false;
+  }
+  if (touch_sensor_disable(touch_sensor_handle) != ESP_OK) {
+    log_e("Touch sensor disable failed!");
     return false;
   }
   enabled = false;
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y (approved):
Nice catch. Thank you for the fix!

cores/esp32/esp32-hal-touch-ng.c
```diff
@@ -129,7 +129,7 @@ bool touchStart() {
   if (running) {  // Already running
     return true;
   }
-  if (enabled && (touch_sensor_start_continuous_scanning(touch_sensor_handle) != ESP_OK)) {
+  if (!enabled && (touch_sensor_start_continuous_scanning(touch_sensor_handle) != ESP_OK)) {
```
P-R-O-C-H-Y: Hi, your approach will create a bug and touch will never start. PLease apply following fix!
```suggestion
  if(!enabled) {
    log_e("Touch sensor not enabled!");
    return false;
  }
  if (touch_sensor_start_continuous_scanning(touch_sensor_handle) != ESP_OK) {
```
+ change the error log message to:
`Touch sensor failed to start continuous scanning!`

AlexAls: Oh, you're right!
I also saw the same logical issue in the `touchDisable` function and fixed it. Please take a look.

PR #11982, opened 2025-11-04
fix(eth): Del mac and phy resources

szerwi:
## Description of Change
Delete `_mac` and `_phy` to free the resources if initialization of ETH failed or when calling `ETHClass::end()` and ETH is not fully initialized.

I found a bug that leads to `w5500_tsk` running when `begin()` for W5500 ETH module fails. When `begin()` fails, it makes it impossible to free the resources allocated in `_mac` and `_phy` and calling `begin()` one more time makes those resources to allocate again leading to a memory leak.

This PR adds clearing the `_mac` and `_phy` handles when one of them or `_eth_handle` fails to initialize during `begin()`.
The fix also changes `end()` function to free those handles even if `_eth_handle` is null.

## Test Scenarios
Tested on arduino-esp32 v3.3.2, ESP32 Dev Board and W5500 Ethernet module.
Tested on both scenarios, with Ethernet module connected to ESP32 and disconnected from power.
When the module is not powered, `ETHClass::begin()` fails with:
```
E (17:34:56.299) w5500.mac: w5500_reset(308): reset timeout
E (17:34:56.300) w5500.mac: emac_w5500_init(893): reset w5500 failed
E (17:34:56.310) esp_eth: esp_eth_driver_install(250): init mac failed
E (17:34:56.311) ARDUINO: SPI Ethernet driver install failed: 263
```
and deletes `_mac` and `_phy` handles, which results in the removal of `w5500_tsk`.

Before the fix, the `w5500_tsk` kept running resulting in constant `w5500.mac: w5500_get_rx_received_size(238): read RX RSR failed` errors.

## Related links

### DIFF
--- a/cores/esp32/HardwareSerial.cpp
+++ b/cores/esp32/HardwareSerial.cpp
@@ -572,8 +572,20 @@ HardwareSerial::operator bool() const {
   return uartIsDriverInstalled(_uart);
 }
 
-void HardwareSerial::setRxInvert(bool invert) {
-  uartSetRxInvert(_uart, invert);
+bool HardwareSerial::setRxInvert(bool invert) {
+  return uartSetRxInvert(_uart, invert);
+}
+
+bool HardwareSerial::setTxInvert(bool invert) {
+  return uartSetTxInvert(_uart, invert);
+}
+
+bool HardwareSerial::setCtsInvert(bool invert) {
+  return uartSetCtsInvert(_uart, invert);
+}
+
+bool HardwareSerial::setRtsInvert(bool invert) {
+  return uartSetRtsInvert(_uart, invert);
 }
 
 // negative Pin value will keep it unmodified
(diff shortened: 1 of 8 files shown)


TD-er:
Purely based on your description, I think I may have ran into the issues you describe.
So that would be great if those are fixed by this PR.
Will try and test tomorrow.

libraries/Ethernet/src/ETH.cpp
```diff
@@ -281,6 +281,7 @@ bool ETHClass::begin(eth_phy_type_t type, int32_t phy_addr, int mdc, int mdio, i
   _mac = esp_eth_mac_new_esp32(&mac_config, &eth_mac_config);
   if (_mac == NULL) {
     log_e("esp_eth_mac_new_esp32 failed");
+    _delMacAndPhy();
```
me-no-dev: mac is null and phy has not been initialized yet. This is not needed here

me-no-dev (approved):

PR #11985, opened 2025-11-05
feat(board): add support for Arduino Nesso N1

facchinm:
-----------
## Description of Change
This PR adds support for upcoming Arduino Nesso N1 board. The board is a boxed product, containing an ESP32-C6 module, a [small display](https://github.com/m5stack/M5GFX/blob/6c93c6836f11a5abad67c0ff35ff75050e636abc/src/M5GFX.cpp#L2203-L2208) , a buzzer, QWIIK and GROVE connectors, a battery + its charger/fuel gauge and a LoRa SX1262 radio.

Most of the internal GPIOs are exposed via I2C expanders, so they have been described and added to a couple of extra files in `variant` folder. 
If this is not advisable, let me know and I'll drop the last couple of commits, turning them into a library (this wouldn't be ideal since also LED_BUILTIN pin is on an expander).

## Test Scenarios
We tested the board with:
* M5GFX
* https://github.com/beegee-tokyo/SX126x-Arduino with this config
```
hwConfig.CHIP_TYPE = SX1262_CHIP;
hwConfig.PIN_LORA_RESET = -1; // LORA RESET
hwConfig.PIN_LORA_NSS = LORA_CS;	  // LORA SPI CS
hwConfig.PIN_LORA_SCLK = SCK;	  // LORA SPI CLK
hwConfig.PIN_LORA_MISO = MISO;	  // LORA SPI MISO
hwConfig.PIN_LORA_DIO_1 = LORA_IRQ; // LORA DIO_1
hwConfig.PIN_LORA_BUSY = LORA_BUSY;	  // LORA SPI BUSY
hwConfig.PIN_LORA_MOSI = MOSI;	  // LORA SPI MOSI
hwConfig.RADIO_TXEN = -1;		  // LORA ANTENNA TX ENABLE
hwConfig.RADIO_RXEN = -1;		  // LORA ANTENNA RX ENABLE
hwConfig.USE_DIO2_ANT_SWITCH = true;
hwConfig.USE_DIO3_TCXO = true;
hwConfig.USE_DIO3_ANT_SWITCH = false;
hwConfig.USE_RXEN_ANT_PWR = true;
```

## Related links
LInks will be added as soon as the product pages are up.

### DIFF
--- a/variants/arduino_nesso_n1/expander.cpp
+++ b/variants/arduino_nesso_n1/expander.cpp
@@ -53,7 +53,7 @@ void pinMode(ExpanderPin pin, uint8_t mode) {
   }
   writeBitRegister(pin.address, 0x3, pin.pin, mode == OUTPUT);
   if (mode == OUTPUT) {
-    // remove high impedence
+    // remove high impedance
     writeBitRegister(pin.address, 0x7, pin.pin, false);
   } else if (mode == INPUT_PULLUP) {
     // set pull-up resistor


me-no-dev:
@facchinm we have just released core version 3.3.3 that includes support for this board.

If you need to add similar support in the future, I suggest to use the I2C C API from `esp32-hal-i2c.h`, because it does not need any tricks, it's lightweight and most of all, is thread-safe. `Wire.read()` is not.

PR #12003, opened 2025-11-10
feat(udp): Allow to change the async_udp task stack size

mathieucarbou:
Example:  `-D CONFIG_ARDUINO_UDP_TASK_STACK_SIZE=2048`

The default is to much conservative since a correctly developed app probably won't allocate too much on stack on the callbacks. I currently see a waste of memory:
`I (1222501) MONITOR: async_udp  (p=3) 3416 bytes`

### DIFF
--- a/Kconfig.projbuild
+++ b/Kconfig.projbuild
@@ -147,6 +147,12 @@ config ARDUINO_UDP_TASK_PRIORITY
     help
         Select at what priority you want the UDP task to run.
 
+config ARDUINO_UDP_TASK_STACK_SIZE
+    int "UDP task stack size"
+    default 4096
+    help
+        Amount of stack available for the UDP task.
+
 config ARDUINO_ISR_IRAM
     bool "Run interrupts in IRAM"
     default "n"
--- a/libraries/AsyncUDP/src/AsyncUDP.cpp
+++ b/libraries/AsyncUDP/src/AsyncUDP.cpp
@@ -15,6 +15,27 @@ extern "C" {
 
 #include "lwip/priv/tcpip_priv.h"
 
+#ifndef CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
+#define CONFIG_ARDUINO_UDP_TASK_STACK_SIZE 4096
+#endif
+#ifndef ARDUINO_UDP_TASK_STACK_SIZE
+#define ARDUINO_UDP_TASK_STACK_SIZE CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
+#endif
+
+#ifndef CONFIG_ARDUINO_UDP_TASK_PRIORITY
+#define CONFIG_ARDUINO_UDP_TASK_PRIORITY 3
+#endif
+#ifndef ARDUINO_UDP_TASK_PRIORITY
+#define ARDUINO_UDP_TASK_PRIORITY CONFIG_ARDUINO_UDP_TASK_PRIORITY
+#endif
+
+#ifndef CONFIG_ARDUINO_UDP_RUNNING_CORE
+#define CONFIG_ARDUINO_UDP_RUNNING_CORE -1
+#endif
+#ifndef ARDUINO_UDP_RUNNING_CORE
+#define ARDUINO_UDP_RUNNING_CORE CONFIG_ARDUINO_UDP_RUNNING_CORE
+#endif
+
 #ifdef CONFIG_LWIP_TCPIP_CORE_LOCKING
 #define UDP_MUTEX_LOCK()                                \
   if (!sys_thread_tcpip(LWIP_CORE_LOCK_QUERY_HOLDER)) { \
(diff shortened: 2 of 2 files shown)


me-no-dev:
@mathieucarbou `CONFIG_ARDUINO_UDP_TASK_STACK_SIZE` might not be normally present when compiling in Arduino IDE. I suggest to check if it's not defined and define it to default value on the top of the file (or header)

mathieucarbou:
> @mathieucarbou `CONFIG_ARDUINO_UDP_TASK_STACK_SIZE` might not be normally present when compiling in Arduino IDE. I suggest to check if it's not defined and define it to default value on the top of the file (or header)

Hello,

That's what I did first if you look here: https://github.com/espressif/arduino-esp32/compare/61aaa44254c1ccc4e45eaecbee1cd268e7324795..b3c3a31df445ce1e6990b412e0b1740a50187930

But then I saw that all the macros are only defined in `Kconfig.projbuild`...

So for this one exceptionally, I should do both ? Is it what you mean ?

Thanks :-)

me-no-dev:
Since this has not been merged and the lib-builder did not get a chance to run Kconfig and generate the defines, they do not currently exists in Arduino at all. That is the reason why CI is failing and anyone that would normally try that code now will also get a compile error. Adding check on top will ensure that the code will compile in all cases. The other values are missing because before the libs used to be built in a different way and included in the core. Later the define check could be removed if need be.

There are also better ways to handle this and allow overwrite in Arduino or pioarduino
```cpp
#ifndef CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
#define CONFIG_ARDUINO_UDP_TASK_STACK_SIZE 4096
#endif
#ifndef ARDUINO_UDP_TASK_STACK_SIZE
#define ARDUINO_UDP_TASK_STACK_SIZE CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
#endif

xTaskCreateUniversal(
      _udp_task, "async_udp", ARDUINO_UDP_TASK_STACK_SIZE, NULL, CONFIG_ARDUINO_UDP_TASK_PRIORITY, (TaskHandle_t *)&_udp_task_handle, CONFIG_ARDUINO_UDP_RUNNING_CORE
    );
```

This way you can `-D ARDUINO_UDP_TASK_STACK_SIZE=8192` either in `build_opt.h` in the sketch folder or in platformio config to change the value and not cause redefine warning

mathieucarbou:
> This way you can `-D ARDUINO_UDP_TASK_STACK_SIZE=8192` either in `build_opt.h` in the sketch folder or in platformio config to change the value and not cause redefine warning

Thanks for your help!

Should I do the same for CONFIG_ARDUINO_UDP_RUNNING_CORE and CONFIG_ARDUINO_UDP_TASK_PRIORITY then ?

In my app I set CONFIG_ARDUINO_UDP_RUNNING_CORE to core 0 for example and on purpose since it is running critical in the callback and code I do not want to be impacted by the loop.

me-no-dev:
> Should I do the same for CONFIG_ARDUINO_UDP_RUNNING_CORE and CONFIG_ARDUINO_UDP_TASK_PRIORITY then ?

Please do. We have overlooked this and since nobody asked for it, we did not catch it

mathieucarbou:
> > Should I do the same for CONFIG_ARDUINO_UDP_RUNNING_CORE and CONFIG_ARDUINO_UDP_TASK_PRIORITY then ?
> 
> Please do. We have overlooked this and since nobody asked for it, we did not catch it

done!

```c++
#ifndef CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
#define CONFIG_ARDUINO_UDP_TASK_STACK_SIZE 4096
#endif
#ifndef ARDUINO_UDP_TASK_STACK_SIZE
#define ARDUINO_UDP_TASK_STACK_SIZE CONFIG_ARDUINO_UDP_TASK_STACK_SIZE
#endif

#ifndef CONFIG_ARDUINO_UDP_TASK_PRIORITY
#define CONFIG_ARDUINO_UDP_TASK_PRIORITY 3
#endif
#ifndef ARDUINO_UDP_TASK_PRIORITY
#define ARDUINO_UDP_TASK_PRIORITY CONFIG_ARDUINO_UDP_TASK_PRIORITY
#endif

#ifndef CONFIG_ARDUINO_UDP_RUNNING_CORE
#define CONFIG_ARDUINO_UDP_RUNNING_CORE -1
#endif
#ifndef ARDUINO_UDP_RUNNING_CORE
#define ARDUINO_UDP_RUNNING_CORE CONFIG_ARDUINO_UDP_RUNNING_CORE
#endif
```

I set -1 like it is done at other places, hopping this is ok for async udp too.

me-no-dev:
> I set -1 like it is done at other places, hopping this is ok for async udp too.

It is. `xTaskCreateUniversal` will turn it to 0 if chip is single core

PR #12043, opened 2025-11-19
fix(wifi): Do not use persistent mode with ESP-Hosted

me-no-dev:
This pull request introduces a configuration change to the WiFi initialization process to improve compatibility when using esp-hosted mode. The main update ensures that NVS (Non-Volatile Storage) is disabled and persistent storage is turned off when the remote WiFi feature is enabled.

WiFi initialization compatibility:

* In `WiFiGeneric.cpp`, added a conditional block that disables NVS (`cfg.nvs_enable = false`) and sets `persistent = false` when `CONFIG_ESP_WIFI_REMOTE_ENABLED` is defined, ensuring proper operation with esp-hosted setups.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -19,6 +19,7 @@ menu.EraseFlash=Erase All Flash Before Sketch Upload
 menu.JTAGAdapter=JTAG Adapter
 menu.ZigbeeMode=Zigbee Mode
 menu.PinNumbers=Pin Numbering
+menu.ChipVariant=Chip Variant
 
 # Custom options
 menu.Revision=Board Revision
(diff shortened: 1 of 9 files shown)


me-no-dev:
> What is the issue when persistence is enabled ?

credentials are stored on the C6 and that causes issues in some cases. @TD-er had a case also. Change was recommended from the hosted team

TD-er:
Yep, the C6 could even (re)connect when I just wiped the flash of the P4 and tried to start fresh without any wifi credentials present in ESPEasy.
With this fix (and a test build of ESP-hosted-mcu), the false reconnect issue is fixed.

PR #12062, opened 2025-11-24
fix(wifi): Fix printDiag() when WiFi not initialized

szerwi:
## Description of Change
Initialize local variables at declaration in `WiFiClass::printDiag(Print &p)` to prevent crash when WiFi is not initialized.
Previously, when WiFi was not initialized, `mode` could be any value leading to crash when trying to print its value.

Also added `NAN` value to `modes[]`, as `wifi_mode_t` can have following values:
```
typedef enum {
    WIFI_MODE_NULL = 0,  /**< Null mode */
    WIFI_MODE_STA,       /**< Wi-Fi station mode */
    WIFI_MODE_AP,        /**< Wi-Fi soft-AP mode */
    WIFI_MODE_APSTA,     /**< Wi-Fi station + soft-AP mode */
    WIFI_MODE_NAN,       /**< Wi-Fi NAN mode */
    WIFI_MODE_MAX
} wifi_mode_t;
```

## Test Scenarios
Tested on arduino-esp32 v3.3.4 and ESP32 DevKit

### DIFF
--- a/libraries/WiFi/src/WiFi.cpp
+++ b/libraries/WiFi/src/WiFi.cpp
@@ -45,13 +45,13 @@ extern "C" {
  * @param p Print interface
  */
 void WiFiClass::printDiag(Print &p) {
-  const char *modes[] = {"NULL", "STA", "AP", "STA+AP"};
+  const char *modes[] = {"NULL", "STA", "AP", "STA+AP", "NAN"};
 
-  wifi_mode_t mode;
+  wifi_mode_t mode = WIFI_MODE_NULL;
   esp_wifi_get_mode(&mode);
 
-  uint8_t primaryChan;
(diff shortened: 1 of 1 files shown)


me-no-dev (approved):

me-no-dev:
This is old API. I suggest you use `Serial.println(WiFi.STA);`

szerwi:
> This is old API. I suggest you use `Serial.println(WiFi.STA);`

Is `Serial.println(WiFi.STA);` same as `WiFi.STA.printTo(Serial)`?
If yes, then it gives different information than `printDiag()`.

I use both for debugging my app and I have noticed that bug with `printDiag()`, so I thought I will patch it.

me-no-dev:
BTW NAN is not enabled

PR #12067, opened 2025-11-25
refactor(wifi): Refactor WiFi sleep modes

szerwi:
## Description of Change
Refactor `WiFiGenericClass::setSleep()` and `WiFiGenericClass::getSleep()` functions.

`WiFiGenericClass::setSleep()` was returning false when requested mode was the same as already set, which is not expected behavior. 

I have refactored functions logic, unified variables names, removed unused variable `_forceSleepLastMode` from `WiFiGenericClass`.

---

I am not sure if [this ifdef](https://github.com/espressif/arduino-esp32/compare/master...szerwi:refactor/wifi_sleep?expand=1#diff-0aa87997f6f48b5397f969144b34ce0ad4da8e173aee854c1cf2e1bf69ecf3aaL378-L383) should be kept. Is there any reason to disable modem sleep for ESP32 S2? I haven't found any info about it in ESP-IDF docs.

## Test Scenarios
Tested on v3.3.4 and ESP32 DevKit

### DIFF
--- a/libraries/WiFi/src/WiFiGeneric.cpp
+++ b/libraries/WiFi/src/WiFiGeneric.cpp
@@ -375,7 +375,6 @@ static bool espWiFiStop() {
 
 bool WiFiGenericClass::_persistent = true;
 bool WiFiGenericClass::_long_range = false;
-wifi_mode_t WiFiGenericClass::_forceSleepLastMode = WIFI_MODE_NULL;
 #if CONFIG_IDF_TARGET_ESP32S2
 wifi_ps_type_t WiFiGenericClass::_sleepEnabled = WIFI_PS_NONE;
 #else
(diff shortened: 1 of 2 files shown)


me-no-dev:
most of the changes are renaming variables. That makes reading your changes really hard. Please revert var names!

szerwi:
> most of the changes are renaming variables. That makes reading your changes really hard. Please revert var names!

@me-no-dev sure, I will revert that. Should I use old var names, which are not accurate in some cases, or can I correct them and split those changes into 2 commits (logic change & vars renaming)?

me-no-dev:
please use the old names and only provide the logic changes. In a separate PR you can rename the variables later

szerwi:
@me-no-dev Var names reverted

PR #12079, opened 2025-11-29
feat(spi): Set power channel on ESP32P4

lbernstone:
## Description of Change
Set power channel on ESP32P4 so 3V3 is supplied if the SDMMC pins are used for SPI. This allows you to use old SD based libraries.

## Test Scenarios
ESP32-P4-EYE w v3.3
I think my intersection is correct :)

### DIFF
--- a/cores/esp32/esp32-hal-spi.c
+++ b/cores/esp32/esp32-hal-spi.c
@@ -35,7 +35,7 @@
 #include "esp_system.h"
 #include "esp_intr_alloc.h"
 
-#ifdef SOC_SDMMC_IO_POWER_EXTERNAL //ESP32-P4
+#ifdef SOC_SDMMC_IO_POWER_EXTERNAL  //ESP32-P4
 #include "sd_pwr_ctrl_by_on_chip_ldo.h"
 #include "soc/sdmmc_pins.h"
 #endif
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y:
@lbernstone I think we should move it to `sdcard_init` here if its meant for the SD SPI. https://github.com/espressif/arduino-esp32/blob/e7df08a65241930b188a48ffa0cac9551029db6a/libraries/SD/src/sd_diskio.cpp#L737

me-no-dev:
@lbernstone and if this is meant for general SPI, I will suggest to move the code to `esp32-hal-spi.c`.  We leave the class for very high level implementation only

lbernstone:
I still need to test this on a couple different configs. I have a P4 dev board on the way.

cores/esp32/esp32-hal-spi.c
```diff
@@ -256,6 +261,48 @@ static bool spiDetachBus_SS(void *bus) {
   return true;
 }
 
+#ifdef SOC_SDMMC_IO_POWER_EXTERNAL //ESP32-P4
+void setLDOPower(int8_t pin) {
```
me-no-dev: if this function is used only in this file, please make it static

lbernstone:
Ok, I think this is ready to merge

lbernstone:
This failed on a network timeout. Can someone restart the CI? @P-R-O-C-H-Y

Jason2866:
@me-no-dev do you see a reason not to merge?

me-no-dev:
@Jason2866 needs to be tested on old and new chip. That is it.

P-R-O-C-H-Y:
Assigning myself to test that.

PR #12107, opened 2025-12-05
revert: nesso_n1: move expander and battery to external library

facchinm:
We noticed that adding the expander code in the core has been a bad idea :upside_down_face: 
Moving it out to https://github.com/arduino-libraries/Arduino_Nesso_N1 , while leaving here only an explicit error message (much more informative than just failing at link time).

@ubidefeo
@me-no-dev

### DIFF
--- a/variants/arduino_nesso_n1/pins_arduino.h
+++ b/variants/arduino_nesso_n1/pins_arduino.h
@@ -46,7 +46,7 @@ static const uint8_t LCD_RS = 16;
 
 class ExpanderPinError {
 public:
-  ExpanderPinError(uint16_t p) {};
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y (approved):

facchinm:
@me-no-dev sorry to bother you, just need to know if 3.3.5 is supposed to be released before the end of this week (we have to push the documentation changes and winter vacation time is not ideal :upside_down_face: )

me-no-dev:
we will release Friday the latest. Holding because of a serious bug in one of the external components. We will disable it if it's not fixed by then

Issue #12050, opened 2025-11-21
BLE_for_ESP32_Arduino Manufacturer's data with a null character


profor-cloud:
### Board

ESP32

### Device Description

LOLIN32

### Hardware Configuration

none

### Version

v3.3.4

### Type

Task

### IDE Name

Arduino IDE 2.3.6

### Operating System

Windows 10

### Flash frequency

40

### PSRAM enabled

yes

### Upload speed

115200

### Description

Hi,
While copying the BLE ManufacturerData from a real device to simulate the device,
I noticed the data is treated like a string, so a zero char half-way is treated like an end of string.
But in my case a zero is in the Manufacturor data, and can't set vendor_ID or additional vendor specific data properly when it contains a zero with this code.

### Sketch

```cpp
String manufacturerData = "\xEC\xB1\xAC\x00\xB6\x7c";
advertisementData.setManufacturerData(manufacturerData);
pAdvertising->setAdvertisementData(advertisementData);
// When looking with a BLE 'scanner' all after the first zero is basically chopped off.
```

### Debug Message

```plain
Done
```

### Other Steps to Reproduce

Arduino_ESP32 2.0.17 the function is declared as:
void setManufacturerData(std::string data);
and the code:
std::string manufacturerData = { 0xEC,  0xB1 , 0xAC,  0x00,  0xB6,  0x7C};
advertisementData.setManufacturerData(manufacturerData);
pAdvertising->setAdvertisementData(advertisementData);
 is executed correctly.
starting with library version 3.0.0 the function is declared as:
void setManufacturerData(String data);

Please change the library to return the ability to work with characters 0x00.

### I have checked existing issues, online documentation and the Troubleshooting Guide

- [x] I confirm I have checked existing issues, online documentation and Troubleshooting guide.

lbernstone:
I know it's not really the answer you want, but [NimBLE](https://github.com/h2zero/NimBLE-Arduino) has multiple methods for setManufacturerData such that you could use a char* with a length. I'm not sure if the bluedroid library will be getting any more updates.

lucasssvaz:
Opened a PR to allow for initializer lists in the Arduino string.
This will make it possible to create strings just like in C++:
```
// Just like std::string
String manufacturerData = {'H', 'i', '!'}
String manufacturerData = {0xEC, 0xB1, 0xAC, 0x00, 0xB6, 0x7C};
```

PR #12109, opened 2025-12-05
feat(string): Add initializer list constructor

lucasssvaz:
## Description of Change

This pull request adds support for initializing a `String` object using an initializer list of characters, making it easier to construct `String` instances with a list of characters in C++11 and newer environments. The main changes include updating the header to include the necessary library and adding the new constructor implementation.

This will make it possible to create strings like in C++:
```
// Just like std::string
String manufacturerData = {'H', 'i', '!'}
String manufacturerData = {0xEC, 0xB1, 0xAC, 0x00, 0xB6, 0x7C};
```

**Initializer list support for `String`:**

* Added `#include <initializer_list>` to `WString.h` to support initializer lists.
* Added a new constructor `String(std::initializer_list<char> list)` to the `String` class declaration in `WString.h`.
* Implemented the new initializer list constructor in `WString.cpp`, allowing `String` objects to be created from a list of characters.

## Test Scenarios

Tested locally

## Related links

Fixes https://github.com/espressif/arduino-esp32/issues/12050

### DIFF
--- a/cores/esp32/MacAddress.h
+++ b/cores/esp32/MacAddress.h
@@ -23,6 +23,9 @@
 #include <stdint.h>
 #include <WString.h>
 #include <Printable.h>
+#ifdef __GXX_EXPERIMENTAL_CXX0X__
+#include <initializer_list>
+#endif
 
 enum MACType {
   MAC6,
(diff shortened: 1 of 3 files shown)


cores/esp32/MacAddress.h
```diff
@@ -23,6 +23,7 @@
 #include <stdint.h>
 #include <WString.h>
 #include <Printable.h>
+#include <initializer_list>
```
me-no-dev, line 27: shouldn't this include be guarded too?

cores/esp32/WString.h
```diff
@@ -29,6 +29,7 @@
 #include <stdint.h>
 #include <string.h>
 #include <ctype.h>
+#include <initializer_list>
```
me-no-dev, line 33: same question

PR #12152, opened 2025-12-18
fix(matter_docs): fixes the Matter features table for the ESP32-C5

SuGlider:
## Description of Change
ESP32-C5 has been configured in Lib Builder to support Matter over Thread (no WiFi).
This is due to restrictions of SRAM space when running Thread + WiFi and Matter all together as done for the ESP32-C6.

This is fixed in all ESP32 Arduino Matter README,md files for the examples.

## Test Scenarios
CI only

## Related links
None


lucasssvaz:
Shouldn't we also update the table in libraries.rst ?

PR #12196, opened 2026-01-02
Add partition schema definition optimized for local storage

frdoni:
## Description of Change
Add partition schema definition optimized for local storage. This permit to serve many web pages from SPIFFS and store a lot of information in NVS partition

## Test Scenarios
 I have tested my Pull Request on Arduino-esp32 core v3.3.5 with ESP32 Board


tools/partitions/storage_4MB_noota.csv
```diff
@@ -0,0 +1,5 @@
+# Name,   Type, SubType, Offset,   Size,     Flags
+app0,     app,  ota_0,   0x10000,  0x200000,
```
P-R-O-C-H-Y: When no OTA is used, use the `factory` SubType instead of `ota_0`.
```suggestion
app0,     app,  factory,   0x10000,  0x200000,
```

frdoni:
I've uploaded new storage_4MB_noota.csv version

Issue #12198, opened 2026-01-03
BLECharacteristic onRead() callbacks never run on NimBLE


langufo:
### Board

ESP32-S3-DevKitC-1-N8R2

### Device Description

ESP32-S3-DevKitC-1-N8R2

### Hardware Configuration

- The development board, connected to a PC via the USB COM port.
- An Android phone with the Renesas SmartBond app to read/write BLE characteristics.

### Version

v3.3.5

### Type

Bug

### IDE Name

Arduino IDE

### Operating System

Fedora 42

### Flash frequency

QIO 80 MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

I cannot see the `onRead` callback for a `BLECharacteristic` running upon a read request. No issue is observed instead for `onWrite` callbacks.

When running the provided sketch, compiled with a Core Debug Level set to "Debug", I can see the following through the serial port:
```
[   338][I][BLEDevice.cpp:246] init(): Initializing BLE stack: NimBLE
```
```
[ 43751][D][BLECharacteristic.cpp:959] handleGATTServerEvent(): Characteristic a93b4dd8-10f0-4d07-b045-68d492a8968c Read event
[ 43841][D][BLECharacteristic.cpp:959] handleGATTServerEvent(): Characteristic a93b4dd8-10f0-4d07-b045-68d492a8968c Read event
[ 70212][D][BLECharacteristic.cpp:959] handleGATTServerEvent(): Characteristic a93b4dd8-10f0-4d07-b045-68d492a8968c Write event
[ 70273][D][onRead_callback_bug_3p3p5.ino:14] onWrite(): 
[ 83757][D][BLECharacteristic.cpp:959] handleGATTServerEvent(): Characteristic a93b4dd8-10f0-4d07-b045-68d492a8968c Read event
```
1. When reading the characteristic, two `Read event`s are logged (because the value of the characteristic does not fit in a single response), but for neither the `onRead()` call is logged.
2. When writing (blanking) the same characteristic, both the `Write event` and the `onWrite()` call are logged as expected.
3. When reading again the same characteristic, the `Read event` (just one this time, as the value is empty) is again logged, and the `onRead()` is not.

After modifying the code of the library with the addition of extra logging in `BLECharacteristic::handleGATTServerEvent` for NimBLE, I can see that `ctxt->om->om_pkthdr_len == 8` for all read requests, meaning `onRead` is never called as it requires `ctxt->om->om_pkthdr_len > 8`.

[...]

PR #12234, opened 2026-01-13
fix(ble): Fix onRead callback

lucasssvaz:
## Description of Change

This pull request updates the logic for handling GATT server read events in both `BLECharacteristic` and `BLEDescriptor` to ensure that the `onRead()` callback is only called for the initial read request, and not for follow-up packets during long reads. This prevents unnecessary repeated callbacks and makes the behavior more robust.

**Improvements to GATT read event handling:**

* Updated `BLECharacteristic::handleGATTServerEvent` to call `onRead()` only if the buffer length is greater than 0 and the connection handle is valid, ensuring callbacks are not triggered for follow-up packets in long reads.
* Updated `BLEDescriptor::handleGATTServerEvent` to use the same logic, only calling `onRead()` for the initial read request and not for follow-up packets, by checking buffer length and connection handle validity.

## Test Scenarios

Tested locally

## Related links

Fixes https://github.com/espressif/arduino-esp32/issues/12198

### DIFF
--- a/libraries/BLE/src/BLECharacteristic.cpp
+++ b/libraries/BLE/src/BLECharacteristic.cpp
@@ -963,9 +963,9 @@ int BLECharacteristic::handleGATTServerEvent(uint16_t conn_handle, uint16_t attr
     switch (ctxt->op) {
       case BLE_GATT_ACCESS_OP_READ_CHR:
       {
-        // If the packet header is only 8 bytes this is a follow up of a long read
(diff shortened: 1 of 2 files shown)


P-R-O-C-H-Y:
LGTM, CI is failing due to unused variable.

lucasssvaz:
> LGTM, CI is failing due to unused variable.

Fixed, thanks!

SuGlider (approved):

PR #12258, opened 2026-01-19
feat(wifi): Add WPA2-Enterprise support to WiFiMulti

lglina:
## Description of Change
`WiFiMulti`, which connects to the strongest available wireless access point based on a user-supplied list, currently only supports open networks (with no authentication) or networks using passphrase-based authentication (e.g. WPA2-PSK). This change adds support for WPA2-Enterprise networks, such as those used in offices and schools, which require username and password authentication, e.g. via PEAPv0/EAP-MSCHAPv2.

`WiFiMulti::add()` is modified to optionally accept a `username` string, in addition to the existing `passphrase` string, and optionally an `identity` string where the network requires a separate "anonymous identity" for PEAP. This change is backward compatible with existing code utilising only open networks and/or networks using passphrase-based authentication.

Connecting to WPA2-Enterprise networks using other forms of phase 1 or phase 2 authentication (e.g. WPA2-Enterprise with EAP-TLS, using certificates) is not yet supported. Nor is supplying a CA certificate to verify the server certificate provided during phase 1 authentication (PEAP). These could be added in a future change, although there would need to be a mechanism to store certificates within `WiFiMulti`. Although widely used, PEAPv0/EAP-MSCHAPv2 may be vulnerable to credential theft, particularly when the server certificate is not verified, and these networks may in future migrate to EAP-TLS which provides stronger security. This change notes in the documentation that considering the adequacy of connection security is the responsibility of the user.

`CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT` is required to connect to networks using WPA2-Enterprise. `WiFiMulti` will fail to connect if an enterprise network is `add()`'ed and this option is not enabled.

## Test Scenarios
This change has been tested on an ESP32-C3-MINI-1-N4 against networks using WPA2-Enterprise (PEAPv0/EAP-MSCHAPv2, new functionality) as well as those using WPA2-PSK (existing functionality).

An additional example sketch is included showing usage of the new functionality.

### DIFF
--- a/libraries/WiFi/src/WiFiMulti.cpp
+++ b/libraries/WiFi/src/WiFiMulti.cpp
@@ -114,20 +114,26 @@ bool WiFiMulti::addAP(const char *ssid, const char *passphrase) {
     if (!newAP.username) {
       log_e("[WIFI][APlistAdd] fail newAP.username == 0");
       free(newAP.ssid);
-      if (newAP.passphrase) free(newAP.passphrase);
+      if (newAP.passphrase) {
+        free(newAP.passphrase);
+      }
       return false;
     }
   } else {
     newAP.username = NULL;
   }
 
-  if (identity) { // Allow empty identity as not all WPA2-Enterprise networks require it
+  if (identity) {  // Allow empty identity as not all WPA2-Enterprise networks require it
     newAP.identity = strdup(identity);
     if (!newAP.identity) {
       log_e("[WIFI][APlistAdd] fail newAP.identity == 0");
       free(newAP.ssid);
-      if (newAP.passphrase) free(newAP.passphrase);
-      if (newAP.username) free(newAP.username);
+      if (newAP.passphrase) {
+        free(newAP.passphrase);
+      }
+      if (newAP.username) {
+        free(newAP.username);
+      }
       return false;
     }
   } else {
@@ -264,11 +270,10 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout, bool scanHidden) {
               }
               known = true;
               log_v("rssi_scan: %" PRIi32 ", bestNetworkDb: %d", rssi_scan, bestNetworkDb);
-              if (rssi_scan > bestNetworkDb) {                         // best network
+              if (rssi_scan > bestNetworkDb) {  // best network
 #if CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT
-                if (sec_scan == WIFI_AUTH_OPEN ||
-                    ((sec_scan != WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase) ||
-                    ((sec_scan == WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase && entry.username && entry.identity)) {
+                if (sec_scan == WIFI_AUTH_OPEN || ((sec_scan != WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase)
+                    || ((sec_scan == WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase && entry.username && entry.identity)) {
 #else
                 if (sec_scan == WIFI_AUTH_OPEN || entry.passphrase) {  // check for passphrase if not open wlan
 #endif
@@ -330,7 +335,9 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout, bool scanHidden) {
       if (bestNetworkSec != WIFI_AUTH_WPA2_ENTERPRISE) {
         WiFi.begin(bestNetwork.ssid, (bestNetworkSec == WIFI_AUTH_OPEN) ? NULL : bestNetwork.passphrase, bestChannel, bestBSSID);
       } else {
-        WiFi.begin(bestNetwork.ssid, WPA2_AUTH_PEAP, bestNetwork.identity, bestNetwork.username, bestNetwork.passphrase, NULL, NULL, NULL, -1, bestChannel, bestBSSID);
+        WiFi.begin(
+          bestNetwork.ssid, WPA2_AUTH_PEAP, bestNetwork.identity, bestNetwork.username, bestNetwork.passphrase, NULL, NULL, NULL, -1, bestChannel, bestBSSID
+        );
       }
 #else
       WiFi.begin(bestNetwork.ssid, (bestNetworkSec == WIFI_AUTH_OPEN) ? NULL : bestNetwork.passphrase, bestChannel, bestBSSID);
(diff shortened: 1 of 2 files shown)


me-no-dev:
Please make all your additions depend on `CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT`

lglina:
> Please make all your additions depend on `CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT`

I can do, but using a lot of preprocessor directives might get messy. Is the concern that:

- We want to produce a compile-time error if the caller is using the version of `add()` with `username` and `identity` without `CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT`? I suppose that's better than a runtime error. Or,
- We don't want the overhead of extra code and memory if `CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT` is disabled?
- Both? :)

me-no-dev:
Both, yes :)

libraries/WiFi/examples/WiFiMultiEnterprise/ci.yml
```diff
@@ -0,0 +1,4 @@
+requires_any:
+  - CONFIG_SOC_WIFI_SUPPORTED=y
+  - CONFIG_ESP_WIFI_REMOTE_ENABLED=y
+  - CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT=y
```
lucasssvaz:
```suggestion
requires:
  - CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT=y

requires_any:
  - CONFIG_SOC_WIFI_SUPPORTED=y
  - CONFIG_ESP_WIFI_REMOTE_ENABLED=y
```

Jason2866:
@lglina Are you going to do the requested changes? Without the PR will not get merged. If not please close. Thx!

lglina:
Yes, apologies for letting this hang around. I've made the changes but need to test. I'll see if I can get this done by the end of the week, else I'll close it and resubmit it later.

libraries/WiFi/src/WiFiMulti.cpp
```diff
@@ -207,14 +262,22 @@ uint8_t WiFiMulti::run(uint32_t connectTimeout, bool scanHidden) {
               known = true;
               log_v("rssi_scan: %d, bestNetworkDb: %d", rssi_scan, bestNetworkDb);
               if (rssi_scan > bestNetworkDb) {                                            // best network
-                if (_bAllowOpenAP || (sec_scan == WIFI_AUTH_OPEN || entry.passphrase)) {  // check for passphrase if not open wlan
+#if CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT
+                if ( sec_scan == WIFI_AUTH_OPEN ||
+                     ((sec_scan != WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase) ||
+                     ((sec_scan == WIFI_AUTH_WPA2_ENTERPRISE) && entry.passphrase && entry.username && entry.identity ) ) {
+#else
+                if ( sec_scan == WIFI_AUTH_OPEN || entry.passphrase ) {  // check for passphrase if not open wlan
```
Jason2866: Original code is altered, why? It was
```
if (_bAllowOpenAP || (sec_scan == WIFI_AUTH_OPEN || entry.passphrase)) {  // check for passphrase if not open wlan
```

lglina:
@me-no-dev OK, all changes are now dependent on `CONFIG_ESP_WIFI_ENTERPRISE_SUPPORT`. With enterprise support you get the new `add()` function prototype with optional `username` and `identity` parameters, and without enterprise support you get the existing function prototype (and save on the space of two char pointers in each entry in the AP list).

I've made a further change such that we never attempt to connect to an enterprise network without `username` and `identity` set (although `identity` can be an empty string). I realised that the enterprise versions of `STAClass::connect()` and `STAClass::begin()` (which calls `connect()`) both have the username and identity as `NULL`-default optional parameters, but they will in fact crash if you don't supply a username and/or identity if required (see `STA.cpp` from line 504). This should probably be a separate PR to fix, but in the meantime we can ensure `WiFiMulti` always provides those parameters.

Therefore we will ignore APs that have been found in the scan and are in our AP list where insufficient credentials have been provided via `add()` (passphrase for WPA2/3-PSK or username/password/identity for WPA2-Enterprise) and will `log_v` to say so.

As part of that change I've also removed the check against `_bAllowOpenAP` at line 266. This looks like a bug, as the effect in the current code would be to try to connect to passphrase-protected networks even without a passphrase being set whenever `_bAllowOpenAP` is true. We still, in previous lines, add all open APs to the AP list when `_bAllowOpenAP` is true. On that note, the existing functionality seems to be broken where we call `WiFi.begin()`, in that we require both `_bAllowOpenAP` and `WIFI_AUTH_OPEN` to send `NULL` as the passphrase to `begin()`, which implies if we add open APs to the AP list without setting `_bAllowOpenAP` that we'll call `begin()` with a passphrase instead. This probably works, though, as the passphrase will either be `NULL` or it will just be ignored for open APs. We should probably also remove the check for `_bAllowOpenAP` here (lines 327 and 332). The only thing `_bAllowOpenAP` should be doing is adding every open AP to our match list, and so it only ever needs to be checked at that point. Hope that's not too confusing!

libraries/WiFi/examples/WiFiMultiEnterprise/ci.yml
lglina: Done.

libraries/WiFi/src/WiFiMulti.cpp
lglina: @Jason2866 See my comment below.

Jason2866:
Thx for the explanation. Please do not mix a bug fix change in a PR with a new feature. Please open a PR to fix the bug and reference this PR to the bug fix PR which needs to be merged first.

lglina:
Will do.

me-no-dev:
@lglina will you please proceed with the bugfix PR?

lglina:
Bug fix PR opened as #12429

me-no-dev:
@lglina there are some merge conflicts. Please fix :) your other PR is merged

me-no-dev:
@lglina PTAL at the copilot suggestions

me-no-dev:
If you free something, please make sure that it points to NULL after its freed.

lglina:
The potential memory leak is something I missed, although it shouldn't normally happen. I'll fix that. I'd set the fields to `NULL` except that it's just a stack local and we `return` immediately after `free()`.

As for the other comments, the signature for `addAP()` now considers the case where the user is adding an open AP, WPA2/3 AP, or an enterprise AP. It's up to the user to pass the correct parameters, and a sensible error message is generated if they haven't provided the correct credentials. Treating `NULL` identity as an empty string is possible, but while that allows the connection to work where identity is not required and no string is provided, it also means the user may be confused when the identity is defaulted to the empty string when the network actually requires one. I'd argue for leaving it as-is.

lglina:
@me-no-dev @Jason2866 Thanks for your patience and assistance getting this merged :smiley:

PR #12295, opened 2026-01-30
ci(push): Optimize chunk counting to avoid empty chunks

lucasssvaz:
## Description of Change

This pull request updates the chunking logic in the `set_ci_output` function within `.github/scripts/get_affected.py` to better distribute sketches across CI chunks. The new approach ensures that chunks are created only as needed, preventing empty chunks and aligning with the logic used in `sketch_utils.sh`.

**Improvements to CI chunk calculation:**

* Updated the chunking algorithm to create one chunk per sketch when the number of sketches is less than or equal to the maximum allowed chunks, and to distribute sketches evenly across chunks when there are more sketches than the maximum, using ceiling division to avoid empty chunks.
* Added informative print statements to clarify how chunks are being allocated based on the number of sketches and maximum chunks.

## Test Scenarios

Tested locally

### DIFF
--- a/.github/scripts/get_affected.py
+++ b/.github/scripts/get_affected.py
@@ -940,12 +940,24 @@ def set_ci_output(print_vars: bool=True):
     try:
         # Write all output values
         f.write(f"recompile_preset={recompile_preset}\n")
-        chunk_count = len(affected_sketches)
-        if chunk_count > 0:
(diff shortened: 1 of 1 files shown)


P-R-O-C-H-Y (approved):
Looking good.

me-no-dev (approved):

PR #12314, opened 2026-02-04
PPP: Add burst of AT commands to lock autobaudrate on begin

gonzabrusco:
## Description of Change
This Pull Request improves the robustness of PPPClass::begin() when used with cellular modems that boot in autobaud mode, such as SIMCom modules.

After a power-on reset, these modems often require multiple "AT\r" commands before they can lock onto the host UART baud rate. The current implementation attempts to synchronize immediately using esp_modem_sync(), which may fail if the modem is still determining the baud rate, leading to unnecessary retries and mode switching.

This change introduces a short burst of "AT\r" commands before calling esp_modem_sync(), allowing the modem to reliably lock the baud rate first. In practice, this significantly improves synchronization reliability without affecting modems that do not use autobaud.

## Test Scenarios
I tested this change on a custom ESP32-based board equipped with a SIM7070 modem. Using a logic analyzer on the UART interface, I verified that the added AT command burst improves baud-rate locking and results in a more reliable initialization sequence.

### DIFF
--- a/.github/scripts/sketch_utils.sh
+++ b/.github/scripts/sketch_utils.sh
@@ -98,6 +98,10 @@ function build_sketch { # build_sketch <ide_path> <user_path> <path-to-ino> [ext
             shift
             debug_level="DebugLevel=$1"
             ;;
+        -td )
+            shift
+            ci_yml_dir=$1
+            ;;
         * )
             break
             ;;
@@ -300,9 +304,12 @@ function build_sketch { # build_sketch <ide_path> <user_path> <path-to-ino> [ext
                 exit "$exit_status"
             fi
 
-            # Copy ci.yml alongside compiled binaries for later consumption by reporting tools
-            if [ -f "$sketchdir/ci.yml" ]; then
-                cp -f "$sketchdir/ci.yml" "$build_dir/ci.yml" 2>/dev/null || true
+            # Copy ci.yml alongside compiled binaries for later consumption by reporting tools.
+            # For multi-device tests, ci.yml lives in the parent test directory (-td),
+            # not in individual sketch directories.
+            local ci_yml_source="${ci_yml_dir:-$sketchdir}"
+            if [ -f "$ci_yml_source/ci.yml" ]; then
+                cp -f "$ci_yml_source/ci.yml" "$build_dir/ci.yml" 2>/dev/null || true
             fi
 
             if [ -n "$log_compilation" ]; then
@@ -348,9 +355,12 @@ function build_sketch { # build_sketch <ide_path> <user_path> <path-to-ino> [ext
                 echo "ERROR: Compilation failed with error code $exit_status"
                 exit $exit_status
             fi
-            # Copy ci.yml alongside compiled binaries for later consumption by reporting tools
-            if [ -f "$sketchdir/ci.yml" ]; then
-                cp -f "$sketchdir/ci.yml" "$build_dir/ci.yml" 2>/dev/null || true
+            # Copy ci.yml alongside compiled binaries for later consumption by reporting tools.
+            # For multi-device tests, ci.yml lives in the parent test directory (-td),
+            # not in individual sketch directories.
+            local ci_yml_source="${ci_yml_dir:-$sketchdir}"
+            if [ -f "$ci_yml_source/ci.yml" ]; then
+                cp -f "$ci_yml_source/ci.yml" "$build_dir/ci.yml" 2>/dev/null || true
             fi
             # $ide_path/arduino-builder -compile -logger=human -core-api-version=10810 \
             #     -fqbn=\"$currfqbn\" \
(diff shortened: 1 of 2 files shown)


libraries/PPP/src/PPP.cpp
```diff
@@ -299,6 +299,11 @@ bool PPPClass::begin(ppp_modem_model_t model, uint8_t uart_num, int baud_rate) {
   }
 
   esp_modem_set_error_cb(_dce, _ppp_error_cb);
+  
+  // Send burst of AT to lock auto baurate on modem
+  for (int i = 0; i < 50; i++) {
+    esp_modem_at(_dce, "AT", NULL, 10);
+  }
```
lucasssvaz: IMO this should be only enabled if an argument of the constructor is set. @me-no-dev What do you think ?

gonzabrusco: @lucasssvaz Just to add a data point: sending a short burst of "AT" commands is safe and has no side effects on modems that do not rely on autobaud.
In practice, this acts as a no-op for fixed-baud devices while greatly improving initialization reliability for autobaud-based modems, so I don’t think it needs to be gated behind an additional constructor option.

me-no-dev: Can you please give models of modems that do require such burst?
I also agree that this should be optional or at least connected to the models that need it

TD-er: One more argument for making it optional.
When using flow-control the UART buffer may fill up and at some point may become blocking until it is able to flush its data.

gonzabrusco: My modem is SIM7070

gonzabrusco:
I took the time to capture a few traces so you can verify this behavior yourself. I’m using a SIM7070, and this is the exact test sketch I used:

```cpp
#include "Arduino.h"
#include "PPP.h"

#define UART_NUMBER 0
#define UART_RX 3
#define UART_TX 1
#define MODEM_PWRKEY 2 // IO2
#define PWRKEY_PULSE_MS 1200
#define MODEM_TURN_OFF_WAIT_MS 3300
#define MODEM_BOOT_WAIT_MS 7000

void setup() {
    pinMode(MODEM_PWRKEY, OUTPUT);
    digitalWrite(MODEM_PWRKEY, HIGH);

    // === WAIT TO BOOT ===
    delay(MODEM_BOOT_WAIT_MS);

    // === TURN OFF MODEM ===
    digitalWrite(MODEM_PWRKEY, LOW);
    delay(PWRKEY_PULSE_MS);
    digitalWrite(MODEM_PWRKEY, HIGH);

    // === WAIT TO OFF ===
    delay(MODEM_TURN_OFF_WAIT_MS);

    // === TURN ON MODEM ===
    digitalWrite(MODEM_PWRKEY, LOW);
    delay(PWRKEY_PULSE_MS);
    digitalWrite(MODEM_PWRKEY, HIGH);

    // === WAIT TO BOOT ===
    delay(MODEM_BOOT_WAIT_MS);

    PPP.setPins(UART_TX, UART_RX); // Not using control flow

    PPP.setApn("");

    PPP.setPin("");
   
    PPP.begin(PPP_MODEM_SIM7070, UART_NUMBER, 115200);
}

void loop() {
    // Nothing
}
```

The first capture shows the behavior without sending an initial burst of AT commands. As you can see, PPPClass struggles to synchronize with the modem because the modem does not respond until AT+CMUX=0 is issued. Most likely, the modem is still trying to lock the autobaud rate at that point.

Without burst

The second capture shows the behavior with an AT burst sent prior to synchronization. In this case, the modem synchronizes immediately, and `esp_modem_sync` succeeds on the first attempt.

With burst

For reference, here are the _Saleae Logic_ capture files in case you want to dig deeper into this:
[Captures.zip](https://github.com/user-attachments/files/25209585/Captures.zip)

me-no-dev:
we agree with all. We are just asking that this is made optional :)

gonzabrusco:
> we agree with all. We are just asking that this is made optional :)

How would you suggest implementing this as an optional feature? At construction time, as an argument to begin(), or enabled by default for SIMCom modules?

I would like to emphasize that sending an initial AT command is effectively a no-op and should not have any negative impact on other modems. That said, if you prefer to keep this behavior optional, I’m happy to implement it accordingly—just let me know how you envision the option and I can update this PR.

me-no-dev:
```cpp
class PPPClass {
  public:
    void sendAtBurst(bool en) {
      _sendAtBurst = en;
    }
  private:
    bool _sendAtBurst;
}
```

Then check in begin if `_sendAtBurst` is `true` and do the burst

TD-er:
> ```c++
> bool _sendAtBurst;
> ```

And don't forget to initialize at false.

gonzabrusco:
@me-no-dev PTAL

lucasssvaz (approved):
LGTM

PR #12319, opened 2026-02-04
Add Serial1 TX1/RX1 pin definitions for RAK3112

Change1213qwer:
## Description
Added TX1 (GPIO43) and RX1 (GPIO44) pin definitions for Serial1 in `pins_arduino.h` for the RAKwireless RAK3112 variant.

## Changes
- Added `static const uint8_t TX1 = 43;` 
- Added `static const uint8_t RX1 = 44;`
- These pins are available on RAK3112 for additional UART

## Testing
Tested with RAK5860 module and other serial devices. Confirmed that existing Serial (UART0) functionality remains unaffected.


P-R-O-C-H-Y:
Closing this PR, as there is no addition to the RAK variant, defining same TX and RX pins twice.

PR #12336, opened 2026-02-09
fix(ppp): Use proper response length for AT commands

me-no-dev:
This pull request increases the buffer size used for AT command responses in the `PPPClass` methods to improve compatibility with longer modem responses. Instead of using a fixed-size buffer, it now uses a buffer sized according to the `CONFIG_ESP_MODEM_C_API_STR_MAX` configuration value.

**Buffer size improvements:**

* Updated the buffer allocation in `PPPClass::cmd(const char *at_command, int timeout)` to use `CONFIG_ESP_MODEM_C_API_STR_MAX+1` instead of a fixed size of 128.
* Updated the buffer allocation in `PPPClass::cmd(const char *at_command, String &response, int timeout)` to use `CONFIG_ESP_MODEM_C_API_STR_MAX+1` instead of a fixed size of 128.

### DIFF
--- a/libraries/PPP/src/PPP.cpp
+++ b/libraries/PPP/src/PPP.cpp
@@ -737,7 +737,7 @@ bool PPPClass::sms(const char *num, const char *message) {
 String PPPClass::cmd(const char *at_command, int timeout) {
   PPP_CMD_MODE_CHECK(String());
 
-  char out[CONFIG_ESP_MODEM_C_API_STR_MAX+1] = {0};
+  char out[CONFIG_ESP_MODEM_C_API_STR_MAX + 1] = {0};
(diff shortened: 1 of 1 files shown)


lucasssvaz (approved):

TD-er:
I wonder how `esp_modem_at` knows the `out` array is now longer.
If this array was too short before, wouldn't that have lead to crashes?

Or does this function internally also use the `CONFIG_ESP_MODEM_C_API_STR_MAX` ?


Edit:
Never mind....
Found it
https://github.com/espressif/esp-protocols/blob/c3bedf042afd1128055a1f5f7d682f45cbd7a37d/components/esp_modem/src/esp_modem_c_api.cpp#L221

PR #12372, opened 2026-02-17
feat(format): Add example to print proper format specifiers

lucasssvaz:
## Description of Change

Format specifiers depend on the architecture implementation.

This pull request adds a new example sketch, `FormatIdentifier.ino`, to the ESP32 Arduino library. The sketch demonstrates which format specifiers to use with `Serial.printf` for various C++ data types, helping developers correctly print values of different types.

New example for format specifiers:

* Added `FormatIdentifier.ino`, which reports the appropriate `printf` format specifiers, type names, and sizes for native integer types, fixed-width integer types, floating-point types, pointers, and strings using helper template functions.
* The sketch outputs a table to the serial monitor, making it easier to identify the correct format specifier for each type on the ESP32.

## Test Scenarios

Tested locally

### DIFF
--- a/libraries/ESP32/examples/FormatSpecifier/FormatSpecifier.ino
+++ b/libraries/ESP32/examples/FormatSpecifier/FormatSpecifier.ino
@@ -1,0 +1,181 @@
+/*
+ * FormatSpecifier Example
+ *
+ * Format specifiers can vary depending on the architecture and the compiler.
+ *
+ * This sketch will print the format specifiers for the different types for
+ * the current architecture and compiler.
+ */
+
+#include <Arduino.h>
+#include <inttypes.h>
+#include <type_traits>
+#include <stddef.h>
+
+template<typename T> void reportIntegral(const char *name) {
+  Serial.printf("%-18s | size: %2u | format: ", name, (unsigned)sizeof(T));
+
+  // Character types
+  if constexpr (std::is_same_v<T, char>) {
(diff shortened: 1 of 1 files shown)


SuGlider:
I just don't see how is it deeply related to Arduino API.
"Aduinish way" for printing is simpler... This is very C++

lucasssvaz:
> I just don't see how is it deeply related to Arduino API.
> "Aduinish way" for printing is simpler... This is very C++

The code itself is not for printing. It will create a table at runtime with the proper format specifiers.

The format specifiers depend on the compiler and architecture so with this sketch we can find the proper specifier to use for each type no matter what architecture and compiler our chips use, past and future.

SuGlider:
> > I just don't see how is it deeply related to Arduino API.
> > "Aduinish way" for printing is simpler... This is very C++
> 
> The code itself is not for printing. It will create a table at runtime with the proper format specifiers.
> 
> The format specifiers depend on the compiler and architecture so with this sketch we can find the proper specifier to use for each type no matter what architecture and compiler our chips use, past and future.

OK. It is still just C++ example. Not Arduino.

me-no-dev:
I also agree that this does not bring much value. Proper thing would be to write documentation on how to use the `PRIx` types (and actually convert all our prints to them)

lucasssvaz:
Ok, will open a PR for that

PR #12385, opened 2026-02-21
fix(webserver): reset _contentLength after _streamFileCore send

Sick-E:
## Description of Change
`_streamFileCore` calls `setContentLength(fileSize)` before `send()`, which sets the member variable `_contentLength` to the file size. After `send()` completes, `_contentLength` is never reset, causing it to leak into subsequent response handlers. Any handler calling `send()` after a `streamFile` response will incorrectly inherit the previous file's size as `Content-Length`, resulting in `ERR_CONTENT_LENGTH_MISMATCH` in the browser.

Fix: reset `_contentLength` to `CONTENT_LENGTH_NOT_SET` at the end of `_streamFileCore`.

## Test Scenarios
Tested on ESP32 with framework-arduinoespressif32 via PlatformIO.

Reproduction scenario:
1. HTML page with `<link rel="icon">` pointing to a file served via `streamFile`
2. On page load, browser automatically fetches favicon via GET — `_contentLength` is set to favicon file size
3. Subsequent POST request handler calls `send(200, "text/plain", "Success")` without explicitly calling `setContentLength`
4. Response incorrectly contains `Content-Length` equal to favicon size → `ERR_CONTENT_LENGTH_MISMATCH`

After fix: `_contentLength` is reset after each `streamFile` response, subsequent handlers receive correct `Content-Length`.

## Related links
None

### DIFF
--- a/.github/workflows/tests_hw_wokwi.yml
+++ b/.github/workflows/tests_hw_wokwi.yml
@@ -13,6 +13,7 @@ permissions:
 jobs:
   get-artifacts:
     name: Get required artifacts
+    if: github.event.workflow_run.conclusion != 'cancelled'
     runs-on: ubuntu-latest
     permissions:
       actions: read
--- a/.gitlab/workflows/hw_test_template.yml
+++ b/.gitlab/workflows/hw_test_template.yml
@@ -46,7 +46,6 @@ hw-test-template:
 
   script:
     - echo "Using binaries for $TEST_CHIP"
-    - ls -laR ~/.arduino/tests || true
     - |
       set -e
       rc=0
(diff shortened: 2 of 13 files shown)


.gitignore
```diff
@@ -63,3 +63,4 @@ libraries/Insights/examples/*/*.ino.zip
 /affected_examples.txt
 /ctags_*
 .github/copilot-instructions.md
+/.idea/
```
lucasssvaz: Please move it to the proper section

Sick-E: Moved to IDE section.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -536,6 +536,10 @@ void WebServer::setContentLength(const size_t contentLength) {
   _contentLength = contentLength;
 }
 
+void WebServer::resetContentLength() {
+  _contentLength = CONTENT_LENGTH_NOT_SET;
+}
+
```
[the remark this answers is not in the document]
Sick-E: Thanks for catching that. I've reconsidered the approach — instead of introducing a new method, I'm reusing the existing setContentLength(CONTENT_LENGTH_NOT_SET) call directly at the end of _streamFileCore. No new declaration in WebServer.h needed.

PR #12419, opened 2026-03-02
adding xteink x4 board

BlitzCityDIY:
## Description of Change

adding Xteink X4 eReader board def. uses esp32-c3

## Test Scenarios

tested various sketches and onboard peripherals

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -513,9 +513,9 @@ esp32p4.menu.PartitionScheme.app5M_little24M_32MB.upload.maximum_size=4718592
 esp32p4.menu.PartitionScheme.app13M_data7M_32MB=32M Flash (13MB APP/6.75MB SPIFFS)
 esp32p4.menu.PartitionScheme.app13M_data7M_32MB.build.partitions=default_32MB
 esp32p4.menu.PartitionScheme.app13M_data7M_32MB.upload.maximum_size=13107200
-esp32p4.menu.PartitionScheme.esp_sr_16=ESP SR 16M (3MB APP/7MB SPIFFS/2.9MB MODEL)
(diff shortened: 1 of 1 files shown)


BlitzCityDIY:
@brentru fyi

Jason2866:
Since the device has only a CDC interface and has 16MB flash it makes no sense to provide not valid menu selection entrys.

brentru:
> to provide not valid menu selection entrys.

@Jason2866 Could you elaborate? Partition entries are provided.

Jason2866:
@brentru Device needs no selection for CDC modes. Flash size should be not available to select too, set fix to 16MB

EDIT: upload speed is a useless selection too. CDC is fixed in speed.

brentru:
@Jason2866  Gotcha, thanks for clarifying. 

@BlitzCityDIY  - Are you OK with these suggested changes?

boards.txt
tonhuisman, line 518: Why do these settings (and others below) need to be changed? They don't address the new board being added.
If you have personal changes/preferences, other users probably don't want these (I for sure don't want this changed...)

BlitzCityDIY: this was not intentional. i had tried to pull in the boards.txt from main here in the PR because i realized the branch was out of date but it did not work as expected. i'm going to close and rebase to fix this, thank you

tonhuisman: The essence is in pulling (aka syncing or rebasing) the latest `develop`/`main` or whatever branch you want to work on, before creating the branch and PR.
It's usually easily fixable, I'm surprised how often ppl close their PR after a friendly comment to nudge them in another, (more desirable), direction 🤔

PR #12420, opened 2026-03-02
feat(webserver): Add send() overload for Stream objects

### LINKED ISSUE #12275: ESP32 WebServer ==> Send gzip File
### Related area

Library Support: WebServer

### Hardware specification

ESP32

Hi,

I wanted to use a WebServer code used on the ESP8266 similar on the ESP32.
I want the same usage in same way than on ESP8266 via gzip file.

In the ESP8266 I have a code part where I send a gzip file like this:

```
server.on("/", [] () {
      File file = FILESYSTEM.open("index.html.gz", "r");
      if(!file) {
         server.send(500, "text/plain", "File not found!");
         return;
      }
      server.sendHeader(F("Content-Encoding"), "gzip");
      server.send(200, "text/html", file);
      file.close();
   });

```
 which uses this implementation.

```
void send(int code, const char* content_type, Stream* stream, size_t content_length = 0);
  void send(int code, const char* content_type, Stream& stream, size_t content_length = 0) {
    send(code, content_type, &stream, content_length);
  }
```

Could you please take over this methods to be able to use this in same way on the ESP32 core, please?

_No response_

### Additional context

_No response_

bittoby:
## Description of Change

Add a `send()` overload to `WebServer` that accepts a `Stream&` parameter, matching the ESP8266 WebServer API.

This enables direct portability of ESP8266 code that sends file/stream content via `server.send(200, "text/html", file)` without requiring users to rewrite their code to use `streamFile()`.

**New signature:**

```cpp
void send(int code, const char *content_type, Stream &stream, size_t content_length = 0);
```

When `content_length` is omitted or zero, it falls back to `stream.available()`. Only a `Stream&` (reference) overload is provided - a `Stream*` (pointer) overload was intentionally excluded to avoid C++ overload ambiguity with the existing `const char*` overload when `NULL`/`nullptr` is passed.

## Changed Files

| File | Change |
|---|---|
| `libraries/WebServer/src/WebServer.h` | Add `send(Stream&)` method declaration |
| `libraries/WebServer/src/WebServer.cpp` | Add `send(Stream&)` method implementation - builds HTTP headers, then streams body via `_currentClient.write(stream)` |
| `tests/validation/webserver/webserver.ino` | New Unity-based validation test sketch with 5 test cases |
| `tests/validation/webserver/test_webserver.py` | New pytest wrapper for Unity output |
| `tests/validation/webserver/ci.yml` | New CI config requiring WiFi support, no QEMU |

## Test Scenarios

- Validation test added at `tests/validation/webserver/` using Unity framework on WiFi softAP (self-contained, no external router required)
- Tested scenarios: text stream, binary stream (data integrity), explicit content_length, empty stream (headers only), and regression for existing `send(String)` overload
- Requires `CONFIG_SOC_WIFI_SUPPORTED=y`

## Related links

Closes #12275


Crystora:
@lucasssvaz Could you please review this PR? I think the testing error doesn't seem to relate my code.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -694,6 +694,18 @@ void WebServer::send(int code, const char *content_type, const char *content) {
   send(code, content_type, passStr);
 }
 
+void WebServer::send(int code, const char *content_type, Stream &stream, size_t content_length) {
+  String header;
+  if (content_length == 0) {
+    content_length = stream.available();
```
me-no-dev, line 699: is it worth continuing execution if content length here returns 0? Maybe check and just return.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -694,6 +694,18 @@ void WebServer::send(int code, const char *content_type, const char *content) {
...
+  }
+  _prepareHeader(header, code, content_type, content_length);
+  _currentClientWrite(header.c_str(), header.length());
+  if (content_length) {
```
me-no-dev: following the comment above, this check should not be needed?

libraries/WebServer/src/WebServer.cpp
```diff
@@ -694,6 +694,18 @@ void WebServer::send(int code, const char *content_type, const char *content) {
...
+    _currentClient.write(stream);
```
me-no-dev: this method will again check how much is available and will be sending data as long as there is something in the buffer. It is totally possible that the amount available has changed in between your check above and what happens in Client::write(Stream&). For this PR to make sense, a new method needs to be added to Client that accepts two arguments `Client::write(Stream &s, size_t length)`

Crystora:
@me-no-dev Thanks for your feedback. I updated. Please check again.

libraries/WebServer/src/WebServer.cpp
```diff
@@ -694,6 +694,20 @@ void WebServer::send(int code, const char *content_type, const char *content) {
   send(code, content_type, passStr);
 }
 
+void WebServer::send(int code, const char *content_type, Stream &stream, size_t content_length) {
+  if (!content_length) {
+    content_length = stream.available();
+    if (!content_length) {
+      log_e("Stream has no data available");
+      return;
```
me-no-dev, line 702: maybe instead of just returning, you could `return send(204);` to let the application know that you do not have any data and close the request. My idea of just returning is not great, because the request will not be correctly answered

Crystora:
@me-no-dev Updated!

Crystora:
Any update for me, please

Crystora:
@me-no-dev I have added status code following Copilot feedback. It's ready to merge now. Thanks

me-no-dev:
@bittoby some patience from your side is required

Crystora:
This testing error doesn't seem to relate my code. I'd appreciate you check testing log when you have time.

Crystora:
@Lucasczm @me-no-dev Need to update more? please let me know. thanks

Crystora:
@me-no-dev please give me feedbacks

Crystora:
@lucasssvaz Some testing fails doesn't not relate to my changes. Please check the error log. 
thank you

tests/validation/webserver/test_webserver.py
lucasssvaz, line 1: This needs to be a multi-dut test as the tests are automatically run autonomously in our internal CI. Take a look at the wifi_ap test as an example:

https://github.com/espressif/arduino-esp32/tree/master/tests/validation/wifi_ap
and
https://docs.espressif.com/projects/arduino-esp32/en/latest/contributing.html#adding-a-multi-device-test-suite

lucasssvaz:
@bittoby Don't worry. It is related to changes in IDF and not your code.

Crystora:
@lucasssvaz I updated test to multi DUT format. Please review again

lucasssvaz:
> @lucasssvaz I updated test to multi DUT format. Please review again

I will review and test when I have free time. I am currently on vacation.

Crystora:
@lucasssvaz Thanks for approving.  Would you merge this PR? 🙏

lucasssvaz:
@bittoby Usually the PRs are merged in batches after the merge is approved. It will probably be merged in the next week with other fixes.

me-no-dev:
@bittoby there is something to fix: https://github.com/espressif/arduino-esp32/actions/runs/23025192533/job/66871245880#step:8:27

PR #12439, opened 2026-03-10
fix(board): update pins and OTA of senseBox Eye

PaulaScharf:
## Description of Change
We mad some changes to the pin definition and also the OTA script for our board.

## Test Scenarios
I have tested this pull request on the senseBox Eye board.

### DIFF
--- a/variants/sensebox_eye/pins_arduino.h
+++ b/variants/sensebox_eye/pins_arduino.h
@@ -10,10 +10,10 @@
 #define USB_SERIAL       ""  // Empty string for MAC address
 
 // Default USB FirmwareMSC Settings
-#define USB_FW_MSC_VENDOR_ID        "senseBox"     // max 8 chars
-#define USB_FW_MSC_PRODUCT_ID       "Eye"  // max 16 chars
-#define USB_FW_MSC_PRODUCT_REVISION "1.4"      // max 4 chars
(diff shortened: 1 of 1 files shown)


variants/sensebox_eye/APOTA.ino
```diff
@@ -2,7 +2,6 @@
 // APOTA opens an access point which waits to receive a .bin file on /sketch.
 // After successful upload, the file is written to OTA0_Partition, and the microcontroller reboots to the newly uploaded sketch.
 
-#include <Arduino.h>
```
lucasssvaz, line 5: Leave the Arduino include in so code scanners can properly trace the code.

variants/sensebox_eye/APOTA.ino
```diff
@@ -213,7 +212,7 @@ void setupOTA() {
         if (server.clientContentLength() > 0) {
           fsize = server.clientContentLength();
         }
-        Serial.printf("Receiving Update: %s, Size: %lu\n", upload.filename.c_str(), (unsigned long)fsize);
```
lucasssvaz, line 216: This is a regression on the format specifiers. Please check https://docs.espressif.com/projects/arduino-esp32/en/latest/guides/coding_guidelines.html#format-specifiers

PaulaScharf: thank you for the review. I reverted both changes

PR #12466, opened 2026-03-20
feat(USB): Add support for USB Audio

### LINKED ISSUE #12053: Add USB Audio Class support (UAC2) for ESP32-S2/S3
### Related area

USB

### Hardware specification

ESP32-S2, ESP32-S3

Yes. The Arduino ESP32 core provides wrappers for USB device classes like CDC, HID, and MSC, but USB Audio Class is completely unavailable.

While TinyUSB supports USB Audio via CFG_TUD_AUDIO, this option is not enabled in the arduino-esp32 TinyUSB build. The audio class code is simply not compiled in.

  **This means:**
  - Audio callbacks like tud_audio_rx_done_post_read_cb() don't exist
  - Audio descriptors cannot be registered
  - There is no workaround within the Arduino framework

  Even with custom descriptor code, USB Audio cannot be implemented without rebuilding the ESP32 Arduino core from source with CFG_TUD_AUDIO=1 — effectively forcing users to abandon Arduino for ESP-IDF.

 Add a `USBAudio` class following the same pattern as `USBCDC`, `USBHIDKeyboard`, etc., supporting **input, output, and bidirectional audio**.

  #### Desired API

  **USB Speaker (host → ESP32 → I2S DAC):**

  ```cpp
  #include "USB.h"
  #include "USBAudio.h"

  USBAudioOutput Speaker;

  void onSpeakerData(uint8_t* data, size_t len) {
    i2s_write(I2S_NUM_0, data, len, &written, portMAX_DELAY);
  }

  void setup() {
    Speaker.begin(44100, 16, 2);  // 44.1kHz, 16-bit, stereo
    Speaker.onReceive(onSpeakerData);
    USB.begin();
  }

  USB Microphone (I2S ADC → ESP32 → host):

  #include "USB.h"
  #include "USBAudio.h"

  USBAudioInput Microphone;

  void loop() {
    i2s_read(I2S_NUM_0, buffer, sizeof(buffer), &read, portMAX_DELAY);
    Microphone.write(buffer, read);
  }

  void setup() {
    Microphone.begin(48000, 16, 1);  // 48kHz, 16-bit, mono
    USB.begin();
  }

  Full USB Audio Interface (bidirectional):

  #include "USB.h"
  #include "USBAudio.h"

  USBAudioInterface AudioCard;

  void setup() {
    AudioCard.beginOutput(44100, 16, 2);  // Speaker
    AudioCard.beginInput(44100, 16, 2);   // Microphone
    AudioCard.onReceive(onSpeakerData);
    USB.begin();
  }

  Key features

  - Separate classes for output (USBAudioOutput), input (USBAudioInput), and combined (USBAudioInterface)
  - Simple begin(sampleRate, bitDepth, channels) initialization
  - Callback-based receive, buffer-based transmit
  - Automatic UAC2 descriptor generation
  - Optional volume/mute control support
  - Works seamlessly with existing USB.begin() flow



 1. **Raw TinyUSB implementation** — Not possible within Arduino. Even with custom descriptor code, `CFG_TUD_AUDIO` is not enabled in the arduino-esp32 TinyUSB build, so audio callbacks and functions don't exist.

  2. **Rebuilding arduino-esp32 from source** — Users could clone the core and enable `CFG_TUD_AUDIO=1`, but this defeats the purpose of using Arduino and requires maintaining a custom fork.

  3. **External libraries** — No maintained Arduino ESP32 USB Audio library exists. The [ESP32-USB-Audio](https://github.com/pschatzmann/ESP32-USB-Audio) repo by pschatzmann is experimental and requires ESP-IDF modifications.

  4. **ESP-IDF directly** — The only currently working option, but loses all Arduino ecosystem benefits (libraries, simple API, community support) and requires significant migration effort.

  None of these alternatives provide the Arduino experience that makes ESP32 accessible to hobbyists and rapid prototyping.

### Additional context

#### Use cases this would enable

  - **USB DAC** — ESP32-S3 as plug-and-play USB sound card → I2S DAC (PCM5102, MAX98357, ES9038)
  - **USB Microphone** — ESP32-S3 as USB recording device from I2S/PDM mic (INMP441, SPH0645)
  - **USB Audio Interface** — Bidirectional audio for effects processors, synthesizers, loopers
  - **USB MIDI + Audio** — Combined MIDI controller with audio capabilities

  #### Technical foundation already exists

  - TinyUSB has mature Audio Class support (`CFG_TUD_AUDIO`)
  - ESP-IDF's TinyUSB component includes all necessary audio headers
  - Similar wrappers already exist for HID/CDC/MSC in arduino-esp32
  - The implementation pattern is well-established in the codebase

  #### Community interest

  USB Audio on ESP32 is frequently requested in forums and discussions. Native support would unlock a whole category of audio projects for the Arduino community.

me-no-dev:
This pull request adds support for USB Audio Class (UAC) devices to the ESP32 Arduino core, including a new `USBAudioCard` API, descriptors, and an example. The changes introduce a new USB interface type for audio, provide the necessary descriptors for both UAC1 and UAC2, and include an example sketch demonstrating usage.

**USB Audio Class (UAC) Support:**

* Added a new `USB_INTERFACE_AUDIO` to the `tinyusb_interface_t` enum and updated the `tinyusb_interface_names` array to include "AUDIO" for improved interface identification. [[1]](diffhunk://#diff-dd9bbc60d81f2e139c1f70d3460754740eb104c1f7eb6470ee0e81a0e35515e4R100) [[2]](diffhunk://#diff-d097e4c789a457b03434bb0c64f666c27ec3dda012b39dbf036924904fe95bacL794-R794)
* Introduced `USBAudioCard` class in `USBAudioCard.h` to provide an API for USB audio device functionality, including event handling, volume/mute controls, data streaming, and configuration.
* Added `USBAudioCardDescriptors.h` containing descriptor templates for both USB Audio Class 1 (UAC1) and Class 2 (UAC2), enabling flexible device descriptor definitions for headsets and other audio peripherals.

**Example and Build Configuration:**

* Added a new example sketch `AudioCard.ino` demonstrating how to use the `USBAudioCard` API for a USB headset with stereo speaker and microphone, including event handling and I2S integration.
* Included a `ci.yml` file for the example, specifying required configuration flags for USB OTG and TinyUSB audio support.

Closes https://github.com/espressif/arduino-esp32/issues/12053


sauloverissimo:
Great PR! 👏 I just read through the whole code and having USB Audio directly in arduino-esp32 opens up so many possibilities, especially for composite devices. I've been working on a project with USB MIDI/MIDI 2.0 on the ESP32-S3 and a composite Audio + MIDI device would be amazing.

Looking at the descriptors with that composite scenario in mind, I noticed a few things that will break when `_itfnum > 0`:

The IAD in USBAudioCardDescriptors.h uses `(_itfnum+3)` for `_nitfs`, but that field expects the count of interfaces, not an index. So if there's a CDC before it (occupying 0 and 1), `_itfnum` comes in as 2 and the IAD ends up declaring 5 interfaces instead of 3. I think it should just be a fixed `3`.

In `tud_audio_set_itf_cb`, the comparison `itf == 1` is hardcoded, so it only works when Audio is the first interface. It should use `_itf_num + 1` for speaker and `_itf_num + 2` for mic. Also the current logic never really distinguishes mic from speaker when both are active, the else always falls into MIC.

And in the UAC1 path, the `_totallen` of the CS AC Header includes `TUD_AUDIO10_DESC_FEATURE_UNIT_LEN(1)` for a mic Feature Unit that's never actually emitted in the descriptor sequence. The declared length won't match the actual bytes, so the host may reject the enumeration.

I've run into similar issues with composite descriptors while working with tusb_ump (USB MIDI 2.0 on TinyUSB). I'd like to help with this, if you want support, just let me know and authorize me.

me-no-dev:
Thanks @sauloverissimo ! I'll take all those into account. Keep in mind this is still in early WIP state. Started from somewhere and slowly progressing through the USB specs involved. Everything used to be hardcoded and I have not yet tested with other devices in the mix. When I open this for review, I hope all those things will be taken care of and Audio will work with other things in the mix. How do you see this as an Arduino API otherwise? I also plan for SPK-Only and MIC-Only options (with separate descriptors). Will see how that will go :smile:

me-no-dev:
> Also the current logic never really distinguishes mic from speaker when both are active, the else always falls into MIC.

If SPK is active, it's always the first interface. Logic might be a bit flawed, but that will get fixed in the process of implementing all options.

me-no-dev:
> And in the UAC1 path, the _totallen of the CS AC Header includes TUD_AUDIO10_DESC_FEATURE_UNIT_LEN(1) for a mic Feature Unit that's never actually emitted in the descriptor sequence.

Yeah. Copilot warned me of that and it seems it's an error in the original TinyUSB example. Funny enough my computer did not complain 🤷 will be fixed though

sauloverissimo:
@me-no-dev Man, so many things came to my mind but let me tell you a few so you can see how what you're working on is awesome and completely changes the MIDI scenario for ESP32.

The most obvious one: composite Audio+MIDI. Today if you want to build a synth with ESP32-S3, you need two USB cables or build the descriptors by hand. With an Audio class in arduino-esp32, a single USB cable and the computer sees the device as a sound card and a MIDI controller at the same time. Open Ableton, audio already plays and the knobs already send CC. Think of a homemade Teenage Engineering OP-1. Same case for a guitar pedalboard: the pedal receives audio over USB, processes it on the ESP32 (delay, reverb), sends the processed audio back, and at the same time you control the parameters via MIDI over the same cable. With your work, ESP32-S3 and P4 enter a game that was restricted to Daisy Seed and Teensy. This is really strong for the MIDI and MIDI 2.0 scenario. You know what I mean?

And the composite fixes that will come out of this work benefit everyone. The issues I mentioned (IAD, hardcoded indices) are generic problems of "two devices on the same USB". If this gets solved well for Audio, any class that comes after (MIDI, HID, CDC) already inherits the correct pattern. It's like fixing the plumbing of the building, it benefits all the apartments. Think of a DJ controller: Audio for headphone pre-listen, MIDI for faders, HID for the jog wheel. Three classes on the same device. If composite doesn't work right, none of that happens.

Another thing that came to mind: audio as sonic feedback for MIDI. A USB MIDI keyboard that has no sound output of its own depends on the computer to hear what was played. But if the device has an Audio class, it can generate a local sound preview and send it to the host as USB audio. You're on stage, the computer crashed, the MIDI keyboard alone is silent. With audio composite at least a basic sonic fallback exists.

Now a more technical one: clock sync. USB Audio uses the USB SOF to synchronize samples, it's a 1ms or 125us clock coming from the host. If the MIDI device uses that same clock to timestamp messages, audio and MIDI stay perfectly aligned. Think of an ESP32 drum machine: the kick fires as a MIDI note and the sample goes out through USB Audio. If the two aren't on the same clock, the visual kick in the DAW arrives a few ms before or after the sound. With shared clock, zero drift. That's the difference between "tight" and "flamming". MIDI 2.0 UMP has timestamp fields precisely for this, having the audio clock available on the same device makes those timestamps actually useful.

And the Feature Unit you're implementing for volume and mute is conceptually very similar to Property Exchange in MIDI 2.0. The host asks "what's the volume?", the device responds, the host sends "set it to 80%". In MIDI 2.0 PE the host asks "what patches do you have?", the device responds with JSON. The get/set with callbacks pattern is the same. A synth editor on the computer that pulls all parameters automatically, patch names, knob values, firmware version. Today this is done with messy SysEx. With PE it's structured, and the API design you do for Feature Unit serves as a reference.

There's more: sample rate negotiation is analogous to Protocol Negotiation in MIDI 2.0. When you connect an audio interface, the host and device negotiate "do you support 44.1? 48? 96?". In MIDI 2.0 it's "do you speak MIDI 1.0? MIDI 2.0? UMP?". A controller that works both in modern DAWs (per-note expression, 32-bit velocity) and in vintage synths that only speak MIDI 1.0 needs that negotiation. The API design for one informs the other, both need a begin() that declares capabilities and callbacks for when the host chooses.

Lastly, Audio streaming and MIDI 2.0 in volume. USB Audio handles continuous streams of isochronous data, samples arriving non-stop in real time. MIDI 2.0 with UMP can generate large volumes of data (Property Exchange sending JSON, MIDI-CI Discovery with multiple endpoints). Think of an Ableton Push that needs to update 64 pads with colors, receive state of 200 parameters, sync everything in real time. The data volume approaches a stream, and the buffer management patterns from Audio apply.

So in short, what you're doing with USB Audio isn't just audio, it's the infrastructure that's missing for the ESP32 to be a serious platform for complete musical devices. I'm working on the MIDI 2.0 side in TinyUSB and these two things together are the combo this ecosystem is missing.

sauloverissimo:
@me-no-dev Sorry for the wall of text. I really got enthusiastic about this one hahahah

sauloverissimo:
@me-no-dev About the API question, here's how I see it.

The `USBAudioCard` pattern is solid: `begin()` to set up, callbacks for events, stream methods for data. I'd keep that same shape for SPK-Only and MIC-Only too. Maybe something like `begin(UAC_SPK_STEREO)` vs `begin(UAC_HEADSET)` so the user picks the mode and the descriptors adjust behind the scenes. One class, different modes, less API surface.

The first thing people will want to do with USB Audio on ESP32 is combine it with MIDI. A synth that shows up as a sound card AND a MIDI controller on one USB cable. Today you need a Teensy for that. If arduino-esp32 makes Audio + MIDI work together with two `begin()` calls and that's it, the ESP32-S3 (and especially the P4 with High-Speed) becomes the go-to platform for musical devices.

And this connects to something bigger. The MIDI 2.0 spec is now finalized and already working on Windows 11. The embedded side is almost empty, nobody has a complete implementation yet. Think about how big that market is. The USB transport for MIDI 2.0 uses the same TinyUSB class driver structure, same alt-setting mechanism, same descriptor pattern. So any composite work that lands for Audio + MIDI 1.0 will naturally extend to MIDI 2.0 when it arrives in TinyUSB.

I've been working on MIDI 2.0 drivers. And the ESP32 has something no other musical platform has: WiFi, BLE, ESP-NOW built in. I've been doing some experiments with that, used it for MIDI wireless and mesh between devices. Put that together with Audio + USB + MIDI composite and the ESP32 is in a category Teensy and Daisy can't reach. Would be amazing to unlock this! Happy to help on the composite + MIDI side.

me-no-dev:
> Maybe something like begin(UAC_SPK_STEREO) vs begin(UAC_HEADSET) so the user picks the mode and the descriptors adjust behind the scenes.

This will happen in the constructor where you define number of input and output channels. This happens there because descriptor needs to be picked before USB starts and if CDC is enabled for Serial communication, it starts before `setup()` runs.
```cpp
USBAudioCard uac(48000, UAC_BPS_16, UAC_SPK_STEREO, UAC_MIC_MONO); //Headset
USBAudioCard uac(48000, UAC_BPS_16, UAC_SPK_STEREO, UAC_MIC_STEREO); //Stereo IN+OUT
USBAudioCard uac(48000, UAC_BPS_16, UAC_SPK_STEREO, UAC_MIC_NONE); //Stereo speaker 
USBAudioCard uac(48000, UAC_BPS_16, UAC_SPK_MONO, UAC_MIC_NONE); //Mono speaker
USBAudioCard uac(48000, UAC_BPS_16, UAC_SPK_NONE, UAC_MIC_MONO); //Microphone only
```

me-no-dev:
Tested on Windows 11, Ubuntu and Mac. Works as expected

Decimon1223:
I see light! Cant wait to finally diy my own usb DAC hsing esp32 and pcm5102a

sauloverissimo:
Been quietly working on integrating this with the ESP32 MIDI ecosystem on my side. USB Audio + MIDI composite devices is exactly the kind of thing that opens up a whole new category of projects.

Great to see it tested and working across all three platforms already @me-no-dev. Looking forward to the merge!

sauloverissimo:
Hey @me-no-dev, I've been building a project using this branch with arduino as an ESP-IDF component (USB Audio + MIDI composite device) and it's looking great!

While setting up the TinyUSB integration, I noticed a tiny thing in `cores/esp32/esp32-hal-tinyusb.c` line 215:

```c
.bMaxPacketSize0 = CFG_TUD_ENDOINT0_SIZE,
```

The macro name in TinyUSB's `tusb_option.h` is `CFG_TUD_ENDPOINT0_SIZE` (with the 'P'). Easy to miss -- I only caught it because the compiler flagged it when I was wiring up the composite device.

Just wanted to flag it while it's fresh. Thanks for all the work on the UAC support, it's really solid.

P-R-O-C-H-Y (approved):
LGTM!

MarcFinns:
Hi there, is this happening for real...?

me-no-dev:
> The macro name in TinyUSB's tusb_option.h is CFG_TUD_ENDPOINT0_SIZE (with the 'P')

in ours its without P, so we need to fix it there first

nesror:
https://github.com/espressif/arduino-esp32/issues/12586  ESP32-S3 USBAudioCard microphone cannot record on Windows with Arduino core 3.3.8

PR #12477, opened 2026-03-24
fix(boards.txt):  USB-CDC menu entry for Codecell C3.

Robotto:
Before this fix both options under the Tools -> "USB-CDC on boot:" menu were named "Enabled", and the default option was actually set to disable USB-CDC on boot..

Default is now "Enabled" with correct results.


Robotto:
Hey @CarlBugeja 
Just letting you know i found this small bug affecting the codecell. It's an easy fix, but I'm unsure if it's gonna pass review for unclear formatting reasons. 🤔

Issue #12478, opened 2026-03-25
update from 3.3.6 to 3.3.7 - Guru Meditation Error: Core  1 panic'ed (LoadProhibited)


HubertPOP:
### Board

esp32S3

### Device Description

XIAO esp32S3, 240x320 IPS TFT LCD Display Module for Arduino, ICM42688 

### Hardware Configuration

no

### Version

v3.3.7

### Type

Bug

### IDE Name

Arduino IDE 2.x

### Operating System

Win 10

### Flash frequency

QIO 80 MHz

### PSRAM enabled

no

### Upload speed

921600

### Description

Receiving "Guru Meditation Error: Core  1 panic'ed (LoadProhibited)" when starting running code
I think happens when trying to start BLE, because neither "Bluetooth failed!" or "Bluetooth OK" gets written, see code below

### Sketch

```cpp
if (!BLE.begin()) {
    Serial.println("starting Bluetooth failed!");
    bleTaskState= 0;

  } else {
    Serial.println("Bluetooth OK");
    
    startScanUUID= 0; 
    startScanName= 0; 
    disconnectDevice= false;
    bleTaskState= 1;

  }
```

### Debug Message

```plain
ICM42688 OK
btdm: bss start 0x3fcef180, len 36
btdm: data start 0x3fcef174, data start rom 0x40057350, len 12
MAGIC fadebead VERSION 0001000a
MAGIC fadebead VERSION 00010001
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x40386005  PS      : 0x00060833  A0      : 0x80385944  A1      : 0x3fcebd30  
A2      : 0x949c83c2  A3      : 0x000000e8  A4      : 0x3fceffa0  A5      : 0x00060823  
A6      : 0x3fceff68  A7      : 0x3fca1004  A8      : 0x000000f0  A9      : 0x00000003  
A10     : 0x00000008  A11     : 0x3fcef96c  A12     : 0x3fce0365  A13     : 0x3fcef990  
A14     : 0x3fce02e5  A15     : 0x00000100  SAR     : 0x00000019  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x949c83c6  LBEG    : 0x400570e8  LEND    : 0x400570f3  LCOUNT  : 0x00000000  


[...]

Issue #12383, opened 2026-02-21
[Arduino Core 3.3.7 - Matter] Guru Meditation Error in new 3.3.7 release


yaroslav-hub:
### Board

Xiao ESP32C6

### Device Description

Fan block (coolers module)

### Hardware Configuration

GPIO pins are connected to fan PWM, fan LEDs, external reset button and cascade of transistors for supplying power to the fans. In my case I think it does not matter. But I can provide my scheme if needed.

### Version

v3.3.7

### Type

Bug

### IDE Name

Arduino IDE

### Operating System

Windows 10

### Flash frequency

80 MHz, as default

### PSRAM enabled

yes

### Upload speed

921600, as default

### Description

After resetting my device infinitely rebooting

### Sketch

```cpp
void setup() {
  // Environment
  Serial.begin(115200);

  button.setPressTicks(BUTTON_LONG_PRESS_TIME);
  FastLED.addLeds<LED_TYPE, LED_PIN, LED_COLOR_ORDER>(leds, LED_COUNT);

  pinMode(FAN_POWER_PIN, OUTPUT);
  pinMode(FAN_TACHOMETER_PIN, INPUT_PULLUP);

  ledcAttach(FAN_PWM_PIN, FAN_PWM_FREQUENCY, FAN_PWM_SPEED_RESOLUTION);
  attachInterrupt(digitalPinToInterrupt(FAN_TACHOMETER_PIN), countFanPulse, FALLING);

  digitalWrite(FAN_POWER_PIN, LOW);
  ledcWrite(FAN_PWM_PIN, 0);

  // ESP Memory
  prefs.begin("MatterPrefs", false);
  prefs.getBytes("light", &light, sizeof(light));
  prefs.getBytes("fan", &fan, sizeof(fan));

  // Button
  button.attachLongPressStart(handleButtonLongPress);

  // Matter
  Light.begin();
  Light.setOnOff(light.isOn);
  Light.setColorHSV({ light.color, light.saturation, light.brightness });
  Light.onChange(onLightChange);

  Fan.begin(0,
            MatterFan::FAN_MODE_OFF,
            MatterFan::FAN_MODE_SEQ_OFF_LOW_MED_HIGH);
  Fan.setMode(fan.mode);
  Fan.onChangeMode(onFanModeChange);
  Fan.onChange(onFanAnyChange);

  FanRPM.begin();
  FanRPM.setPressure(fanCurrentRPM);

  Matter.begin();

  renderLight();
[...]

PR #12515, opened 2026-04-13
fix(matter): BLE memory releasing model adjustment

SuGlider:
## Description of Change
This PR adjusts ESP32 Arduino Matter Library to work with Arduino BLE changes doen to the way how BLE Memory is manged within Arduino Core.

## Test Scenarios
Executing Arduino example https://github.com/espressif/arduino-esp32/blob/master/libraries/Matter/examples/MatterMinimum/MatterMinimum.ino

## Related links
#12383
#12403 
#12478


libraries/Matter/src/Matter.h
```diff
@@ -40,6 +40,10 @@
 #include <MatterEndpoints/MatterThermostat.h>
 #include <MatterEndpoints/MatterWindowCovering.h>
 
+#if CONFIG_ENABLE_CHIPOBLE
+#include "esp32-hal-bt-mem.h"
+#endif
```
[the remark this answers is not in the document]
SuGlider: Yeap.... constructors will eat memory...

libraries/Matter/src/Matter.cpp
```diff
@@ -22,6 +22,14 @@
 #include "platform/ESP32/OpenthreadLauncher.h"
 #endif
 
+// Strong function that is overridden by Arduino layer when Matter.cpp is linked
+// This doesn't allow initArduino() to release BLE memory before Matter stack can use it 
+#if CONFIG_ENABLE_CHIPOBLE
+extern "C" bool btInUse(void) {
+  return true;
+}
```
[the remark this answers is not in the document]
SuGlider: No include here.

libraries/Matter/src/Matter.cpp
```diff
@@ -22,6 +22,15 @@
...
 #endif
 
+// Strong definition that overrides the core's weak btInUse() implementation.
+// This prevents initArduino() from releasing BTDM controller memory before the
+// Matter stack can use Bluetooth transport.
+#if CONFIG_ENABLE_CHIPOBLE
+extern "C" bool btInUse(void) {
+  return true;
+}
+#endif
```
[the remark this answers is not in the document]
SuGlider: it won't change. Don't worry about it.

libraries/Matter/src/Matter.cpp
```diff
@@ -22,6 +22,15 @@
 #include "platform/ESP32/OpenthreadLauncher.h"
 #endif
 
+// Strong definition that overrides the core's weak btInUse() implementation.
+// This prevents initArduino() from releasing BTDM controller memory before the
+// Matter stack can use Bluetooth transport.
```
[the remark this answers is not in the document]
SuGlider: too verbose. No need for that.

libraries/Matter/src/Matter.cpp
```diff
@@ -22,6 +22,15 @@
...
+#if CONFIG_ENABLE_CHIPOBLE
+extern "C" bool btInUse(void) {
+  return true;
+}
+#endif
+
```
lucasssvaz, line 30: This will cause "multiple definitions" errors with user defined `btInUse`. Why not just include `esp32-hal-bt-mem.h` ?

SuGlider: Multiple definition would only happen if the user actually declares such function. I really don't think that it would be done within an Arduino Matter application.

The `esp32-hal-bt-mem.h` injects a constructor that changes `_btLibraryInUse` default `false`value to `true`.

A strong `btInUse(void)` from the user would override the `esp32-hal-bt-mem.h` constructor along the execution.
In case the user makes this fucntion return `false` it would crash the application.

A linker error would be a stronger and better way to avoid it, I think.

lucasssvaz: But in the case the memory was released, the lib should gracefully fail rather than crashing. I will try adding a function to check the status of the bt memory.

SuGlider: At this time, BLE is initialized by the Matter layer, internally in the ESP-Matter event loop.
If BLE memory is released, the application crashes with Panic message.

For an ESP-Matter application that uses CHIPOBLE (BLE Commissioning), BLE must work correctly.

There is another issue that makes the user application crash when ESP-Matter tries to release BLE memory at the end of the commissioning phase. Apparenlty it works fine when the application is built as IDF Component, and fails with the BLE pre compiled libraries.

At this time, for the correct Matter application execution, Arduino Matter Library must force that the BLE memory is never released.

SuGlider: > But in the case the memory was released, the lib should gracefully fail rather than crashing. I will try adding a function to check the status of the bt memory.

This new function won't help at all Arduino Matter. 
BLE is managed internally by ESP-Matter. Arduino Layer has no control over it.

lucasssvaz: But in this case the new function will still help as you just need to check if the memory is available before calling any matter API.

SuGlider: > But in this case the new function will still help as you just need to check if the memory is available before calling any matter API.

Arduino Matter calls ESP-Matter initialization function when `Matter.begin()` is called by the sketch.
At this point if BLE Memory was released, the application crashes.

Your suggestion is to replace the strong `btInUse()` declaration with `#include "esp32-hal-bt-mem.h"` and when the Arduino Matter function `Matter.begin()`  is called, check if BLE memory has not been released before initializing ESP Matter Stack?

If it has been released? -->`Matter.begin()` shall not initialize ESP Matter stack and just print "oops... matter has failed... and then forever Loop..."?

What about Arduino as IDF Component? The user may just use a own `main_app()` and never run the Arduino Initializer. There are many possible ways to create an Arduino APP using Matter.

IHO, failing in linking is a good option... Such failure will never happen by default, only when the user explicitaly declares another string `btInUse()` or uses some external BLE library that does such declaration.

lucasssvaz: No need to loop forever. Just return an error like all the other libraries do. Please try `#include "esp32-hal-bt-mem.h"` with https://github.com/espressif/arduino-esp32/pull/12574. It will work no matter where the the release was called even in component mode. You can also check if the memory was released before initializing matter. This way it solves the issue without removing the user's flexibility.

PR #12558, opened 2026-04-29
Use arduino-cmake to build tests in CI

me-no-dev:
Arduino-cmake significantly speeds up compilation of sketches in CI. This makes a full test to pass in a bit over half an hour, compared to almost two and a half hours. It has a few downsides, but the team accepts them in favor of having a much faster CI.


tools/arduino_cmake.py
```diff
@@ -0,0 +1,1029 @@
...
+# -*- coding: utf-8 -*-
+"""Portable single-file Arduino CMake builder (bundled acmake package).
+
+Copy ``arduino_cmake.py`` anywhere and run:
+  python3 arduino_cmake.py compile --fqbn ... --sketch ...
+  chmod +x arduino_cmake.py && ./arduino_cmake.py board
+
+The embedded bundle is extracted once under ``~/.cache/arduino_cmake_portable/<id>/``.
+Regenerate this file from the repository with:
+  python3 tools/gen_arduino_cmake_portable.py
```
lucasssvaz:
```suggestion
"""
Portable single-file Arduino CMake builder (bundled acmake package).
Generated from https://github.com/me-no-dev/ArduinoCmake

Copy ``arduino_cmake.py`` anywhere and run:
  python3 arduino_cmake.py compile --fqbn ... --sketch ...
  chmod +x arduino_cmake.py && ./arduino_cmake.py board

The embedded bundle is extracted once under ``~/.cache/arduino_cmake_portable/<id>/``.
Regenerate this file from the ArduinoCmake repository with:
  python3 tools/gen_arduino_cmake_portable.py
```

me-no-dev: Note this should also be changed in the generator [here](https://github.com/me-no-dev/ArduinoCmake/blob/master/tools/gen_arduino_cmake_portable.py#L37)

lucasssvaz (approved):

SuGlider (approved):

P-R-O-C-H-Y (approved):

PR #12597, opened 2026-05-16
Enhance RequestArgument struct with constructors

Nezaemmy:
## Summary
Fix HTTPClient build error caused by missing constructor in RequestArgument when compiling with GCC 14 (ESP-IDF 5.5.4).

## Problem
Compilation fails with:

error: no matching function for call to
HTTPClient::RequestArgument::RequestArgument(String&, String&)

This happens because RequestArgument struct does not define a constructor accepting (String, String), while std::vector::emplace_back() requires it.

## Solution
Added a constructor to RequestArgument to support direct initialization.

### Checklist
5. [ ] Please **confirm option to "Allow edits and access to secrets by maintainers"** when opening a Pull Request

-----------
## Description of Change

## Test Scenarios

## Related links


karandancer07-coder (approved):

lucasssvaz:
Where do you see this error ? The entire CI is already compiling all the examples with GCC 14 and IDF 5.5. No compilation fails.

Nezaemmy:
> Where do you see this error ? The entire CI is already compiling all the examples with GCC 14 and IDF 5.5. No compilation fails.

 When I tried to build the Matter example using Arduino as an ESP-IDF component, the build failed with that error.
With Arduino core version 3.3.8 and Esp matter 1.4.1.

SuGlider:
It depends on the C++ standard being used. This is the root cause, and it explains why we don't see it in CI while @Nezaemmy does in the Matter / Arduino-as-IDF-component setup:

- **_std::vector::emplace_back(args...) performs ::new(p) T(std::forward<Args>(args)...)_** — i.e. direct-initialization with parentheses, not braces.
- RequestArgument (as it stands today on master) is an aggregate with no user-declared constructor.
- C++17 and earlier: parentheses cannot aggregate-initialize → emplace_back(headerName, headerValue) is ill-formed → compile error (exactly the one the author reports).
- C++20 (P0960R3, "parenthesized aggregate initialization"): parentheses now also perform aggregate-init → it compiles. GCC 11+ implements this.

**Arduino-ESP32's CI builds with -std=gnu++20 (or newer)** from compiler.sdk.path/flags/cpp_flags, so it never trips.

As soon as someone re-compiles HTTPClient.cpp from within an ESP-IDF project that defaults its component build to C++17 (or pulls in a sub-component that does), GCC 14 raises the error @Nezaemmy  quoted.

So the report is real, just not visible from the CI's point of point.

SuGlider:
Project's CMake enforces C++ 2a. This error never happens.

`idf_build_set_property(CXX_COMPILE_OPTIONS "-std=gnu++2a;-Os;-DCHIP_HAVE_CONFIG_H" APPEND)`


@Nezaemmy - Do you need to use GCC 14?

SuGlider:
@Nezaemmy - 

A simpler alternative exists (zero touch to the struct, one-line change at the call site):

`_currentHeaders.push_back({headerName, headerValue});`

This keeps `RequestArgument` a true aggregate and works on every C++ standard from C++11 onward.
It's arguably a cleaner fix than introducing constructors just to satisfy `emplace_back`'s parentheses semantics.

-----------------------

### My suggested change:

Only one line in one file. The struct in `HTTPClient.h` stays completely untouched.
In `libraries/HTTPClient/src/HTTPClient.cpp`, change:


HTTPClient.cpp
Lines 1246-1248
``` cpp
        if (_collectAllHeaders && headerName.length() > 0) {
          _currentHeaders.emplace_back(headerName, headerValue);
        } else {
```
to:
``` cpp
        if (_collectAllHeaders && headerName.length() > 0) {
          _currentHeaders.push_back({headerName, headerValue});
        } else {
```

That'd be the entire patch. HTTPClient.h is not modified at all.

SuGlider (changes requested):
@Nezaemmy - Please review my notes.

me-no-dev:
@SuGlider I suggest you propose a PR with the change so we can close this one

Nezaemmy:
> Project's CMake enforces C++ 2a. This error never happens.
> 
> `idf_build_set_property(CXX_COMPILE_OPTIONS "-std=gnu++2a;-Os;-DCHIP_HAVE_CONFIG_H" APPEND)`
> 
> @Nezaemmy - Do you need to use GCC 14?

In my setup I am using Arduino as an ESP-IDF component together with ESP-Matter on ESP-IDF 5.5.4, where GCC 14 is used by the toolchain. I do not specifically need GCC 14 by choice; it comes from the ESP-IDF 5.5.4 environment. The issue appears because this component path can be built with a C++17 mode, while Arduino-ESP32 CI uses gnu++2a, so CI does not reproduce the error.

Nezaemmy:
> @Nezaemmy -
> 
> A simpler alternative exists (zero touch to the struct, one-line change at the call site):
> 
> `_currentHeaders.push_back({headerName, headerValue});`
> 
> This keeps `RequestArgument` a true aggregate and works on every C++ standard from C++11 onward. It's arguably a cleaner fix than introducing constructors just to satisfy `emplace_back`'s parentheses semantics.
> 
> ### My suggested change:
> Only one line in one file. The struct in `HTTPClient.h` stays completely untouched. In `libraries/HTTPClient/src/HTTPClient.cpp`, change:
> 
> HTTPClient.cpp Lines 1246-1248
> 
> ```c++
>         if (_collectAllHeaders && headerName.length() > 0) {
>           _currentHeaders.emplace_back(headerName, headerValue);
>         } else {
> ```
> 
> to:
> 
> ```c++
>         if (_collectAllHeaders && headerName.length() > 0) {
>           _currentHeaders.push_back({headerName, headerValue});
>         } else {
> ```
> 
> That'd be the entire patch. HTTPClient.h is not modified at all.

This keeps RequestArgument as an aggregate and works across older C++ standards as well.

SuGlider:
@Nezaemmy - to make it easier I have added the suggested change into the PR #12639
Thanks!

SuGlider:
> @SuGlider I suggest you propose a PR with the change so we can close this one

Done in https://github.com/espressif/arduino-esp32/pull/12639

Issue #12517, opened 2026-04-15
NetworkClientRxBuffer::r_available() may cause crash when network interface is down or disconnected


TD-er:
### Board

ESP32-xx, reproducible on ESP32C6

### Device Description

Tried here on ESP32-C6 DevKit-C-1, but have seen similar crashes on other boards

### Hardware Configuration

-

### Version

latest master (checkout manually)

### Type

Bug

### IDE Name

VSCode

### Operating System

Windows 11 + WSL2

### Flash frequency

-

### PSRAM enabled

no

### Upload speed

115200

### Description

I've seen these crashes often but rarely had a proper stack trace.
Now I've had some which consistently point to `NetworkClientRxBuffer::r_available()` so I started digging there.

Current implementation.
```c++
  size_t r_available() {
    if (_fd < 0) {
      return 0;
    }
    int count;
#ifdef ESP_IDF_VERSION_MAJOR
    int res = lwip_ioctl(_fd, FIONREAD, &count);
#else
    int res = lwip_ioctl_r(_fd, FIONREAD, &count);
#endif
    if (res < 0) {
      _failed = true;
      return 0;
    }
    return count;
  }
```

One potential issue here is that `count` is not initialized before calling `lwip_ioctl`.
I've not looked through all LwIP code to see if the `count` may remain untouched while `res >= 0`, but it wouldn't harm if it was initialized. For example to `-1`.
Then we can have the code something like this:

```c++
    if (res < 0 || count < 0) {
      _failed = true;
      return 0;
    }
```

Here the code for [lwip_ioctl](https://github.com/espressif/esp-lwip/blob/392707e5c3a5d8e2531461db77843b9759cd5a43/src/api/sockets.c#L3786)
`int recv_avail;` is also not initialized and I don't know whether `SYS_ARCH_GET(sock->conn->recv_avail, recv_avail);` could also leave this value untouched.

Part of a stack trace from my own software can be seen below, but I guess the top is the most interesting and I have seen very similar traces in other issues here.

At least it looks like `NetworkClientRxBuffer::r_available()` might sometimes return some values which cause issues when trying to allocate buffers for it.
Especially right at the moment the network connection is lost.

### Sketch

```cpp
-
```

### Debug Message

```plain
Core: 0
PC: 0x00000000
Fault Code: 1
Stack Trace
#	Address	Function																					Location
0	0x420e0476	esp_pbuf_free																			esp_pbuf_ref.c.obj
[...]

PR #12614, opened 2026-05-25
NetworkClient: guard RX available count

dhimasardinata:
## Summary
- Initialize the `FIONREAD` count in `NetworkClientRxBuffer::r_available()`.
- Treat negative counts as failed reads before converting the value to `size_t`.

## Verification
- `git diff --check`
- `pio run` in `/tmp/arduino-esp32-rx-count-check` with this checkout as a symlinked `framework-arduinoespressif32` package

Refs #12517

### DIFF
--- a/tests/validation/network_client/ci.yml
+++ b/tests/validation/network_client/ci.yml
@@ -1,0 +1,4 @@
+platforms:
+  hardware: true
+  qemu: false
+  wokwi: false
--- a/tests/validation/network_client/ci.yml
+++ b/tests/validation/network_client/ci.yml
@@ -1,4 +1,0 @@
-platforms:
-  hardware: true
-  qemu: false
-  wokwi: false
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -49,13 +49,13 @@ private:
     if (_fd < 0) {
       return 0;
     }
-    int count;
+    int count = -1;
 #ifdef ESP_IDF_VERSION_MAJOR
     int res = lwip_ioctl(_fd, FIONREAD, &count);
 #else
     int res = lwip_ioctl_r(_fd, FIONREAD, &count);
 #endif
-    if (res < 0) {
+    if (res < 0 || count < 0) {
       _failed = true;
       return 0;
     }
--- a/tests/validation/network_client/network_client.ino
+++ b/tests/validation/network_client/network_client.ino
@@ -1,0 +1,72 @@
+#include <Arduino.h>
+#include <NetworkClient.h>
+#include <unity.h>
+
+#include <esp_netif.h>
+#include <lwip/inet.h>
+#include <lwip/sockets.h>
+#include <unistd.h>
+
+static int create_loopback_client(int *peer_fd) {
+  int listen_fd = socket(AF_INET, SOCK_STREAM, IPPROTO_IP);
+  TEST_ASSERT_GREATER_OR_EQUAL(0, listen_fd);
+
+  sockaddr_in addr = {};
+  addr.sin_family = AF_INET;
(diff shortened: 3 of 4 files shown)


TD-er:
@dhimasardinata Did you also see my [last comment](https://github.com/espressif/arduino-esp32/issues/12517#issuecomment-4434299581) on the issue which will be closed by this PR (if merged)?

dhimasardinata:
yes, i saw it. good point! thanks for pointing out
this PR only covers the NetworkClienTxBuffer::r_available() path. the NetworkUdp::stop() looks related, but it is a separate teardown path. so i dont want to claim this PR fixes that too. i can change Closes to Refs if this issue should stay open for the UDP/lwIP teardown crash

TD-er:
@dhimasardinata Since I don't know what the fix would be, so I don't know for sure if it should actually be part of the same issue.
So if you can, please have a look and we can see if it should be considered a separate issue.

dhimasardinata:
I updated this PR to use `Refs #12517` rather than closing that issue, and I also ran the related fixes on real ESP32 hardware.

Hardware: ESP32-D0WD-V3 rev v3.1 on `/dev/ttyUSB0` (MAC `b0:cb:d8:03:aa:a8`).

Arduino PR branch (`896b62a78c56`):
- Added `tests/validation/network_client` so runtime tests cover `NetworkClient::available()` over TCP loopback without WiFi credentials or external network dependencies.
- Built and uploaded the same validation sketch with PlatformIO using this PR branch as the local `framework-arduinoespressif32` package.
- UART result:
  - `test_network_client_available_counts_loopback_bytes:PASS`
  - `1 Tests 0 Failures 0 Ignored`
  - `OK`

Lower-level esp-netif PR ([espressif/esp-idf#18656](https://github.com/espressif/esp-idf/pull/18656), `a6b6b06b2a31`):
- Test app allocates an esp-netif custom pbuf, clears the netif driver config, then calls `pbuf_free()`. This reproduces the lifetime problem where a queued RX pbuf can be freed after driver teardown.
- On the unpatched IDF base (`c1b7f38212`), the same ESP32 test crashes immediately after `PBUF_HW_TEST_FREE_AFTER_CLEAR`:
  - `Guru Meditation Error: Core 0 panic'ed (InstrFetchProhibited)`
  - `PC: 0x00000000`
  - `EXCVADDR: 0x00000000`
- On the patched branch (`a6b6b06b2a31`), the same ESP32 test passes:
  - `PBUF_HW_TEST_RESULT free_result=1 l2_free_calls=1`
  - `PBUF_HW_TEST_PASS`

So I think this Arduino PR should stay as the small defensive `NetworkClientRxBuffer::r_available()` guard. The shared TCP/UDP `pbuf_free()` / `esp_pbuf_free()` teardown crash is covered separately by [espressif/esp-idf#18656](https://github.com/espressif/esp-idf/pull/18656) if that approach is accepted.

TD-er:
Great!
This would be really great if these crashes could finally be resolved, as I have been experiencing these crashes a lot.

TD-er:
By the way, really nice to see someone with a fresh view diving into these hard to reproduce issues/errors and finding matching issues along the way while fixing these bugs.  (new hard-to-reproduce issues will be added soon, maybe even today ;) )
At some point one may have looked at the code too often to become blind for bugs, so a new set of eyes going through the code every now and then is very welcome.

me-no-dev (approved):

me-no-dev:
@lucasssvaz PTAL at the test

me-no-dev (approved):

P-R-O-C-H-Y (approved):

tests/validation/network_client/ci.yml
lucasssvaz, line 1: No need for a ci.yml file. It should run in any platform without a problem.

lucasssvaz (approved):

PR #12621, opened 2026-05-26
fix(matter): Fix internal state synchronization

lucasssvaz:
## Description of Change

This pull request improves the reliability and clarity of attribute and internal state management for Matter endpoints, aligning all endpoint setters with the "Ember pattern" described in the new documentation. The changes ensure that internal state variables are only updated after successfully updating the Matter attribute store, preventing inconsistencies between device state and protocol state. The PR also adds a comprehensive developer guide to document these patterns and clarifies the use of attribute update APIs.

**Documentation and Developer Guidance**

* Added a detailed `README.md` explaining the architecture, state management patterns, setter/getter rules, attribute update APIs, and endpoint class inventory for the Arduino Matter library. This guides developers in implementing robust and consistent endpoint classes.

**Consistent Ember Pattern for State Updates**

* Refactored all endpoint setter methods (e.g., in `MatterTemperatureControlledCabinet`, `MatterWindowCovering`, `MatterOccupancySensor`) to update internal state variables only after the attribute store update succeeds. This prevents internal state from diverging from the protocol-level attribute store. [[1]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L367-R370) [[2]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L420-R421) [[3]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L473-R472) [[4]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L526-R523) [[5]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L593-R586) [[6]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L641-R643) [[7]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfR364-L379) [[8]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfR412-R414) [[9]](diffhunk://#diff-53209a569e471fcf1c819c32ab80fb1c1837f89109e3d10df671fd07ec323521R448-R456) [[10]](diffhunk://#diff-6220b46c8a219ee62f6c140cb9fe2a2e93b0be5bc391fea262da16ed848d0f6fR194)

* For attributes that may fail to update via `updateAttributeVal()` (e.g., nullable or dynamic attributes), added a fallback to `setAttributeVal()` as documented in the new guide. [[1]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L367-R370) [[2]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L420-R421) [[3]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L473-R472) [[4]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L526-R523) [[5]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L593-R586) [[6]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfR364-L379) [[7]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfR412-R414)

**Getter and Internal State Simplification**

* Simplified getter methods for affected endpoints to always return the internal state variable, removing unnecessary reads from the attribute store, since the internal state now reliably reflects the last committed value. [[1]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L391-L394) [[2]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L444-L447) [[3]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L497-L500) [[4]](diffhunk://#diff-8c96ca12e0bcc01959c9eab72a617864e544bbe017d457a3c49b54e6365acec3L612-L615) [[5]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfR364-L379) [[6]](diffhunk://#diff-daf0d19ced10da9bb9246b8f84beee1fa5c19106af3a7327c11ba4f0832bdddfL421-L424)

**Atomicity and Buffer Handling Improvements**

* In setters that update arrays (e.g., `setSupportedTemperatureLevels`), now copy data into internal buffers only after successfully updating the attribute store, ensuring the internal buffer is not modified on failure.

**Event Loop Scheduling and State Commit**

* For `MatterOccupancySensor`, internal state for hold time limits is now updated only after successfully scheduling the attribute update on the Matter event loop, further aligning with the Ember pattern. [[1]](diffhunk://#diff-53209a569e471fcf1c819c32ab80fb1c1837f89109e3d10df671fd07ec323521L410-L414) [[2]](diffhunk://#diff-53209a569e471fcf1c819c32ab80fb1c1837f89109e3d10df671fd07ec323521R427-L438) [[3]](diffhunk://#diff-53209a569e471fcf1c819c32ab80fb1c1837f89109e3d10df671fd07ec323521R448-R456)

These changes collectively make endpoint implementations more robust and maintainable, and the new documentation provides a clear reference for future development.

## Test Scenarios

Tested locally with WIP Matter test. All 33 Phase 0 unit tests pass (previously 2 failures in  test_matter_window_covering` and `test_matter_cabinet`). Phase 1 controller-driven tests (commissioning, multi-endpoint control) also pass.

### DIFF
--- a/libraries/Matter/src/MatterEndpoints/MatterEnhancedColorLight.cpp
+++ b/libraries/Matter/src/MatterEndpoints/MatterEnhancedColorLight.cpp
@@ -191,6 +191,7 @@ bool MatterEnhancedColorLight::begin(bool initialState, espHsvColor_t _colorHSV,
 
   light_config.level_control.current_level = brightness;
   light_config.level_control.lighting.start_up_current_level = nullptr;
+  brightnessLevel = brightness;
 
   light_config.color_control.enhanced_color_mode = (uint8_t)ColorControl::ColorMode::kColorTemperature;
   light_config.color_control.color_temperature.color_temperature_mireds = ColorTemperature;
--- a/libraries/Matter/src/MatterEndpoints/MatterOccupancySensor.cpp
+++ b/libraries/Matter/src/MatterEndpoints/MatterOccupancySensor.cpp
@@ -407,11 +407,6 @@ bool MatterOccupancySensor::setHoldTimeLimits(uint16_t _holdTimeMin_seconds, uin
     return false;
   }
 
-  // Update member variables immediately
-  holdTimeMin_seconds = _holdTimeMin_seconds;
-  holdTimeMax_seconds = _holdTimeMax_seconds;
-  holdTimeDefault_seconds = _holdTimeDefault_seconds;
-
   // Check if current HoldTime is outside the new limits and adjust if necessary
   uint16_t adjustedHoldTime = holdTime_seconds;
   bool holdTimeAdjusted = false;
@@ -429,13 +424,14 @@ bool MatterOccupancySensor::setHoldTimeLimits(uint16_t _holdTimeMin_seconds, uin
   uint16_t endpoint_id = getEndPointId();
   CHIP_ERROR schedule_err;
 
+  // Schedule the attribute store update on the Matter event loop.
+  // Lambdas capture all values by copy, so they don't depend on member state.
   if (holdTimeAdjusted) {
     // Schedule both limits and HoldTime updates together
     schedule_err = chip::DeviceLayer::SystemLayer().ScheduleLambda([endpoint_id, min = _holdTimeMin_seconds, max = _holdTimeMax_seconds,
                                                                     def = _holdTimeDefault_seconds, holdTime = adjustedHoldTime]() {
       SetHoldTimeLimitsAndHoldTimeInEventLoop(endpoint_id, min, max, def, holdTime);
     });
-    holdTime_seconds = adjustedHoldTime;
   } else {
     // No adjustment needed, just schedule the limits update
     schedule_err =
@@ -447,6 +443,15 @@ bool MatterOccupancySensor::setHoldTimeLimits(uint16_t _holdTimeMin_seconds, uin
   if (schedule_err != CHIP_NO_ERROR) {
     log_e("Failed to schedule HoldTimeLimits update: %" CHIP_ERROR_FORMAT, schedule_err.Format());
     return false;
+  }
+
+  // Commit to internal state only after scheduling succeeds,
+  // so that on failure the member variables remain unchanged (Ember pattern).
+  holdTimeMin_seconds = _holdTimeMin_seconds;
+  holdTimeMax_seconds = _holdTimeMax_seconds;
+  holdTimeDefault_seconds = _holdTimeDefault_seconds;
+  if (holdTimeAdjusted) {
+    holdTime_seconds = adjustedHoldTime;
   }
 
   log_v("HoldTimeLimits scheduled for update: Min=%u, Max=%u, Default=%u seconds", _holdTimeMin_seconds, _holdTimeMax_seconds, _holdTimeDefault_seconds);
--- a/libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
+++ b/libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
@@ -361,11 +361,15 @@ bool MatterWindowCovering::setLiftPosition(uint16_t liftPosition) {
   if (val.val.u16 != liftPosition) {
     val.val.u16 = liftPosition;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val);
+    }
     if (ret) {
       currentLiftPosition = liftPosition;
-      // Also update CurrentPositionLiftPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's LiftPositionSet() behavior
-      setLiftPercentage((uint8_t)(liftPercent100ths / 100));
+      if (!setLiftPercentage((uint8_t)(liftPercent100ths / 100))) {
+        log_e("Failed to sync Lift Percentage from Lift Position.");
+        return false;
+      }
     }
     return ret;
   }
@@ -373,10 +377,6 @@ bool MatterWindowCovering::setLiftPosition(uint16_t liftPosition) {
 }
 
 uint16_t MatterWindowCovering::getLiftPosition() {
-  esp_matter_attr_val_t val = esp_matter_invalid(NULL);
-  if (getAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val)) {
-    currentLiftPosition = val.val.u16;
-  }
   return currentLiftPosition;
 }
 
@@ -409,6 +409,9 @@ bool MatterWindowCovering::setLiftPercentage(uint8_t liftPercent) {
   if (currentVal.val.u16 != liftPercent100ths) {
     currentVal.val.u16 = liftPercent100ths;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLiftPercent100ths::Id, &currentVal);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLiftPercent100ths::Id, &currentVal);
+    }
     if (ret) {
       currentLiftPercent = liftPercent;
     }
@@ -418,10 +421,6 @@ bool MatterWindowCovering::setLiftPercentage(uint8_t liftPercent) {
 }
 
 uint8_t MatterWindowCovering::getLiftPercentage() {
-  esp_matter_attr_val_t val = esp_matter_invalid(NULL);
-  if (getAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLiftPercent100ths::Id, &val)) {
-    currentLiftPercent = (uint8_t)(val.val.u16 / 100);
-  }
   return currentLiftPercent;
 }
 
@@ -516,11 +515,15 @@ bool MatterWindowCovering::setTiltPosition(uint16_t tiltPosition) {
   if (val.val.u16 != tiltPosition) {
     val.val.u16 = tiltPosition;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val);
+    }
     if (ret) {
       currentTiltPosition = tiltPosition;
-      // Also update CurrentPositionTiltPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's TiltPositionSet() behavior
-      setTiltPercentage((uint8_t)(tiltPercent100ths / 100));
+      if (!setTiltPercentage((uint8_t)(tiltPercent100ths / 100))) {
+        log_e("Failed to sync Tilt Percentage from Tilt Position.");
+        return false;
+      }
     }
     return ret;
   }
@@ -528,10 +531,6 @@ bool MatterWindowCovering::setTiltPosition(uint16_t tiltPosition) {
 }
 
 uint16_t MatterWindowCovering::getTiltPosition() {
-  esp_matter_attr_val_t val = esp_matter_invalid(NULL);
-  if (getAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val)) {
-    currentTiltPosition = val.val.u16;
-  }
   return currentTiltPosition;
 }
 
@@ -564,6 +563,9 @@ bool MatterWindowCovering::setTiltPercentage(uint8_t tiltPercent) {
   if (currentVal.val.u16 != tiltPercent100ths) {
     currentVal.val.u16 = tiltPercent100ths;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTiltPercent100ths::Id, &currentVal);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTiltPercent100ths::Id, &currentVal);
+    }
     if (ret) {
       currentTiltPercent = tiltPercent;
     }
@@ -573,10 +575,6 @@ bool MatterWindowCovering::setTiltPercentage(uint8_t tiltPercent) {
 }
 
 uint8_t MatterWindowCovering::getTiltPercentage() {
-  esp_matter_attr_val_t val = esp_matter_invalid(NULL);
-  if (getAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTiltPercent100ths::Id, &val)) {
-    currentTiltPercent = (uint8_t)(val.val.u16 / 100);
-  }
   return currentTiltPercent;
 }
 
--- a/libraries/Matter/README.md
+++ b/libraries/Matter/README.md
@@ -1,0 +1,133 @@
+# Arduino Matter Library
+
+Arduino-friendly wrapper around [ESP-Matter](https://docs.espressif.com/projects/esp-matter/en/latest/) (Espressif's SDK for Matter), providing high-level endpoint classes for common Matter device types.
+
+## Architecture
+
+Each Matter device type is represented by a C++ class under `src/MatterEndpoints/` (e.g., `MatterOnOffLight`, `MatterTemperatureSensor`, `MatterFan`). These classes manage:
+
+- **Internal state variables** — C++ members that cache the device's current state (e.g., `onOffState`, `brightnessLevel`, `rawTemperature`).
+- **Matter attribute store** — The ESP-Matter SDK's attribute database, which is the protocol-level representation read by controllers and used for subscriptions/reporting.
+
+## Attribute Update Pattern
+
+This library is built on ESP-Matter, which uses an Ember-based attribute store. The attribute store is the protocol-level source of truth: when a Matter controller reads an attribute, it reads from this store.
+
+All endpoint setters and getters **must** follow one of the two patterns below.
+
+### Setter Pattern
+
+When the device application (Arduino sketch) wants to change state, the setter **must** update the attribute store first and only commit to internal state after the store confirms success.
+
+```cpp
+bool MyEndpoint::setValue(int newValue) {
+  if (internalValue == newValue) {
+    return true;  // No change needed
+  }
+
+  esp_matter_attr_val_t val = esp_matter_invalid(NULL);
+  if (!getAttributeVal(ClusterId, AttributeId, &val)) {
+    log_e("Failed to get attribute.");
+    return false;
+  }
+
+  if (val.val.i16 != newValue) {
+    val.val.i16 = newValue;
+    bool ret = updateAttributeVal(ClusterId, AttributeId, &val);
+    if (!ret) {
+      log_e("Failed to update attribute.");
+      return false;
+    }
+    // Internal state committed ONLY after attribute store update succeeds
+    internalValue = newValue;
+  }
+
+  return true;
+}
+```
+
+Key rules:
+- **Never assign internal state before the attribute store update.** If `updateAttributeVal()` fails and the function returns `false`, the internal state must remain unchanged.
+- **Internal state assignment goes inside the `if (value changed)` block**, after the successful update call.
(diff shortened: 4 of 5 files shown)


libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -361,10 +361,11 @@ bool MatterWindowCovering::setLiftPosition(uint16_t liftPosition) {
...
     val.val.u16 = liftPosition;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val);
+    }
     if (ret) {
       currentLiftPosition = liftPosition;
-      // Also update CurrentPositionLiftPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's LiftPositionSet() behavior
       setLiftPercentage((uint8_t)(liftPercent100ths / 100));
```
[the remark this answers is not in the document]
lucasssvaz: fixed

libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -516,10 +511,11 @@ bool MatterWindowCovering::setTiltPosition(uint16_t tiltPosition) {
...
     val.val.u16 = tiltPosition;
     bool ret = updateAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val);
+    if (!ret) {
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val);
+    }
     if (ret) {
       currentTiltPosition = tiltPosition;
-      // Also update CurrentPositionTiltPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's TiltPositionSet() behavior
       setTiltPercentage((uint8_t)(tiltPercent100ths / 100));
```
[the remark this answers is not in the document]
lucasssvaz: fixed

libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -395,6 +392,8 @@ bool MatterWindowCovering::setLiftPercentage(uint8_t liftPercent) {
     return true;
   }
 
+  currentLiftPercent = liftPercent;
+
```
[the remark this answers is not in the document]
lucasssvaz: This is by design. The setXxx() methods are the device-side API -- called from the Arduino sketch to assert the device's own state. The _onChangeCB callback is specifically intended for controller-originated changes flowing through attributeChangeCB() (PRE_UPDATE), giving the sketch a chance to accept/reject external commands. When the sketch itself calls setLiftPercentage(), it is the authoritative source of truth and vetoing its own call would be counter-intuitive. This pattern is consistent across all other Matter endpoints (e.g., MatterThermostat, MatterDimmableLight, MatterContactSensor).

libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -550,6 +542,8 @@ bool MatterWindowCovering::setTiltPercentage(uint8_t tiltPercent) {
     return true;
   }
 
+  currentTiltPercent = tiltPercent;
+
```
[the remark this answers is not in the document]
lucasssvaz: This is by design. The setXxx() methods are the device-side API -- called from the Arduino sketch to assert the device's own state. The _onChangeCB callback is specifically intended for controller-originated changes flowing through attributeChangeCB() (PRE_UPDATE), giving the sketch a chance to accept/reject external commands. When the sketch itself calls setLiftPercentage(), it is the authoritative source of truth and vetoing its own call would be counter-intuitive. This pattern is consistent across all other Matter endpoints (e.g., MatterThermostat, MatterDimmableLight, MatterContactSensor).

libraries/Matter/src/MatterEndpoints/MatterContactSensor.cpp
```diff
@@ -97,7 +99,6 @@ bool MatterContactSensor::setContact(bool _contactState) {
       log_e("Failed to update Contact Sensor Attribute.");
       return false;
     }
-    contactState = _contactState;
```
SuGlider, line 100: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

libraries/Matter/src/MatterEndpoints/MatterEnhancedColorLight.cpp
```diff
@@ -191,6 +191,7 @@ bool MatterEnhancedColorLight::begin(bool initialState, espHsvColor_t _colorHSV,
 
   light_config.level_control.current_level = brightness;
   light_config.level_control.lighting.start_up_current_level = nullptr;
+  brightnessLevel = brightness;
```
SuGlider, line 194: There is no sense doing it given that this is done by setting the level internally in the Matter Data structure `light_config.level_control.current_level = brightness;` which is used as the only source of Thruth.

libraries/Matter/src/MatterEndpoints/MatterOccupancySensor.cpp
```diff
@@ -314,7 +316,6 @@ bool MatterOccupancySensor::setOccupancy(bool _occupancyState) {
       log_e("Failed to update Occupancy Sensor Attribute.");
       return false;
     }
-    occupancyState = _occupancyState;
```
SuGlider, line 317: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

libraries/Matter/src/MatterEndpoints/MatterOnOffPlugin.cpp
```diff
@@ -120,7 +122,6 @@ bool MatterOnOffPlugin::setOnOff(bool newState) {
       log_e("Failed to update On-Off Plugin Attribute.");
       return false;
     }
-    onOffState = newState;
```
SuGlider, line 123: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

libraries/Matter/src/MatterEndpoints/MatterRainSensor.cpp
```diff
@@ -97,7 +99,6 @@ bool MatterRainSensor::setRain(bool _rainState) {
       log_e("Failed to update Rain Sensor Attribute.");
       return false;
     }
-    rainState = _rainState;
```
SuGlider, line 100: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

libraries/Matter/src/MatterEndpoints/MatterWaterFreezeDetector.cpp
```diff
@@ -97,7 +99,6 @@ bool MatterWaterFreezeDetector::setFreeze(bool _freezeState) {
       log_e("Failed to update Water Freeze Detector Attribute.");
       return false;
     }
-    freezeState = _freezeState;
```
SuGlider, line 100: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

libraries/Matter/src/MatterEndpoints/MatterWaterLeakDetector.cpp
```diff
@@ -97,7 +99,6 @@ bool MatterWaterLeakDetector::setLeak(bool _leakState) {
       log_e("Failed to update Water Leak Detector Attribute.");
       return false;
     }
-    leakState = _leakState;
```
SuGlider, line 100: This is correct. It only changes the internal value after updating the Matter internal data structure, otherwise it returns false and keeps last valid state.

SuGlider (commented):
The PR seems incorrect.
The new value can only be set after successfully changing the internal Matter server records.

It sounds like this PR was the result of an AI agent analysis. 
I'd say to just close it.

SuGlider:
Analysis of another AI about this PR, taking the `MatterOnOffPlugin::setOnOff(bool)` change example:

# Analysis: `onOffState = newState` placement in `MatterOnOffPlugin::setOnOff(bool)`

**File:** [`libraries/Matter/src/MatterEndpoints/MatterOnOffPlugin.cpp`](libraries/Matter/src/MatterEndpoints/MatterOnOffPlugin.cpp)
**Function:** `bool MatterOnOffPlugin::setOnOff(bool newState)` (lines 98-127)

## The user's proposal

Move the assignment

```cpp
onOffState = newState;
```

from **line 123** (after a successful `updateAttributeVal()`) to **line 108**
(right after the no-op short-circuit on line 105-107, before any interaction
with the Matter cluster).

## Verdict

**The proposal is incorrect and would break the failure-mode semantics of the
class.** Keep the assignment at line 123.

## What the current placement actually means

`setOnOff()` is structured as a classic **read-modify-write with
commit-on-success**:

```cpp
bool MatterOnOffPlugin::setOnOff(bool newState) {
  if (!started) {                                                   // (1) guard
    log_e("Matter On-Off Plugin device has not begun.");
    return false;
  }

  if (onOffState == newState) {                                     // (2) no-op
    return true;
  }

  esp_matter_attr_val_t onoffVal = esp_matter_invalid(NULL);

  if (!getAttributeVal(OnOff::Id,                                   // (3) READ
                       OnOff::Attributes::OnOff::Id, &onoffVal)) {
    log_e("Failed to get On-Off Plugin Attribute.");
    return false;                                                   //   fail: do NOT update shadow
  }
  if (onoffVal.val.b != newState) {
    onoffVal.val.b = newState;                                      // (4a) MODIFY local copy
    bool ret;
    ret = updateAttributeVal(OnOff::Id,                             // (4b) WRITE cluster
                             OnOff::Attributes::OnOff::Id, &onoffVal);
    if (!ret) {
      log_e("Failed to update On-Off Plugin Attribute.");
      return false;                                                 //   fail: do NOT update shadow
    }
    onOffState = newState;                                          // (5) COMMIT shadow
  }
  log_v("Plugin OnOff state set to %s", newState ? "ON" : "OFF");
  return true;
}
```

`onOffState` is a **shadow** of what the Matter data model holds for the
cluster attribute. The class's invariant is:

> `onOffState` mirrors the cluster attribute that a Matter controller would
> read.

Line 123 is the *commit* step that restores that invariant after the Matter
cluster has been successfully updated.

## Why moving the assignment to line 108 breaks the class

If line 123 moves to line 108, the failure paths become inconsistent:

| Failure                       | Current behavior (line 123)                                          | Proposed behavior (line 108)                                                  |
| ----------------------------- | -------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
| `getAttributeVal` fails       | Returns `false`. `onOffState` keeps OLD value. Matter cluster keeps OLD value. **Consistent.** | Returns `false`. `onOffState` is NEW. Matter cluster is OLD. **Inconsistent.** |
| `updateAttributeVal` fails    | Returns `false`. `onOffState` keeps OLD value. Matter cluster keeps OLD value. **Consistent.** | Returns `false`. `onOffState` is NEW. Matter cluster is OLD. **Inconsistent.** |
| Both Matter calls succeed     | Returns `true`. `onOffState` is NEW. Matter cluster is NEW. **Consistent.** | Same as current. **Consistent.**                                              |

A transient Matter SDK failure (memory pressure, cluster locked by an
ongoing operation, plugin tear-down in progress, ...) would leave the
wrapper **lying** about the actual cluster value:

* `MatterOnOffPlugin::getOnOff()` (which returns `onOffState`) would report
  `true`.
* A Matter controller subscribed to the same attribute would still read
  `false`, because the cluster write failed.

From the controller's point of view nothing changed; from the local
application's point of view the device is "on". That is exactly the
split-brain state the current ordering is designed to prevent.

Knock-on consequences:

* `toggle()` is implemented as `setOnOff(!onOffState)`. If a failed write
  ever flips `onOffState` without flipping the cluster, the next `toggle()`
  computes "the opposite of what we *thought* we set" rather than "the
  opposite of what is actually in the cluster".
* `operator bool()` returns `getOnOff()`, so user code that does
  `if (myPlugin) { ... }` would also believe the wrong state.

## Doesn't `updateAttributeVal()` need `onOffState` to be updated first?

No.

* Line 118 passes `&onoffVal` to the SDK - a **separate local variable**
  that is filled with `newState` on line 116.
* `updateAttributeVal` never reads `onOffState`.

Moving line 123 to line 108 therefore has **zero effect** on what the SDK
sees. It only affects what the *application* sees if it asks `getOnOff()`
(or `operator bool()`) between line 108 and line 123, and that is only
possible from inside a callback that the SDK itself fired during line 118.
See next section.

## Interaction with `attributeChangeCB`

`onOffState` is also written in a second place, the SDK-fired attribute
change callback (`MatterOnOffPlugin::attributeChangeCB`, lines 26-54):

```cpp
if (attribute_id == OnOff::Attributes::OnOff::Id) {
  if (_onChangeOnOffCB != NULL) {
    ret &= _onChangeOnOffCB(val->val.b);
  }
  if (_onChangeCB != NULL) {
    ret &= _onChangeCB(val->val.b);
  }
  if (ret == true) {
    onOffState = val->val.b;     // <-- second writer
  }
}
```

There are two timing scenarios:

* **SDK calls back synchronously inside `updateAttributeVal()`**. The
  callback writes `onOffState = val->val.b` before line 123 runs. Line 123
  then writes the same value - redundant but harmless.
* **SDK queues the callback for later (async)**. Line 123 is then the only
  path that updates `onOffState` *before `setOnOff()` returns*. This is
  precisely what callers expect: "if `setOnOff()` returned `true`,
  `getOnOff()` already reflects the new state."

In both scenarios the current placement is correct.

## The "no-change" guard on line 105 is unrelated

`if (onOffState == newState) return true;` is on line 105 - **before** line
108 in either proposal. Both the current code and the user's proposal pass
through it identically. So this guard is **not** a justification for moving
line 123 earlier.

The guard does protect against re-entrant cycles: the user's
`_onChangeOnOffCB` / `_onChangeCB` might in turn call `setOnOff(...)`. By
the time the inner `setOnOff()` runs, `attributeChangeCB` has already done
`onOffState = val->val.b`, so the inner `setOnOff(sameValue)` short-circuits
and returns `true` without re-entering `updateAttributeVal()`.

## When would the proposal be *partially* right?

There is exactly one practical scenario where the early update would
*appear* helpful:

> "I want `getOnOff()` (called from inside my `_onChangeCB`) to already
> return the new state."

The proper fix for that is **not** to move line 123 earlier, because doing
so still leaves the Matter cluster write uncommitted. The proper fix is one
of:

1. Use the `bool` argument the callback already receives (`val->val.b`),
   instead of calling `getOnOff()` inside the callback.
2. In `attributeChangeCB`, update `onOffState = val->val.b` **before**
   invoking `_onChangeOnOffCB` / `_onChangeCB`. The user callbacks then see
   a consistent shadow whether they read the parameter or call
   `getOnOff()`.

Option 2 is a small, localized change inside `attributeChangeCB` and it
does not break the failure-mode invariant of `setOnOff()`.

## Conclusion

* **Keep line 123 where it is.** It is the commit step of a successful
  write to the Matter cluster.
* **Do NOT move it to line 108.** That would let the local shadow advance
  even when the Matter SDK rejected the write, producing a permanent
  inconsistency between what `getOnOff()` returns and what a Matter
  controller sees on the cluster attribute.
* If the underlying complaint is "`getOnOff()` returns stale data inside
  my callback", the fix lives in `attributeChangeCB` (update `onOffState`
  before invoking the user callbacks), **not** in `setOnOff()`.

libraries/Matter/src/MatterEndpoints/MatterContactSensor.cpp
lucasssvaz: Indeed, esp-matter uses the old ember storage pattern. I missed that. Will revert these.

libraries/Matter/src/MatterEndpoints/MatterEnhancedColorLight.cpp
lucasssvaz: This is not redundant. `getBrightness()` returns `brightnessLevel`, not a value from the attribute store. Without this initialization, `getBrightness()` returns `0` immediately after `begin(true, 128)`.
Both `MatterDimmableLight::begin()` and `MatterColorTemperatureLight::begin()` already have this exact initialization:
```cpp
// MatterDimmableLight::begin() — line 91-93
light_config.level_control.current_level = brightness;
light_config.level_control.lighting.start_up_current_level = nullptr;
brightnessLevel = brightness;  // already exists
// MatterColorTemperatureLight::begin() — line 106-108
light_config.level_control.current_level = brightness;
light_config.level_control.lighting.start_up_current_level = nullptr;
brightnessLevel = brightness;  // already exists
```
`MatterEnhancedColorLight` was the only light endpoint missing it. This fix aligns it with the existing pattern.

SuGlider:
**Where the PR over-reaches** - the seven simple setters' reorder is a silent behavioural change that disables the existing user-callback veto.

lucasssvaz:
@SuGlider Indeed the order change is a mistake and esp-matter uses the old ember pattern. I will revert it.

Either way there are inconsistencies across the EPs that this PR fixes. PTAL.

libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -516,22 +515,22 @@ bool MatterWindowCovering::setTiltPosition(uint16_t tiltPosition) {
...
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionTilt::Id, &val);
+    }
     if (ret) {
       currentTiltPosition = tiltPosition;
-      // Also update CurrentPositionTiltPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's TiltPositionSet() behavior
-      setTiltPercentage((uint8_t)(tiltPercent100ths / 100));
+      if (!setTiltPercentage((uint8_t)(tiltPercent100ths / 100))) {
+        log_e("Failed to sync Tilt Percentage from Tilt Position.");
+        return false;
```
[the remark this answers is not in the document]
lucasssvaz: currentLiftPosition is updated after its own attribute store write succeeds, consistent with the Ember pattern. The percentage sync is a secondary attribute — deferring the position assignment would create a worse inconsistency (store has new value, getter returns old). The function returns false to signal the partial failure to the caller.

libraries/Matter/src/MatterEndpoints/MatterWindowCovering.cpp
```diff
@@ -361,22 +361,22 @@ bool MatterWindowCovering::setLiftPosition(uint16_t liftPosition) {
...
+      ret = setAttributeVal(WindowCovering::Id, WindowCovering::Attributes::CurrentPositionLift::Id, &val);
+    }
     if (ret) {
       currentLiftPosition = liftPosition;
-      // Also update CurrentPositionLiftPercent100ths to keep attributes in sync
-      // This matches ESP-Matter's LiftPositionSet() behavior
-      setLiftPercentage((uint8_t)(liftPercent100ths / 100));
+      if (!setLiftPercentage((uint8_t)(liftPercent100ths / 100))) {
+        log_e("Failed to sync Lift Percentage from Lift Position.");
+        return false;
```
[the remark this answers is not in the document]
lucasssvaz: currentLiftPosition is updated after its own attribute store write succeeds, consistent with the Ember pattern. The percentage sync is a secondary attribute — deferring the position assignment would create a worse inconsistency (store has new value, getter returns old). The function returns false to signal the partial failure to the caller.

libraries/Matter/src/MatterEndpoints/MatterEnhancedColorLight.cpp
SuGlider: This change is approved.

SuGlider (approved):
This last version is fine. Approved.

PR #12623, opened 2026-05-27
fix(hosted): Fix hosted slave chip null if slave version is old

me-no-dev:
This pull request introduces a conditional check in the `hostedGetSlaveTargetName()` function to ensure compatibility with different slave firmware versions. The main change is to verify the slave version before proceeding, which helps prevent issues when interacting with older firmware.

**Compatibility improvements:**

* Added a version check for the slave firmware in the `hostedGetSlaveTargetName()` function to ensure that if the slave version is older than `2.12.2`, the function returns `CONFIG_IDF_SLAVE_TARGET` early, maintaining compatibility with older firmware versions.


me-no-dev:
@lucasssvaz PTAL

P-R-O-C-H-Y (approved):

lucasssvaz (approved):
Works as expected

PR #12645, opened 2026-06-02
fix(usb_cdc): Fix blocking write() when timeout is zero

### LINKED ISSUE #12634: fix infinite-blocking write() when setTxTimeoutMs(0) is used
### Board

esp32s2

### Device Description

heltec wireless tracker

### Hardware Configuration

usb-serial via usb-c plug

### Version

### Type

Task

### IDE Name

platformio

### Operating System

macosx

### Flash frequency

80MHtz I think

### PSRAM enabled

yes

### Upload speed

115200

## Bug

  `Serial.setTxTimeoutMs(0)` is documented (and named) as a way to make
  USB CDC writes non-blocking. In practice, on ESP32-S3 with
  `ARDUINO_USB_CDC_ON_BOOT=1` and no USB host attached, `Serial.write()`
  blocks for ~49.7 days.

  ## Cause

  In `cores/esp32/HWCDC.cpp::write()`, the `tries` counter (uint32_t)
  is initialized from `tx_timeout_ms` and decremented before the
  `tries == 0` break check. When `tx_timeout_ms == 0`, the first
  "no progress" iteration underflows `tries` to `0xFFFFFFFF` and the
  break condition is never reached. The loop continues spinning with
  `delay(1)` per iteration for 2^32 ms.

  ## Repro

  ESP32-S3 board (e.g. Heltec Wireless Tracker), no USB host attached:

  ```cpp
  void setup() {
    Serial.begin(115200);
    Serial.setTxTimeoutMs(0);   // expect: non-blocking writes
    Serial.println("hello");    // blocks until USB host connects
  }

  Fix

  Move the tries == 0 check before the decrement. Behavior with
  tx_timeout_ms > 0 is unchanged. With tx_timeout_ms == 0 the loop
  now aborts on the first stuck iteration, as expected.

  flush() is not affected -- its loop condition
  (while(connected && tries && uxItemsWaiting)) already guards against
  entering with tries == 0.



patch:

diff --git a/cores/esp32/HWCDC.cpp b/cores/esp32/HWCDC.cpp
--- a/cores/esp32/HWCDC.cpp
+++ b/cores/esp32/HWCDC.cpp
@@ -454,17 +454,16 @@ size_t HWCDC::write(const uint8_t *buffer, size_t size)
                 usb_serial_jtag_ll_txfifo_flush();
                 if(connected) usb_serial_jtag_ll_ena_intr_mask(USB_SERIAL_JTAG_INTR_SERIAL_IN_EMPTY);
                 if(last_toSend == to_send) {
-                    // no progress in sending data... USB CDC is probably unplugged
-                    tries--;
+                    // no progress in sending data... USB CDC is probably unplugged.
+                    // Check BEFORE decrementing so tx_timeout_ms == 0 aborts
+                    // immediately instead of underflowing 'tries' to 0xFFFFFFFF.
+                    if (tries == 0) { // CDC isn't connected (or non-blocking requested)
+                        size = so_far;
+                        log_w("write failed due to waiting USB Host - timeout");
+                        connected = false;
+                        break;
+                    }
+                    tries--;
                     delay(1);
                 } else {
                     last_toSend = to_send;
                     tries = tx_timeout_ms; // reset the timeout
                 }
-                if (tries == 0) {  // CDC isn't connected anymore...
-                    size = so_far;
-                    log_w("write failed due to waiting USB Host - timeout");
-                    connected = false;
-                }
             }
         }
         // CDC was diconnected while sending data ==> flush the TX buffer keeping the last data



### Sketch

```cpp
void setup() {
    Serial.begin(115200);
    Serial.setTxTimeoutMs(0);   // expect: non-blocking writes
    Serial.println("hello");    // blocks until USB host connects
  }
```

### Debug Message

```plain
In my case, my firmware wrote, after setting Serial.setTxTimeoutMs(0); via Serial.println() a message. Firmware waited forever. 
A minute later I connected to the device via cu -s 115200 -l /dev/cu.usbmodem14401 and it continued, and cu showed the one minute old message (write() had decided to wait up to 49.7 days).
```

### Other Steps to Reproduce

_No response_

SuGlider:
## Description of Change
With USB OTG CDC, when timeout is set to zero, the interface gets locked in an infinite loop.
This PR fixes it.

## Test Scenarios
ESP32-S3 | ESP32-S2 using USB OTG (TinyUSB) CDC.

```cpp
void setup() {
  Serial.begin();
  delay(2000);
  Serial.println("It will set timeout to 0 and try to print something...");
  Serial.setTxTimeoutMs(0);   // expect: non-blocking writes
  Serial.println("hello");    // blocks until USB host connects
}

void loop() {
  Serial.println("loop...");
  delay(1000);
}
```

## Related links
Fix #12634

### DIFF
--- a/cores/esp32/USBCDC.cpp
+++ b/cores/esp32/USBCDC.cpp
@@ -406,9 +406,10 @@ size_t USBCDC::write(const uint8_t *buffer, size_t size) {
   }
   size_t to_send = size, so_far = 0;
   // writeTimeout will prevent that TinyUSB failure locks the while(to_send) loop
+  bool non_blocking = tx_timeout_ms == 0;
   uint32_t writeTimeout = millis() + tx_timeout_ms;
   while (to_send) {
-    if (!tud_cdc_n_connected(itf) || (int32_t)(millis() - writeTimeout) >= 0) {
+    if (!tud_cdc_n_connected(itf) || (!non_blocking && (int32_t)(millis() - writeTimeout) >= 0)) {
       log_e("USB is disconnected or CDC writing has timed out.");
       size = so_far;
       break;
@@ -416,6 +417,10 @@ size_t USBCDC::write(const uint8_t *buffer, size_t size) {
     size_t space = tud_cdc_n_write_available(itf);
     if (!space) {
       tud_cdc_n_write_flush(itf);
+      if (non_blocking) {
+        size = so_far;
+        break;
+      }
       continue;
     }
     if (space > to_send) {


lucasssvaz (approved):
LGTM

P-R-O-C-H-Y (approved):

me-no-dev (approved):

PR #12662, opened 2026-06-05
feat(esp_video): Add the ESP_Video library, example and document.

donghengqaz:
# Description of Change

This pull request introduces the ESP_Video Arduino library, along with API documentation and a basic capture example. It wraps the ESP-IDF esp_video component and exposes a V4L2-style capture API for camera sensors on supported SoCs, covering MIPI-CSI (with ISP) and DVP interfaces. The new library and documentation make it easier for Arduino developers to initialize camera hardware, open video device nodes, allocate buffers, dequeue frames, and tune sensor controls without dealing with low-level V4L2 ioctl calls directly.

### ESP_Video Library:

Added ESP_Video.hpp / ESP_Video.cpp with a four-layer API: hardware configuration (ESPVideCamConfigClass, ESPVideoCSIConfigClass, ESPVideoDVPConfigClass), hardware initialization (ESPVideoClass::init()), capture device (ESPVideoCaptureDevClass), and RAII frame buffer (ESPVideoBufferClass).
Supports SCCB/I2C camera configuration, CSI and DVP pin setup, V4L2 MMAP buffer management, and sensor controls (gain, exposure, JPEG quality, flip, test pattern, etc.).
Added library.properties, keywords.txt, and build integration via CMakeLists.txt, Kconfig.projbuild (selective compilation), and idf_component.yml (espressif/esp_video dependency for esp32s3, esp32p4).

### API Documentation:

Added docs/en/api/esp_video.rst (~430 lines) describing supported targets, video device nodes (/dev/video0–/dev/video2), the recommended capture flow, configuration classes, capture device methods, sensor control APIs, and ESPVideoBufferClass lifetime semantics.

### Basic Capture Example:

Added libraries/ESP_Video/examples/Basic/Basic.ino demonstrating end-to-end camera capture on ESP32-P4 (MIPI-CSI and/or DVP, ESP32-P4-Function-EV-Board V1.5 pin defines) and ESP32-S3 (DVP, ESP32-S3-EYE pin defines).
Added libraries/ESP_Video/examples/Basic/ci.yml specifying required FQBN options and CONFIG_ESP_VIDEO_ENABLE=y for CI builds on esp32s3 and esp32p4.
These additions provide a complete starting point for camera preview, frame grabbing, and sensor tuning workflows on Arduino-ESP32.

# Test Scenarios

Build and run the Basic example on ESP32-P4-Function-EV-Board V1.5 with DVP camera sensor enabled.
Build and run the Basic example on ESP32-S3-EYE with DVP camera sensor enabled.
Verify serial output reports captured buffer address and size in a continuous capture loop.

# Related links
https://github.com/espressif/esp32-arduino-lib-builder/pull/375


libraries/ESP_Video/src/ESP_Video.h
lucasssvaz, line 1: Please use the `.h` extension for the header to keep in line with the rest of the core

libraries/ESP_Video/examples/dvp_camera/dvp_camera.ino
lucasssvaz, line 1: Add a header comment in the example to explain what it does and how it works.

libraries/ESP_Video/library.properties
```diff
@@ -0,0 +1,10 @@
+name=ESP_Video
+version=1.0.0
```
lucasssvaz: Keep the version the same as the core.

```suggestion
version=3.3.10
```

libraries/ESP_Video/src/ESP_Video.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+private:
+  void reclaim();
+
+  int buf_fd = -1;
+  uint8_t *buf_ptr = nullptr;
+  size_t buf_size = 0;
+  int buf_index = -1;
+  __u32 buf_type = 0;
+  __u32 memory_type = 0;
```
lucasssvaz: AFAIK the `__u32` notation is from the Linux kernel. Prefer using `uint32_t` if possible to make it uniform across the core.

```suggestion
  uint32_t buf_type = 0;
  uint32_t memory_type = 0;
```

libraries/ESP_Video/src/ESP_Video.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+private:
+  uint32_t xclk_freq;
+};
+
+class ESPVideoBufferClass {
+public:
+  ESPVideoBufferClass() = default;
+
+  ESPVideoBufferClass(int fd, uint8_t *ptr, size_t n, int index, __u32 type, __u32 memory);
```
lucasssvaz: AFAIK the `__u32` notation is from the Linux kernel. Prefer using `uint32_t` if possible to make it uniform across the core.

```suggestion
  ESPVideoBufferClass(int fd, uint8_t *ptr, size_t n, int index, uint32_t type, uint32_t memory);
```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,446 @@
...
+#if CONFIG_ESP_VIDEO_ENABLE_DVP_VIDEO_DEVICE
+  cam_config.dvp = &dvp_config;
+#endif
+
+  if (esp_video_init_with_flags(&cam_config, target_flags) == ESP_OK) {
+    flags_ |= target_flags;
+  } else {
+    log_e("failed to init DVP");
+    flags_ &= ~target_flags;
+  }
```
lucasssvaz: The pin assignments need to use the Peripheral Manager to avoid conflicts. @P-R-O-C-H-Y might be able to help and explain better

libraries/ESP_Video/examples/Basic/ci.yml
```diff
@@ -0,0 +1,16 @@
...
+requires:
+  - CONFIG_ESP_VIDEO_ENABLE=y
+
+targets:
+  esp32: false
+  esp32c3: false
+  esp32c6: false
+  esp32h2: false
+  esp32s2: false
+  esp32c5: false
```
lucasssvaz: No need to add this if `CONFIG_ESP_VIDEO_ENABLE` only is defined for S3 and P4.

libraries/ESP_Video/examples/Basic/ci.yml
```diff
@@ -0,0 +1,16 @@
+fqbn:
+  esp32s3:
+    - espressif:esp32:esp32s3:USBMode=default,PartitionScheme=esp_sr_16,FlashSize=16M,FlashMode=dio
+  esp32p4:
+    - espressif:esp32:esp32p4:USBMode=default,ChipVariant=postv3,PartitionScheme=esp_sr_16,FlashSize=16M,FlashMode=qio
```
lucasssvaz: Any specific reason why one is running in `dio` mode and the other in `qio` ? Also, is using the Speech Recognition partition correct ? You might want to simplify this by doing:


```suggestion
fqbn_append: PartitionScheme=esp_sr_16,FlashSize=16M
```

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  ESPVideoBufferClass buffer = capture_dev->CaptureBuffer();
+  if (!buffer.valid()) {
+    Serial.println("failed to capture buffer");
+    delay(10);
+    return;
+  }
+
+  Serial.printf("captured buffer: %p size: %zu\n", buffer.data(), buffer.size());
```
lucasssvaz: To work with all the std libs as described [here](https://docs.espressif.com/projects/arduino-esp32/en/latest/guides/coding_guidelines.html#format-specifiers)

```suggestion
  Serial.printf("captured buffer: %p size: %lu\n", buffer.data(), (unsigned long)buffer.size());
```

libraries/ESP_Video/examples/Basic/ci.yml
```diff
@@ -0,0 +1,16 @@
...
+
+requires:
+  - CONFIG_ESP_VIDEO_ENABLE=y
```
lucasssvaz: Where does this Kconfig option comes from ? I can't find it in https://github.com/espressif/esp-video-components/blob/master/esp_video/Kconfig

donghengqaz: Yes, it is not defined. I will remove this.

libraries/ESP_Video/src/ESP_Video.cpp
donghengqaz: I try to add the Peripheral Manager for sensor interface I/Os.

libraries/ESP_Video/examples/Basic/ci.yml
lucasssvaz: Maybe you can use a regex to avoid having to hardcode anything:

```
requires:
  - CONFIG_CAMERA_.+=y
```

donghengqaz: I checked some examples' CI, they mostly use the CONFIG_SOC_XXX_SUPPORTED, I think it is also OK for this, because the example needs CONFIG_SOC_LCDCAM_CAM_SUPPORTED and CONFIG_SOC_MIPI_CSI_SUPPORTED.

donghengqaz:
@lucasssvaz  The esp-video requires esp-idf > v5.4. How could I add this limitation to the library and test example?

I add ESP_IDF_VERSION to manage these.

lucasssvaz:
@donghengqaz For the IDF version you can use the `idf_component.yml`. For example: https://github.com/espressif/arduino-esp32/blob/master/idf_component.yml#L81-L85

donghengqaz:
> @donghengqaz For the IDF version you can use the `idf_component.yml`. For example: https://github.com/espressif/arduino-esp32/blob/master/idf_component.yml#L81-L85

@lucasssvaz  Yes, I added this. But the library example compiles failed, PTAL: 

 - https://github.com/espressif/arduino-esp32/actions/runs/27180159763/job/80267592996?pr=12662

  - https://github.com/espressif/arduino-esp32/actions/runs/27180159763/job/80267593022?pr=12662

The release/v6.0 toolchain is older; it lacks some libc/libm functions: https://github.com/espressif/arduino-esp32/actions/runs/27180159763/job/80267593233?pr=12662

The built library [MR](https://github.com/espressif/esp32-arduino-lib-builder/pull/375) should be merged first, to fix the compilation issues: https://github.com/espressif/arduino-esp32/actions/runs/27180159744/job/80267620518?pr=12662.

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,191 @@
...
+ *    destroyed at the end of each iteration.
+ *
+ * Open Serial Monitor at 115200 baud to see capture status and frame metadata.
+ * Adapt the EXAMPLE_* pin defines if your camera is wired to a different board.
+ */
+#include "Arduino.h"
+#include <array>
+#include <memory>
+#include <ESP_Video.h>
+#include "esp_idf_version.h"
```
me-no-dev: this should be already included by including Arduino.h

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,191 @@
...
+#define EXAMPLE_DVP_D6_PIN           17
+#define EXAMPLE_DVP_D7_PIN           16
+#define EXAMPLE_DVP_XCLK_FREQ        10000000
+
+#define EXAMPLE_ENABLE_DVP_CAM_SENSOR 1
+#else
+#error "The selected target SoC is not supported"
+#endif
+
+namespace {
```
me-no-dev: avoid namespaces, statics, unique pointers and such in Arduino sketches. They need to be in simpler C++ language.
```cpp
ESPVideoClass video;
ESPVideoCaptureDevClass capture_dev;
const char *kCaptureDevicePath = ESP_VIDEO_MIPI_CSI_DEVICE_NAME;
```

Then instead of make unique, you will call `capture_dev.begin()` to initialize it.

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,191 @@
...
+#endif
+
+static constexpr size_t kCaptureBufferCount = 2;
+
+bool init_camera_hardware() {
+#if EXAMPLE_ENABLE_MIPI_CSI_CAM_SENSOR
+  ESPVideoCSIConfigClass csi_config(ESPVideoCamConfigClass(EXAMPLE_MIPI_CSI_SCCB_I2C_PORT, EXAMPLE_MIPI_CSI_SCCB_I2C_SCL_PIN, EXAMPLE_MIPI_CSI_SCCB_I2C_SDA_PIN)
+  );
+
+  if (!video.init(csi_config)) {
```
me-no-dev: This should be `video.begin(csi_config);` Arduino uses `begin()` and `end()` to initialize/deinit things

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,717 @@
...
+    sda_pin = config.sda_pin;
+    i2c_freq = config.i2c_freq;
+    reset_pin = config.reset_pin;
+    pwdn_pin = config.pwdn_pin;
+    i2c_handle = config.i2c_handle;
+  }
+  return *this;
+}
+
+bool ESPVideoCamConfigClass::NeedInitI2C() const {
```
me-no-dev: methods in Arduino start with lower case -> `needInitI2C`

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,717 @@
...
+      log_e("failed to queue video frame");
+    }
+
+    buf_fd = -1;
+    buf_index = -1;
+  }
+}
+
+ESPVideoCaptureDevClass::ESPVideoCaptureDevClass(const char *path, size_t buffer_count) {
+  if (!Open(path)) {
```
me-no-dev: all that needs to happen in a `begin()` method, so that class instance can be global and would not try to access devices before the sketch has started.

me-no-dev:
> The built library https://github.com/espressif/esp32-arduino-lib-builder/pull/375 should be merged first, to fix the compilation issues: https://github.com/espressif/arduino-esp32/actions/runs/27180159744/job/80267620518?pr=12662

You should test if the include exists and only then compile the code. For IDF5.3 that will not be true for sure and we keep the core compatible with IDF 5.3 onward

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,178 @@
...
+    return;
+  }
+#elif EXAMPLE_ENABLE_DVP_CAM_SENSOR
+  ESPVideoCamConfigClass cam_config;
+  cam_config.begin(EXAMPLE_DVP_SCCB_I2C_PORT, EXAMPLE_DVP_SCCB_I2C_SCL_PIN, EXAMPLE_DVP_SCCB_I2C_SDA_PIN);
+
+  ESPVideoDVPPinsConfigClass dvp_pins;
+  dvp_pins.begin(
+    EXAMPLE_DVP_VSYNC_PIN, EXAMPLE_DVP_DE_PIN, EXAMPLE_DVP_PCLK_PIN, EXAMPLE_DVP_XCLK_PIN,
+    std::array<int8_t, 8>{
```
me-no-dev: do not use array, instead have the pins as separate arguments. Yes, the arguments list becomes long, but this will be easier for users to understand.

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,178 @@
...
+#define EXAMPLE_DVP_D4_PIN           3
+#define EXAMPLE_DVP_D5_PIN           6
+#define EXAMPLE_DVP_D6_PIN           5
+#define EXAMPLE_DVP_D7_PIN           21
+#define EXAMPLE_DVP_XCLK_FREQ        20000000
+
+/**
+ * @brief Enable the MIPI-CSI camera sensor or DVP camera sensor
+ */
+#define EXAMPLE_ENABLE_MIPI_CSI_CAM_SENSOR 0
```
me-no-dev: I suggest to enable the MIPI sensor for P4 by default. This will be actually what users expect. Also note which sensors are supported. Same should be true for DVP as well. For S3 we also have the other camera driver example, so users should be able to choose which one to use depending on the supported sensors.

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,178 @@
...
+#define CAPTURE_DEVICE_PATH ESP_VIDEO_MIPI_CSI_DEVICE_NAME
+#elif EXAMPLE_ENABLE_DVP_CAM_SENSOR
+#define CAPTURE_DEVICE_PATH ESP_VIDEO_DVP_DEVICE_NAME
+#else
+#error "No camera sensor interface selected"
+#endif
+
+ESPVideoClass video;
+ESPVideoCaptureDevClass capture_dev;
+const char *kCaptureDevicePath = CAPTURE_DEVICE_PATH;
```
me-no-dev: could this be something different than the pre-defined values? If not, then this should be part of the library code and selected there internally. No need to bother the user with it.

libraries/ESP_Video/examples/dvp_camera/dvp_camera.ino
```diff
@@ -0,0 +1,178 @@
...
+#elif EXAMPLE_ENABLE_DVP_CAM_SENSOR
+#define CAPTURE_DEVICE_PATH ESP_VIDEO_DVP_DEVICE_NAME
+#else
+#error "No camera sensor interface selected"
+#endif
+
+ESPVideoClass video;
+ESPVideoCaptureDevClass capture_dev;
+const char *kCaptureDevicePath = CAPTURE_DEVICE_PATH;
+const size_t kCaptureBufferCount = 2;
```
me-no-dev, line 86: add comment what this value means and why it is selected to be `2`

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,178 @@
...
+const size_t kCaptureBufferCount = 2;
+
+void setup() {
+  Serial.begin(115200);
+
+#if EXAMPLE_ENABLE_MIPI_CSI_CAM_SENSOR
+  ESPVideoCamConfigClass cam_config;
+  cam_config.begin(EXAMPLE_MIPI_CSI_SCCB_I2C_PORT, EXAMPLE_MIPI_CSI_SCCB_I2C_SCL_PIN, EXAMPLE_MIPI_CSI_SCCB_I2C_SDA_PIN);
```
me-no-dev: you can omit the `_MIPI_CSI_` `_DVP_` parts of the SCCB port. They are the same values for both. `EXAMPLE_SCCB_I2C_*` will be enough

libraries/ESP_Video/examples/Basic/Basic.ino
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  ESPVideoBufferClass buffer = capture_dev.captureBuffer();
+  if (!buffer.valid()) {
+    Serial.println("failed to capture buffer");
+    delay(10);
+    return;
+  }
+
+  Serial.printf("captured buffer: %p size: %lu\n", buffer.data(), (unsigned long)buffer.size());
```
me-no-dev: What is the format of the captured buffer?

libraries/ESP_Video/src/ESP_Video.h
```diff
@@ -0,0 +1,248 @@
...
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+#include "Arduino.h"
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+#include <memory>
+#include "linux/videodev2.h"
+#include "esp_video_init.h"
+#include "esp_video_device.h"
```
me-no-dev, line 18: include i the header only things required by the header. All else should be included in the implementation.

libraries/ESP_Video/src/ESP_Video.h
```diff
@@ -0,0 +1,248 @@
...
+public:
+  ESPVideoCamConfigClass() = default;
+  ESPVideoCamConfigClass(const ESPVideoCamConfigClass &config);
+
+  ESPVideoCamConfigClass operator=(const ESPVideoCamConfigClass &config);
+
+  virtual ~ESPVideoCamConfigClass() = default;
+
+  bool begin(i2c_port_num_t port, int8_t scl_pin, int8_t sda_pin, uint32_t i2c_freq = i2c_freq_hz, int8_t reset_pin = -1, int8_t pwdn_pin = -1);
+  bool begin(i2c_master_bus_handle_t i2c_handle, uint32_t i2c_freq = i2c_freq_hz, int8_t reset_pin = -1, int8_t pwdn_pin = -1);
```
me-no-dev, line 79: what is the value of `i2c_freq_hz`? Where does it come from?

idf_component.yml
```diff
@@ -161,6 +161,10 @@ dependencies:
     version: "^1.0.0"
     rules:
       - if: "idf_version >= 6.0 && target != esp32c2"
+  espressif/esp_video:
+    version: "^2.2.0"
+    rules:
+      - if: "target in [esp32s3, esp32p4]"
```
me-no-dev: `- if: "idf_version >= 5.4 && target in [esp32s3, esp32p4]"`

libraries/ESP_Video/examples/Basic/Basic.ino
me-no-dev: Also unknown is the frame width and height. How to get/control that?

libraries/ESP_Video/examples/Basic/ci.yml
```diff
@@ -0,0 +1,5 @@
+fqbn_append: PartitionScheme=esp_sr_16,FlashSize=16M
+
+requires:
+  - CONFIG_SOC_LCDCAM_CAM_SUPPORTED=y
+  - CONFIG_SOC_MIPI_CSI_SUPPORTED=y
```
me-no-dev: shouldn't this be based on `CONFIG_ESP_VIDEO_ENABLE_MIPI_CSI_VIDEO_DEVICE` and `CONFIG_ESP_VIDEO_ENABLE_DVP_VIDEO_DEVICE`?

idf_component.yml
me-no-dev: Since you also include this in your lib-builder PR, I suggest to remove it from here. If implemented properly, the library and example should not be compiled if the component is missing

libraries/ESP_Video/examples/Basic/Basic.ino
donghengqaz: Yes, I enable the MIPI and DVP for P4 by default, and enable DVP for S3 by default. Separating the example to be DVP or MIPI is also well.

libraries/ESP_Video/examples/Basic/ci.yml
donghengqaz: Yes, directly using this definition is well.

libraries/ESP_Video/src/ESP_Video.h
donghengqaz: `i2c_freq_hz` is the default I2C frequency in HZ, it is defined in the class ESPVideoCamConfigClass.

libraries/ESP_Video/examples/Basic/Basic.ino
donghengqaz: Separating the example into DVP and MIPI is better to avoid using many predefined values.

libraries/ESP_Video/examples/Basic/Basic.ino
donghengqaz: This format is based on the sensor configuration for the DVP camera sensor; the format is fixed, and for the MIPI, the users can select the target format.

docs/en/api/esp_video.rst
```diff
@@ -0,0 +1,965 @@
...
+        // use frame; buffer is returned to the driver when `frame` is destroyed
+    }
+
+Design notes
+************
+
+* **Configuration objects are value types.** They store pin and bus settings only and can be copied or assigned. Pin reservation through ``periman`` is performed by ``ESPVideoClass::begin()`` and released in ``ESPVideoClass::end()`` or the destructor.
+* **CSI and DVP can coexist** on SoCs that support both (for example ESP32-P4). Each interface is tracked independently via internal flags and released in ``end()`` or the destructor.
+* **Capture device opening is unified.** ``ESPVideoCaptureDevClass::begin()`` opens the device, reads and re-applies the current V4L2 format (``VIDIOC_G_FMT`` / ``VIDIOC_S_FMT``), stores width and height, and requests capture buffers.
+* **Format and resolution are shared base classes.** ``ESPVideoFormatClass`` and ``ESPVideoSolutionClass`` hold pixel format and frame dimensions. Both ``ESPVideoCaptureDevClass`` and ``ESPVideoBufferClass`` inherit from them so metadata is available on the device and on each dequeued frame.
```
pedrominatel: ...inherit from them, so metadata is available...
Missing comma.

docs/en/api/esp_video.rst
```diff
@@ -0,0 +1,965 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: cpp
+
+    bool begin(i2c_master_bus_handle_t i2c_handle,
+               uint32_t i2c_freq = i2c_freq_hz,
+               int8_t reset_pin = -1, int8_t pwdn_pin = -1);
+
+Use this when SCCB should reuse an I2C bus already created by your application.
+In that case the library does not initialize a new I2C driver for SCCB.
```
pedrominatel: In that case, the library does not initialize a new I2C driver for SCCB.
Missing comma.

docs/en/api/esp_video.rst
```diff
@@ -0,0 +1,965 @@
...
+^^^^^
+
+.. code-block:: cpp
+
+    bool begin(const char *path, size_t buf_count = 2);
+
+Opens the device at ``path`` if not already open, reads the driver format (``VIDIOC_G_FMT``), re-applies it (``VIDIOC_S_FMT``), stores width, height, and pixel format, and requests ``buf_count`` MMAP capture buffers.
+The detected format and resolution are exposed through ``getFormat()``, ``getFormatName()``, ``getWidth()``, and ``getHeight()``.
+If the device is already open, returns ``true`` immediately without changing the path.
+On any failure the device is closed and the instance remains unopened.
```
pedrominatel: On any failure, the device is closed and the instance remains unopened.
Missing comma.

docs/en/api/esp_video.rst
```diff
@@ -0,0 +1,965 @@
...
+
+Sets ``V4L2_CID_TEST_PATTERN``.
+
+Each control method returns ``true`` if the driver accepted the value.
+
+ESPVideoBufferClass
+*******************
+
+RAII wrapper for one dequeued frame.
+Inherits ``ESPVideoSolutionClass`` and ``ESPVideoFormatClass`` so each frame carries its own format and resolution metadata.
```
pedrominatel: Inherits ESPVideoSolutionClass and ESPVideoFormatClass, so each frame carries its own...
Missing comma.

docs/en/api/esp_video.rst
```diff
@@ -0,0 +1,965 @@
...
+
+Move constructor and move assignment operator
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. code-block:: cpp
+
+    ESPVideoBufferClass(ESPVideoBufferClass &&other) noexcept;
+    ESPVideoBufferClass &operator=(ESPVideoBufferClass &&other) noexcept;
+
+Transfer buffer ownership to another instance.
```
pedrominatel: Transfer -> Transfers

pedrominatel (approved):
On the docs side, just a few nitpicking, otherwise, looks good to me.

SuGlider (commented):
@donghengqaz - It could be added to docs/en/api/libraries.rst into the feature table

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,984 @@
...
+    req.memory = memory_type_;
+    ioctl(fd, VIDIOC_REQBUFS, &req);
+  }
+}
+
+bool ESPVideoCaptureDevClass::requestBuffer(size_t buf_count) {
+  if (fd < 0) {
+    return false;
+  }
+
```
SuGlider, line 702:
```suggestion
if (buf_count == 0) {
    log_w("requested zero buffers");
    return false;
}

```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,984 @@
...
+  struct v4l2_requestbuffers req = {};
+  memory_type_ = V4L2_MEMORY_MMAP;
+  req.count = buf_count;
+  req.type = buffer_type_;
+  req.memory = memory_type_;
+  if (ioctl(fd, VIDIOC_REQBUFS, &req) != 0) {
+    log_e("failed to require buffer");
+    return false;
+  }
+
```
SuGlider, line 719:
```suggestion
if (req.count == 0) {
    log_e("driver granted zero buffers");  // failed to allocate the buffers
    return false;
}

// ===>>>> Pick one possible behaviour option:

// Option1
buf_count = req.count;  // in case it grants less tan the requested number of buffers

// Option 2 - fail and log it
if (req.count != buf_count) {
    log_e("driver granted less than requested (%u) buffers", buf_count);
    return false;
}

```

SuGlider: I understand that it may not allocate all the requested buffers, therefore it shall be treated.
Above I have suggested 2 possible options to treat the exception.

SuGlider (changes requested):

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,984 @@
...
+    }
+
+    if (buf_count_ > 0) {
+      if (!requestBuffer(buf_count_)) {
+        log_e("failed to request buffer");
+        releaseBuffers();
+        return false;
+      }
+    }
+  }
```
lucasssvaz, line 865: `setFormat()` can release and re-request buffers without stopping the stream first. If called after `startCapture()`, behavior is undefined. Either document “must call before startCapture()” (docs partly say this) and enforce it in code, or stop/restart capture inside `setFormat()`.

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,984 @@
...
+  buf.memory = memory_type_;
+  if (ioctl(fd, VIDIOC_DQBUF, &buf) != 0) {
+    log_e("failed to receive video frame");
+    return ESPVideoBufferClass();
+  }
+
+  if (buf.index >= buf_ptr_.size()) {
+    log_e("invalid buffer index %u", buf.index);
+    return ESPVideoBufferClass();
+  }
```
lucasssvaz, line 814: After a successful `VIDIOC_DQBUF`, an out-of-range `buf.index` returns an empty buffer without `VIDIOC_QBUF`. Either re-queue the buffer before returning, or treat this as a fatal error and call `stopCapture()`.

lucasssvaz (changes requested):

libraries/ESP_Video/src/ESP_Video.cpp
donghengqaz: The buffer is dynamically allocated. This error log is used to inform users that there is insufficient memory to allocate the video buffer.

libraries/ESP_Video/src/ESP_Video.cpp
donghengqaz: Yes, the `setFormat` function should be called before starting stream. I will add check `capture_started_` in this function.

libraries/ESP_Video/src/ESP_Video.cpp
donghengqaz: Yes, this may be a fatal error, so re-queuing isn't a good option, and applications should call `stopCapture`.

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,994 @@
...
+}
+
+ESPVideoDVPPinsConfigClass &ESPVideoDVPPinsConfigClass::operator=(const ESPVideoDVPPinsConfigClass &config) {
+  if (this != &config) {
+    dvp_pin = config.dvp_pin;
+  }
+  return *this;
+}
+
+bool ESPVideoDVPPinsConfigClass::setPinsBus() {
```
P-R-O-C-H-Y: Pass the ESPVideoClass instance as bus handle to Permian pins, so its needed as parameter here.
```suggestion
bool ESPVideoDVPPinsConfigClass::setPinsBus(void *bus) {
```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,994 @@
...
+  if (!clearPinsBus()) {
+    log_e("failed to clear pins bus");
+    return false;
+  }
+
+  if (!perimanSetPinBus(dvp_pin.xclk_io, ESP32_BUS_TYPE_LCDCAM_CAM_XCLK, (void *)(dvp_pin.xclk_io + 1), -1, -1)) {
```
P-R-O-C-H-Y: Change Perlman handle to bus - please apply to all pins.
```suggestion
  if (!perimanSetPinBus(dvp_pin.xclk_io, ESP32_BUS_TYPE_LCDCAM_CAM_XCLK, bus, -1, -1)) {
```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,994 @@
...
+  if ((flags_ & target_flags) == target_flags) {
+    return true;
+  }
+
+  esp_video_init_dvp_config_t dvp_config = config.getDvpInitConfig();
+  esp_video_init_config_t cam_config = {};
+  cam_config.dvp = &dvp_config;
+
+  dvp_config_ = std::make_unique<ESPVideoDVPConfigClass>(config);
+  if (!dvp_config_->acquirePins()) {
```
P-R-O-C-H-Y: acquirePins() should take (void *)this here so LCDCAM pins use ESPVideoClass* as the periman bus handle; register perimanSetBusDeinit() for all ESP32_BUS_TYPE_LCDCAM_CAM_* types before this call (see esp32-hal-hosted.c).

```suggestion
  if (!dvp_config_->acquirePins((void *)this)) {
```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,994 @@
...
+    flags_ = 0;
+  }
+
+#if CONFIG_ESP_VIDEO_ENABLE_MIPI_CSI_VIDEO_DEVICE
+  csi_config_.reset();
+#endif
+#if CONFIG_ESP_VIDEO_ENABLE_DVP_VIDEO_DEVICE
+  dvp_config_.reset();
+#endif
+}
```
P-R-O-C-H-Y, line 942: With a blocking `videoDetachBus()`, `releasePins()` while `flags_ != 0` will fail because `perimanClearPinBus()` calls the detach callback first.
Correct order:
Deinit driver → clear `flags_` → then `releasePins()`, like `hostedDeinit()`.

the videoDetachBus is missing, should look like this:
```cpp
static bool videoDetachBus(void *bus_pointer) {
  ESPVideoClass *video = static_cast<ESPVideoClass *>(bus_pointer);
  if (video == nullptr || video->isActive()) {
    log_e("ESP-Video: call video.end() before reusing camera pins");
    return false;
  }
  return true;
}
```

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,994 @@
...
+    log_e("failed to clear pins bus");
+    return false;
+  }
+
+  if (reset_pin != GPIO_NUM_NC) {
+    if (!perimanSetPinBus(reset_pin, ESP32_BUS_TYPE_GPIO, (void *)this, -1, -1)) {
+      log_e("failed to set reset pin");
+      goto cleanup;
+    }
+  }
```
P-R-O-C-H-Y, line 241: Permian bus extra type can be added here for both reset and pwdn pins.
```suggestion
  if (reset_pin != GPIO_NUM_NC) {
    if (!perimanSetPinBus(reset_pin, ESP32_BUS_TYPE_GPIO, (void *)this, -1, -1)) {
      log_e("failed to set reset pin");
      goto cleanup;
    }
    perimanSetPinBusExtraType(reset_pin, "CSI_RESET");
  }
```

P-R-O-C-H-Y (changes requested):
I have added some comments to the Peripheral manager implementation PTAL.

donghengqaz:
> I have added some comments to the Peripheral manager implementation PTAL.

@P-R-O-C-H-Y  It looks like the Peripheral manager used to check the peripheral driver state before deinitializing the peripheral driver.

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    flags_ = 0;
+
+#if CONFIG_ESP_VIDEO_ENABLE_MIPI_CSI_VIDEO_DEVICE
+    if (flags_ & ESP_VIDEO_INIT_FLAGS_MIPI_CSI && csi_config_) {
+      csi_config_->releasePins();
+    }
+#endif
+#if CONFIG_ESP_VIDEO_ENABLE_DVP_VIDEO_DEVICE
+    if (flags_ & ESP_VIDEO_INIT_FLAGS_DVP && dvp_config_) {
+      dvp_config_->releasePins();
```
P-R-O-C-H-Y: `flags_` is set to 0 on L917, then checked on L920/L925 — those conditions are always false, so releasePins() never runs. Its best when pins are cleared on `end()`.

P-R-O-C-H-Y:
@donghengqaz I have pushed small fix for the peripheral manager in VideoCam. Everything should be good now.

donghengqaz:
> @donghengqaz I have pushed small fix for the peripheral manager in VideoCam. Everything should be good now.

@P-R-O-C-H-Y thanks, I see it.

P-R-O-C-H-Y (approved):
LGTM, but there is still an open comment from @lucasssvaz not solved.

libraries/ESP_Video/src/ESP_Video.cpp
lucasssvaz: Shouldn't it call `stopCapture` before returning rather than leaving it to the application ?

SuGlider (approved):

libraries/ESP_Video/src/ESP_Video.cpp
donghengqaz: The right buffer index should be less than the size(), so if the value is larger or equal, some fatal error triggers, so I think re-queue also will fail, and calling `stopCapture` automatically is not a good choice, maybe users should call `stopCapture` and free all resources.

lucasssvaz: Why isn't it a good idea to call `stopCapture` before returning ? Is there any case where a deadlock might happen ?

donghengqaz: If the `index >= buf_ptr_.size()`, as you mentioned, calling the `stopCapture`, if calling `stopCapture ` also fails, how to do?

lucasssvaz: If it fails there will be an error message to let the user know. Either way if it fails inside the function it will probably also fail when the user calls it in the main application.

donghengqaz: OK, this is a solution.

lucasssvaz (approved):

P-R-O-C-H-Y:
@donghengqaz you have reverted my changes to Permiman here: https://github.com/espressif/arduino-esp32/compare/2cee14426816e898a8a005786d52c2737d681ad1..a2a6f96b478786b27ac330868ab74c32ac2d255e
Please revert those changes. Thanks

libraries/ESP_Video/src/ESP_Video.cpp
```diff
@@ -0,0 +1,1016 @@
...
+    return ESPVideoBufferClass();
+  }
+
+  if (buf.index >= buf_ptr_.size()) {
+    log_e("invalid buffer index %u", buf.index);
+    if (stopCapture()) {
+      log_w("stopped capture");
+    } else {
+      log_e("failed to stop capture");
+    }
```
lucasssvaz, line 812: Currently `stopCapture` is `void stopCapture()`. We need to make it `bool stopCapture()` to be able to check for errors.

lucasssvaz (changes requested):

donghengqaz:
> @donghengqaz you have reverted my changes to Permiman here: https://github.com/espressif/arduino-esp32/compare/2cee14426816e898a8a005786d52c2737d681ad1..a2a6f96b478786b27ac330868ab74c32ac2d255e Please revert those changes. Thanks

OK, I will do this.

lucasssvaz (approved):

PR #12691, opened 2026-06-18
fix(sta): Only disconnect if actually started

lucasssvaz:
## Description of Change

This pull request introduces a small but important change to the `STAClass::disconnect` method in `STA.cpp`. Now, the method will immediately return `true` if the WiFi station has not been started, preventing crashes. 

* Added a check in `STAClass::disconnect` to return early if the WiFi station is not started, improving method robustness.

## Test Scenarios

Tested locally


me-no-dev:
There is a test if STA is started further down in the code. It makes no sense to be there anymore

lucasssvaz:
@me-no-dev Yeah, I missed that. This is some change I did when adding the wifi tests to check some issue but it is indeed useless. I though it was some fix I forgot to push XD Will close it now

Issue #11810, opened 2025-09-09
ESP-IDF v5.5.x / Arduino esp32 v3.3.x build error


r2d2r01:
### Board

LOLIN D32 ESP32 WROOM

### Device Description

LOLIN D32 V1.0.0 (WEMOS) WiFi,BT ESP-32 ESP-WROOM-32 4MB FLASH 

### Hardware Configuration

Nothing connected

### Version

latest stable Release (if not listed below)

### Type

Task

### IDE Name

Eclipse / VS / Notepad++ / Shell

### Operating System

W11pro / Dell Xps13 / 

### Flash frequency

40

### PSRAM enabled

yes

### Upload speed

460800 <-> 921600

### Description

Hi i install (try) to deploy a new platform esp-idf v5 / arduino v3 to restart a project which turn on esp-idf v4 / arduino v2. I work only on shell (cmd) to stabilize the environent before switch project to eclipse / vs. I experience some build error with some version, so thank's for helping.

Target is set to "esp32"

1. Matrix tests acted
================
ESP-IDF v5.5.1 / Arduino 3.3.0-alpha1 or 3.2.X or 3.1.X
ESP-IDF v5.5   / Arduino 3.3.0-alpha1 or 3.2.X or 3.1.X
ESP-IDF v5.5.-rc1 / Arduino 3.3.0-alpha1 or 3.2.X or 3.1.X

2. Results
==========
After executing idf.py fullclean + build

Test 5.5.1 / origin/release/v3.3.0-alpha1 
----------------------------------------------------------------
C:/_DEV/eclipse-workspace_ESP-IDE/hello_world_02/components/arduino/cores/esp32/esp32-hal-i2c-slave.c:340:3: error: implicit declaration of function 'i2c_ll_slave_init'; did you mean 'i2c_ll_slave_rw'? [-Wimplicit-function-declaration]
  340 |   i2c_ll_slave_init(i2c->dev);
      |   ^~~~~~~~~~~~~~~~~
      |   i2c_ll_slave_rw

Test 5.5.1 / origin/release/v3.2.x
------------------------------------------------------
ERROR: Because project depends on idf (>=5.3,<5.5)

Test 5.5.1 / origin/release/v3.1.x
----------------------------------------------
ERROR: Because project depends on idf (>=5.3,<5.4)

Test 5.5 / origin/release/v3.3.0-alpha1 
--------------------------------------------------------------
set ARDUINO_SKIP_IDF_VERSION_CHECK=0
[...]

PR #12708, opened 2026-06-24
ETH: include driver/spi_master.h in ETH.h (fix spi_host_device_t undeclared in from-source builds)

thehellmaker:
## Description

`libraries/Ethernet/src/ETH.h` uses `spi_host_device_t` in the **public** method signatures `ETHClass::begin(...)` and `ETHClass::beginSPI(...)`, but it does not include the header that declares that type (`driver/spi_master.h`, which provides `hal/spi_types.h`).

`ETH.cpp` includes `ETH.h` (line ~24) **before** `driver/spi_master.h` (line ~32). So when the Ethernet library is **compiled from source**, `ETH.h` is parsed while `spi_host_device_t` is still undeclared. The parameter then mis-parses, yielding:

```
ETH.h:184: error: 'spi_host_device_t' has not been declared
ETH.cpp:578: error: no declaration matches 'bool ETHClass::beginSPI(..., spi_host_device_t, uint8_t)'
ETH.cpp:960: error: no declaration matches 'bool ETHClass::begin(..., spi_host_device_t, int, int, int, uint8_t)'
```

Pre-built-lib builds don't hit this because `ETH.cpp` isn't recompiled, so the header/source pair is never compiled together in a context that lacks the type. It reproduces reliably with a **from-source** Arduino build (e.g. PlatformIO / pioarduino with `custom_sdkconfig`) on ESP-IDF 5.5.x / arduino-esp32 3.3.x.

## Fix

Include `driver/spi_master.h` in `ETH.h` so the type it uses in its own public API is always declared (include-what-you-use). One line, no behavior change.

## Testing

Built `esp32dev` (ESP32) from source via pioarduino with a `custom_sdkconfig` (forces the framework/library to recompile). Before: fails compiling `ETH.cpp`. After: `ETH.cpp` compiles and the full image links/builds successfully.

## Related

- pioarduino/platform-espressif32#504
- espressif/arduino-esp32#11810

### DIFF
--- a/libraries/Ethernet/src/ETH.h
+++ b/libraries/Ethernet/src/ETH.h
@@ -76,6 +76,7 @@
 #include "esp_system.h"
 #include "esp_eth.h"
 #include "esp_netif.h"
+#include "driver/spi_master.h"  // for spi_host_device_t (used in begin()/beginSPI())
 
 #if CONFIG_ETH_USE_ESP32_EMAC
 #if defined __has_include && __has_include("esp_eth_phy_lan867x.h")


libraries/Ethernet/src/ETH.h
```diff
@@ -76,6 +76,7 @@
 #include "esp_system.h"
 #include "esp_eth.h"
 #include "esp_netif.h"
+#include "driver/spi_master.h"  // for spi_host_device_t (used in begin()/beginSPI())
```
me-no-dev, line 79: can you try including `hal/spi_types.h` instead?

me-no-dev:
You need to sign CLA before we process the PR

me-no-dev:
closing due to no response

PR #12723, opened 2026-06-29
fix(wire): Clamp requestFrom() larger than the configured buffer

### LINKED ISSUE #11180: I2C Wire causes getMaxAllocHeap exception 'Interrupt wdt timeout' when reading more than 132 bytes
### Board

ESP32 S3 ESP-4827S043

### Device Description

Board HMI 4.3" with ESP32 S3 revision 0.1. I2C (pin 19 SDA, 20 SCL) connected to touch GT911.

### Hardware Configuration

No.

### Version

latest development Release Candidate (RC-X)

### IDE Name

Arduino IDE

### Operating System

Ubuntu Desktop 22.04

### Flash frequency

80MHz

### PSRAM enabled

yes

### Upload speed

1152000

### Description

The test based on WireMaster example. If reading bytes more than 132 bytes, `Wire.requestFrom(I2C_DEV_ADDR, 133)`, then exception occurs if calling `ESP.getMaxAllocHeap()` after the call `requestFrom` anywhere in the program.

Arduino ESP32: 3.2.0-RC2, IDF 5.4

### Sketch

```cpp
#include "Wire.h"

#define I2C_DEV_ADDR 0x5D

uint32_t i = 0;

void setup() {
  Serial.begin(115200);
  Serial.setDebugOutput(true);
  Wire.begin(19, 20);
}

void loop() {
  delay(5000);

  //Write message to the slave
  Wire.beginTransmission(I2C_DEV_ADDR);
  Wire.write(0x8047 >> 8);
  Wire.write(0x8047 & 0x00ff);
  uint8_t error = Wire.endTransmission(true);
  Serial.printf("endTransmission: %u\n", error);

  //Read 16 bytes from the slave
  // 132 bytes is OK. 133 failed
  uint8_t bytesReceived = Wire.requestFrom(I2C_DEV_ADDR, 133);
  Serial.printf("requestFrom: %u\n", bytesReceived);
  if ((bool)bytesReceived) {  //If received more than zero bytes
    uint8_t temp[bytesReceived];
    Wire.readBytes(temp, bytesReceived);
    log_print_buf(temp, bytesReceived);
  }

  Serial.printf("Max alloc heap: %u\n", ESP.getMaxAllocHeap());
}
```

### Debug Message

```plain
Core  0 register dump:
PC      : 0x4202aad2  PS      : 0x00060833  A0      : 0x8202a9e7  A1      : 0x3fca4380  
A2      : 0xeb8465ec  A3      : 0x4202aa24  A4      : 0x3fca43e4  A5      : 0x3fca4400  
A6      : 0xeb8465f4  A7      : 0xeb8465f0  A8      : 0xabba1234  A9      : 0x3fca4360  
A10     : 0x00000001  A11     : 0xabba1234  A12     : 0x00000001  A13     : 0x3fca43e4  
A14     : 0x3fca4740  A15     : 0x0000cdcd  SAR     : 0x00000004  EXCCAUSE: 0x0000001c  
EXCVADDR: 0xeb8465f0  LBEG    : 0x400570e8  LEND    : 0x400570f3  LCOUNT  : 0x00000000  


Backtrace: 0x4202aacf:0x3fca4380 0x4202a9e4:0x3fca43a0 0x4202aa69:0x3fca43c0 0x4200bfef:0x3fca43e0 0x4200c03d:0x3fca4420 0x42002b6d:0x3fca4460 0x42001fe5:0x3fca4480 0x42007270:0x3fca44b0 0x4037d272:0x3fca44d0
```

### Other Steps to Reproduce

_No response_

ACEFGI:
## Description

`TwoWire::requestFrom()` passed the caller's requested length directly to the I2C driver without validating it against the allocated RX buffer. Requesting more bytes than `setBufferSize()` reserved caused the driver to write past the end of `rxBuffer`, corrupting the heap.

This adds a `size > bufferSize` guard that **clamps** the request to the configured buffer size before any transfer is started, matching the Arduino Wire API convention (AVR/Renesas/ESP8266 all clamp). The call then reads up to the buffer size and returns the actual number of bytes read, so callers that check the returned count behave the same across cores and can issue a follow-up request for the remainder.

The pre-existing `NULL` buffer check is also moved inside the lock, so a `setBufferSize()` running on another task cannot free the buffers between the check and the transfer.

The 32-byte minimum enforced by `setBufferSize()` (the I2C hardware FIFO length) is unchanged; this only clamps requests that exceed whatever buffer size is currently configured.

The guard logs both values at error level so the caller can see the mismatch directly while still getting their (partial) data:

```
Requested %zu bytes but the buffer holds only %zu
Set the buffer size with TwoWire::setBufferSize() before begin().
Set the buffer size with Wire.setBufferSize() before begin().
```

Closes #11180

## Tests

Added a `request_from_undersized_buffer` case to the existing `tests/validation/i2c_master` suite. It shrinks the buffer to its 32-byte minimum, then asserts that an oversized `requestFrom()` (33 bytes) is clamped: it returns `32` and leaves `available() == 32`. Verified locally against the repo `pre-commit` hooks (clang-format, codespell, trailing-whitespace all pass); the suite itself runs on Wokwi in CI.

## Authorship

In the interest of fair attribution:

- The `Wire.cpp` fix is my own work (Zac Anderson).
- The regression test was authored by Claude Opus.
- Both were reviewed by me and with AI assistance.

## Checklist

- [x] Change is scoped to a single logical fix (guard + its regression test)
- [x] Written in English, commented
- [x] `pre-commit` formatting/style hooks pass
- [ ] Runtime suite executed on hardware/Wokwi (left to CI)

### DIFF
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -510,7 +510,8 @@ size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop) {
   }
   if (size > bufferSize) {
     log_e("Requested %zu bytes but the buffer holds only %zu", size, bufferSize);
-    log_e("Either use TwoWire::setBufferSize() or Wire.setBufferSize()");
+    log_e("Set the buffer size with TwoWire::setBufferSize() before begin().");
+    log_e("Set the buffer size with Wire.setBufferSize() before begin().");
     size = bufferSize;  // limit to the buffer size
   }
   if (nonStop) {
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -486,10 +486,6 @@ size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop) {
     return 0;
   }
 #endif /* SOC_I2C_SUPPORT_SLAVE */
-  if (rxBuffer == NULL || txBuffer == NULL) {
-    log_e("NULL buffer pointer");
-    return 0;
-  }
   esp_err_t err = ESP_OK;
 #if !CONFIG_DISABLE_HAL_LOCKS
   TaskHandle_t task = xTaskGetCurrentTaskHandle();
@@ -502,6 +498,26 @@ size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop) {
     currentTaskHandle = task;
   }
 #endif
+  // check after the lock is acquired to avoid asynchronous buffer changes
+  if (rxBuffer == NULL || txBuffer == NULL) {
+    log_e("NULL buffer pointer");
+#if !CONFIG_DISABLE_HAL_LOCKS
+    currentTaskHandle = NULL;
+    //release lock
+    xSemaphoreGive(lock);
+#endif
+    return 0;
+  }
+  if (size > bufferSize) {
+    log_e("Requested %zu bytes but the buffer holds only %zu", size, bufferSize);
+    log_e("Set the buffer size with TwoWire.setBufferSize() before begin().");
+#if !CONFIG_DISABLE_HAL_LOCKS
+    currentTaskHandle = NULL;
+    //release lock
+    xSemaphoreGive(lock);
+#endif
+    return 0;
+  }
   if (nonStop) {
     if (address != txAddress) {
       log_e("Unfinished Repeated Start transaction! Expected address do not match! %u != %u", address, txAddress);
--- a/libraries/Wire/src/Wire.cpp
+++ b/libraries/Wire/src/Wire.cpp
@@ -510,13 +510,8 @@ size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop) {
   }
   if (size > bufferSize) {
     log_e("Requested %zu bytes but the buffer holds only %zu", size, bufferSize);
-    log_e("Set the buffer size with TwoWire.setBufferSize() before begin().");
-#if !CONFIG_DISABLE_HAL_LOCKS
-    currentTaskHandle = NULL;
-    //release lock
-    xSemaphoreGive(lock);
-#endif
-    return 0;
+    log_e("Either use TwoWire::setBufferSize() or Wire.setBufferSize()");
+    size = bufferSize;  // limit to the buffer size
   }
   if (nonStop) {
     if (address != txAddress) {
--- a/tests/validation/i2c_master/i2c_master.ino
+++ b/tests/validation/i2c_master/i2c_master.ino
@@ -294,17 +294,16 @@ void scan_bus() {
 void request_from_undersized_buffer() {
   // The RX/TX buffer cannot be shrunk below 32 bytes (the I2C hardware FIFO
   // length): setBufferSize() rejects anything smaller and leaves the size as-is.
-  TEST_ASSERT_EQUAL(0, Wire.setBufferSize(31));
   TEST_ASSERT_EQUAL(32, Wire.setBufferSize(32));
 
   // Start from a known-empty RX state so available() is deterministic.
   Wire.flush();
 
-  // Requesting even one byte more than the buffer holds
-  // must be rejected up front.
-  // returns 0 without touching the bus or the buffer.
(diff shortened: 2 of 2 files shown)


me-no-dev:
Clamping is what Arduino API does. I suggest we follow the same

ACEFGI:
It's tough, see this table:
| Core | Guard in `requestFrom` | Verdict | Buffer | Resizable API |
|------|------------------------|---------|--------|---------------|
| Arduino **AVR** | `if (quantity > BUFFER_LENGTH) { quantity = BUFFER_LENGTH; }` | **CLAMP** | Fixed, 32 B | No |
| Arduino **SAMD** | *(none — 256 B ring buffer wraps on overflow)* | **NEITHER** | Fixed, 256 B ring | No |
| Arduino **Renesas** (UNO R4) | `if (quantity > I2C_BUFFER_LENGTH) { quantity = I2C_BUFFER_LENGTH; }` | **CLAMP** | Fixed, 255 B | No |
| **ESP8266** | `if (size > I2C_BUFFER_LENGTH) { size = I2C_BUFFER_LENGTH; }` | **CLAMP** | Fixed static array | No |
| **STM32duino** | *(none — `allocateRxBuffer(len)` reallocs to fit)* | **GROW** | Dynamic (auto-grow) | No explicit API (auto-grow only) |
| **arduino-pico** (RP2040) | `if (... \|\| (quantity > _buffSize)) { return 0; }` | **REJECT** | Dynamic, default 256 B | **Yes — `setBufferSize()`, `WIRE_HAS_BUFFER_SIZE`** |
| **ESP32** (arduino-esp32) | *(the bug — no guard; fix under review)* | — | Dynamic, default 128 B | **Yes — `setBufferSize()`, `WIRE_HAS_BUFFER_SIZE`** |

It looks to me like this:
- Fixed-buffer cores: (AVR, Renesas, ESP8266). The caller cannot enlarge the
  buffer, so a best-effort short read is the only non-overflow option available.
- Resizable-buffer cores do not clamp: arduino-pico rejects & STM32duino grows.
  When the buffer size is something the caller deliberately set (via `setBufferSize()`),
  requesting more than that is a caller bug, and silently returning partial data hides it.
  
  I'm open to just saying "clamp is more common" but my rationale going into this was that the best thing for a caller is to know they made a mistake ASAP.

libraries/Wire/src/Wire.cpp
```diff
@@ -502,6 +498,26 @@ size_t TwoWire::requestFrom(uint8_t address, size_t size, bool sendStop) {
...
+#if !CONFIG_DISABLE_HAL_LOCKS
+    currentTaskHandle = NULL;
+    //release lock
+    xSemaphoreGive(lock);
+#endif
+    return 0;
+  }
+  if (size > bufferSize) {
+    log_e("Requested %zu bytes but the buffer holds only %zu", size, bufferSize);
+    log_e("Set the buffer size with TwoWire.setBufferSize() before begin().");
```
me-no-dev: Either use `TwoWire:setBufferSize()` or `Wire.setBufferSize()`

ACEFGI: Like this?
`log_e("Either use TwoWire::setBufferSize() or Wire.setBufferSize()");`

me-no-dev:
People should check the amount of returned bytes either way. Rationale to follow official Arduino API is to be compatible with as many cores as possible and especially with the official ones. Going that way also allows for a follow-up request for the rest of the bytes (at least that is how most I2C slaves behave). While we can make an exception in this case, I doubt it's really needed. Error can stay as error and still be clamped

libraries/Wire/src/Wire.cpp
me-no-dev: no no.
```cpp
log_e("Set the buffer size with TwoWire:setBufferSize() before begin().");
log_e("Set the buffer size with Wire.setBufferSize() before begin().");
```

ACEFGI:
Changed to clamp - tests updated - error messages updated. Out for lunch for 30 mins now. Will get this sorted after I'm back if further comments.

libraries/Wire/src/Wire.cpp
ACEFGI: Now matches as requested.

me-no-dev (approved):

P-R-O-C-H-Y (approved):

Issue #12163, opened 2025-12-21
Support for Matter 1.5 device types


lboue:
### Related area

Matter

### Hardware specification

ESP32-C6

### Is your feature request related to a problem?

I would like to know from which version of arduino-esp32 the new Matter 1.5 device types will be supported.

### Describe the solution you'd like

Support for new 1.5 device types:
- Closures
- Soil sensors
- Chime
- Cameras
- Doorbell
- Intercom
- Irrigation System
- Electrical Meter

### Additional context

[Matter 1.5 Introduces Cameras, Closures, and Enhanced Energy Management Capabilities](https://csa-iot.org/newsroom/matter-1-5-introduces-cameras-closures-and-enhanced-energy-management-capabilities/)

### I have checked existing list of Feature requests and the Contribution Guide

- [x] I confirm I have checked existing list of Feature requests and Contribution Guide.

lbernstone:
esp-matter has it's own [repository](https://github.com/espressif/esp-matter) with that information. arduino-esp32 is downstream of them, so about 3 months after they complete the API.

SuGlider:
@lboue - ESP32 Arduino Matter Library depends on ESP-Matter version that is available in the ESP Registry (https://components.espressif.com/components/espressif/esp_matter/)

At this time the latest version available is 1.4.2

lboue:
Now Matter 1.5 is supported and there is a new release:
https://github.com/espressif/esp-matter/tree/release/v1.5

A new version of the component should be available soon.

lboue:
Version 1.5 of the component has still not been released, even though it appears in the table here:
https://github.com/espressif/esp-matter

I don't understand why.

SuGlider:
It is currently being updated.
We are working to get Version 1.5 within Arduino Core 3.3.11 and 4.x

https://github.com/espressif/arduino-esp32/pull/12730

lboue:
Thanks

lboue:
Status update / expanded gap list, cross-referencing the other open Matter device-type issues:

**Done:**
- Soil Sensor (cluster `0x0430`, device type `0x0045`) — implemented against the currently vendored `esp_matter` 1.5.x (no wrapper exists in esp_matter for it yet, so the endpoint is built from the low-level API + a hand-registered `SoilMeasurementCluster`, the same technique `boolean_state_integration.cpp` uses upstream). See branch `feat/matter-soil-sensor` / #12854.

**In progress:**
[...]

PR #12730, opened 2026-07-01
feat(matter): update to ESP-Matter 1.5

SuGlider:
## Description of Change
ESP Matter Component is now on version 1.5.
This PR changes necessary settings and Arduino Matter Library code and examples in order to update it to ESP Matter 1.5
It also patches ESP Matter in order to be built using c++ 20.

## Test Scenarios
CI only

## Related links
Related to https://github.com/espressif/arduino-esp32/issues/12163


SuGlider:
@me-no-dev | @P-R-O-C-H-Y | @lucasssvaz 

This PR depends on a Lib Builder change and will only compile after a new branch is created with the ESP Matter v1.5 Component

lucasssvaz:
Shouldn't this have waited for the libs first ? Now the entire CI will fail every time until the new libs are merged.

me-no-dev:
no, because the libs are failing without it. I'm regenerating them now and all will be good by the time you wake up

libraries/Matter/src/matter_closure_patch.h
```diff
@@ -0,0 +1,33 @@
...
+// ClosureControl structs only define operator==(const BaseType&), which is not
+// sufficient. This header adds free-function operators via ADL without modifying
+// managed_components.
+//
+// Included from Matter.h for user code that compares ClosureControl Nullable types.
+// Also injected into espressif__esp_matter via CMake -include when Matter is enabled.
+
+#ifdef __cplusplus
+
+#include "app/clusters/closure-control-server/closure-control-cluster-objects.h"
```
Jason2866, line 13: @SuGlider The include breaks compile Arduino as an component of IDF in pioarduino.
The file is not found. https://github.com/pioarduino/platform-espressif32/actions/runs/33094899638/job/98597121219

SuGlider: @Jason2866 - Can you verify if pioarduino has updated the esp-matter component to version 1.5?

Jason2866: @SuGlider Yes, i have updated the matter component to v1.5. I fear, since the file is completely new, the `#ifdef __cplusplus` is not working here with pioarduino.

Is the include path correct at all?

SuGlider: Have you been able to solve it?

Your job break is this include from Matter.h:

Matter.h Ln 46–46
`#include "matter_closure_patch.h"`
That header then pulls a CHIP-internal file:

matter_closure_patch.h Ln 13–13
`#include "app/clusters/closure-control-server/closure-control-cluster-objects.h"`

It is there just to support c++ 20.
You can remove it and then use C++ 17.

Jason2866:
@SuGlider pioarduino uses original releases / sources from espressif (loaded from github releases). So no way to remove this file.

Jason2866:
Regarding to https://stackoverflow.com/questions/2324658/how-to-determine-the-version-of-the-c-standard-used-by-the-compiler the check in the file should be done for C++ 20.
Currently the check fires for all versions, even when not needed for C++ 17

PR #12752, opened 2026-07-10
fix(esp32p4): Allow selecting the USB device port

tanakamasayuki:
## Description of Change

This PR allows selecting the USB device controller on ESP32-P4 at build time using `ARDUINO_USB_DEVICE_PORT`.

The default remains port 1 to preserve the existing high-speed USB behavior.

- `ARDUINO_USB_DEVICE_PORT=0`
  - Uses the full-speed USB controller
  - Configures the USB PHY and TinyUSB for full speed
  - Uses a maximum endpoint size of 64 bytes
  - Uses the endpoint limits of the full-speed controller

- `ARDUINO_USB_DEVICE_PORT=1`
  - Preserves the existing high-speed USB behavior
  - Uses the existing TinyUSB endpoint size
  - Uses the endpoint limits of the high-speed controller

Values other than `0` or `1` produce a compile-time error.

A separate `ARDUINO_USB_ENDPOINT_SIZE` macro is used for descriptors generated by the Arduino USB implementations. This keeps TinyUSB's compile-time `CFG_TUD_ENDPOINT_SIZE` configuration unchanged across translation units.

Existing ESP32-P4 builds are unaffected unless `ARDUINO_USB_DEVICE_PORT` is explicitly defined.

Example using `build_opt.h`:

```text
-DARDUINO_USB_DEVICE_PORT=0
```

No board menu option is added by this PR to keep the change limited to the USB implementation.

## Test Scenarios

Tested using:

- ESP32-P4 development board
- `ESP32P4 Dev Module` board configuration
- Arduino-ESP32 built from this PR branch
- ESP32-P4 Arduino libraries/toolchain version 3.3.10

The following configurations were compiled:

- Default configuration (`ARDUINO_USB_DEVICE_PORT=1`)
- Full-speed controller (`ARDUINO_USB_DEVICE_PORT=0`)
- Invalid controller value (`ARDUINO_USB_DEVICE_PORT=2`), confirming that compilation fails as expected

The affected CDC, MSC, HID, MIDI, Vendor, and TinyUSB HAL source files were compiled for both port 0 and port 1.

Hardware testing was performed with:

```text
-DARDUINO_USB_DEVICE_PORT=0
```

A USB HID keyboard sketch enumerated and operated successfully through the ESP32-P4 full-speed USB device port.

## Related links

N/A

### DIFF
--- a/cores/esp32/USBCDC.cpp
+++ b/cores/esp32/USBCDC.cpp
@@ -30,7 +30,7 @@ USBCDC *devices[CFG_TUD_CDC];
 static uint16_t load_cdc_descriptor(uint8_t *dst, uint8_t *itf) {
   uint8_t str_index = tinyusb_add_string_descriptor("TinyUSB CDC");
   uint8_t descriptor[TUD_CDC_DESC_LEN] = {// Interface number, string index, EP notification address and size, EP data address (out, in) and size.
-                                          TUD_CDC_DESCRIPTOR(*itf, str_index, 0x85, CFG_TUD_ENDPOINT_SIZE, 0x03, 0x84, CFG_TUD_ENDPOINT_SIZE)
+                                          TUD_CDC_DESCRIPTOR(*itf, str_index, 0x85, ARDUINO_USB_ENDPOINT_SIZE, 0x03, 0x84, ARDUINO_USB_ENDPOINT_SIZE)
   };
   *itf += 2;
   memcpy(dst, descriptor, TUD_CDC_DESC_LEN);
@@ -47,7 +47,9 @@ static uint16_t load_cdc_descriptor2(uint8_t *dst, uint8_t *itf) {
   TU_VERIFY(ep_out != 0);
   uint8_t descriptor[TUD_CDC_DESC_LEN] = {
     // Interface number, string index, EP notification address and size, EP data address (out, in) and size.
-    TUD_CDC_DESCRIPTOR(*itf, str_index, (uint8_t)(0x80 | ep_ntfy), CFG_TUD_ENDPOINT_SIZE, ep_out, (uint8_t)(0x80 | ep_in), CFG_TUD_ENDPOINT_SIZE)
+    TUD_CDC_DESCRIPTOR(
+      *itf, str_index, (uint8_t)(0x80 | ep_ntfy), ARDUINO_USB_ENDPOINT_SIZE, ep_out, (uint8_t)(0x80 | ep_in), ARDUINO_USB_ENDPOINT_SIZE
+    )
   };
   *itf += 2;
   memcpy(dst, descriptor, TUD_CDC_DESC_LEN);
--- a/cores/esp32/USBMSC.cpp
+++ b/cores/esp32/USBMSC.cpp
@@ -24,7 +24,7 @@ extern "C" uint16_t tusb_msc_load_descriptor(uint8_t *dst, uint8_t *itf) {
   uint8_t ep_num = tinyusb_get_free_duplex_endpoint();
   TU_VERIFY(ep_num != 0);
   uint8_t descriptor[TUD_MSC_DESC_LEN] = {// Interface number, string index, EP Out & EP In address, EP size
-                                          TUD_MSC_DESCRIPTOR(*itf, str_index, ep_num, (uint8_t)(0x80 | ep_num), CFG_TUD_ENDPOINT_SIZE)
+                                          TUD_MSC_DESCRIPTOR(*itf, str_index, ep_num, (uint8_t)(0x80 | ep_num), ARDUINO_USB_ENDPOINT_SIZE)
   };
   *itf += 1;
   memcpy(dst, descriptor, TUD_MSC_DESC_LEN);
--- a/cores/esp32/esp32-hal-tinyusb.c
+++ b/cores/esp32/esp32-hal-tinyusb.c
@@ -136,7 +136,11 @@ esp_err_t init_usb_hal(bool external_phy) {
     .target = USB_PHY_TARGET_INT,
     .otg_mode = USB_OTG_MODE_DEVICE,
 #if CONFIG_IDF_TARGET_ESP32P4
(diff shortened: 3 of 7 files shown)


tanakamasayuki:
# Simple testing method
arduino-cli core uninstall esp32:esp32
arduino-cli core install esp32:esp32
cp -v ~/dev/arduino-esp32/cores/esp32/esp32-hal-tinyusb.* ~/.arduino15/packages/esp32/hardware/esp32/3.3.10/cores/esp32
cp -v ~/dev/arduino-esp32/cores/esp32/USB*.* ~/.arduino15/packages/esp32/hardware/esp32/3.3.10/cores/esp32
cp -v ~/dev/arduino-esp32/libraries/USB/src/USB*.* ~/.arduino15/packages/esp32/hardware/esp32/3.3.10/libraries/USB/src

# Tested example
https://github.com/espressif/arduino-esp32/tree/master/libraries/USB/examples/Keyboard

Verified that the USB Keyboard example builds and works correctly with the following changes:

Added -DARDUINO_USB_DEVICE_PORT=0 to build_opt.h
Changed the button pin to GPIO35, which is connected to the BOOT button
Swapped the USB CDC and USB OTG ports

```c
#include <Arduino.h>
#include "USB.h"
#include "USBHIDKeyboard.h"
#include "hal/usb_wrap_ll.h"
#include "soc/usb_wrap_struct.h"

USBHIDKeyboard Keyboard;

const int buttonPin = 35;       // input pin for pushbutton
int previousButtonState = HIGH; // for checking the state of a pushButton
int counter = 0;                // button push counter

void setup()
{
    usb_wrap_ll_phy_select(&USB_WRAP, 0);

    // make the pushButton pin an input:
    pinMode(buttonPin, INPUT_PULLUP);
    // initialize control over the keyboard:
    Keyboard.begin();
    USB.begin();
}

void loop()
{
    // read the pushbutton:
    int buttonState = digitalRead(buttonPin);
    // if the button state has changed,
    if ((buttonState != previousButtonState)
        // and it's currently pressed:
        && (buttonState == LOW))
    {
        // increment the button counter
        counter++;
        // type out a message
        Keyboard.print("You pressed the button ");
        Keyboard.print(counter);
        Keyboard.println(" times.");
    }
    // save the current button state for comparison next time:
    previousButtonState = buttonState;
}
```

me-no-dev:
What is the point of using the slower port? TinyUSB is compiled with all values to use the full speed port.

tanakamasayuki:
@me-no-dev
Thank you for checking.

For devices like the M5Stack TAB5, the HS port is Type-A and the FS port is Type-C.
https://docs.m5stack.com/en/core/Tab5

This is also convenient when you want to use the HS port for USB Host functionality.

However, the current use cases are limited.
For this reason, it has not been added to the board menu.

me-no-dev:
Tab5 FS port is actually connected to the CDC/JTAG and not to the FS-OTG port. That one runs on different pins. I've yet to see a board that exposes the correct pins for the FS-OTG

SuGlider:
Based on the [ESP32-P4 Datasheet](https://documentation.espressif.com/[esp32-p4_datasheet_en.pdf](https://documentation.espressif.com/esp32-p4_datasheet_en.pdf)), on page 77 (pin assignment of both USB 2.0 FS and USB_SERIAL_JTAG), it is possible to select which pair of pins is attached to USEB_SERIAL_JTAG (Hardware USB CDC/ACM) and USB1P1 (TinyUSB Full Speed  for Host and Device).

Therefore, in theory, it would be possible to assign the TAB 5 USB C port (GPIOs 24 and 25) to the USB 2.0 OTG Peripheral.
By default in boot, those pins are assigned to the USB_SERIAL_JTAG peripheral, but it can be changed by software after boot.

The other GPIOs availble for USB OTG | USB_SERIAL are the GPIOs 26 and 27. The TAB 5 uses those GPIOs to control the ES8388 DSIN + SCLK and ES7210 SCLK. Therefore, those pins are not exposed and can't be used as USB.

The TAB USB A (HS USB OTG) has dedicated pins that can't be used for any thing else.

This PR addresses specific need for using the TAB 5 GPIOs 24/25 (USB C port) for USB OTG | USB_SERIAL_JTAG.
This is not a generic solution. Other boards may use other GPIO combinations.

@tanakamasayuki, I think that these code change should be more generic in order to allow **Arduino Core API to do a pin selection for the ESP32-P4 (GPIOs 24, 25 | GPIOs 26, 27) for the USB 2.0 OTG function or for the USB_SERIAL_JTAG function**. It shall also reflect the pin function within Peripheral Manager, specially when the Debug Level reports the pins after end of `setup()`.

Arduino Core uses the `ARDUINO_USB_MODE` and `ARDUINO_USB_CDC_ON_BOOT` macros that will add Serial USB diferent drivers for those pins as well. All of it must be coordinated for a generic good PR.

SuGlider (changes requested):
@tanakamasayuki - thanks for the PR. This is good starting point for a broader and generic USB pin assignment for the ESP32-P4 (GPIOs 24/25 | GPIOs 26/27 using USB 2.0 FS OTG | USB SERIAL JTAG).

In order to move forward, please modify the PR for a broader generic scope.
The PR solves an specific ESP32-P4 USB pins use case.

me-no-dev:
In general, this should not be done at all, because TinyUSB is compiled to be used in FullSpeed mode. HighSpeed has different constraints and we do not have the option to have it both ways at the same time.

tanakamasayuki:
Hi.

The ESP32-P4-Function-EV-Board is essentially the only ESP32-P4 board that exposes all the USB interfaces.

Often, the Full-Speed ​​(FS) interface is configured as CDC, similar to TAB5.
However, it is possible to swap between CDC and OTG using `usb_wrap_ll_phy_select()`.
The swapping functionality itself is outside the scope of this PR.
That feature is also applicable to USB Host mode.

This PR focuses on the ability to select whether the USB device operates in HS (High-Speed) or FS (Full-Speed) mode.
ESP-IDF supports this selection.

PR #12754, opened 2026-07-11
fix(network): don't inspect stale errno when recv() reports EOF in connected()

### LINKED ISSUE #12751: half-closed TCP connection results in premature closure
### Board

Adafruit Feather ESP32-C6

### Device Description

Adafruit Feather ESP32-C6

### Hardware Configuration

Nothing of note.

### Version

v3.3.10

### Type

Bug

### IDE Name

Mostly arduino-cli but sometimes arduino-ide, rarely PlatformIO

### Operating System

Linux

### Flash frequency

40 MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

I ran into an issue recently while developing some code that uses `NetworkClient`.
After much debugging, I at least one issue which I will describe here.

Background:

In a normal TCP connection, one of the participants (e.g., the client) may elect to use `shutdown` to close _one-half_ of the connection. For example, it is not uncommon for HTTP clients to send their request and then issue a `shutdown(file_descriptor, SHUT_WR)` which indicates that the client will no longer issue _write_ operations to the socket but will (may) furthermore only issue _read_ operations.

Unfortunately, there appears to be a bug where clients that behave this way experience a prematurely closed connection. 

The diagnostic message printed by the arduino-esp32 code (on or around line 593 of `NetworkClient.cpp`) is:
```
[215692][D][NetworkClient.cpp:593] connected(): Disconnected: RES: 0, ERR: 128
```

128 translates to `ENOTCONN`, which isn't exactly true. The connection is normally full duplex, and the shutdown has closed one half (the write half from the client's perspective, the read half from the server's perspective).

Using `shutdown` is commonly considered part of "graceful" shutdown.

The following is - I believe - the expected behavior (assume client is not a microcontroller, e.g. Python and the server is leveraging arduino-esp32 for network connectivity).

From the client's point of view:

1. client `connect`s
2. client sends request
3. client issues a `shutdown(SHUT_WR)` to shutdown the _write_ side of the socket.  Notably, this should cause the client's TCP stack to send a `FIN` (and an `ACK`) to the server.
4. client goes into a read loop until it receives an empty read
5. client `close`es the connection

From the server's point of view:
1. server `accept`s new connection
2. server reads request until it receives 0 bytes <- this is where things go wrong
3. server is free to issue a `shutdown(SHUT_RD)` but that is less common
4. server sends response
5. server may use `shutdown(SHUT_WR)`
6. server closes socket

What's happening instead is that step (2) from the server side treats the received FIN as an error which then shuts the entire connection down.

I have used tcpdump to confirm that the client sends a `FIN, ACK` (one packet) to the server (the ESP32C6), the server response with an `ACK` and then a `FIN` (two packets), to which the client response with an `ACK`. So it _looks_ like an orderly shutdown but it's not the expected behavior.

More notes: this occurs with or without a non-zero timeout (via `client.setTimeout`).

### Sketch

```cpp
/*
 A simple sketch based on "SimpleWiFiServer".
 */

#include <WiFi.h>

const bool nonBlocking = true;

NetworkServer server(80);

void setup() {
  Serial.begin(115200);
  Serial.print("Connecting to WiFi");

  WiFi.begin(YOUR_SSID_HERE, YOUR_PASSWORD_HERE);

    delay(500); 
    Serial.print(".");
  }

  Serial.println("");
  Serial.println("WiFi connected.");
  Serial.println("IP address: ");

  server.begin();
}

enum Mode {
  READ, WRITE, WAIT_FOR_CLOSE
};

void loop() {
  static const size_t bufSz = 1024;
  static char buf[bufSz];
  NetworkClient client = server.accept();

  if (client) {
    Serial.println("New Client.");
    if (nonBlocking) {
      client.setTimeout(0); // non-blocking
    }
    Mode mode = READ;
    while (client) { // same as while(client.connected())       
      switch(mode) {
        case READ:
          // note that we aren't using .available() here, because we
          // *want* to get a read of 0 bytes.
          {
            size_t bytesRead = client.readBytes(buf, bufSz);
            Serial.print("Read ");
            Serial.print(bytesRead);
            Serial.print(" bytes.");
            Serial.println();
            if (bytesRead == 0) {
              mode = WRITE;
            }
          }
          break;
        case WRITE:
          {          
            client.write("ok");
            Serial.println("Sent payload.");
            mode = WAIT_FOR_CLOSE;
          }
          break;
        case WAIT_FOR_CLOSE:
          delay(10);
          break;                  
      } // switch  
    } // while client
    client.stop();
    Serial.println("Client Disconnected.");
  }
}
```

### Debug Message

```plain
[ 24256][D][NetworkClient.cpp:593] connected(): Disconnected: RES: 0, ERR: 128
```

### Other Steps to Reproduce

The following program should be invoked thusly:

`python3 path_to_program.py HOST_IP 80 0`
to test the _non_-shutdown path, or 
`python3 path_to_program.py HOST_IP 80 1`
to test the _shutdown_ path.

```python
#! /usr/bin/env python3
import sys
import socket
import time

NAPTIME = 5

to_send = b"GET / HTTP/1.0\r\n\r\n"
(addr, port, use_shutdown) = sys.argv[1:4]
port = int(port)
use_shutdown = bool(int(use_shutdown))

print("Connecting to %s:%d" % (addr, port))
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect( (addr, port) )

print("Will try to send",len(to_send), "bytes.")
m = memoryview(to_send)

while len(m):
  sent = sock.send(m)
  print("Sent", sent, "bytes")
  m = m[sent:]

if use_shutdown:
  # now *half-close* the socket
  print("Half-closing the socket.")
  sock.shutdown(socket.SHUT_WR)
  print("Taking a %d second nap." % (NAPTIME,))
else:
  print("NOT Half-closing the socket.")
time.sleep(NAPTIME)

print("Reading...")
while True:
  d = sock.recv(1024)
  print("Read", len(d), "bytes.")
  print(repr(d))
  if d == b"":
    break
  elif d == b"ok":
    break
```

94xhn:
### Checklist
- [x] Please provide specific title of the PR describing the change, including the component name
- [x] Please provide related links (Issue which will be closed by this Pull Request)
- [x] Please **update relevant Documentation** if applicable (no user-facing API/behavior change requiring docs)
- [x] Please check [Contributing guide](https://docs.espressif.com/projects/arduino-esp32/en/latest/contributing.html)
- [x] Please **confirm option to "Allow edits and access to secrets by maintainers"** when opening a Pull Request

-----------
## Description of Change

`NetworkClient::connected()` (`libraries/Network/src/NetworkClient.cpp`) peeks the
socket to detect disconnects:

```cpp
int res = recv(fd(), &dummy, 1, MSG_DONTWAIT | MSG_PEEK);
// recv only sets errno if res is <= 0
if (res <= 0) {
  switch (errno) {
    case ENOTCONN:
    case EPIPE:
    case ECONNRESET:
    case ECONNREFUSED:
    case ECONNABORTED:
      _connected = false;
      ...
```

`recv()` only sets `errno` when it returns **-1** (an actual error). When a stream
peer performs an *orderly* shutdown (e.g. it calls `shutdown(fd, SHUT_WR)` then
closes, as many HTTP clients do to signal "I'm done sending, keep reading my
response"), `recv()` returns **0** — a *successful* call reporting EOF, not an
error. In that case `errno` is left completely untouched and may still hold a
stale value from any unrelated earlier syscall on the same thread.

The existing code runs the `switch (errno)` for `res <= 0`, i.e. for both the
error case (`res < 0`) *and* the EOF case (`res == 0`). So whenever a half-closed
connection is polled and the leftover `errno` happens to coincide with
`ENOTCONN`/`EPIPE`/`ECONNRESET`/`ECONNREFUSED`/`ECONNABORTED`, the still-usable,
half-closed socket gets marked `_connected = false` and torn down — even though
nothing is actually wrong with it.

This exactly matches the symptom in #12751: the reporter's own debug log shows

```
[NetworkClient.cpp:593] connected(): Disconnected: RES: 0, ERR: 128
```

`RES: 0` proves `recv()` succeeded (EOF / orderly shutdown), yet the code
inspected `errno` (128 == `ENOTCONN` on the target's newlib) and incorrectly
closed the connection before the sketch could write its response.

### Fix

Only consult `errno` when `res < 0`. When `res == 0`, treat it deterministically
as "peer is done sending for now" and keep `_connected` as `true`, since a
half-closed connection can still be written to. This removes the dependency on
whatever `errno` happens to be lying around and matches the documented POSIX
`recv()`/`errno` contract that ESP-IDF's lwIP socket layer is built to mirror.

## Test Scenarios

I do not have ESP32 hardware set up for this at the moment, so I could not
verify this end-to-end on target. Instead, I wrote a standalone, hardware-free
C reproduction that exercises the *exact* statement and switch-logic from
`NetworkClient::connected()` against **real loopback TCP sockets** (not a mock)
on Linux, to validate the underlying POSIX semantics the fix relies on:

1. Open a real TCP connection over `127.0.0.1`.
2. The "client" sends a request, then calls `shutdown(SHUT_WR)` — the exact
   half-close pattern from the issue.
3. The "server" reads the request to completion.
4. `errno` is poisoned to `ENOTCONN` via a genuine failing syscall (`recv()`
   on a fresh, never-connected socket) — a stand-in for "some unrelated
   syscall earlier in the same thread left `errno` set", which is exactly
   what happens on real hardware per the reporter's log.
5. The exact statement `recv(fd, &dummy, 1, MSG_DONTWAIT | MSG_PEEK)` is
   called on the half-closed socket: it returns `res == 0`, and `errno`
   is confirmed **unchanged** (still `ENOTCONN`) — i.e. real, unmodified
   kernel behavior, not an assumption.
6. Running the **current** switch-statement logic verbatim on `(res=0,
   errno=ENOTCONN)` produces `_connected = false` — the bug, reproduced.
7. Running the **fixed** logic on the same inputs produces
   `_connected = true` — correct.
8. To prove the connection really is still usable (true half-close, not a
   dead socket): the server calls `send()` on the "disconnected" socket and
   the client genuinely receives the bytes.

Output (abridged):

```
=== Step 5: the exact statement from NetworkClient::connected() ===
recv(server_fd, &dummy, 1, MSG_DONTWAIT|MSG_PEEK) -> res = 0
errno BEFORE this call : 107
errno AFTER  this call : 107  (unchanged by a successful/EOF recv, per POSIX)

=== Step 6: run CURRENT (buggy) connected() logic ===
    [current logic] Disconnected: RES: 0, ERR: 107
    current logic result: _connected = false (WRONG - socket is only half-closed)

=== Step 7: run FIXED connected() logic ===
    fixed logic result:   _connected = true (correct)

=== Step 8: prove the socket is genuinely still writable (true half-close) ===
Server send() on the 'disconnected' socket returned 21 bytes (errno=107)
Client still receives it: 21 bytes: "HTTP/1.0 200 OK

ok"
```

(`errno 107` is `ENOTCONN` on Linux glibc; the reporter's hardware log shows
`128`, which is `ENOTCONN` on ESP32/newlib — same condition, different libc
numbering.)

I'd very much appreciate it if a maintainer (or @P-R-O-C-H-Y / @me-no-dev, who
were involved in the issue thread) could additionally confirm this against
real hardware with the reporter's original sketch, since I could only validate
the socket-layer logic on host.

## Related links

Closes #12751

## Disclosure

This fix was investigated and drafted with AI assistance (Claude), based on
reading the issue thread and the source, and validated with the host-side
reproduction described above. I've reviewed the change and reasoning myself
before opening this PR.

### DIFF
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -575,15 +575,8 @@ uint8_t NetworkClient::connected() {
   if (_connected) {
     uint8_t dummy;
     int res = recv(fd(), &dummy, 1, MSG_DONTWAIT | MSG_PEEK);
-    if (res == 0) {
-      // The peer performed an orderly shutdown (FIN received, e.g. via shutdown(SHUT_WR)).
-      // This is a *successful* recv() call, not an error, so errno is not set by it and
-      // must not be inspected here (it may still hold a stale value from an unrelated,
-      // earlier syscall). A half-closed connection can still be perfectly writable, so
-      // do not tear it down based on leftover errno.
-      _connected = true;
-    } else if (res < 0) {
-      // recv only sets errno if res is < 0
+    // recv only sets errno when it returns -1
+    if (res < 0) {
       switch (errno) {
         case EWOULDBLOCK:
         case ENOENT:  //caused by vfs
--- a/libraries/Network/src/NetworkClient.cpp
+++ b/libraries/Network/src/NetworkClient.cpp
@@ -575,10 +575,15 @@ uint8_t NetworkClient::connected() {
   if (_connected) {
     uint8_t dummy;
     int res = recv(fd(), &dummy, 1, MSG_DONTWAIT | MSG_PEEK);
-    // avoid unused var warning by gcc
-    (void)res;
-    // recv only sets errno if res is <= 0
-    if (res <= 0) {
+    if (res == 0) {
+      // The peer performed an orderly shutdown (FIN received, e.g. via shutdown(SHUT_WR)).
+      // This is a *successful* recv() call, not an error, so errno is not set by it and
+      // must not be inspected here (it may still hold a stale value from an unrelated,
+      // earlier syscall). A half-closed connection can still be perfectly writable, so
+      // do not tear it down based on leftover errno.
+      _connected = true;
+    } else if (res < 0) {
+      // recv only sets errno if res is < 0
       switch (errno) {
         case EWOULDBLOCK:
         case ENOENT:  //caused by vfs


libraries/Network/src/NetworkClient.cpp
```diff
@@ -575,10 +575,15 @@ uint8_t NetworkClient::connected() {
...
-    (void)res;
-    // recv only sets errno if res is <= 0
-    if (res <= 0) {
+    if (res == 0) {
+      // The peer performed an orderly shutdown (FIN received, e.g. via shutdown(SHUT_WR)).
+      // This is a *successful* recv() call, not an error, so errno is not set by it and
+      // must not be inspected here (it may still hold a stale value from an unrelated,
+      // earlier syscall). A half-closed connection can still be perfectly writable, so
+      // do not tear it down based on leftover errno.
+      _connected = true;
```
JonNelson: I think this line (`_connected = true;`) is redundant.

JonNelson: Technically, each of the times that `_connected` is set back to `true` in that function (::connected) are redundant.  That said, this PR does not resolve the problem for me, although it may be part of the answer. The issue is that `recv` is still setting `errno` and returning -1. This suggests that the core issue isn't in arduino-esp32 but elsewhere. I did potentially find another related issue in `read`: the value of -1 is returned from read even when the underlying call returns 0.

me-no-dev (approved):

libraries/Network/src/NetworkClient.cpp
me-no-dev: read is a different story. Returns -1 because there is no byte to be read. If it returns 0, then this is data with value `0x00`

PR #12774, opened 2026-07-21
Fix rgbLedWrite typo in MiniMain ESP32-S2 variant comment

HarryXin0919:
## What changed

Corrected `rgbLedWritee()` to `rgbLedWrite()` in the MiniMain ESP32-S2 board variant comment.

## Why

The comment referenced a non-existent API name because of an extra `e`. This could confuse users looking for the documented RGB LED helper.

## Impact

Documentation-only change; runtime behavior is unchanged.

## Validation

- Ran `git diff --check`.
- Searched the repository to confirm no remaining `rgbLedWritee` occurrences.

### DIFF
--- a/variants/department_of_alchemy_minimain_esp32s2/pins_arduino.h
+++ b/variants/department_of_alchemy_minimain_esp32s2/pins_arduino.h
@@ -16,7 +16,7 @@
 
 // RGB LED
 #define PIN_RGB_LED 33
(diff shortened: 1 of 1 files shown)


eeminionn (approved):
Confirmed that `rgbLedWrite` is the public helper used by the ESP32 core and that `rgbLedWritee` has no valid declaration. This is a focused comment-only correction with no runtime impact, and the repository search/checks support the change.

PR #12780, opened 2026-07-24
feat(esp32): Add setCiphers to NetworkClientSecure

soundstorm:
## Description of Change
This PR allows setting a subset of cipher suits. This is sometimes necessary, as some providers expect specific suits for specific user agents.

## Test Scenarios
I've tested this on a WROOM32/CYD with DBAPI 

## Related links
This will allow DBAPI to use the mainline core again (https://github.com/ArduinoHannover/DBAPI/commit/229196ff50ff7a407efc3456f302fb65a8c566e1)

### DIFF
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -456,19 +456,19 @@ void NetworkClientSecure::setAlpnProtocols(const char **alpn_protos) {
 }
 
 bool NetworkClientSecure::setCiphers(const int *list, size_t count) {
-  if (list == nullptr || count == 0) {
-    log_e("setCiphers: list is null or count is zero");
-    return false;
-  }
-
-  if (connected()) {
+  if (_connected) {
     log_e("setCiphers: cannot change ciphersuites while a connection is active; "
           "call this before connect() or after stop()");
     return false;
   }
 
+  if (list == nullptr || count == 0) {
+    sslclient->cipher_list.reset();
+    return true;
+  }
+
   sslclient->cipher_list = std::shared_ptr<int>(new (std::nothrow) int[count + 1], std::default_delete<int[]>());
-  if (!sslclient->cipher_list.get()) {
+  if (!sslclient->cipher_list) {
     return false;
   }
   memcpy(sslclient->cipher_list.get(), list, count * sizeof(int));
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -456,6 +456,17 @@ void NetworkClientSecure::setAlpnProtocols(const char **alpn_protos) {
 }
 
 bool NetworkClientSecure::setCiphers(const int *list, size_t count) {
+  if (list == nullptr || count == 0) {
+    log_e("setCiphers: list is null or count is zero");
+    return false;
+  }
+
+  if (connected()) {
+    log_e("setCiphers: cannot change ciphersuites while a connection is active; "
+          "call this before connect() or after stop()");
+    return false;
+  }
+
   sslclient->cipher_list = std::shared_ptr<int>(new (std::nothrow) int[count + 1], std::default_delete<int[]>());
   if (!sslclient->cipher_list.get()) {
     return false;
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
@@ -455,6 +455,16 @@ void NetworkClientSecure::setAlpnProtocols(const char **alpn_protos) {
   _alpn_protos = alpn_protos;
 }
 
+bool NetworkClientSecure::setCiphers(const int *list, size_t count) {
+  sslclient->cipher_list = std::shared_ptr<int>(new (std::nothrow) int[count + 1], std::default_delete<int[]>());
+  if (!sslclient->cipher_list.get()) {
+    return false;
+  }
+  memcpy(sslclient->cipher_list.get(), list, count * sizeof(int));
+  sslclient->cipher_list.get()[count] = 0;
+  return true;
+}
+
 int NetworkClientSecure::fd() const {
   return sslclient->socket;
 }
--- a/libraries/NetworkClientSecure/src/NetworkClientSecure.h
+++ b/libraries/NetworkClientSecure/src/NetworkClientSecure.h
@@ -79,6 +79,7 @@ public:
   bool verify(const char *fingerprint, const char *domain_name);
   void setHandshakeTimeout(unsigned long handshake_timeout);
   void setAlpnProtocols(const char **alpn_protos);
+  bool setCiphers(const int *list, size_t count);
 
   // Certain protocols start in plain-text; and then have the client
   // give some STARTSSL command to `upgrade' the connection to TLS
--- a/libraries/NetworkClientSecure/src/ssl_client.cpp
+++ b/libraries/NetworkClientSecure/src/ssl_client.cpp
@@ -48,8 +48,8 @@ static int _handle_error(int err, const char *function, int line) {
 #define handle_error(e) _handle_error(e, __FUNCTION__, __LINE__)
 
 void ssl_init(sslclient_context *ssl_client) {
-  // reset embedded pointers to zero
-  memset(ssl_client, 0, sizeof(sslclient_context));
+  // reset ssl_client by creating new (empty) context, as shared_ptr is not safe for memset
+  *ssl_client = sslclient_context();
   mbedtls_ssl_init(&ssl_client->ssl_ctx);
   mbedtls_ssl_config_init(&ssl_client->ssl_conf);
 #if MBEDTLS_VERSION_MAJOR < 4
--- a/libraries/NetworkClientSecure/src/ssl_client.cpp
+++ b/libraries/NetworkClientSecure/src/ssl_client.cpp
@@ -409,8 +409,9 @@ void stop_ssl_socket(sslclient_context *ssl_client) {
   int last_err = ssl_client->last_error;
   crt_bundle_attach_cb bundle_attach_cb = ssl_client->bundle_attach_cb;
 
-  // reset embedded pointers to zero
-  memset(ssl_client, 0, sizeof(sslclient_context));
+  
+  // reset ssl_client by creating new (empty) context, as shared_ptr is not safe for memset
+  *ssl_client = sslclient_context();
 
   ssl_client->handshake_timeout = handshake_timeout;
   ssl_client->socket_timeout = socket_timeout;
--- a/libraries/NetworkClientSecure/src/ssl_client.cpp
+++ b/libraries/NetworkClientSecure/src/ssl_client.cpp
@@ -196,6 +196,10 @@ int start_ssl_client(
       return handle_error(ret);
     }
   }
+  
+  if (ssl_client->cipher_list) {
+    mbedtls_ssl_conf_ciphersuites(&ssl_client->ssl_conf, ssl_client->cipher_list.get());
+  }
 
   // MBEDTLS_SSL_VERIFY_REQUIRED if a CA certificate is defined on Arduino IDE and
   // MBEDTLS_SSL_VERIFY_NONE if not.
--- a/libraries/NetworkClientSecure/src/ssl_client.cpp
+++ b/libraries/NetworkClientSecure/src/ssl_client.cpp
@@ -408,6 +408,7 @@ void stop_ssl_socket(sslclient_context *ssl_client) {
   unsigned long socket_timeout = ssl_client->socket_timeout;
   int last_err = ssl_client->last_error;
   crt_bundle_attach_cb bundle_attach_cb = ssl_client->bundle_attach_cb;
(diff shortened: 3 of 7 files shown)


libraries/NetworkClientSecure/src/ssl_client.h
```diff
@@ -40,6 +41,8 @@ typedef struct sslclient_context {
   int last_error;
   int peek_buf;
 
+  std::shared_ptr<int> cipher_list;
```
lucasssvaz, line 44: Putting a non-trivial C++ object into `sslclient_context` is what makes the existing `memset()` unsafe, and that is what forces the `*ssl_client = sslclient_context()` rewrite in `ssl_init()` and `stop_ssl_socket()`. Measured cost of that rewrite is ~2.2 KB of stack in each (numbers in my comment on `stop_ssl_socket()`), plus about 1 KB of flash for every TLS user, which is why the memory report shows every `HTTPClient` example growing even though none of them touch ciphersuites.

There is also a latent use-after-free in the current shape. `sslclient` is a `std::shared_ptr<sslclient_context>` and `NetworkClientSecure` declares no copy constructor, so copies share one context while each keeps its own `_connected`:

```cpp
NetworkClientSecure a;
a.setCiphers(list, n);
a.connect(host, 443);       // a._connected == true; ssl_conf holds the raw pointer

NetworkClientSecure b = a;  // shares sslclient; b._connected == false
b.setCiphers(other, m);     // guard passes, frees the array a is still using
```

`mbedtls_ssl_conf_ciphersuites()` stores the pointer without copying, so `a`'s live config is left dangling.

What to do instead: revert this file entirely (drop this member and the `<memory>` include at L13), and take the list as a parameter, the same way ALPN already works:

```cpp
int start_ssl_client(
  sslclient_context *ssl_client, const IPAddress &ip, uint32_t port, const char *hostname, int timeout, const char *rootCABuff, bool useRootCABundle,
  const char *cli_cert, const char *cli_key, const char *pskIdent, const char *psKey, bool insecure, const char **alpn_protos, const int *ciphersuites
);
```

The list itself then lives in `NetworkClientSecure` next to `_alpn_protos`, as a plain `const int *_cipher_list`. That keeps `sslclient_context` a POD, keeps both `memset()`s, and removes the ownership question entirely, which matches mbedTLS's own contract, since `mbedtls_ssl_conf_ciphersuites()` never copies the array and requires it to outlive the config regardless.

One caution if you are tempted to keep a copy instead: do **not** make the member a `std::unique_ptr<int[]>`. `NetworkClient` and `NetworkClientSecure` are both implicitly copyable today (neither declares a copy constructor or assignment operator), and a `unique_ptr` member would delete the implicit copy constructor and silently break user code that copies these objects.

libraries/NetworkClientSecure/src/ssl_client.cpp
```diff
@@ -404,14 +408,17 @@ void stop_ssl_socket(sslclient_context *ssl_client) {
   unsigned long socket_timeout = ssl_client->socket_timeout;
   int last_err = ssl_client->last_error;
   crt_bundle_attach_cb bundle_attach_cb = ssl_client->bundle_attach_cb;
+  std::shared_ptr<int> saved_ciphers = ssl_client->cipher_list;
```
lucasssvaz, line 411: **Please revert this whole hunk back to the `memset()`.** With the list held in `NetworkClientSecure` there is nothing here left to preserve, so both the `saved_ciphers` copy and the restore below become unnecessary.

For the record, this is the measurement that motivates the request. `*ssl_client = sslclient_context()` has to materialize a fully value-initialized `sslclient_context` temporary on the stack before the memberwise assignment; the compiler can no longer fold it into a `memset()` because `std::shared_ptr`'s assignment operator is non-trivial. Built with `xtensa-esp32-elf-g++ 14.2.0` at `-Os -std=gnu++2b` against the real IDF mbedTLS headers, using `-fstack-usage`:

| | master | this PR |
|---|---|---|
| `stop_ssl_socket()` stack | 32 B | **2224 B** |
| `ssl_init()` stack | 32 B | **2208 B** |
| `stop_ssl_socket()` `.text` | 108 B | 386 B |

`sizeof(sslclient_context)` is 2168 bytes on ESP32. I re-measured with the `shared_ptr` removed from the struct and `stop_ssl_socket()` goes straight back to 32 bytes.

This is not a cold path: `stop_ssl_socket()` is reached through `NetworkClientSecure::stop()`, which runs on every failed `read()`, `write()` and `available()`. So it fires deep in the call stack, on the 8 KB Arduino loop task or on user tasks that are commonly created with 4–6 KB.

libraries/NetworkClientSecure/src/ssl_client.cpp
```diff
@@ -48,8 +48,8 @@ static int _handle_error(int err, const char *function, int line) {
 #define handle_error(e) _handle_error(e, __FUNCTION__, __LINE__)
 
 void ssl_init(sslclient_context *ssl_client) {
-  // reset embedded pointers to zero
-  memset(ssl_client, 0, sizeof(sslclient_context));
+  // reset ssl_client by creating new (empty) context, as shared_ptr is not safe for memset
+  *ssl_client = sslclient_context();
```
lucasssvaz, line 52: Please revert this to the original `memset()` as well, same 2.2 KB stack temporary as in `stop_ssl_socket()` (2208 bytes measured here). Once `sslclient_context` has no `shared_ptr` member, `memset()` is safe again and this hunk is no longer needed at all.

libraries/NetworkClientSecure/src/ssl_client.cpp
```diff
@@ -196,6 +196,10 @@ int start_ssl_client(
       return handle_error(ret);
     }
   }
+  
+  if (ssl_client->cipher_list) {
+    mbedtls_ssl_conf_ciphersuites(&ssl_client->ssl_conf, ssl_client->cipher_list.get());
+  }
```
lucasssvaz, line 202: This becomes a check on the new parameter rather than on struct state:

```cpp
if (ciphersuites != NULL) {
  mbedtls_ssl_conf_ciphersuites(&ssl_client->ssl_conf, ciphersuites);
}
```

libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
```diff
@@ -455,6 +455,27 @@ void NetworkClientSecure::setAlpnProtocols(const char **alpn_protos) {
...
+  }
+
+  sslclient->cipher_list = std::shared_ptr<int>(new (std::nothrow) int[count + 1], std::default_delete<int[]>());
+  if (!sslclient->cipher_list) {
+    return false;
+  }
+  memcpy(sslclient->cipher_list.get(), list, count * sizeof(int));
+  sslclient->cipher_list.get()[count] = 0;
+  return true;
+}
```
lucasssvaz, line 477: With the list moved out of `sslclient_context`, this whole function collapses to a plain setter in the same style as `setAlpnProtocols()` right above it:

```cpp
void NetworkClientSecure::setCiphers(const int *ciphersuites) {
  _cipher_list = ciphersuites;
}
```

Everything else falls away with it:

- No allocation, so no failure mode, so no `bool` return and no missing `log_e` on the failure path.
- No `count`, so no risk of `count + 1` or `count * sizeof(int)` wrapping into a zero-sized `new` followed by a huge `memcpy`. A 0-terminated array matches both mbedTLS and the `NULL`-terminated `const char **` that `setAlpnProtocols()` already takes.
- No `_connected` guard needed. mbedTLS snapshots the pointer into `ssl_conf` at connect time, so a later call simply applies to the next connection, the same semantics every other setter in this class already has. It also removes the use-after-free I described on `ssl_client.h`, since nothing gets freed out from under a live config any more.

Don't forget the two `start_ssl_client()` call sites need the new argument: L148 for the certificate path and L196 for the PSK path.

libraries/NetworkClientSecure/src/NetworkClientSecure.h
```diff
@@ -79,6 +79,7 @@ class NetworkClientSecure : public NetworkClient {
   bool verify(const char *fingerprint, const char *domain_name);
   void setHandshakeTimeout(unsigned long handshake_timeout);
   void setAlpnProtocols(const char **alpn_protos);
+  bool setCiphers(const int *list, size_t count);
```
lucasssvaz, line 82: Suggested signature, to line up with `setAlpnProtocols()` on the line above:

```cpp
void setCiphers(const int *ciphersuites);  // 0-terminated, caller-owned
```

and a `const int *_cipher_list;` member next to `_alpn_protos` at L43.

Documentation is still needed for the new public API:

- No Doxygen block. Please document that the values are mbedTLS `MBEDTLS_TLS_*` ciphersuite IDs, that the array must be 0-terminated, and that passing `nullptr` restores the default list.
- With a caller-owned pointer, the array has to stay valid for as long as the client may reconnect (a `static const` or global array). That constraint needs to be stated explicitly, the same way it applies to `setAlpnProtocols()` and `setCACert()`.
- `libraries/NetworkClientSecure/README.md` documents `setAlpnProtocols()` with a short usage snippet; `setCiphers()` should get an equivalent section next to it.
- Two footguns worth calling out: ciphersuite IDs not enabled in the build are silently dropped by mbedTLS and surface later as a handshake failure rather than an error from this call, and a restricted list combined with `setPreSharedKey()` can filter out every PSK suite.

A validation test under `tests/validation/tls_http/` would also be welcome, since nothing currently exercises this path.

lucasssvaz (changes requested):
Thanks for the PR, and sorry for the slow review. The feature is welcome, we just need a different implementation shape before we can take it.

The main ask is to **drop the `std::shared_ptr` from `sslclient_context`**. That member is the root of everything else in the diff: it is what makes the existing `memset()` unsafe, which forced the `*ssl_client = sslclient_context()` rewrite in `ssl_init()` and `stop_ssl_socket()`, which in turn costs about **2.2 KB of stack** in each of those functions. `stop_ssl_socket()` runs from `NetworkClientSecure::stop()` on every failed read/write, so as it stands this is a stack regression for every TLS user, not only for people calling `setCiphers()`. Measurements are in the inline comment on `stop_ssl_socket()`.

Instead, please store the list in `NetworkClientSecure` and pass it into `start_ssl_client()` as a parameter, exactly how `_alpn_protos` already works in this class. With that change the rest of the diff collapses nicely: both `memset()`s stay, the save/restore dance in `stop_ssl_socket()` disappears, `ssl_client.h` no longer needs `<memory>`, there is no per-connection allocation, and the ~1 KB of flash growth the memory-usage report shows on every TLS example goes away. The inline comments walk through what each hunk should become.

lucasssvaz (commented):
Also please add the new methods to `keywords.txt`

me-no-dev:
@lucasssvaz PTAL

libraries/NetworkClientSecure/src/NetworkClientSecure.h
```diff
@@ -79,6 +80,24 @@ class NetworkClientSecure : public NetworkClient {
...
+   * Note: ciphersuite IDs that are not enabled in the mbedTLS build are silently ignored and will
+   * only surface later as a handshake failure. A restrictive list combined with `setPreSharedKey()`
+   * can also filter out every PSK-capable ciphersuite.
+   */
+  void setCiphers(const int *ciphersuites);
+  /**
+   * Overload for source compatibility with BearSSL-based cores (e.g. ESP8266's
+   * WiFiClientSecure::setCiphers()). `ciphersuites` must still be 0-terminated; `count` is unused.
+   */
+  void setCiphers(const int *ciphersuites, size_t count);
```
lucasssvaz, line 100: Drop this overload as BearSSL requires `const uint16_t *` vs mbedtls' `const int *`. Either way conversion will be needed from BearSSL code and this overload can be misleading.

```suggestion
```

libraries/NetworkClientSecure/src/NetworkClientSecure.cpp
```diff
@@ -460,6 +466,15 @@ void NetworkClientSecure::setAlpnProtocols(const char **alpn_protos) {
...
 
+void NetworkClientSecure::setCiphers(const int *ciphersuites) {
+  _cipher_list = ciphersuites;
+}
+
+void NetworkClientSecure::setCiphers(const int *ciphersuites, size_t count) {
+  (void)count;
+  setCiphers(ciphersuites);
+}
+
```
lucasssvaz, line 477:
```suggestion

```

libraries/NetworkClientSecure/README.md
```diff
@@ -94,6 +94,29 @@ const char *aws_protos[] = {"mqtt", NULL};
...
+
+```
+static const int cipher_list[] = {MBEDTLS_TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, 0};
+...
+wiFiClient.setCiphers(cipher_list);
+```
+
+Call `setCiphers(nullptr)` to restore mbedTLS's default ciphersuite list. An overload
+`setCiphers(cipher_list, count)` is also provided for source compatibility with BearSSL-based cores
+(e.g. ESP8266); `count` is ignored and the array must still be 0-terminated.
```
lucasssvaz, line 114:
```suggestion
Call `setCiphers(nullptr)` to restore mbedTLS's default ciphersuite list.
```

lucasssvaz (changes requested):
Thanks for the changes. Just one small adjustment and it is ready to merge.

lucasssvaz (approved):

Issue #12831, opened 2026-08-17
adding std::string to the Preferences object getter and putter.


SuGlider:
#### Originally posted by **nielsnl68** August 11, 2026

Hello,

In my firmware i use only the `std::string` as string storage's.
When i want to use the Preferences object i found out that only the Arduino String variable is offered.
Would it be interesting when i create a PR to add this option as well.

TD-er:
When adding a getter/setter for `str::string`, the `String` class will still be linked in as it is used in many more classes.
So wouldn't it make more sense to add a `std::string String::to_stdstring() const;` and a `String& String::from_stdstring();` in the `String` class?

Edit:
And constructor for `String(const std::string& str);`

PR #12834, opened 2026-08-17
feat(Preferences): Add std::string to the Preferences object.

nielsnl68:

### Checklist

-----------
## Description of Change
In my firmware i use only the std::string as string storage's.
When i want to use the Preferences object i found out that only the Arduino String variable is offered.
This PR add this option. It is basically a copy of the `putString()` and `getString()` but returning `std::string` instead.

## Test Scenarios
I have tested my Pull Request on Arduino-esp32 core v3.3.10 with an ESP32-S3.
```cpp
std::string current_name = "";

void load() {
  Preferences prefs;
  prefs.begin("testing", true);
  current_name = prefs.getStdString("name", current_name);
  prefs.end();
}

void load() {
  Preferences prefs;
  prefs.begin("testing");
  prefs.putStdString("name", current_name);
  prefs.end();
}
```

## Related links
https://github.com/espressif/arduino-esp32/issues/12831
https://github.com/espressif/arduino-esp32/discussions/12820

### DIFF
--- a/docs/en/tutorials/preferences.rst
+++ b/docs/en/tutorials/preferences.rst
@@ -78,9 +78,11 @@ Preferences directly supports the following data types:
    +-------------------+-------------------+---------------+
    | Double            | double_t          | 8             |
    +-------------------+-------------------+---------------+
-   |                   | const char*       | variable      |
-   | String            +-------------------+               |
+   | String            | const char*       | variable      |
+   |                   +-------------------+               |
    |                   | String            |               |
+   |                   +-------------------+               |
+   |                   | std::string       |               |
    +-------------------+-------------------+---------------+
    | Bytes             | uint8_t           | variable      |
    +-------------------+-------------------+---------------+
@@ -618,6 +620,8 @@ With Preferences, the ``getX`` methods listed in Table 2 below will return a def
    | Double           |                 |
    +------------------+-----------------+
    | String (String)  | ""              |
+   |                  |                 |
+   | std::string      |                 |
    +------------------+-----------------+
    | String (* buf)   | \\0             |
    +------------------+-----------------+
--- a/libraries/Preferences/src/Preferences.cpp
+++ b/libraries/Preferences/src/Preferences.cpp
@@ -282,6 +282,10 @@ size_t Preferences::putString(const char *key, const String value) {
   return putString(key, value.c_str());
 }
 
+size_t Preferences::putStdString(const char *key, const std::string value) {
+  return putString(key, value.c_str());
+}
+
 size_t Preferences::putBytes(const char *key, const void *value, size_t len) {
   if (!_started || !key || !value || !len || _readOnly) {
     return 0;
@@ -514,6 +518,27 @@ String Preferences::getString(const char *key, const String defaultValue) {
     return String(defaultValue);
   }
   return String(buf);
+}
+
+std::string Preferences::getStdString(const char *key, const std::string defaultValue) {
+  char *value = NULL;
+  size_t len = 0;
+  if (!_started || !key) {
+    return std::string(defaultValue);
+  }
+  esp_err_t err = nvs_get_str(_handle, key, value, &len);
+  if (err) {
+    log_e("nvs_get_str len fail: %s %s", key, nvs_error(err));
+    return std::string(defaultValue);
+  }
+  char buf[len];
+  value = buf;
+  err = nvs_get_str(_handle, key, value, &len);
+  if (err) {
+    log_e("nvs_get_str fail: %s %s", key, nvs_error(err));
+    return std::tring(defaultValue);
+  }
+  return std::string(buf);
 }
 
 size_t Preferences::getStringLength(const char *key) {
--- a/libraries/Preferences/src/Preferences.cpp
+++ b/libraries/Preferences/src/Preferences.cpp
@@ -282,9 +282,11 @@ size_t Preferences::putString(const char *key, const String value) {
   return putString(key, value.c_str());
 }
 
+#ifdef USE_STD_STRING
 size_t Preferences::putStdString(const char *key, const std::string value) {
   return putString(key, value.c_str());
 }
+#endif
 
 size_t Preferences::putBytes(const char *key, const void *value, size_t len) {
   if (!_started || !key || !value || !len || _readOnly) {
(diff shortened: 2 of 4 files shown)


docs/en/tutorials/preferences.rst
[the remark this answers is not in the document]
nielsnl68: link does not work properly.

nielsnl68:
Sorry, Altrough i understand why this PR fails, i do not know how resolve the issue correctly. 
When someone likes to jump in then i'm really happy,

libraries/Preferences/src/Preferences.h
```diff
@@ -16,6 +16,10 @@
 
 #include "Arduino.h"
 
+#if __has_include(<string>)
```
robertlipe, line 19: I'm not sure you need this test. In all but the most restrictive environments (freestanding, which is already doomed with Arduino) you're going to have <string> - that's been part of the STL since the early 90's.

libraries/Preferences/src/Preferences.cpp
```diff
@@ -516,6 +522,29 @@ String Preferences::getString(const char *key, const String defaultValue) {
   return String(buf);
 }
 
+#ifdef USE_STD_STRING
+std::string Preferences::getStdString(const char *key, const std::string defaultValue) {
```
robertlipe, line 526: Make that a string& and you'll save a copy on every call.

libraries/Preferences/src/Preferences.cpp
```diff
@@ -516,6 +522,29 @@ String Preferences::getString(const char *key, const String defaultValue) {
...
+  if (!_started || !key) {
+    return std::string(defaultValue);
+  }
+  esp_err_t err = nvs_get_str(_handle, key, value, &len);
+  if (err) {
+    log_e("nvs_get_str len fail: %s %s", key, nvs_error(err));
+    return std::string(defaultValue);
+  }
+  char buf[len];
+  value = buf;
```
robertlipe, line 538: IMO, it's clearer not reuse value here. Just pass &buf.

libraries/Preferences/src/Preferences.cpp
```diff
@@ -516,6 +522,29 @@ String Preferences::getString(const char *key, const String defaultValue) {
...
+  err = nvs_get_str(_handle, key, value, &len);
+  if (err) {
+    log_e("nvs_get_str fail: %s %s", key, nvs_error(err));
+    return std::tring(defaultValue);
```
robertlipe, line 542: How does that compile?

I'm not saying it's a great idea, but you might be able to save one trip through nvs_get_str just by deciding that MOST keys will be shorter than some reasonable value and at least trying to shove it into that buffer, just reallocating on failure.

const int kDefaultSize  = 65;
char* buf = alloca(kDefaultSize); // Why alloca? so you can reallocate it.
// try to get it. If len didn't come back 65, return std::string(buf); and you're done.
Now you have the real size of len you can use for nvs_get_str()

It's a real annoyance to bolt C++ APIs atop C for this kind of thing.

Maybe instead of alloca, you can malloc it and then std::move the result into a string.

It's definitely the kind of thing I would wrap up in a unit test that exercised BOTH paths of that with values below, at, and above the knee.

I wouldn't fake shock if the C API had edge cases involving null termination. Test it with a long key to see.           

While I'm all about abolishing String from the world, if this is an Arduino library and not a 'normal' C++ one, should we accept the self-loathing and make that a String instead of a std::string? (Insert "Why not both?" meme)

robertlipe (commented):
The hint you probably need is buried inside the error itself.


/__w/arduino-esp32/arduino-esp32/components/arduino-esp32/libraries/Preferences/src/Preferences.h:69:40: error: 'std::string' has not been declared
   69 |   size_t putStdString(const char *key, std::string value);
      |                                        ^~~
/__w/arduino-esp32/arduino-esp32/components/arduino-esp32/libraries/Preferences/src/Preferences.h:91:8: error: 'string' in namespace 'std' does not name a type
   91 |   std::string getStdString(const char *key, std::string defaultValue = std::string());
      |        ^~~~~~
/__w/arduino-esp32/arduino-esp32/components/arduino-esp32/libraries/Preferences/src/Preferences.h:18:1: note: 'std::string' is defined in header '<string>'; this is probably fixable by adding '#include <string>'
   17 | #include "Arduino.h"
  +++ |+#include <string>

You're testing for the presence of string (why? it's been part of the STL since...forever) but never actually including it.

libraries/Preferences/src/Preferences.cpp
robertlipe: Another idea. Eliminate the special case. The code you need is in the very next function.

need_len = getStringLength(key);
buf[need_len + 1];
  esp_err_t err = nvs_get_str(_handle, key, &buf, &len); // Again, consider zero terminatein rules
return std::string(buf)

Done. This works because the next function IS the "call it with zero just so it can tell me the size" case. 

This means you're still making two nvs_get_str API calls, but the code becomes dead easy. (Well, other than the ambiguity on whether 'len' includes the null terminator or not.)

nielsnl68: Nice idea. But that is not fully the issue here.
It seams that **std::string** is not included by default. I do not want to include the `<string>` header by default just to allow nice need.

robertlipe: It's not optional. That's what the error is telling you quite clearly. In C++, you have to include what you use; hoping that someone else included it for you rarely pays off.

And the "have a small buffer, if it doesn't fit, make a bigger one" is in code that I *just* had to write and test. It's algorthimically perfectly to drop in here if you choose to go that way.

```
  static string asprintf(const std::string& fmt, ...) {
+    char buf[1024];
+    va_list args;
+    va_start(args, fmt);
+    int n = vsnprintf(buf, sizeof(buf), fmt.c_str(), args);
+    va_end(args);
+    if (n < 0) {
+      return string();
+    }
+    if (n < static_cast<int>(sizeof(buf))) {
+      return string(buf);
+    }
+    string result(n, '\0');
+    va_start(args, fmt);
+    vsnprintf(result.data(), n + 1, fmt.c_str(), args);
+    va_end(args);
+    return result;
+  }
```

PR #12837, opened 2026-08-18
feat(usb): Add USB Host library (HID,MSC,Serial)

### LINKED ISSUE #10978: Add support for TinyUSB Host
### Related area

USB

### Hardware specification

ESP32-S3

I would like to use USB Host on S3.

I hear that USB Host support is ready from tinyusb, which is very exciting.
https://github.com/hathach/tinyusb/issues/2634#issuecomment-2646978422

Would it be possible to confirm this?
Will we see some example code at some point?
I am particularly interested in MIDI USB Host.
thank you!

Currently I use esp-idf for USB MIDI Host, but I would like to use Arduino, because I produce development boards, and my users prefer Arduino.

### Additional context

_No response_

### LINKED ISSUE #12286: USB Host MSC Support for ESP32-S3 to read external USB Storage
### Related area

USB Host MSC 

### Hardware specification

ESP32-S3-DEVKITC1

For 1 of my production solution, i want to access pendrive's files and read them and do some computation in ESP32-S3. 

Hi,

I have a pendrive with csv files connected to ESP32-S3-D3VKITC1 over USB OTG . 
I want to read the files of pendrive in s3. 
Is there a way to achieve this using arduino ide?

_No response_

### Additional context

_No response_

### LINKED ISSUE #12400: USB MassStorage support like SD_MMC for ESP32-P4
### Related area

USB

### Hardware specification

ESP32-P4-Module-DEV-KIT

For example you want to create a mp3 music player using a ESP32-P4 it would be nice to be able to read mp3 files from a FAT32 formatted USB stick instead of the SD card. The ESP32-P4-Module-DEV-KIT comes with 4 USB 2.0 ports and the NANO with 1. 

Image

Reading/Writing files on USB sticks just like SD_MMC.

_No response_

### Additional context

_No response_

### LINKED ISSUE #12449: Restore USB Host MSC support for ESP32-S3 on Arduino-ESP32 Roadmap
### Related area

USB Host MSC functionality (Mass Storage Class host mode) support for ESP32-S3 in Arduino-ESP32 core.

### Hardware specification

ESP32-S3, ESP32-S2 (native USB OTG capable boards)

USB Host MSC support was previously discussed, requested and appeared to be scheduled in the Arduino-ESP32 roadmap for January 2025. However, it is now missing from both the public roadmap and recent communications. This removes hopes for a production-ready ability to connect USB flash drives directly to ESP32-S3/S2 boards under Arduino, as requested in Issue #9276 (https://github.com/espressif/arduino-esp32/issues/9276). Multiple developers have been monitoring the official roadmap and documentation, and despite requests and forum posts, there is no clear update. It appears the feature has been deprioritized or removed altogether, yet its absence was not publicly acknowledged in roadmap changes.

USB Host MSC Host support should be restored as a priority to the Arduino-ESP32 roadmap, with transparent communication about timeline, known obstacles, and implementation progress. Documentation needs to clarify roadmap changes. At minimum, official rationale should be published for its removal or delay.

Workarounds via ESP-IDF exist but are not ideal for most Arduino users. The only viable current solution requires direct ESP-IDF development and TinyUSB use, which is beyond the scope for most Arduino users. Community forks and code demos (e.g. https://github.com/nathalislight/esp32s3_USB_Drive/tree/main) exist, but none offer stable feature support from Arduino-ESP32. USB Device MSC mode is supported (ESP32 can appear as storage to a PC), but USB Host MSC mode (reading USB flash drive data) is missing.

### Additional context

- Issue #9276 requested USB Host MSC Host support and links community efforts.
- Related forums: https://esp32.com/viewtopic.php?t=32991
- Documentation review and roadmap history show removal:
  https://github.com/espressif/arduino-esp32/wiki/Release-Roadmap-and-Management/_history
- Official documentation: https://espressif-docs.readthedocs-hosted.com/projects/arduino-esp32/en/latest/api/usb_msc.html
- USB Device mode is NOT the same as USB Host mode (feature distinction)
- Most up-to-date ESP-IDF USB Host examples: https://github.com/espressif/esp-idf/tree/master/examples/peripherals/usb/host/usb_host_lib
- Please restore roadmap transparency and provide user-visible updates.

P-R-O-C-H-Y:
## Description of Change
This pull request adds initial support for USB Host mode on ESP32-S2, S3, and P4 chips, enabling Arduino sketches to interact with USB devices such as serial adapters, flash drives (MSC), mice, keyboards, and gamepads. The changes include new core files, updates to the build system, enhancements to the TinyUSB integration, library Host APIs (`USBHostSerial`, `USBHostMSC` / `USBMSCFS`, `USBHostHID` + mouse/keyboard/gamepad), and example sketches demonstrating USB Host functionality.

**Key changes:**

**USB Host Core Implementation:**
- Added new core files `USBHost.cpp` and `USBHost.h` implementing the `USBHostClass`, which manages USB Host initialization, background task handling, and event processing for TinyUSB host mode. (`cores/esp32/USBHost.cpp`, `cores/esp32/USBHost.h`) [[1]](diffhunk://#diff-3f45b629e805b689753516aec37a9fd0834d2faa0d7a9f7864bf8c5fc1b20684R1-R123) [[2]](diffhunk://#diff-f064a350f2da756b5614f43da05d6078c8243aadf0330d2ee310263aa4042ed3R1-R62)
- Updated `CMakeLists.txt` to include the new core and library source files required for USB Host support, such as HID, MSC, serial, USBMSCFS, and report map parsing/dumping. [[1]](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR77) [[2]](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR262-R273)

**TinyUSB Host Integration:**
- Extended the TinyUSB HAL (`esp32-hal-tinyusb.c`, `esp32-hal-tinyusb.h`) to support host mode initialization and deinitialization, including proper PHY bring-up and configuration for different ESP32 targets. Added `tinyusb_host_init()` and related host config structures. [[1]](diffhunk://#diff-d097e4c789a457b03434bb0c64f666c27ec3dda012b39dbf036924904fe95bacR126-R143) [[2]](diffhunk://#diff-d097e4c789a457b03434bb0c64f666c27ec3dda012b39dbf036924904fe95bacR178-R217) [[3]](diffhunk://#diff-d097e4c789a457b03434bb0c64f666c27ec3dda012b39dbf036924904fe95bacL179-R235) [[4]](diffhunk://#diff-dd9bbc60d81f2e139c1f70d3460754740eb104c1f7eb6470ee0e81a0e35515e4R78-R86) [[5]](diffhunk://#diff-d097e4c789a457b03434bb0c64f666c27ec3dda012b39dbf036924904fe95bacR943-R977)

**Examples and Documentation:**
- Added new example sketches demonstrating USB Host usage:
  - [`USBHostCDC_MSC_HID.ino`](diffhunk://#diff-443d594224e2fbf14d6a15760971db632bd43c33b00b735f757ecfb4502ae3bfR1-R182): Shows how to use CDC (serial), MSC (flash drive), and HID (mouse/keyboard) devices.
  - [`USBHostGamepad.ino`](diffhunk://#diff-65224c1e0c0480718d563f20a6ce4eb6b5f59c41c9db50028eff2b979d11c42aR1-R79): Demonstrates reading gamepad/joystick HID devices.
  - Also: `USBHostSerial`, `USBHostMSC_Test`, `USBHostMouse` / `USBHostMouseCallback`, `USBHostKeyboard`, `USBHostHIDCombo`, plus per-example READMEs where present. [[1]](diffhunk://#diff-dfd02f3c18e0e34ad1b0f51322589544c8293056cb6893acb6ae8509eba80321R1-R33) [[2]](diffhunk://#diff-0c6bab4c2d6f94dbbc9bcc50cfab38988ca90d1f4dcb6b64f4b714f6a4e63adbR1-R29)
- MSC: FatFS mount via `USBMSCFS` (including GPT flash drives via a synthetic MBR for Arduino FatFs).
- HID mouse: switches to Report protocol on claim so wheel data is available (boot mouse reports are 3 bytes and have no wheel).

These changes collectively enable USB Host support in the Arduino-ESP32 core for common peripherals on supported OTG boards.

**Known limitations:** ESP32-S3 has a limited host channel pool — hub + multiple HID (and MSC) is best-effort. Global handlers are effectively one device/interface each. Keyboard host path targets boot-protocol keyboards.

## Test Scenarios
New USB Host sketches tested on ESP32-S3 (USB-OTG board) and ESP32-P4.
On S3 with a hub, the stack can hang when unplugging/replugging peripherals multiple times (see known limitations above).
Mouse wheel verified after the Report-protocol switch.

## Related links
Closes #12286
Closes #12449
Closes #10978
Closes #12400


cores/esp32/esp32-hal-tinyusb.c
```diff
@@ -162,6 +175,46 @@ esp_err_t deinit_usb_hal() {
...
 
+#if CFG_TUH_ENABLED
+esp_err_t init_usb_hal_host(bool external_phy) {
+  if (phy_handle_host != NULL) {
+    return ESP_OK;  // already initialized
+  }
+  (void)external_phy;
+  usb_phy_config_t phy_config = {
+    .controller = USB_PHY_CTRL_OTG,
+#if CONFIG_IDF_TARGET_ESP32P4
```
me-no-dev: would this need a change for S31?

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+
+/* Do not #include USBHostHID.h here — library paths are often missing from core builds.
+ * HID provides a strong arduino_usb_host_hid_service() instead. */
+extern "C" void arduino_usb_host_hid_service(void) __attribute__((weak));
+extern "C" void arduino_usb_host_hid_service(void) {}
+
+static TaskHandle_t s_tuh_worker = nullptr;
```
me-no-dev: use NULL for C pointers

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,123 @@
...
+  return s_tuh_worker != nullptr;
+}
+
+bool USBHostClass::begin() {
+  if (_started) {
+    return true;
+  }
+
+  /* ESP32-S3-USB-OTG host mux + VBUS (GPIO here; variant helpers may not link into core). */
+#if defined(USB_HOST_EN) && defined(DEV_VBUS_EN) && defined(LIMIT_EN)
```
me-no-dev: this block seems specific for particular board. There should be a better way to implement this, like using weak function that is overwritten by the variant. Different boards might need different init procedures here.

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,123 @@
...
+  pinMode(DEV_VBUS_EN, OUTPUT);
+  digitalWrite(DEV_VBUS_EN, HIGH);
+  pinMode(LIMIT_EN, OUTPUT);
+  digitalWrite(LIMIT_EN, HIGH);
+  delay(10);
+#endif
+
+  tinyusb_host_config_t host_config = {
+    .rhport = 0, /* P4 remaps 0 → HS rhport 1 inside tinyusb_host_init() */
+  };
```
me-no-dev: what is the purpose of providing host config if it will be overwritten later? Is there a case for chips to use rhport 1 (except P4)?

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+  _started = true;
+
+  if (s_tuh_worker == nullptr) {
+    /* Above Arduino loop priority so MSC/FatFs waits still make progress. */
+#if defined(CONFIG_FREERTOS_UNICORE) && CONFIG_FREERTOS_UNICORE
+    const BaseType_t ok = xTaskCreate(arduino_usb_host_tuh_worker, "usbhTuh", 8192, nullptr, 17, &s_tuh_worker);
+#else
+    const BaseType_t ok =
+      xTaskCreatePinnedToCore(arduino_usb_host_tuh_worker, "usbhTuh", 8192, nullptr, 17, &s_tuh_worker, 0);
```
me-no-dev: core should be selectable somewhere. Think of using `xTaskCreateUniversal` that comes with Arduino.

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return true;
+}
+
+void USBHostClass::task() {
+  if (!_started) {
+    return;
+  }
+  if (s_tuh_worker == nullptr) {
```
me-no-dev: NULL instead

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+ *     (Some mice send reports very often; that is normal.)
+ *
+ * Like TinyUSB’s host demo "cdc_msc_hid", but using the Arduino USB library classes.
+ */
+
+#include <Arduino.h>
+#include <USBHost.h>
+#include <USBHostSerial.h>
+#include <USBHostMSC.h>
+#include <USBMSCFS.h>
```
me-no-dev: why separate HostMSC and MSCFS? Shouldn't MSCFS be included by HostMSC?

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+#include <FS.h>
+#include <USBHostHID.h>
+#include <USBHostHIDMouse.h>
+#include <USBHostHIDKeyboard.h>
+#include <USBHostHIDKeyboardDecode.h>
```
me-no-dev: Similar question about those two

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * Like TinyUSB’s host demo "cdc_msc_hid", but using the Arduino USB library classes.
+ */
+
+#include <Arduino.h>
+#include <USBHost.h>
+#include <USBHostSerial.h>
+#include <USBHostMSC.h>
+#include <USBMSCFS.h>
+#include <FS.h>
```
me-no-dev: FS maybe should be included by MSCFS?

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+  for (int i = 0; i < 6 && !any; i++) {
+    any = (keys[i] != 0);
+  }
+  if (!any) {
+    USBHostKeyboard.clear();
+    return;
+  }
+
+  char ascii[8];
+  usbHostHidBootReportAppendAscii(ascii, sizeof(ascii), modifiers, keys, KeyboardLayout_en_US);
```
me-no-dev: maybe this method should be part of USBHostKeyboard class and have simpler name?

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+  bool any = modifiers != 0;
+  for (int i = 0; i < 6 && !any; i++) {
+    any = (keys[i] != 0);
+  }
+  if (!any) {
+    USBHostKeyboard.clear();
+    return;
+  }
+
+  char ascii[8];
```
me-no-dev: What happens with multiple modifiers and 6 keys?

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
```diff
@@ -0,0 +1,182 @@
...
+    s_usb_mount_attempted = false;
+    if (s_usb_stick_mounted) {
+      USBMSCFS.end();
+      s_usb_stick_mounted = false;
+      Serial.println(F("[msc] USB drive removed."));
+    }
+  }
+
+  // --- HID: mouse (simple print) ---
+  if (USBHostMouse.mounted() && USBHostMouse.available()) {
```
me-no-dev: why not use callback on change, like the keyboard?

libraries/USB/src/USBHostHIDGamepad.cpp
```diff
@@ -0,0 +1,327 @@
...
+#if CFG_TUH_HID
+
+#include <string.h>
+#include "tusb.h"
+#include "class/hid/hid.h"
+#include "Arduino.h"
+#include "esp32-hal-log.h"
+
+/** True if collection looks like a Generic Desktop Mouse (usage 0x02). */
+static bool desc_is_desktop_mouse(const uint8_t *d, uint16_t len) {
```
me-no-dev, line 36: these helpers are not pretty. Find a header with HID report value definitions and replace the hex values with those.

cores/esp32/esp32-hal-tinyusb.c
P-R-O-C-H-Y: @me-no-dev Yes, should I target 4.0 branch instead master and prepare the S31?

me-no-dev: There is nothing stopping you to add the define check in master. It will never be true in it, but that is about it. Keep targeting master

cores/esp32/USBHost.cpp
P-R-O-C-H-Y: done

cores/esp32/USBHost.cpp
P-R-O-C-H-Y: Fixed, now its a universal solution with a variant overwrite of weak method.

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
P-R-O-C-H-Y: So the worst case is 6 characters + NUL. ascii[8] is enough, and the helper also stops if w + 1 >= buf_cap, so it will not overflow even with a smaller buffer. A chord like Ctrl+Shift+A is still a single character if the layout matches.

libraries/USB/examples/USBHostCDC_MSC_HID/USBHostCDC_MSC_HID.ino
P-R-O-C-H-Y: done

libraries/USB/src/USBHostHIDGamepad.cpp
P-R-O-C-H-Y: done

cores/esp32/USBHost.cpp
P-R-O-C-H-Y: Agreed, the config is not a real API today. `USBHost.begin()` always passes rhport = 0, and P4 remaps that to 1 so TinyUSB hits OTG HS (_dwc2_controller[1]). The PHY is already forced to UTMI/HS, so the field cannot select P4 FS. No other current chip uses TinyUSB rhport 1. S2/S3 stay on 0. S31 is UTMI-only but that controller is rhport 0 in TinyUSB, not 1.

Dropping `tinyusb_host_config_t`. Default stays: S2/S3 → 0, P4 → 1.

cores/esp32/esp32-hal-tinyusb.c
P-R-O-C-H-Y: Done

cores/esp32/esp32-hal-tinyusb.c
```diff
@@ -176,7 +229,10 @@ void deinit_usb_hal() {
 #endif
 
 esp_err_t tinyusb_driver_install(const tinyusb_config_t *config) {
-  init_usb_hal(config->external_phy);
+  esp_err_t hal_err = init_usb_hal(config->external_phy);
+  if (hal_err != ESP_OK) {
+    return hal_err;
+  }
```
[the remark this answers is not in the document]
P-R-O-C-H-Y: Both live PHY paths return `esp_err_t`. The void stubs sit after #error No way to initialize USP PHY, so they never compile.

libraries/USB/src/USBHostHIDKeyboard.cpp
```diff
@@ -0,0 +1,165 @@
...
+  (void)dev_addr;
+  (void)idx;
+  const uint8_t *boot = report;
+  uint16_t boot_len = len;
+
+  /* Many keyboards prefix the 8-byte boot report with a report ID. */
+  if (len > 8u) {
+    boot = report + 1;
+    boot_len = len - 1u;
+  }
```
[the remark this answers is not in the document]
P-R-O-C-H-Y: Done

libraries/USB/src/USBHIDReportMapParse.c
```diff
@@ -0,0 +1,459 @@
...
+} s_parse_step_t;
+
+static s_parse_step_t s_parse_step = PARSE_WAIT_USAGE_PAGE;
+static uint8_t s_collection_depth = 0;
+static hid_report_params_t s_report_params = {0,};
+static uint16_t s_report_size = 0;
+static uint16_t s_report_count = 0;
+
+static bool s_new_map = false;
+static temp_hid_report_map_t *s_temp_hid_report_map;
```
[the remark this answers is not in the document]
P-R-O-C-H-Y: Agreed it is not reentrant (same as the ESP-IDF source). In this PR the only caller is the optional descriptor dumper, from the host worker, one claim at a time. I will leave the parser as-is and treat a stack/heap parse context as a follow-up if we use it from more than one task.

me-no-dev (approved):

cores/esp32/USBHost.cpp
```diff
@@ -0,0 +1,162 @@
...
+      log_w("[USBHost] worker resumed after stall (iter=%u)", (unsigned)iter);
+    }
+    last_iter = iter;
+    last_change_ms = now;
+    return;
+  }
+  if (!stalled && (now - last_change_ms) > 1000u) {
+    stalled = true;
+    const uint8_t phase = s_tuh_phase;
+    const char *where = (phase == TUH_PHASE_TASK) ? "tuh_task" : (phase == TUH_PHASE_HID) ? "hid service" : "idle";
```
me-no-dev: You need to fix [this](https://github.com/espressif/arduino-esp32/actions/runs/33175568561/job/98871809578?pr=12837#step:8:11819) by guarding it only if log level is above error (or just have it directly in the log_e command)

P-R-O-C-H-Y: Fixed all warnings. Thanks for checking that!

P-R-O-C-H-Y:
PR is ready and in final state. Only needs a new libs with fixes on TinyUSB side -> https://github.com/hathach/tinyusb/pull/3864

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+  if (_pdrv == 0xFF || buffer == nullptr) {
+    return false;
+  }
+  return ff_usbmsc_write(_pdrv, buffer, sector, 1u) == RES_OK;
+}
+
+}  // namespace fs
+
+#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_USB_MSC_FS)
+fs::USBMSCFS USBMSCFS(FSImplPtr(new VFSImpl()));
```
lucasssvaz, line 592: Seven new globals (USBHostHID.cpp:503, USBHostHIDMouse.cpp:361, USBHostHIDKeyboard.cpp:422, USBHostHIDGamepad.cpp:334, USBHostMSC.cpp:469, USBHostSerial.cpp:253, and this one) each emit a static-init root. arduino-cli links every object file of a library the sketch touches, so --gc-sections can't drop them.

I've re-measured locally on the device-side Gamepad example against master with identical libs: 365,618 → 390,110 B flash, 54,680 → 55,272 B RAM. This global causes the most impact as it drags VFSImpl, VFSFileImpl, FatFs and VolToPart in behind it. The existing device-side USB library defines no globals at all. Users write `USBHIDGamepad Gamepad;` in the sketch.

It would be ideal if the host libs didn't increase the device-side sketches sizes. If possible, move the host classes to their own libraries/USBHost/. A device sketch then never references the library, so arduino-cli won't compile or link it.

libraries/USB/src/USBHIDReportMapParse.c
```diff
@@ -0,0 +1,459 @@
...
+        } else if (cmd->cmd == USBHID_RM_REPORT_COUNT) {
+            s_report_count = cmd->value;
+        } else if (cmd->cmd == USBHID_RM_INPUT) {
+            s_report_params.input_len += (s_report_size * s_report_count);
+        } else if (cmd->cmd == USBHID_RM_OUTPUT) {
+            s_report_params.output_len += (s_report_size * s_report_count);
+        } else if (cmd->cmd == USBHID_RM_FEATURE) {
+            s_report_params.feature_len += (s_report_size * s_report_count);
+        } else if (cmd->cmd == USBHID_RM_COLLECTION) {
+            s_collection_depth += 1;
```
lucasssvaz, line 308: `s_report_size` and `s_report_count` are both `uint16_t` filled from device bytes (`memcpy(cmd->data, dp + 1, cmd->len)` at line 220), so each can be `0xFFFF`. The product is computed in `int` and `0xFFFF * 0xFFFF` overflows. Accumulating in `uint32_t` and rejecting anything over `UINT16_MAX` keeps the existing field widths, which is the right ceiling since these are bit counts. The COLLECTION branch is folded in because `s_collection_depth` is `uint8_t` and wraps at 256 nested collections. Far less likely, but it's the same edit.

```suggestion
        } else if (cmd->cmd == USBHID_RM_INPUT || cmd->cmd == USBHID_RM_OUTPUT || cmd->cmd == USBHID_RM_FEATURE) {
            uint16_t *acc = (cmd->cmd == USBHID_RM_INPUT)  ? &s_report_params.input_len
                          : (cmd->cmd == USBHID_RM_OUTPUT) ? &s_report_params.output_len
                                                           : &s_report_params.feature_len;
            /* size and count are both device-controlled and reach 0xFFFF, so the
               uint16_t product promotes to int and overflows. Accumulate in 32 bits. */
            const uint32_t total = (uint32_t)*acc + (uint32_t)s_report_size * (uint32_t)s_report_count;
            if (total > UINT16_MAX) {
                USBHID_PARSE_LOGE("report bit count overflow: size=%u count=%u", (unsigned)s_report_size, (unsigned)s_report_count);
                s_parse_step = PARSE_WAIT_USAGE_PAGE;
                return -1;
            }
            *acc = (uint16_t)total;
        } else if (cmd->cmd == USBHID_RM_COLLECTION) {
            if (s_collection_depth == UINT8_MAX) {
                USBHID_PARSE_LOGE("collection nesting too deep");
                s_parse_step = PARSE_WAIT_USAGE_PAGE;
                return -1;
            }
            s_collection_depth += 1;
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+  end();
+}
+
+static void usbmsc_begin_cleanup(const char *mountpoint, uint8_t pdrv, bool vfs_registered) {
+  if (vfs_registered && mountpoint != nullptr) {
+    esp_vfs_fat_unregister_path(mountpoint);
+  }
+  ff_diskio_register(pdrv, NULL);
+  usbmsc_free_slot(pdrv);
+}
```
lucasssvaz, line 389: `f_mount()` publishes `FatFs[vol] = fs` and, with `FF_FS_REENTRANT = 1`, creates the
per-volume mutex before it attempts the mount, so both survive a failure.
`esp_vfs_fat_unregister_path()` then frees the `FATFS` while the global pointer still
references it, and the mutex is leaked, because only `f_mount(NULL, drv, 0)` destroys
it. That's one leak per failed `begin()`, and `begin()` fails routinely on an
unformatted or unsupported drive.

`end()` at lines 510-514 already does this in the right order, so this just makes the
failure path match. The same fix covers all three callers at 476, 488 and 493. The one
at 449 passes `vfs_registered == false` and correctly skips it.

```suggestion
static void usbmsc_begin_cleanup(const char *mountpoint, uint8_t pdrv, bool vfs_registered) {
  if (vfs_registered) {
    /* f_mount() publishes FatFs[vol] and creates the per-volume mutex before the
     * mount is attempted, so both outlive a failure — and esp_vfs_fat_unregister_path()
     * would free the FATFS out from under the still-registered pointer. */
    char drv[3] = {(char)('0' + pdrv), ':', 0};
    (void)f_mount(NULL, drv, 0);
    if (mountpoint != nullptr) {
      esp_vfs_fat_unregister_path(mountpoint);
    }
  }
  ff_diskio_register(pdrv, NULL);
  usbmsc_free_slot(pdrv);
}
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+        if (lba >= bcount || !USBHostMSC.readBlocks(lba, sec, 1)) {
+          break;
+        }
+      }
+      if (memcmp(sec + ofs, kGuidMsBasicData, 16) != 0) {
+        continue;
+      }
+      const uint64_t first = ld_u64(sec + ofs + 32);
+      const uint64_t last = ld_u64(sec + ofs + 40);
+      if (first == 0 || first >= bcount || last < first) {
```
lucasssvaz, line 205: `last` is only checked against `first`, never against the device's block count, so a bogus GPT entry reaches `*out_gpt_count = (uint32_t)(last - first + 1)` at line 221 and truncates a 64-bit value. Today that only feeds the synthetic MBR's size field, which FatFs ignores, plus a log line, so this is hardening rather than a live bug. 

Still worth it: it's untrusted input flowing into an on-disk structure, and the adjacent first check already establishes the pattern.

```suggestion
      if (first == 0 || first >= bcount || last < first || last >= bcount) {
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+}
+
+static DRESULT ff_usbmsc_read(unsigned char pdrv, unsigned char *buff, uint32_t sector, unsigned count) {
+  if (pdrv >= FF_VOLUMES || s_usbmsc_slots[pdrv] == nullptr || buff == nullptr || count == 0) {
+    return RES_PARERR;
+  }
+  if (!USBHostMSC.mounted()) {
+    return RES_NOTRDY;
+  }
+
```
lucasssvaz, line 277: `sector` and `count` arrive from FatFs, driven by on-disk metadata the device supplies, and go straight to `readBlocks`/`writeBlocks` with no bounds check. A corrupt or hostile BPB can point past the end of the medium. Most drives will reject it, but relying on that puts the safety check on the far side of the untrusted boundary. Returning `RES_PARERR` also gives FatFs a meaningful error instead of a generic RES_ERROR. Placing the check ahead of the synthetic-MBR branch covers that path too, since it reads LBA 1..count-1, a subrange of what's just been validated.

```suggestion
const uint32_t bcount = USBHostMSC.blockCount();
if (sector >= bcount || (uint32_t)count > bcount - sector) {
  log_e("[USBMSCFS] disk_read out of range sector=%" PRIu32 " count=%u blocks=%" PRIu32, sector, (unsigned)count, bcount);
  return RES_PARERR;
}
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+}
+
+static DRESULT ff_usbmsc_write(unsigned char pdrv, const unsigned char *buff, uint32_t sector, unsigned count) {
+  if (pdrv >= FF_VOLUMES || s_usbmsc_slots[pdrv] == nullptr || buff == nullptr || count == 0) {
+    return RES_PARERR;
+  }
+  if (!USBHostMSC.mounted()) {
+    return RES_NOTRDY;
+  }
+
```
lucasssvaz, line 311: Same as previous

```suggestion
const uint32_t bcount = USBHostMSC.blockCount();
if (sector >= bcount || (uint32_t)count > bcount - sector) {
  log_e("[USBMSCFS] disk_write out of range sector=%" PRIu32 " count=%u blocks=%" PRIu32, sector, (unsigned)count, bcount);
  return RES_PARERR;
}
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+#if FF_FS_EXFAT
+      log_e("USBMSCFS: f_mount failed (13 FR_NO_FILESYSTEM) — not FAT32/FAT16/exFAT, or bad sector read");
+#else
+      log_e("USBMSCFS: f_mount failed (13 FR_NO_FILESYSTEM) — use FAT32/FAT16 (exFAT disabled), or check USB read/cache");
+#endif
+    } else {
+      log_e("USBMSCFS: f_mount failed (%d)", (int)res);
+    }
+
+    if (res == FR_NO_FILESYSTEM && format_if_empty) {
```
lucasssvaz, line 472: On a GPT disk `use_synth_mbr` is set, so `ff_usbmsc_write` silently discards any write
to LBA 0 (lines 318-326). `f_mkfs` writes its partition table there and its VBR at the
synthetic offset, so the format half-succeeds. The GPT header and entries survive but
now describe a filesystem that no longer matches, and `f_mkfs` returns `FR_OK`.

This is an opt-in path, since `format_if_empty` defaults to `false`, but the consequence
is data loss on a disk the user expected to be left alone. The `else if` makes the
refusal explicit rather than leaving a confusing partial result.

```suggestion
    if (res == FR_NO_FILESYSTEM && format_if_empty && gpt_lba == 0) {
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+      if (res == FR_OK) {
+        res = f_mount(fs, drv, 1);
+      }
+      if (res != FR_OK) {
+        log_e("USBMSCFS: format/remount failed (%d)", (int)res);
+        usbmsc_begin_cleanup(mountpoint, pdrv, true);
+        _pdrv = 0xFF;
+        return false;
+      }
+    } else {
```
lucasssvaz, line 492: See previous for explanation.

```suggestion
    } else if (res == FR_NO_FILESYSTEM && format_if_empty) {
          log_e("USBMSCFS: refusing to format a GPT disk. f_mkfs would write through the synthetic MBR and destroy the GPT header");
          usbmsc_begin_cleanup(mountpoint, pdrv, true);
          _pdrv = 0xFF;
          return false;
    } else {
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+#else
+                  * 512;
+#endif
+  return size;
+}
+
+bool USBMSCFS::readRAW(uint8_t *buffer, uint32_t sector) {
+  if (_pdrv == 0xFF || buffer == nullptr) {
+    return false;
+  }
```
lucasssvaz, line 578: These are new public API taking a bare `uint8_t *` with no length, while
`ff_usbmsc_read` memsets `blockSize() * count` bytes on the synthetic-MBR path. On a
4K-sector drive that's a 3.5 KB overflow of a caller's 512-byte buffer, and the caller
has no way to signal capacity.

Either add a `size_t buffer_len` parameter or reject non-512 sectors. The guard is the
cheaper option, and worth a matching note on `USBMSCFS.h:54-55`. FatFs itself is safe
here, because `FF_MIN_SS != FF_MAX_SS`, so `mount_volume()` validates `GET_SECTOR_SIZE`
before allocating its window.

```suggestion
  }
  /* The synthetic-MBR path memsets blockSize() bytes and callers cannot signal capacity. */
  if (USBHostMSC.blockSize() != 512) {
    log_e("[USBMSCFS] readRAW needs a 512-byte sector, device reports %" PRIu32, USBHostMSC.blockSize());
    return false;
  }
```

libraries/USB/src/USBMSCFS.cpp
```diff
@@ -0,0 +1,595 @@
...
+  return ff_usbmsc_read(_pdrv, buffer, sector, 1u) == RES_OK;
+}
+
+bool USBMSCFS::writeRAW(uint8_t *buffer, uint32_t sector) {
+  if (_pdrv == 0xFF || buffer == nullptr) {
+    return false;
+  }
```
lucasssvaz, line 585: Same as above

```suggestion
  }
  /* The synthetic-MBR path memsets blockSize() bytes and callers cannot signal capacity. */
  if (USBHostMSC.blockSize() != 512) {
    log_e("[USBMSCFS] writeRAW needs a 512-byte sector, device reports %" PRIu32, USBHostMSC.blockSize());
    return false;
  }
```

libraries/USB/src/USBHostSerial.cpp
```diff
@@ -0,0 +1,276 @@
...
+    } else {
+      if (!USBHost.tuhBackgroundActive()) {
+        tuh_task();
+      } else {
+        yield();
+      }
+    }
+  }
+
+  xSemaphoreGive(s_usb_host_serial_tx_mutex);
```
lucasssvaz, line 186: Four things in this loop.

**`setTxTimeoutMs(0)` returns 0 forever.** The deadline is tested before the first
`tuh_cdc_write`, so with a 0 timeout `deadline == millis()`, the check passes
immediately, and nothing is ever written. Moving the first write ahead of the check
gives 0 the usual "try once" meaning, as in `USBCDC.cpp:409`.

**`tuh_task()` never returns on an idle bus**, since it's `tuh_task_ext(UINT32_MAX,
false)`. This is only reachable when `xTaskCreateUniversal` failed and the code fell
back to pumping from `loop()`, but `USBHost.cpp:137-138` already documents and avoids
exactly this.

**`mounted()` is never rechecked**, so unplug mid-write spins on a stale `_cdc_idx` for
the full 250 ms default.

**The flush inside the success branch** makes `write(uint8_t)` one bulk transfer per
character. Moving it after the loop keeps the FIFO-full flush, which is what makes room,
while batching the common case.

One judgement call in the suggested version. The `continue` skips the deadline while
progress is being made, so a device dribbling bytes can exceed `_tx_timeout_ms`. I think
that's the better semantic, since a deadline shouldn't truncate a write that's
succeeding, but say if you'd rather bound the call duration instead.

```suggestion
  const bool non_blocking = (_tx_timeout_ms == 0);
  size_t so_far = 0;
  uint32_t deadline = millis() + _tx_timeout_ms;
  while (so_far < size) {
    uint32_t n = tuh_cdc_write(_cdc_idx, buffer + so_far, (uint32_t)(size - so_far));
    if (n) {
      /* Flush once after the loop, not per chunk. Otherwise write(uint8_t) is
       * one bulk transfer per character. */
      so_far += n;
      continue;
    }
    /* FIFO full or endpoint busy: nothing more fits until it drains. */
    if (non_blocking || (int32_t)(millis() - deadline) >= 0) {
      break;
    }
    /* The device can vanish mid-write, leaving _cdc_idx stale. */
    if (!mounted()) {
      break;
    }
    (void)tuh_cdc_write_flush(_cdc_idx);
    if (!USBHost.tuhBackgroundActive()) {
      /* Loop fallback must not block: tuh_task() waits on the event queue forever. */
      tuh_task_ext(0, false);
    } else {
      yield();
    }
  }


  if (so_far) {
    (void)tuh_cdc_write_flush(_cdc_idx);
  }
  xSemaphoreGive(s_usb_host_serial_tx_mutex);
```

libraries/USB/src/USBHostSerial.cpp
```diff
@@ -0,0 +1,276 @@
...
+  _dev_addr = info.daddr;
+  _itf_num = info.desc.bInterfaceNumber;
+  _binding_valid = true;
+  _mounted = true;
+
+  applyBeginLineCodingIfNeeded();
+}
+
+void USBHostSerialClass::onCdcUnmount(uint8_t idx) {
+  (void)idx;
```
lucasssvaz, line 246: Not a live bug, since `CFG_TUH_CDC` is 1 on both targets and `cdc_host.c` indexes a flat
`[CFG_TUH_CDC]` array, so `idx` is always 0 and only one CDC interface is ever claimed.
Filing it as hygiene rather than a defect.
`(void)idx;` means any CDC interface unmounting would tear down this object's binding,
even one it never owned, and `onCdcMount` at line 236 has the mirror problem in that it
rebinds to whichever interface mounted last. Both become real the moment someone
rebuilds with `CFG_TUH_CDC > 1`, which the header at line 35 contemplates with "prebuilt
often 1", and neither would fail loudly. A two-line guard here plus an early return in
`onCdcMount` when `_binding_valid` is already set would make the single-binding
assumption explicit instead of implicit:

```suggestion
  if (!_binding_valid || idx != _cdc_idx) {
    return;
  }
```

Alternatively, if the singleton is meant to stay single-binding permanently, a
`static_assert(CFG_TUH_CDC == 1)` would document that better than the current silent
dependency.

libraries/USB/src/USBHostMSC.cpp
```diff
@@ -0,0 +1,483 @@
...
+static bool wait_scsi(uint32_t timeout_ms, const char *op_tag) {
+  ensure_scsi_sem();
+  const uint32_t start = millis();
+  while ((millis() - start) <= timeout_ms) {
+    const TickType_t wait = USBHost.tuhBackgroundActive() ? pdMS_TO_TICKS(100) : 0;
+    if (xSemaphoreTake(s_scsi_done_sem, wait) == pdTRUE) {
+      return s_scsi_ok;
+    }
+    if (!USBHost.tuhBackgroundActive()) {
+      tuh_task();
```
lucasssvaz, line 209: `pump_tuh()` already exists at line 109 for precisely this, and its comment says why:
*"Non-blocking: tuh_task() would wait on the event queue past our timeout."* It's used
at 120 and 145, so this site looks like an oversight.

As written, the `timeout_ms` parameter is unenforceable here. An idle bus parks inside
`tuh_task()` indefinitely, so the timeout log below is unreachable.

```suggestion
      pump_tuh();
```

libraries/USB/src/USBHostMSC.cpp
```diff
@@ -0,0 +1,483 @@
...
+}
+
+USBHostMSCClass::USBHostMSCClass() : _mounted(false), _dev_addr(0), _lun(0), _itf_num(0), _ep_in(0), _ep_out(0), _block_count(0), _block_size(0) {}
+
+bool USBHostMSCClass::cacheEndpoints(uint8_t dev_addr) {
+  _itf_num = 0;
+  _ep_in = 0;
+  _ep_out = 0;
+
+  uint8_t cfg[256];
```
lucasssvaz, line 227: `tuh_descriptor_get_configuration_sync()` returns only a `tusb_xfer_result_t`, with no
way to learn how many bytes actually arrived, so a short transfer leaves the tail of
`cfg` as stack garbage. The `total > sizeof(cfg)` clamp at lines 234-236 bounds reads to
the array but not to the bytes received, since `wTotalLength` is device-supplied.

Zeroing makes a truncated descriptor terminate the walk at the `len < 2` check on line
243 instead of parsing residue as descriptors. This is an error path only, reached from
`recoverBot()` at line 298, hence the cheap fix rather than reworking the API.

```suggestion
  uint8_t cfg[256];
  memset(cfg, 0, sizeof(cfg));
```

libraries/USB/src/USBHostMSC.h
```diff
@@ -0,0 +1,85 @@
...
+
+private:
+  bool cacheEndpoints(uint8_t dev_addr);
+  bool recoverBot(const char *reason);
+  bool xferBlocks(bool is_write, uint32_t lba, void *buffer, uint32_t blocks);
+
+  volatile bool _mounted;
+  uint8_t _dev_addr;
+  uint8_t _lun;
+  uint8_t _itf_num;
```
lucasssvaz, line 72: These are written together with `_mounted` and might be cached.

```suggestion
  volatile uint8_t _dev_addr;
  volatile uint8_t _lun;
  volatile uint8_t _itf_num;
```

lucasssvaz (changes requested):

libraries/USB/src/USBMSCFS.cpp
P-R-O-C-H-Y: Taken as suggested.

libraries/USB/src/USBMSCFS.cpp
P-R-O-C-H-Y: Taken as suggested.

libraries/USB/src/USBMSCFS.cpp
P-R-O-C-H-Y: Taken as suggested.

libraries/USB/src/USBHostMSC.cpp
P-R-O-C-H-Y: Taken as suggested.

libraries/USB/src/USBMSCFS.cpp
P-R-O-C-H-Y: Taken as suggested.

libraries/USB/src/USBMSCFS.cpp
P-R-O-C-H-Y: Taken, factored into a shared usbmsc_raw_sector_ok() helper rather than
duplicating the check in both functions. Worth noting the plain readBlocks()
path has the same problem, not just the synthetic-MBR memset — a 4K sector
overruns the caller's buffer either way, so the guard covers both. Added the
note on the declarations as you suggested.

libraries/USB/src/USBHostSerial.cpp
P-R-O-C-H-Y: All four taken, including `tuh_task_ext(0, false)` for the loop fallback.
Kept your semantic. A deadline that truncates a write
which is actively succeeding would be the more surprising behaviour, and
`setTxTimeoutMs()` reads as "how long to wait for a stalled endpoint", not "how
long this call may take". A caller that needs a hard bound has `setTxTimeoutMs(0)`
plus its own loop.

P-R-O-C-H-Y:
Thanks @lucasssvaz for a great review, code improved on many places.
PR is now ready in final state.

lucasssvaz (approved):
LGTM. Great work

pedrominatel (approved):
LGTM! Thanks, @P-R-O-C-H-Y.

PR #12019, opened 2025-11-12
feat(esp32c61): Add support for ESP32-C61

lucasssvaz:
## Description of Change

This pull request adds support for the ESP32-C61 target across the Arduino-ESP32 codebase. The changes ensure that ESP32-C61 is correctly handled in build scripts, documentation, hardware abstraction, and board-specific logic. Additionally, the CPU frequency management code is refactored for better maintainability and extensibility.

**ESP32-C61 Support**

* Added ESP32-C61 to the list of supported targets in build scripts (`.github/scripts/find_all_boards.sh`, `.github/workflows/build_component.yml`) and ensured it is properly skipped where required for library builds. [[1]](diffhunk://#diff-01ca9570a2d7246f395f0c6fbf5fb3d0df01500eae746ae7d8b1f886acb5896cL11-R11) [[2]](diffhunk://#diff-10d3050b713b6a8272bf34026632c4cef626187808d28d4a54cb7543d1b87794L13-R13) [[3]](diffhunk://#diff-10d3050b713b6a8272bf34026632c4cef626187808d28d4a54cb7543d1b87794R40) [[4]](diffhunk://#diff-10d3050b713b6a8272bf34026632c4cef626187808d28d4a54cb7543d1b87794L122-R129)
* Updated documentation (`README.md`) to mention ESP32-C61 support and its requirements.
* Added ESP32-C61-specific hardware definitions for UART pins in `HardwareSerial.h`, and updated related logic for RX/TX pin mappings. [[1]](diffhunk://#diff-ba8df4ae3e43ea0d34307f26d6def560d7c458d88bf6fb2ddd315a94a60dbdaeL153-R188) [[2]](diffhunk://#diff-ba8df4ae3e43ea0d34307f26d6def560d7c458d88bf6fb2ddd315a94a60dbdaeL196-R231)
* Added ESP32-C61-specific handling for flash chip speed and mode in `Esp.cpp`, including magic flash frequency mapping and register usage. [[1]](diffhunk://#diff-0426ace2f3a46b5fe88c9d3936aad8464fa773fd718f739df3acfb6a555da10dL60-R88) [[2]](diffhunk://#diff-0426ace2f3a46b5fe88c9d3936aad8464fa773fd718f739df3acfb6a555da10dL368-R371) [[3]](diffhunk://#diff-0426ace2f3a46b5fe88c9d3936aad8464fa773fd718f739df3acfb6a555da10dR455-R470)
[...]

PR #12838, opened 2026-08-18
fix(cpu): Fix memory leak in getSupportedCpuFrequencyMhz

richardsonnick:
## Description of Change
Minor change.
`getSupportedCpuFrequencyMhz()` allocated a 256-byte buffer with `calloc()` on every call and only freed it on the `#else`/"Unknown" branch. Leaking on every other call. The caller (`setCpuFrequencyMhz()`) never frees the result. The function is not referenced anywhere else in the codebase.

 This PR replaces the heap allocation with a `static char buf[256]`, so no allocation or free is needed. The public signature `const char *getSupportedCpuFrequencyMhz(uint8_t xtal)` is unchanged.
 
 Note that this change is not threadsafe. I figure that this is okay given that this is only called in` setCpuFrequencyMhz()` and is only for logging. 

## Test Scenarios
Compiled against ESP-IDF v6.0.2 for the ESP32 target and flashed to a physical ESP32 where the firmware boots and runs successfully.

## Related links
The function was introduced in https://github.com/espressif/arduino-esp32/pull/12019

### DIFF
--- a/cores/esp32/esp32-hal-cpu.c
+++ b/cores/esp32/esp32-hal-cpu.c
@@ -206,7 +206,7 @@ const char *getClockSourceName(uint8_t source) {
 }
 
 const char *getSupportedCpuFrequencyMhz(uint8_t xtal) {
-  const size_t buf_size = 256;
+  constexpr size_t buf_size = 256;
   static char buf[buf_size];
   int pos = 0;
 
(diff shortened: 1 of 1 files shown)


cores/esp32/esp32-hal-cpu.c
```diff
@@ -206,50 +206,49 @@ const char *getClockSourceName(uint8_t source) {
 }
 
 const char *getSupportedCpuFrequencyMhz(uint8_t xtal) {
-  char *supported_frequencies = (char *)calloc(256, sizeof(char));
+  static char buf[256];
```
me-no-dev: instead of using `sizeof(buf)` constantly, add a constant size definition and use that instead:
```cpp
const size_t buf_size = 256;
static char buf[buf_size];
//.......
pos += snprintf(buf + pos, buf_size - pos, "360");
```

me-no-dev (approved):

PR #12848, opened 2026-08-23
feat(httpupdate): add optional checksum sidecar URLs

### LINKED ISSUE #12826: HTTPUpdate: optional checksum sidecar URL (MD5 / SHA-256)
### Related area

Libraries (HTTPUpdate / OTA)

### Hardware specification

All ESP32 targets. Software-only; no board-specific dependency.

HTTPUpdate can take an expected digest via `setMD5sum()` / `setSHA256sum()` (#12824) or the `x-MD5` / `x-SHA256` response headers. Many firmware hosts (GitHub Releases, simple file servers) cannot set those headers. The usual workaround is a sidecar next to the binary (`firmware.bin` + `firmware.md5` or `firmware.sha256`).

Every project then reimplements the same extra GET: download a small text file, parse the first hex token, call `setMD5sum()` / `setSHA256sum()`. That fetch is generic and belongs next to the existing checksum setters, not in each sketch.

This depends on the SHA-256 API in #12823 / #12824. Please land that first so this can use `setSHA256sum()` the same way as `setMD5sum()`.

Keep the current setters and headers as-is. Add optional sidecar URLs:

- `HTTPUpdate.setMD5sumUrl(const String &url)`
- `HTTPUpdate.setSHA256sumUrl(const String &url)`

Suggested precedence (same idea as today): explicit `setMD5sum()` / `setSHA256sum()` first, then response headers, then sidecar URL if the digest is still empty.

Behavior:

- GET the URL with the same client / redirect settings as the firmware download
- Read the first hex token (32 chars MD5, 64 chars SHA-256, case-insensitive)
- Hand it to the existing `Update.setMD5()` / `setSHA256()`
- On a bad or missing sidecar, fail with `HTTP_UE_SERVER_FAULTY_MD5` / `HTTP_UE_SERVER_FAULTY_SHA256`

No GitHub-specific paths. Sketch still passes the URL.

I can open a PR after #12824 is merged.

- Keep fetching the sidecar in application code. Works, but duplicates what HTTPUpdate already does for header-based checksums.
- Require hosts to send `x-MD5` / `x-SHA256`. Fine for custom servers; not possible on GitHub Releases.
- Auto-derive `firmware.bin` → `firmware.md5`. Too magical and wrong for hosts that use a different name or path.

### Additional context

Related:

- #12823 / #12824 (optional SHA-256 checksum API — prerequisite)
- #7629 (`setMD5sum`)

A sidecar URL is only an integrity check of the downloaded bytes, same threat model as `setMD5sum()` today. It is not code signing.

EyJunge1:
## Description of Change

Adds optional checksum sidecar URLs to `HTTPUpdate` for firmware hosts that cannot provide `x-MD5` or `x-SHA256` response headers.

- Adds `HTTPUpdate.setMD5sumUrl()` and `setSHA256sumUrl()`
- Accepts the first exact 32/64-character hexadecimal token
- Supports fixed-length, unknown-length, and chunked responses
- Reuses the firmware client transport and redirect configuration
- Keeps firmware HTTP authorization and request callbacks isolated from sidecar requests
- Preserves digest precedence: explicit setter > response header > sidecar
- Bounds checksum scanning and sidecar body wait time
- Supports existing `HTTPClient` overloads, including lazy compatibility clients
- Links the sidecar fetch implementation only when a sidecar setter is used

This provides generic support for common files such as:

- `firmware.bin` + `firmware.md5`
- `firmware.bin` + `firmware.sha256`

Sidecar checks provide integrity verification only; they are not code signing.

Fixes #12826

Related: #12824, #7629

## Test plan

- [x] Pre-commit hooks pass
- [x] CMake source-list validation passes
- [x] OTA validation sketch builds successfully for ESP32
- [x] Added coverage for MD5 and SHA-256 sidecars
- [x] Added precedence tests for explicit setters and response headers
- [x] Added malformed, missing, wrong, uppercase, redirect, and combined sidecar cases
- [x] Added fixed-length, unknown-length, chunked, stalled, large-body, and scan-boundary cases
- [x] Added compatibility-client, open-socket, copy-safety, and credential-isolation cases
- [x] Hardware OTA validation runs in CI

### DIFF
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -926,7 +926,7 @@ NetworkClient *HTTPClient::getClient(void) {
   // Reapply transport settings on every retrieval. Deprecated begin() calls can
   // replace the traits while retaining the lazily allocated client.
   if (_tcpDeprecated && !_transportTraits->verify(*_client, _host.c_str())) {
-    log_d("transport level verify failed");
+    log_e("transport level verify failed");
     _client->stop();
     _client = nullptr;
     _tcpDeprecated.reset(nullptr);
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -913,6 +913,27 @@ NetworkClient *HTTPClient::getStreamPtr(void) {
   return nullptr;
 }
 
+NetworkClient *HTTPClient::getClient(void) {
+#ifdef HTTPCLIENT_1_1_COMPATIBLE
+  if (_transportTraits && !_client) {
+    _tcpDeprecated = _transportTraits->create();
+    if (!_tcpDeprecated) {
+      log_e("could not create client");
+      return nullptr;
+    }
+    _client = _tcpDeprecated.get();
+    if (!_transportTraits->verify(*_client, _host.c_str())) {
+      log_d("transport level verify failed");
+      _client->stop();
+      _client = nullptr;
+      _tcpDeprecated.reset(nullptr);
+      return nullptr;
+    }
+  }
+#endif
+  return _client;
+}
+
 /**
  * write all  message body / payload to Stream
  * @param stream Stream *
@@ -1141,28 +1162,10 @@ bool HTTPClient::connect(void) {
     return true;
   }
 
-#ifdef HTTPCLIENT_1_1_COMPATIBLE
-  if (_transportTraits && !_client) {
-    _tcpDeprecated = _transportTraits->create();
-    if (!_tcpDeprecated) {
-      log_e("failed to create client");
-      return false;
-    }
-    _client = _tcpDeprecated.get();
-  }
-#endif
-
-  if (!_client) {
+  if (!getClient()) {
     log_d("HTTPClient::begin was not called or returned error");
     return false;
   }
-#ifdef HTTPCLIENT_1_1_COMPATIBLE
-  if (_tcpDeprecated && !_transportTraits->verify(*_client, _host.c_str())) {
-    log_d("transport level verify failed");
-    _client->stop();
-    return false;
-  }
-#endif
   if (!_client->connect(_host.c_str(), _port, _connectTimeout)) {
     log_d("failed connect to %s:%u", _host.c_str(), _port);
     return false;
--- a/libraries/HTTPClient/src/HTTPClient.cpp
+++ b/libraries/HTTPClient/src/HTTPClient.cpp
@@ -922,13 +922,15 @@ NetworkClient *HTTPClient::getClient(void) {
       return nullptr;
     }
     _client = _tcpDeprecated.get();
-    if (!_transportTraits->verify(*_client, _host.c_str())) {
-      log_d("transport level verify failed");
-      _client->stop();
-      _client = nullptr;
-      _tcpDeprecated.reset(nullptr);
-      return nullptr;
-    }
+  }
+  // Reapply transport settings on every retrieval. Deprecated begin() calls can
+  // replace the traits while retaining the lazily allocated client.
+  if (_tcpDeprecated && !_transportTraits->verify(*_client, _host.c_str())) {
+    log_d("transport level verify failed");
+    _client->stop();
+    _client = nullptr;
+    _tcpDeprecated.reset(nullptr);
+    return nullptr;
   }
 #endif
   return _client;
--- a/libraries/HTTPClient/src/HTTPClient.h
+++ b/libraries/HTTPClient/src/HTTPClient.h
@@ -170,6 +170,8 @@ typedef struct {
 } Cookie;
 typedef std::vector<Cookie> CookieJar;
 
+class HTTPUpdate;
+
 class HTTPClient {
 public:
   HTTPClient();
@@ -251,14 +253,6 @@ public:
 
   NetworkClient &getStream(void);
   NetworkClient *getStreamPtr(void);
-  /**
-   * @brief Return the NetworkClient set by begin(), even before connect().
-   *
-   * Unlike getStreamPtr(), this does not require an active TCP connection.
-   */
-  NetworkClient *getClient(void) {
-    return _client;
-  }
   int writeToStream(Stream *stream);
   String getString(void);
 
@@ -268,6 +262,10 @@ public:
   void setCookieJar(CookieJar *cookieJar);
   void resetCookieJar();
   void clearAllCookies();
+
+private:
+  friend class HTTPUpdate;
+  NetworkClient *getClient(void);
 
 protected:
   struct RequestArgument {
--- a/libraries/HTTPClient/src/HTTPClient.h
+++ b/libraries/HTTPClient/src/HTTPClient.h
@@ -251,6 +251,14 @@ public:
 
   NetworkClient &getStream(void);
   NetworkClient *getStreamPtr(void);
+  /**
+   * @brief Return the NetworkClient set by begin(), even before connect().
+   *
+   * Unlike getStreamPtr(), this does not require an active TCP connection.
+   */
+  NetworkClient *getClient(void) {
+    return _client;
+  }
   int writeToStream(Stream *stream);
   String getString(void);
 
--- a/libraries/HTTPClient/src/HTTPClient.h
+++ b/libraries/HTTPClient/src/HTTPClient.h
@@ -170,8 +170,6 @@ typedef struct {
 } Cookie;
 typedef std::vector<Cookie> CookieJar;
 
-class HTTPUpdate;
-
 class HTTPClient {
 public:
   HTTPClient();
@@ -221,6 +219,9 @@ public:
   // Redirections
   void setFollowRedirects(followRedirects_t follow);
   void setRedirectLimit(uint16_t limit);  // max redirects to follow for a single request
+  uint16_t getRedirectLimit(void) const {
+    return _redirectLimit;
+  }
 
   bool setURL(const String &url);
   void useHTTP10(bool usehttp10 = true);
@@ -253,6 +254,7 @@ public:
 
   NetworkClient &getStream(void);
   NetworkClient *getStreamPtr(void);
+  NetworkClient *getClient(void);
   int writeToStream(Stream *stream);
   String getString(void);
 
@@ -262,10 +264,6 @@ public:
   void setCookieJar(CookieJar *cookieJar);
   void resetCookieJar();
   void clearAllCookies();
-
-private:
-  friend class HTTPUpdate;
-  NetworkClient *getClient(void);
 
 protected:
   struct RequestArgument {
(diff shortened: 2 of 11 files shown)


libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,297 @@
...
+  HTTPClient http;
+  if (!http.begin(client, url)) {
+    return false;
+  }
+
+  // HTTP/1.0 disables keep-alive so the shared NetworkClient is free for the firmware GET.
+  http.useHTTP10(true);
+  http.setTimeout(timeout);
+  http.setFollowRedirects(follow);
+  http.setUserAgent("ESP32-http-Update");
```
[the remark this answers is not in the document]
EyJunge1: Removed the unnecessary sidecar-specific User-Agent call in c7f915707. The focused pre-commit checks and OTA validation build pass.

libraries/HTTPClient/src/HTTPClient.cpp
```diff
@@ -913,6 +913,29 @@ NetworkClient *HTTPClient::getStreamPtr(void) {
...
+    if (!_tcpDeprecated) {
+      log_e("could not create client");
+      return nullptr;
+    }
+    _client = _tcpDeprecated.get();
+  }
+  // Reapply transport settings on every retrieval. Deprecated begin() calls can
+  // replace the traits while retaining the lazily allocated client.
+  if (_tcpDeprecated && !_transportTraits->verify(*_client, _host.c_str())) {
+    log_d("transport level verify failed");
```
me-no-dev: why not log_e?

EyJunge1: Good point. Changed this to `log_e` in 0bc90bb53.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+  HTTPClient http;
+  if (!http.begin(client, url)) {
+    return false;
+  }
+
+  // HTTP/1.0 disables keep-alive so the shared NetworkClient is free for the firmware GET.
+  http.useHTTP10(true);
+  http.setTimeout(timeout);
+  http.setFollowRedirects(follow);
```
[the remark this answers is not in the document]
EyJunge1: Fixed in 0bc90bb53. The sidecar fetch now receives the firmware client’s configured redirect limit and applies it to its `HTTPClient`. I also added a regression test using redirect limit 0.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,296 @@
...
+  bool ok = chunked ? httpUpdateParseChunkedBody(*stream, parser, startedAt, timeoutMs)
+                    : httpUpdateParseIdentityBody(*stream, contentLength, parser, startedAt, timeoutMs);
+  http.end();
+
+  if (!ok) {
+    return false;
+  }
+
+  outDigest = parser.token();
+  return true;
```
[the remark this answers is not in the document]
EyJunge1: Fixed in 0bc90bb53. The fetch now succeeds only when the copied digest has the expected length, so an allocation failure is propagated as a sidecar failure and the update remains fail-closed.

libraries/HTTPClient/src/HTTPClient.h
```diff
@@ -170,6 +170,8 @@ typedef struct {
 } Cookie;
 typedef std::vector<Cookie> CookieJar;
 
+class HTTPUpdate;
```
lucasssvaz: `HTTPClient.h` now forward-declares `HTTPUpdate` and grants it friend access just so `HTTPUpdate::handleUpdate()` can reach the private `getClient()` and `_redirectLimit`. This means every `HTTPClient` user compiles against a dependency on `HTTPUpdate`, inverting the natural stack (`HTTPUpdate` depends on `HTTPClient`, not the other way around).

Suggested fix: make `getClient()` public (or add a public `ensureClient()`/`getRedirectLimit()` accessor) and drop the `friend class HTTPUpdate;` / forward declaration entirely.

```cpp
class HTTPClient {
public:
  // ...
  NetworkClient *getClient(void);   // was private + friended
  uint16_t getRedirectLimit(void) const { return _redirectLimit; }
  // ...
};
```

This keeps `HTTPClient` self-contained and reusable without knowledge of `HTTPUpdate`.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,301 @@
...
+  _md5SumUrl = url;
+  _checksumSidecarFetch = (!_md5SumUrl.isEmpty() || !_sha256SumUrl.isEmpty()) ? httpUpdateFetchChecksumSidecars : nullptr;
+}
+
+void HTTPUpdate::setSHA256sumUrl(const String &url) {
+  _sha256SumUrl = url;
+  _checksumSidecarFetch = (!_md5SumUrl.isEmpty() || !_sha256SumUrl.isEmpty()) ? httpUpdateFetchChecksumSidecars : nullptr;
+}
+
+static bool httpUpdateFetchChecksumSidecar(
```
lucasssvaz, line 228: Every failure path in `httpUpdateFetchChecksumSidecar()` (empty/invalid URL, `begin()` failure, non-200 response, request timeout, missing stream, chunked/identity parse failure, and digest-length mismatch) returns `false` with no logging. On a real device, a bad `setMD5sumUrl()`/`setSHA256sumUrl()` call just produces `HTTP_UE_SERVER_FAULTY_MD5`/`_SHA256` with no way to tell whether the sidecar 404'd, timed out, or was malformed.

Suggested fix: add a `log_e`/`log_d` at each return point with enough context to diagnose remotely

libraries/HTTPUpdate/src/HTTPUpdate.h
```diff
@@ -153,10 +167,18 @@ class HTTPUpdate {
...
   bool _rebootOnUpdate = true;
 
 private:
+  using ChecksumSidecarFetchFn = uint8_t (*)(
+    NetworkClient *client, const String &md5Url, const String &sha256Url, uint8_t requested, String &md5, String &sha256, int timeout, followRedirects_t follow,
+    uint16_t redirectLimit
+  );
+
+  static constexpr uint8_t SIDECAR_MD5_FAILED = 0x01;
+  static constexpr uint8_t SIDECAR_SHA256_FAILED = 0x02;
```
lucasssvaz: `SIDECAR_MD5_FAILED` / `SIDECAR_SHA256_FAILED` are defined once as `HTTPUpdate` private `static constexpr uint8_t` members (`HTTPUpdate.h:179-180`) and again independently as `HTTPUPDATE_SIDECAR_MD5_FAILED` / `HTTPUPDATE_SIDECAR_SHA256_FAILED` in `HTTPUpdateChecksum.cpp:32-33`. They're only compatible because both files happen to use `0x01`/`0x02`. Nothing enforces that if one changes.

Make these public and reuse them in `HTTPUpdateChecksum.cpp`

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,301 @@
...
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ */
+
+#include "HTTPUpdate.h"
+
+// Cap sidecar body scan so a hostile/oversized response cannot grow heap.
+static const size_t HTTPUPDATE_SIDECAR_MAX_SCAN = 512;
+static const uint8_t HTTPUPDATE_SIDECAR_MD5_FAILED = 0x01;
+static const uint8_t HTTPUPDATE_SIDECAR_SHA256_FAILED = 0x02;
```
lucasssvaz: Continuing steps from last suggestion

```suggestion
```

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
```diff
@@ -0,0 +1,301 @@
...
+  NetworkClient *client, const String &md5Url, const String &sha256Url, uint8_t requested, String &md5, String &sha256, int timeout, followRedirects_t follow,
+  uint16_t redirectLimit
+) {
+  uint8_t failures = 0;
+  if ((requested & HTTPUPDATE_SIDECAR_MD5_FAILED) && (!client || !httpUpdateFetchChecksumSidecar(*client, md5Url, 32, md5, timeout, follow, redirectLimit))) {
+    failures |= HTTPUPDATE_SIDECAR_MD5_FAILED;
+  }
+  if ((requested & HTTPUPDATE_SIDECAR_SHA256_FAILED)
+      && (!client || !httpUpdateFetchChecksumSidecar(*client, sha256Url, 64, sha256, timeout, follow, redirectLimit))) {
+    failures |= HTTPUPDATE_SIDECAR_SHA256_FAILED;
```
lucasssvaz: Continuing steps from last suggestion

```suggestion
  if ((requested & HTTPUpdate::SIDECAR_MD5_FAILED) && (!client || !httpUpdateFetchChecksumSidecar(*client, md5Url, 32, md5, timeout, follow, redirectLimit))) {
    failures |= HTTPUpdate::SIDECAR_MD5_FAILED;
  }
  if ((requested & HTTPUpdate::SIDECAR_SHA256_FAILED)
      && (!client || !httpUpdateFetchChecksumSidecar(*client, sha256Url, 64, sha256, timeout, follow, redirectLimit))) {
    failures |= HTTPUpdate::SIDECAR_SHA256_FAILED;
```

lucasssvaz (changes requested):

libraries/HTTPClient/src/HTTPClient.h
EyJunge1: Done in 4ffa63b. `getClient()` and `getRedirectLimit()` are public now, and the `HTTPUpdate` friend/forward declaration is gone.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
EyJunge1: Done in 4ffa63b. Each sidecar failure path now logs with enough context to tell a bad URL, begin failure, HTTP error, timeout, missing stream, parse error, or digest copy failure apart.

libraries/HTTPUpdate/src/HTTPUpdate.h
EyJunge1: Done in 4ffa63b. The flags are public on `HTTPUpdate` and reused from `HTTPUpdateChecksum.cpp`.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
EyJunge1: Applied in 4ffa63b. The local duplicate constants are gone.

libraries/HTTPUpdate/src/HTTPUpdateChecksum.cpp
EyJunge1: Applied in 4ffa63b. The sidecar fetch now uses `HTTPUpdate::SIDECAR_MD5_FAILED` / `SIDECAR_SHA256_FAILED`.

lucasssvaz (approved):

me-no-dev (approved):

Issue #12852, opened 2026-08-24
[Matter] MatterOccupancySensor.cpp fails to build against esp_matter >=1.6.0: occupancy-sensor-server.h removed, replaced by OccupancySensingCluster


lboue:
### Description

Same root cause as #12851 (component manager resolving `espressif/esp_matter` `1.6.0` since `idf_component.yml` has no upper bound), but a bigger break: `libraries/Matter/src/MatterEndpoints/MatterOccupancySensor.cpp` includes:

```cpp
#include <app/clusters/occupancy-sensor-server/occupancy-sensor-server.h>
```

This header **no longer exists** in the connectedhomeip snapshot bundled with `esp_matter` 1.6.0:

```
libraries/Matter/src/MatterEndpoints/MatterOccupancySensor.cpp:24:10: fatal error: app/clusters/occupancy-sensor-server/occupancy-sensor-server.h: No such file or directory
   24 | #include <app/clusters/occupancy-sensor-server/occupancy-sensor-server.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Unlike #12851, this isn't a simple file rename - the whole `OccupancySensing::Instance` / `AttributeAccessInterface`-based API that `MatterOccupancySensor.cpp` relies on (`OccupancySensing::Instance`, `OccupancySensing::SetHoldTime`, `OccupancySensing::GetHoldTimeLimitsForEndpoint`, etc., used by the custom `OccupancySensingAttrAccessWrapper` class) has been migrated to a new code-driven `OccupancySensingCluster : public DefaultServerCluster` class, following the same modernization pattern already seen for `ValveConfigurationAndControl` and `ClosureControl`.

The remaining files are:

```
components/esp_matter/data_model/../connectedhomeip/connectedhomeip/src/app/clusters/occupancy-sensor-server/
├── OccupancySensingCluster.h
├── OccupancySensingCluster.cpp
├── CodegenIntegration.h / .cpp
└── README.md
```

### Impact

[...]

Issue #12851, opened 2026-08-24
[Matter] Build fails against esp_matter >=1.6.0: closure-control-cluster-objects.h renamed to ClosureControlClusterObjects.h


lboue:
### Description

Building any sketch that uses the `Matter` library currently fails when the ESP-IDF Component Manager resolves the latest `espressif/esp_matter` component (`idf_component.yml` only pins `>=1.3.0`, no upper bound, so it resolves to `1.6.0` today).

`libraries/Matter/src/matter_closure_patch.h` includes:

```cpp
#include "app/clusters/closure-control-server/closure-control-cluster-objects.h"
```

In the connectedhomeip snapshot bundled with `esp_matter` 1.6.0, this file was renamed (part of a broader migration of several clusters to the new code-driven `DefaultServerCluster`-based pattern, PascalCase file names):

```
app/clusters/closure-control-server/closure-control-cluster-objects.h  →  app/clusters/closure-control-server/ClosureControlClusterObjects.h
```

This breaks compilation of every Matter endpoint file that transitively includes `Matter.h` (e.g. `MatterGenericSwitch.cpp`, `MatterOnOffLight.cpp`, `MatterDimmableLight.cpp`, `MatterColorTemperatureLight.cpp`, ...):

```
libraries/Matter/src/matter_closure_patch.h:13:10: fatal error: app/clusters/closure-control-server/closure-control-cluster-objects.h: No such file or directory
   13 | #include "app/clusters/closure-control-server/closure-control-cluster-objects.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

### Reproduction

Build any Matter example as an ESP-IDF component (e.g. `idf_component_examples/Arduino_ESP_Matter_over_OpenThread`) for `esp32c6` with ESP-IDF `release/v5.5`, with no upper version bound on `espressif/esp_matter` in `idf_component.yml` (the default), so the component manager resolves `esp_matter` `1.6.0`.

### Fix

One-line path update:

```diff
-#include "app/clusters/closure-control-server/closure-control-cluster-objects.h"
+#include "app/clusters/closure-control-server/ClosureControlClusterObjects.h"
```

[...]

PR #12853, opened 2026-08-24
feat(matter): add MatterWaterValve endpoint

lboue:
## Description of Change

Adds `MatterWaterValve`, a new Matter endpoint (device type `0x0042`) wrapping the **Valve Configuration and Control** cluster - for irrigation valves, main water shutoff valves, or appliance water supply valves. Modeled after the API I contributed for the same feature in [SiliconLabsSoftware/arduino#182](https://github.com/SiliconLabsSoftware/arduino/pull/182), adapted to this repo's `esp_matter`/`MatterEndPoint` architecture.

Waiting for [esp32-arduino-lib-builder](https://github.com/espressif/esp32-arduino-lib-builder) new release.

### API

- `begin(defaultOpenDurationSeconds = 0)` / `end()`
- `open()` / `open(durationSeconds)` / `close()`
- `getCurrentState()` / `getTargetState()` / `isOpen()`
- `getOpenDuration()` / `getDefaultOpenDuration()` / `getRemainingDuration()`
- `getValveFault()` / `setValveFault(fault)`
- `onOpen(cb)` / `onClose(cb)` user callbacks

### Why a Delegate?

Valve Configuration and Control's `CurrentState`/`TargetState`/`OpenDuration`/`RemainingDuration` are managed by the CHIP cluster implementation itself, not through the classic ember attribute store, and there's no ember command path for Open/Close either. So `MatterWaterValve` implements a nested `ValveDelegate` bridging `HandleOpenValve()`/`HandleCloseValve()`/`HandleRemainingDurationTick()` to the `onOpen()`/`onClose()` callbacks - both remote (Matter controller) and local (`open()`/`close()`) commands go through the same underlying cluster functions, so there's a single source of truth for state.

This matches the pattern an Espressif contributor confirmed as the recommended approach in [espressif/esp-matter#1804](https://github.com/espressif/esp-matter/issues/1804), where a user hit exactly the bug this design avoids (`CurrentState` stuck at `Transitioning`).

The SDK's own internal timer drives the `RemainingDuration` countdown and auto-closes the valve when a timed open elapses (calling `HandleCloseValve()` the same way a remote Close command would) - no per-second polling needed in the sketch, unlike the Silicon Labs implementation this was modeled after.

This pattern is new to this repo (only `BooleanState` used a comparable delegate-based approach before), so I documented it in `libraries/Matter/README.md` under a new "Delegate-Based Clusters" section for future contributors/reviewers.

### Targeting esp_matter 1.5.x

The first version of this PR was written against `ValveConfigurationAndControlCluster` / `esp_matter::data_model::provider`, the new CHIP cluster-class architecture introduced in `esp_matter` **1.6.0**. `esp32-arduino-lib-builder`'s `idf_component.yml` currently pins `esp_matter` to `"1.5"`, which doesn't have that architecture, so the endpoint wouldn't have compiled against the toolchain this repo actually ships.

I asked espressif/esp-matter whether Water Valve support could be backported to `release/v1.5` ([espressif/esp-matter#1830](https://github.com/espressif/esp-matter/issues/1830)). Digging into the actually-published `esp_matter` 1.5.0/1.5.1 component packages (not just source branches) showed they already vendor everything this endpoint needs, just under the older ember/delegate-based cluster implementation: `cluster::valve_configuration_and_control`, `endpoint::water_valve`, and `chip::app::Clusters::ValveConfigurationAndControl`'s free-function API (`SetValveLevel()` implements Open, plus `CloseValve()`, `UpdateCurrentState()`, `Attributes::ValveFault`, `EmitValveFault()`) with the exact same `Delegate` interface (`HandleOpenValve`/`HandleCloseValve`/`HandleRemainingDurationTick`).

So this PR now targets that 1.5.x-compatible API instead - no change to the public `MatterWaterValve` API or to the sketch-facing behavior, only to how the cluster is driven internally. `setValveFault()` also now writes the `ValveFault` attribute in addition to emitting the event, since the cluster's own Open command handler reads that attribute to reject Open with `FailureDueToFault` while a fault is active.

### Included in this PR

- `libraries/Matter/src/MatterEndpoints/MatterWaterValve.h` / `.cpp`
- Wiring: `Matter.h` (include + friend class), `keywords.txt`, `CMakeLists.txt` (Arduino as ESP-IDF component build)
- `libraries/Matter/examples/MatterWaterValve/` (sketch, README, `ci.yml`)
- `libraries/Matter/README.md`: new "Delegate-Based Clusters" section

### Testing

- Verified every header path, class/function signature, and delegate interface used (`valve-configuration-and-control-cluster.h`, `ValveConfigurationAndControl::SetValveLevel/CloseValve/UpdateCurrentState/EmitValveFault/Attributes::ValveFault`, `esp_matter::endpoint::water_valve`, `esp_matter::cluster::valve_configuration_and_control`) against the actual `esp_matter` 1.5.0 and 1.5.1 packages published on the ESP Component Registry (downloaded and inspected directly, not assumed from source branches).
- `clang-format --dry-run --Werror` passes on the changed files.
- I did **not** have an ESP-IDF/esp32-arduino-libs toolchain available in this environment to do a real compile or flash test this round, so this still needs CI plus a real build against `esp_matter` 1.5.x, and ideally hardware commissioning/Open/Close/timed-open verification against a Matter controller, before merge.
- espressif/arduino-esp32#12851 and #12852 (build breaks against `esp_matter` >=1.6.0) are no longer relevant to this PR now that it targets the 1.5.x line actually vendored by `esp32-arduino-libs`.

### Test Environment

- [ ] Not yet built or flashed against real hardware - needs CI + a real `esp_matter` 1.5.x build to confirm compilation, and hardware commissioning to confirm Open/Close/timed-open behavior.

### Test Scenario

Not yet exercised against real hardware or a live build in this environment - see "Testing" above.


lboue:
## CI status update: real, deeper blocker found

The last two commits fixed what CI could actually catch:

1. `922cce0` retargeted `MatterWaterValve` to the `esp_matter` 1.5.x delegate/free-function API (was written against 1.6.0's `ValveConfigurationAndControlCluster`/`data_model_provider`, which isn't in the 1.5.x line this repo's toolchain pins - see espressif/esp-matter#1830).
2. `65b7b2c` fixed an unrelated `-Werror=format=` bug in the example sketch that the previous CI run caught.

With both of those in, **`MatterWaterValve.cpp`/`.h` now compile cleanly** against the real `esp32-arduino-libs` toolchain (job `Arduino 10`, run [32757872544](https://github.com/espressif/arduino-esp32/actions/runs/32757872544/job/97529917473)) - all other jobs in that run are green.

But that same run's `Arduino 10` job now fails at **link time**, and not just on my code:

```
libespressif__esp_matter.a(esp_matter_delegate_callbacks.cpp.obj):
  undefined reference to chip::app::Clusters::ValveConfigurationAndControl::SetDefaultDelegate(...)
libespressif__esp_matter.a(esp_matter_cluster.cpp.obj):
  undefined reference to MatterValveConfigurationAndControlPluginServerInitCallback()
libespressif__esp_matter.a(esp_matter_command.cpp.obj):
  undefined reference to emberAfValveConfigurationAndControlClusterOpenCallback(...)
  undefined reference to emberAfValveConfigurationAndControlClusterCloseCallback(...)
```

These undefined references come from **`esp_matter`'s own prebuilt static library**, not from `MatterWaterValve.cpp` - `esp_matter` was clearly built with Valve Configuration and Control enabled (its Kconfig default), so its own wrapper code calls into CHIP's `ValveConfigurationAndControl` free functions and ember callbacks. But the connectedhomeip/CHIP static library bundled in the same `esp32-arduino-libs` toolchain doesn't seem to include the compiled valve-configuration-and-control server object, so those symbols are missing at link time.

This isn't fixable from this repo's side - it's a gap in what's linked into the published `esp32-arduino-libs` binaries. Filed espressif/esp32-arduino-lib-builder#390 with the details.

**So this PR is currently blocked on espressif/esp32-arduino-lib-builder#390**, independent of which `esp_matter` API version `MatterWaterValve` targets. I'll keep the 1.5.x-targeting code as-is since it's correct and now confirmed to compile; it just can't link until the toolchain itself carries the Valve cluster's CHIP object code.

lboue:
Waiting for this PR:
- https://github.com/espressif/esp32-arduino-lib-builder/pull/402

lboue:
## CI failure root cause: `esp_matter` 1.5.x does not have the Valve Configuration and Control cluster at all

Dug into why CI still fails to link (`undefined reference to chip::app::Clusters::ValveConfigurationAndControl::SetValveLevel/CloseValve/UpdateCurrentState`) after fixing the `esp32c5` `MatterNetwork=wifi` FQBN gap.

Checked directly against the `espressif/esp-matter` GitHub branches (since `esp32-arduino-lib-builder`'s `main/idf_component.yml` pins `espressif/esp_matter: "1.5"`):

- `components/esp_matter/data_model_provider/clusters/` on `release/v1.5` — **no `valve_configuration_and_control` entry** (has `boolean_state`, `chime`, `descriptor`, etc., but not valve).
- `components/esp_matter/cluster/` (the "ember-style" path this PR's second commit targeted) — **doesn't exist at all on `release/v1.5`** (404).
- `components/esp_matter/data_model_provider/clusters/valve_configuration_and_control/` **does exist on `release/v1.6`**, under the new code-driven `CodegenIntegration` architecture.
- In `connectedhomeip`, `SetValveLevel`/`CloseValve`/`UpdateCurrentState` only show up in `src/app/clusters/valve-configuration-and-control-server/CodegenIntegration.{h,cpp}` — there's no legacy ember implementation of this cluster to fall back to.

So the premise of this PR's second commit ("esp_matter 1.5.x already vendors this under the older ember/delegate-based implementation") doesn't hold up — Valve Configuration and Control was introduced directly with the new code-driven cluster architecture in esp_matter 1.6.0, with nothing equivalent in 1.5.x. The header declarations compile fine (hence green compile jobs earlier), but the implementing object file is never linked in, hence the flood of undefined references. Since `MatterWaterValve.cpp` is wired into the shared `Matter` library via `CMakeLists.txt`, this currently risks breaking the link step for *any* Matter sketch, not just the `MatterWaterValve` example, as long as `esp32-arduino-lib-builder` stays pinned to `esp_matter 1.5`.

Net effect: this PR is blocked on `esp32-arduino-lib-builder` bumping its `esp_matter` pin to >=1.6.0 (tracked via #12851/#12852 and espressif/esp-matter#1830), not on anything fixable purely within `arduino-esp32`. Will revert to the original 1.6.0-targeting implementation (`ValveConfigurationAndControlCluster` + `esp_matter::data_model::provider`, matching the actual upstream API) and keep this in draft until the dependency bump lands.

lboue:
Tracking the actual dependency: espressif/esp32-arduino-lib-builder#402 ("Add support for Matter 1.6 to IDF 6") is the PR that bumps the `esp_matter` pin. This PR stays blocked/draft until that lands (and until `esp32-arduino-lib-builder` cuts a release the Arduino core picks up).

lboue:
## Update: switching to the release/v1.5-compatible delegate API

Following the investigation in espressif/esp-matter#1830, I'm switching this PR
away from the new `data_model::provider` / `ServerClusterInterface` architecture
(only available starting esp_matter 1.6.0) toward the classic delegate API that's
already shipped in esp_matter 1.5.x — the one currently pinned by
`esp32-arduino-lib-builder`.

Reference implementation: `ValveControlDelegate` in
[connectedhomeip@d144bbb](https://github.com/espressif/connectedhomeip/blob/d144bbb/examples/all-clusters-app/linux/ValveControlDelegate.h),
matching `esp_matter`'s documented delegate list for
`Valve Configuration And Control Cluster` (see [release-v1.5 app guide](https://docs.espressif.com/projects/esp-matter/en/release-v1.5/esp32c5/app_guide.html#valve-configuration-and-control-cluster)).

The `HandleOpenValve` / `HandleCloseValve` / `HandleRemainingDurationTick`
signatures are unchanged between the two architectures, so the driver-level
logic carries over largely as-is. What changes is the registration path:
`esp_matter::cluster::valve_configuration_and_control::create()` + delegate
pointer in `config_t`, instead of the provider registry lookup.

This should compile *and* link cleanly against the currently published
`esp32-arduino-libs` toolchain, without depending on espressif/esp32-arduino-lib-builder#391
or the Matter 1.6 migration (espressif/esp32-arduino-lib-builder#402).

PR #12863, opened 2026-08-29
feat(usb): multi-sample-rate support for UAC1/UAC2

JohannJEG:
## Description of Change

Adds support for configuring multiple USB Audio Class sample rates instead of using a single fixed sample rate.

This allows the device to detect sample-rate changes requested by the host and adjust the I2S configuration accordingly.

### Changes

* Adds support for multiple UAC sample rates.
* Allows the host to select a supported sample rate.
* Allows the selected sample rate to be detected by the device so the I2S configuration can be adjusted accordingly.
* Keeps the existing behavior for single-rate configurations.

## Test Scenarios

Tested on ESP32-S3 with USB Audio Class.

Verified that the sample rate can be switched between **44.1 kHz and 48 kHz** from the host on both Linux and Windows, with the corresponding I2S reconfiguration confirmed through the serial output.

## Related links

None.


me-no-dev:
Not supporting multiple sample rates was a design choice. Anyway, in order to process further this PR, UAC2 support for multiple sample rates also needs to be added

JohannJEG:
> Not supporting multiple sample rates was a design choice. Anyway, in order to process further this PR, UAC2 support for multiple sample rates also needs to be added

Thanks for the clarification. Just to make sure I understand correctly: since not supporting multiple sample rates was a design choice, does that mean this PR would be rejected even if I also added UAC2 support?

I currently don't have a UAC2-capable board available for testing, and I don't think it would be worth buying one just for this if, due to the design decision, the PR would ultimately not be accepted anyway.

If so, I'll continue the development in my own fork for now.

me-no-dev:
UAC2 is actually simpler to add multiple sample rates. UAC1's way of doing it was the reason why I decided to not add the feature. You can see from the size measurements above that your pull request adds 21KB extra flash space for just that feature alone. The best option would be to have support for both UAC1 and UAC2 and that support be guarded by define, so that users who need it, can add a define "UAC_USE_MULTIPLE_RATES" to their project/sketch and enable that feature, accepting the flash use penalty. You can use any ESP32-P4 (or soon ESP32-S31) board to work on UAC2

JohannJEG:
> UAC2 is actually simpler to add multiple sample rates. UAC1's way of doing it was the reason why I decided to not add the feature. You can see from the size measurements above that your pull request adds 21KB extra flash space for just that feature alone. The best option would be to have support for both UAC1 and UAC2 and that support be guarded by define, so that users who need it, can add a define "UAC_USE_MULTIPLE_RATES" to their project/sketch and enable that feature, accepting the flash use penalty. You can use any ESP32-P4 (or soon ESP32-S31) board to work on UAC2

Thank you for the clarification and for explaining in detail the concern regarding the additional Flash memory usage.

Making support for multiple sample rates optional through the `UAC_USE_MULTIPLE_RATES` flag seems like an excellent approach to me. I will start adapting the UAC1 implementation so that the additional code is only included when this option is enabled, while keeping the default behavior unchanged.

My idea is to enable this option through a `build_opt.h` file located in the same directory as the sketch, containing the `-DUAC_USE_MULTIPLE_RATES` flag. Could you please confirm if this is the correct way to implement it in this repository?

I will also work on UAC2 support later. First, I need to acquire an ESP32-P4 to perform the necessary tests, so this will take some additional time. I would appreciate it if the PR could remain open in the meantime, so I can continue making updates.

I am also open to any changes or modifications that may be necessary. My intention is to make this the best possible approach for integrating this feature into the library.

me-no-dev:
> My idea is to enable this option through a build_opt.h file located in the same directory as the sketch, containing the -DUAC_USE_MULTIPLE_RATES flag. Could you please confirm if this is the correct way to implement it in this repository?

This was exactly what I meant. Well... pioarduino has other means, but both will result in the same behavior

JohannJEG:
> > My idea is to enable this option through a build_opt.h file located in the same directory as the sketch, containing the -DUAC_USE_MULTIPLE_RATES flag. Could you please confirm if this is the correct way to implement it in this repository?
> 
> This was exactly what I meant. Well... pioarduino has other means, but both will result in the same behavior

Thanks for your feedback! I implemented it using a build_opt.h file with -DUAC_USE_MULTIPLE_RATES.

Tested multi-sample-rate switching at 44.1 kHz and 48 kHz on:

ESP32-S3 / UAC1: Linux and Windows
ESP32-P4 / UAC2: Linux and Windows

Playback works at both rates, and sample-rate change events correctly reconfigure I2S.

I’m open to any further adjustments or suggestions.

libraries/USB/src/USBAudioCard.cpp
```diff
@@ -52,7 +215,7 @@ uint16_t tusb_audio_load_descriptor(uint8_t *dst, uint8_t *itf) {
     uint8_t descriptor[TUD_AUDIO20_HEADSET_STEREO_DESC_LEN] = {
       // Interface number, string index, EP Out & EP In & EP Interrupt address, max sample rate, speaker channels, mic channels, bytes per sample RX/TX, bits used per sample RX/TX
       TUD_AUDIO20_HEADSET_STEREO_DESCRIPTOR(
-        _itf_num, str_index, ep_num, (uint8_t)(ep_num | 0x80), (uint8_t)(int_ep_num | 0x80), CFG_TUD_AUDIO_MAX_SAMPLE_RATE, _spk_channels, _mic_channels,
+        _itf_num, str_index, ep_num, (uint8_t)(ep_num | 0x80), (uint8_t)(int_ep_num | 0x80), max_sample_rate, _spk_channels, _mic_channels,
```
me-no-dev, line 218: it is OK to keep using `CFG_TUD_AUDIO_MAX_SAMPLE_RATE` for UAC2 descriptor. It ensures that buffers are large enough and chances of issues because of them are minimized

me-no-dev:
small nitpick, otherwise LGTM

JohannJEG:
> small nitpick, otherwise LGTM

Updated, thanks! I restored CFG_TUD_AUDIO_MAX_SAMPLE_RATE for the UAC2 descriptor sizing as requested.

libraries/USB/src/USBAudioCard.cpp
JohannJEG: Done, restored CFG_TUD_AUDIO_MAX_SAMPLE_RATE here. Thanks!

JohannJEG:
I found a limitation in the current UAC2 path while testing more sample rates on ESP32-P4.
Five rates work correctly, including 96 kHz:

`16000, 32000, 44100, 48000, 96000`

But when I add a sixth rate, Linux reports:

`parse_audio_format_rates_v2v3(): unable to retrieve number of sample rates (clock 4)`

I found that CFG_TUD_AUDIO_CTRL_BUF_SZ is 64 bytes. The UAC2 RANGE response fits with 5 rates, but exceeds that size with 6.
UAC1 does not seem to have the same issue because the rates are part of the format descriptor.

Which approach would you prefer for this PR: should I limit UAC2 to 5 rates for now, or handle larger `RANGE` responses differently?

I haven't changed this yet.

me-no-dev:
I think that 5 rates are plenty for now

JohannJEG:
> I think that 5 rates are plenty for now

Done. UAC2 is now limited to five rates; UAC1 remains at eight.

me-no-dev (approved):

PR #12865, opened 2026-08-30
feat(update): add optional SHA512 checksum and sidecar URLs

### LINKED ISSUE #12868: feat(update): optional SHA512 checksum and sidecar URLs for OTA
### Related area

Update / HTTPUpdate (OTA checksum verification)

### Hardware specification

All ESP32 targets supported by Arduino-ESP32 (no hardware-specific dependency)

OTA integrity verification currently supports MD5 and SHA256 (including response headers and sidecar URLs). Some deployments and tooling prefer or require SHA512 digests. Today callers cannot pass a SHA512 expectation through the Arduino Update/HTTPUpdate APIs and must either stay on SHA256 or implement verification outside the core libraries.

Add optional SHA512 support mirroring the existing SHA256 path:

- `Update.setSHA512()` / related verify path
- `HTTPUpdate::setSHA512sum()`
- response header `x-SHA512`
- `setSHA512sumUrl()` sidecar fetch (128-character hex token)

Keep the same digest precedence as SHA256: explicit setter > response header > sidecar URL. MD5/SHA256 behavior should remain unchanged.

Implementation PR: #12865

- Keep verifying SHA512 only in application code after download — duplicates core logic and bypasses the existing header/sidecar integration.
- Force all users onto SHA256 — not ideal when servers or policies already publish SHA512 digests.

### Additional context

Extends the existing checksum/sidecar series (SHA256 + sidecar). Validation coverage should include format/roundtrip, header/setter precedence, and sidecar success/failure cases.

EyJunge1:
## Summary
- Adds optional SHA512 integrity verification mirroring the existing SHA256 path: `Update.setSHA512()`, `HTTPUpdate::setSHA512sum()`, response header `x-SHA512`, and `setSHA512sumUrl()` sidecars (128-character hex token).
- Preserves digest precedence: explicit setter > response header > sidecar URL. Sidecar fetch remains lazily linked when a sidecar setter is used.
- Extends OTA validation coverage for SHA512 format/roundtrip, header/setter precedence, and sidecar success/failure cases. MD5/SHA256 behavior is unchanged.

Fixes #12868


EyJunge1:
@me-no-dev @lucasssvaz The failed check `Build IDF release-v5.4 for esp32p4` looks unrelated to this PR’s code.

The **Build** step itself succeeded. Only the later step
`Upload generated sdkconfig files for debugging` failed with repeated
GitHub Artifact API timeouts (`CreateArtifact` request timeout after 5 attempts).

Could a maintainer please re-run the failed jobs for that workflow?
Thank you

libraries/Update/src/Updater.cpp
```diff
@@ -98,6 +109,72 @@ static void update_sha256_abort(UpdateSHA256Context *ctx) {
...
+  if (psa_ret != PSA_SUCCESS) {
+    log_e("PSA crypto init failed: %d", (int)psa_ret);
+    return false;
+  }
+  ctx->op = psa_hash_operation_init();
+  psa_ret = psa_hash_setup(&ctx->op, PSA_ALG_SHA_512);
+  if (psa_ret != PSA_SUCCESS) {
+    log_e("PSA hash setup failed: %d", (int)psa_ret);
+    return false;
+  }
```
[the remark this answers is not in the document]
EyJunge1: This matches the existing SHA-256 path: `update_sha256_begin()` also returns without `psa_hash_abort()` if `psa_hash_setup()` fails, and `setSHA256()` then just `delete`s the context. SHA-512 copies that.

Should we add the abort in both `update_sha256_begin()` and `update_sha512_begin()` in this PR, or keep the SHA-256 mirror as-is?

libraries/Update/src/Update.h
```diff
@@ -267,6 +268,45 @@ class UpdateClass {
...
+   * @param calc_post_decryption If true, calculate SHA-512 after decryption
+   * @return true if SHA-512 was accepted
+   */
+  bool setSHA512(
+    const char *expected_sha512
+#ifndef UPDATE_NOCRYPT
+    ,
+    bool calc_post_decryption = true
+#endif /* #ifdef UPDATE_NOCRYPT */
+  );
```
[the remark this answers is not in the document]
EyJunge1: This also copies `setSHA256()` / `setMD5()`, which already use `#endif /* #ifdef UPDATE_NOCRYPT */` on an `#ifndef` block. The rest of the file uses `#endif /* UPDATE_NOCRYPT */`.

Should we align the `#endif` comments on `setMD5`, `setSHA256`, and `setSHA512` in this PR, or leave the existing style?

libraries/Update/src/Updater.cpp
```diff
@@ -791,6 +882,14 @@ bool UpdateClass::_writeBuffer() {
...
       return false;
     }
+#ifndef UPDATE_NOCRYPT
+  }
+  if (_target_sha512_decrypted) {
+#endif /* UPDATE_NOCRYPT */
+    if (!_sha512Update(_buffer, _bufferLen)) {
+      _abort(UPDATE_ERROR_SHA512);
+      return false;
+    }
```
lucasssvaz, line 742: This path (and `_sha512FreeContext()` from `_reset`/`begin`) lives in `Updater.cpp`, so `--gc-sections` cannot drop SHA-512.

That pulls `mbedtls_sha512_*` / `psa_hash_*` into every sketch that uses `Update`, including ones that never call `setSHA512()`. CI matches this: `SD_Update` (no TLS) grows about +5K FLASH on ESP32/C3/C6, +7K on C5/P4, and +11K on H2.

This also fights the SHA-256 comment at the top of this file: mbedtls SHA-256 was chosen specifically to avoid linking the SHA-512 software implementation into every Update sketch.

Please lazy-link SHA-512 the same way `HTTPUpdateChecksum.cpp` does (separate TU + function pointer assigned only from `setSHA512()`).

libraries/Update/README.md
```diff
@@ -265,6 +266,33 @@ Return the digest of a completed update that used `setSHA256()`.
...
+bool setSHA512(const char *expected_sha512)
+```
+Sets expected SHA-512 hash for verification of the entire downloaded payload
+(same scope as MD5/SHA-256; not the ESP-IDF partition/image SHA-256 footer).
+
+Call after `begin()` and before writing any payload bytes. The SHA-512 streaming
+context is allocated only when this succeeds, so sketches that never call
+`setSHA512()` (or never receive an `x-SHA512` header / sidecar via `HTTPUpdate`) avoid the
+extra hashing work and keep the smaller static footprint on the global `Update`
+object.
```
lucasssvaz: The heap context is on-demand, but FLASH is not. `_writeBuffer` / `_reset` always reference the SHA-512 helpers, so mbedtls/PSA SHA-512 is linked into every `Update` binary.

Please lazy-link SHA-512.

libraries/Update/src/Update.h
```diff
@@ -403,6 +444,14 @@ class UpdateClass {
   bool _sha256Update(const uint8_t *data, size_t len);
   bool _sha256Finish();
 
+  void *_sha512_ctx;           ///< Opaque streaming SHA-512 context and expected digest (allocated on demand)
+  uint8_t _sha512_result[64];  ///< Final digest kept after context is freed
```
lucasssvaz: This 64-byte buffer is resident on every `UpdateClass` instance, including the global `Update` object, even when SHA-512 is never used. Same pattern as `_sha256_result[32]`.

Fine to keep if we stay with the SHA-256 layout, but the README should not call this a smaller static footprint — it is +64B BSS on top of the FLASH hit from always linking SHA-512.

lucasssvaz (changes requested):
SHA-512 API, precedence, sidecar parsing, and tests look good and correctly mirror the SHA-256 path.

Please address the FLASH always-link before merge like SHA-256.

libraries/Update/src/Updater.cpp
EyJunge1: Moved SHA-512 into `UpdaterSHA512.cpp`. `Updater.cpp` now only calls through `_sha512Ops`, which `setSHA512()` assigns — same pattern as `HTTPUpdateChecksum.cpp` / `_checksumSidecarFetch`.

Sketches that never reference `setSHA512()` should no longer pull mbedtls/PSA SHA-512 (`SD_Update` and similar). `HTTPUpdate` still references the setter so `x-SHA512` / `setSHA512sum()` / sidecar keep working.

libraries/Update/README.md
EyJunge1: Updated. The README no longer claims a smaller FLASH footprint from heap-only allocation. It now describes the separate TU and that linking happens only when `setSHA512()` is referenced.

EyJunge1:
@lucasssvaz SHA-512 is now lazy-linked as requested (separate TU + function pointer assigned only from `setSHA512()`).

One follow-up: SHA-256 is still in `Updater.cpp` (the existing cheaper path). Would you like the same separate-TU treatment for SHA-256 in a follow-up PR, or should it stay where it is?

lucasssvaz:
@EyJunge1 Yes please. I missed that in the previous PR. You can even do in this PR. Whatever you prefer.

tests/validation/ota/ota.ino
lucasssvaz, line 904: Should we add the sha512 here too ?

EyJunge1: Yes — `/ota.ino.bin` already sends both `x-SHA256` and `x-SHA512`. Added `TEST_ASSERT_EQUAL(128, Update.sha512String().length())` next to the SHA256 check.

libraries/Update/src/Update.h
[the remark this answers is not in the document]
EyJunge1: Aligned the `#endif` comments on `setMD5` / `setSHA256` / `setSHA512` to `#endif /* UPDATE_NOCRYPT */`, matching the rest of the file.

lucasssvaz (approved):

EyJunge1:
@me-no-dev @P-R-O-C-H-Y PTAL :)

me-no-dev (approved):

PR #12866, opened 2026-08-30
fix(preferences): return 1 from putString for empty success

### LINKED ISSUE #12869: Preferences::putString("") returns 0 on successful empty write
### Board

ESP32 Dev Module (reproducible on all targets; NVS/Preferences API)

### Device Description

Any board using Preferences / NVS — not board-specific.

### Hardware Configuration

None required.

### Version

### Type

Bug

### IDE Name

PlatformIO / Arduino IDE

### Operating System

macOS / Linux / Windows (host-independent)

### Flash frequency

40Mhz

### PSRAM enabled

no

### Upload speed

921600

### Description

`Preferences::putString(key, "")` successfully stores an empty string via `nvs_set_str`, but the Arduino wrapper returns `strlen(value)`, which is `0` for an empty string.

Callers that treat a `0` return as failure cannot distinguish a successful empty write from an error. Expected: a successful empty-string write should return a non-zero success indicator (e.g. `1`), while non-empty success continues to return `strlen(value)`.

Fix PR: #12866

### Sketch

```cpp
#include <Preferences.h>

Preferences prefs;

void setup() {
  Serial.begin(115200);
  prefs.begin("test", false);

  size_t n = prefs.putString("empty", "");
  Serial.printf("putString(\"\") returned %u (expected non-zero on success)\n", (unsigned)n);

  String back = prefs.getString("empty", "MISSING");
  Serial.printf("read back length=%u value='%s'\n", (unsigned)back.length(), back.c_str());

  prefs.end();
}

void loop() {}
```

### Debug Message

```
putString("") returned 0 (expected non-zero on success)
read back length=0 value=''
```

NVS write succeeds; only the Arduino return value is misleading.

### Other Steps to Reproduce

1. Call `putString` with an empty string.
2. Observe return value `0`.
3. Call `getString` for the same key — value is present and empty, confirming the write succeeded.

EyJunge1:
## Summary
- `Preferences::putString("")` stores successfully but previously returned `0` (`strlen`), indistinguishable from failure.
- On empty-string success, return `1` so callers treating `0` as error still see success; non-empty returns unchanged.
- Document the return value and add `test_string_empty` to the NVS validation suite.

Fixes #12869

### DIFF
--- a/docs/en/api/preferences.rst
+++ b/docs/en/api/preferences.rst
@@ -375,7 +375,8 @@ Arduino-esp32 Preferences API
          - if ``String``, a valid Arduino String type.
 
    **Returns**
-      * if successful: the number of bytes stored; ``0`` otherwise.
+      * if successful: the number of bytes stored; for an empty string ``1`` is returned
+        so success is distinguishable from failure; ``0`` otherwise.
 
    **Notes**
       * Attempting to store a value without a namespace being open in read-write mode will fail.
--- a/libraries/Preferences/src/Preferences.cpp
+++ b/libraries/Preferences/src/Preferences.cpp
@@ -275,7 +275,10 @@ size_t Preferences::putString(const char *key, const char *value) {
     log_e("nvs_commit fail: %s %s", key, nvs_error(err));
     return 0;
   }
-  return strlen(value);
+  // Empty string stores successfully but strlen is 0; return 1 so callers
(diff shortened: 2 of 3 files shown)


me-no-dev:
This is not an acceptable change. I have explained more in the issue that you opened.

Issue #191, opened 2017-02-10
softAPConfig() in WiFiAP.cpp


yukini3:
There are code error in this API. In softAPConfig, tcpip_adapter_set_ip_info() is called and expecting non-zero return value is successful but zero (ESP_OK) is successful value.
Calling tcpip_adapter_dhcps_start() also the same.

me-no-dev:
aaah I though I fixed those a few days ago. Will do

me-no-dev:
on another note :)
code is correct ;)
if tcpip_adapter_set_ip_info returns non-zero, then it failed and dhcps will be started. Error is that it should return false, regardless of what tcpip_adapter_dhcps_start returns.

me-no-dev:
nevermind... too much code in my head... this is AP not STA

Issue #220, opened 2017-02-18
async analogRead


ppisljar:
if i understand correctly we can only do a sync analog read at the moment ?
this has a problem of being super slow and wasting many cpu cycles ...

i tried to split analogRead into 3 functions .... analogReadAsync, analogReadAsyncReady and analogReadAsyncGet : 

```
void IRAM_ATTR __analogReadAsync(uint8_t pin)
{
    int8_t channel = digitalPinToAnalogChannel(pin);
    SET_PERI_REG_BITS(SENS_SAR_MEAS_START1_REG, SENS_SAR1_EN_PAD, (1 << channel), SENS_SAR1_EN_PAD_S);
    CLEAR_PERI_REG_MASK(SENS_SAR_MEAS_START1_REG, SENS_MEAS1_START_SAR_M);
    SET_PERI_REG_MASK(SENS_SAR_MEAS_START1_REG, SENS_MEAS1_START_SAR_M);
}

bool IRAM_ATTR __analogReadAsyncReady()
{
    return (GET_PERI_REG_MASK(SENS_SAR_MEAS_START1_REG, SENS_MEAS1_DONE_SAR) != 0);
}

uint16_t IRAM_ATTR __analogReadAsyncGet()
{
    return GET_PERI_REG_BITS2(SENS_SAR_MEAS_START1_REG, SENS_MEAS1_DATA_SAR, SENS_MEAS1_DATA_SAR_S);
}
```

and then use it like:

```
analogAsyncRead(34);
if (analogReadAsyncReady) myvalue = analogReadAsyncGet();
```

however this is far from perfect and could be implemented better i guess.

also is there an Interrupt i could set to be fired when adc reading is done ? or something like free running mode on atmega chips ? can i somehow change clock freq to make it faster (and less precise i guess) ?

me-no-dev:
Using Async here is really wrong :) this is not async at all. Interrupts are :)
give me some time to fix some other stuff I'm working on and I'll have another look at the ADC hardware and see what I can do.
Probably would not use "analogRead" in the names at all :)

ppisljar:
yeah i agree, its just something i was able to quickly hack together, and it seems its working (instead of waiting for adc finished flag in a loop i just am able to do other things now), but this should be solved with interrupts .... but i don't know enough to do it in the correct way :)

i would appreciate it if you can provide a better way to do it (also timer with interrupt would be nice)

me-no-dev:
[...]

PR #12879, opened 2026-09-02
fix(hosted): Pin version for lower hosted RAM usage

me-no-dev:
This pull request updates the version constraint for the `espressif/esp_hosted` dependency in the `idf_component.yml` file. The new version ensures compatibility with recent changes and bug fixes.

Dependency updates:

* Updated the `espressif/esp_hosted` dependency version from `^2.9.2` to `2.12.3` in `idf_component.yml` to use the latest stable release.

### DIFF
--- a/idf_component.yml
+++ b/idf_component.yml
@@ -120,7 +120,7 @@ dependencies:
       - if: "target in [esp32s3, esp32p4]"
       - if: idf_version < 6.0
   espressif/esp_hosted:
-    version: "^2.9.2"
+    version: "2.12.3"
     rules:
       - if: "target == esp32p4"
   espressif/esp_wifi_remote:


lucasssvaz (approved):

wjduenow:
Please consider pinning to `2.12.12` or above rather than `2.12.3` — `2.12.3` predates the fix for a P4/C6 SDIO deadlock that is, in my experience, unrecoverable in the field.

**What 2.12.3 gives up.** esp-hosted-mcu [#220](https://github.com/espressif/esp-hosted-mcu/issues/220): under sustained inbound traffic the SDIO receive path deadlocks on a dropped RX read. There is no bus error, no transport failure, no restart — the host keeps running and looks healthy while RSSI reads 0 and every RPC times out. Espressif root-caused it and fixed it in **2.12.12** ([`0985253`](https://github.com/espressif/esp-hosted-mcu/commit/098525357e19c81099f2c3769938bd877190a8f5), 2026-07-31, "fix(sdio): recover a dropped RX read instead of deadlocking"), also in 2.12.13 and 3.0.7. That reporter measured **43 wedges in 21 h on 2.12.11 → 0 in 14.5 h patched**, over 196 GB inbound.

Because the failure is silent, it is worse than a crash for an Arduino audience: nothing in the log says the radio died, so the sketch has to grow its own liveness detector before it can even reboot out of it.

**My own data, in case a second report helps.** ESP32-P4 + ESP32-C6 over SDIO (1-bit, 10 MHz), host and slave both on 2.12.11, a wall-mounted Sonos controller pulling JPEGs and SOAP/XML over the link. Bisecting the inbound sources on 2026-09-07:

| inbound source A | inbound source B | result |
|---|---|---|
| on | on | dies within ~1 min of playback — six for six |
| off | on | dies |
| on | off | dies at 3.5 min |
| **off** | **off** | **survives the full 10 min reproduction** |

No single source is the trigger; any sustained inbound stream wedges it. I could not fix it from the application side — pacing, serialising concurrent transfers behind a gate, and cutting 146 KB per track of image traffic each helped and none of them saved the link. The device currently ships with two features disabled and a 7-second dead-link detector that reboots the board, which is the only recovery I have.

**On the RAM regression — it's real, and I don't think this is the way to fix it.** [esp-hosted-mcu#191](https://github.com/espressif/esp-hosted-mcu/issues/191) (`mempool create failed: no mem`, assert at `sdio_drv.c:249`) is DMA-capable internal SRAM, which PSRAM cannot back, and I agree that's painful — internal SRAM is the scarce resource on my board too. But it is still open upstream with no fix in `main` as of 2026-09-10, and pinning the Arduino core nine patch versions back trades a startup failure that is loud and immediate for a runtime deadlock that is silent and intermittent. It also freezes every P4 user at 2.12.3 until someone remembers to unpin, and gives up the other fixes in between (the streaming-mode RX double-free in 2.12.11, among them).

Some alternatives that would keep both properties:

1. `version: ">=2.12.12,<2.13"` — takes the deadlock fix and the RX double-free fix, still excludes 3.x.
2. Fix #191 forward. [esp-hosted-mcu#231](https://github.com/espressif/esp-hosted-mcu/issues/231) proposes capping `ESP_TRANSPORT_SDIO_MAX_BUF_SIZE` at 16384 to match the IDF DMA limit, which looks like it bears directly on the mempool that asserts.
3. If the RAM cost is genuinely unavoidable right now, making the SDIO buffer pool sizing a Kconfig choice would let constrained boards shrink it rather than making every P4 board take an older transport.

For what it's worth, my board runs 2.12.11 without hitting #191 at all, at ~98 KB free internal heap. One thing that bought a lot of that back was moving mbedTLS off internal SRAM (`MBEDTLS_EXTERNAL_MEM_ALLOC=y` plus asymmetric content lengths) — the stock 16 KB in + 16 KB out per TLS connection comes out of exactly the pool that `sdio_mempool_create` is failing to get. That may be a cheaper lever for the reporter of #191 than downgrading the transport.

Happy to test a candidate pin on P4 + C6 hardware with a reproduction that reliably wedges 2.12.11 within a minute, if that would be useful.

me-no-dev:
@wjduenow we will pin to higher version once esp-hosted is fixed to not waste 80KB of RAM

wjduenow:
Correction to my comment above: on re-checking the firmware I flashed, the host was already running esp_hosted **2.12.13**, not 2.12.11. The build rebuilds the Arduino libs locally and resolved the `^2.9.2` caret itself. The deaths in that table therefore happened *with* the #220 fix on the host, so my data does not support 2.12.12 as the fix for this failure, and I withdraw it as an argument against the pin. My C6 slave is still on 2.12.11, which is where I'm looking next. Apologies for the noise.

me-no-dev:
@wjduenow moving to 2.12.13 with mempool disabled [here](https://github.com/espressif/esp32-arduino-lib-builder/pull/398) in order to provide all fixes and lower memory footprint at boot

Issue #402, opened 2017-05-28
SD card library


lutfisan:
I try SD_Test example,
when I want to write string to sd card it become error
```
void loop(){
   // make a string for assembling the data to log:
  String dataString = "";

  // read three sensors and append to the string:
  for (int analogPin = 0; analogPin < 3; analogPin++) {
    int sensor = analogRead(analogPin);
    dataString += String(sensor);
    if (analogPin < 2) {
      dataString += ",";
    }
  }

  Serial.println(dataString);
  writeFile(SD, "/log.txt", dataString);
}
```

jestebanes:
Hi, without reading the error message, I think that you problem could be the way you are passing the argument dataString, becouse the writeFile funtion signature is:

void writeFile(fs::FS &fs, const char * path, const char * message)

and your third argument is a String not a const char*, try

 writeFile(SD, "/log.txt", dataString.c_str());

Bye

lutfisan:
thanks it's works

but when i try this code
```
#include "FS.h"
#include "SD.h"
#include "SPI.h"


void setup(){
    Serial.begin(115200);
    if(!SD.begin()){
        Serial.println("Card Mount Failed");
        return;
    }
    uint8_t cardType = SD.cardType();

    if(cardType == CARD_NONE){
        Serial.println("No SD card attached");
        return;
    }

    Serial.print("SD Card Type: ");
    if(cardType == CARD_MMC){
        Serial.println("MMC");
    } else if(cardType == CARD_SD){
        Serial.println("SDSC");
    } else if(cardType == CARD_SDHC){
        Serial.println("SDHC");
    } else {
        Serial.println("UNKNOWN");
    }

    uint64_t cardSize = SD.cardSize() / (1024 * 1024);
    Serial.printf("SD Card Size: %lluMB\n", cardSize);

}

void loop(){
  // make a string for assembling the data to log:
  String dataString = "";

  // read three sensors and append to the string:
  for (int analogPin = 0; analogPin < 3; analogPin++) {
    int sensor = analogRead(analogPin);
    dataString += String(sensor);
    if (analogPin < 2) {
      dataString += ",";
    }
  }
  
  File file = fs.open("/log.txt", FILE_WRITE);
    if(!file){
        Serial.println("Failed to open file for writing");
        return;
    }
    if(file.println(dataString)){
        Serial.println("File written");
    } else {
        Serial.println("Write failed");
    }

}
```
[...]

PR #12914, opened 2026-09-19
feat(matter): removes matter 1.5 closure patch include

SuGlider:
## Description of Change

This PR is necessary to allow the [PR #402](https://github.com/espressif/esp32-arduino-lib-builder/pull/402) from Arduino Lib Buider to complete the Jobs.
It removes the include of `matter_closure_patch.h` which is not necessary in ESP-Matter 1.6 

Failure:
`#include <Matter.h>` --> `#include "matter_closure_patch.h"` --> ` #include "app/clusters/closure-control-server/closure-control-cluster-objects.h"` :: this last doesn't exist anymore within ESP-Matter v1.6 

## Test Scenarios
CI only

## Related links
https://github.com/espressif/esp32-arduino-lib-builder/pull/402


SuGlider:
@me-no-dev - This is just a quick fix in order to allow the Arduino Lib Builder to clone the Arduino Core 4.0.x and allow it to build all the artifacts. Lib Builder fails because it want to `#include "matter_closure_patch.h"` which has been removed.

This is a pre-requisite to move forward. This PR doesn't pass CI, but must be merged and the Lib Buidler PR shall be restarted.

PR #12929, opened 2026-09-22
fix(partitions): correct littlefs partition subtype in large_littlefs_32MB.csv

### LINKED ISSUE #12898: large_littlefs_32MB.csv contains wrong partition subtype
### Board

ESP32 S3

### Device Description

ESP32 S3

### Hardware Configuration

not needed

### Version

v3.3.11

### Type

Bug

### IDE Name

Arduino IDE 2.3.8

### Operating System

Windows 10

### Flash frequency

not needed

### PSRAM enabled

yes

### Upload speed

not needed

### Description

<APPDATA>\Local\Arduino15\packages\esp32\hardware\esp32\3.3.11\tools\partitions\large_littlefs_32MB.csv
it contains wrong subtype, not littlefs but spiffs.
> # Name,   Type, SubType, Offset,  Size, Flags
> nvs,      data, nvs,     0x9000,  0x5000,
> otadata,  data, ota,     0xe000,  0x2000,
> app0,     app,  ota_0,   0x10000, 0x480000,
> app1,     app,  ota_1,   0x490000,0x480000,
> spiffs,   data, spiffs,  0x910000,0x16E0000,
> coredump, data, coredump,0x1FF0000,0x10000,

### Sketch

```cpp
not needed
```

### Debug Message

```plain
not needed
```

### Other Steps to Reproduce

_No response_

pruthvi828:
## Description

Closes #12898

In `tools/partitions/large_littlefs_32MB.csv`, the data partition entry was incorrectly defined as `spiffs, data, spiffs` instead of `littlefs, data, littlefs`.

This prevented boards using the 32MB LittleFS partition scheme (`app5M_little24M_32MB` across ESP32-S3, ESP32-P4, M5Stack CoreS3, Cardputer, Waveshare, etc.) from mounting LittleFS out of the box when calling `LittleFS.begin()`.

### Key Changes
- Updated `tools/partitions/large_littlefs_32MB.csv` to specify partition name `littlefs` and subtype `littlefs` matching standard LittleFS partition layouts.

---

## Test Configuration
- Verified partition table syntax and board scheme mapping in `boards.txt`.


tools/partitions/large_littlefs_32MB.csv
```diff
@@ -3,5 +3,5 @@ nvs,      data, nvs,     0x9000,  0x5000,
 otadata,  data, ota,     0xe000,  0x2000,
 app0,     app,  ota_0,   0x10000, 0x480000,
 app1,     app,  ota_1,   0x490000,0x480000,
-spiffs,   data, spiffs,  0x910000,0x16E0000,
+littlefs, data, littlefs,0x910000,0x16E0000,
```
P-R-O-C-H-Y, line 6: Thanks for the PR.

`LittleFS.begin()` finds the partition by label, and the default label is `"spiffs"`. esp_littlefs looks it up with `ESP_PARTITION_SUBTYPE_ANY`, so the subtype is not what makes the mount fail. The current name "spiffs" is what the default LittleFS.begin() uses.

Renaming the partition to "littlefs" will make LittleFS.begin() fail with "partition could not be found".

The offsets and the size are already correct. If you only want the subtype to be littlefs, please keep the name:
`spiffs,   data, littlefs, 0x910000,0x16E0000,`

P-R-O-C-H-Y (changes requested):

PR #12930, opened 2026-09-22
fix(rgb_led): use SPI led_strip backend when RMT is unavailable

jajera:
## Description of Change

On SoCs without RMT (`SOC_RMT_SUPPORTED` unset), `rgbLedWrite()` / `rgbLedWriteOrdered()` currently only log an error and return. That breaks onboard WS2812 usage on ESP32-C61 (and ESP32-C2), including boards that define `RGB_BUILTIN` / `PIN_RGB_LED`.

This change adds an SPI backend via [`espressif/led_strip`](https://components.espressif.com/components/espressif/led_strip), matching the ESP-IDF `examples/get-started/blink` path when RMT is not available (`BLINK_LED_STRIP_BACKEND_SPI`).

### Changes
- `cores/esp32/esp32-hal-rgb-led.c`: SPI fallback with a cached `led_strip` handle (SPI owns the bus for the strip lifetime); maps `rgb_led_color_order_t` to `led_strip` color component format; keeps existing RMT path unchanged.
- `idf_component.yml`: depend on `espressif/led_strip` for `esp32c2` and `esp32c61`.
- `CMakeLists.txt`: `PRIV_REQUIRES led_strip` for those targets.

## Test Scenarios

- **Hardware:** ESP32-C61-DevKitC-1 (onboard WS2812 on GPIO8 / `RGB_BUILTIN`)
- **Software:** Arduino-ESP32 core (this branch, based on current `master`), PlatformIO `espressif32` with Arduino as ESP-IDF component so `led_strip` is resolved from `idf_component.yml`
- **Scenario:** Boot sequence calling `rgbLedWrite(RGB_BUILTIN, …)` for red → green → blue → off. Observed correct colors on the addressable LED. Serial showed successful init (`rgb ready`) with no SPI init errors.
- RMT-capable targets are unchanged (compile-time `#if SOC_RMT_SUPPORTED`).

**Note:** Uses `SPI2_HOST` with DMA, same as the IDF blink example. Applications that already initialize SPI2 for other devices may need to choose a different host or share carefully.

## Related links

- ESP-IDF blink LED strip backend selection: RMT when supported, otherwise SPI
- Component: https://components.espressif.com/components/espressif/led_strip


P-R-O-C-H-Y (commented):
Thanks for the proposal, but I don't think we should merge this.

We do not ship the `led_strip` component for any SoC. Adding it only for ESP32-C2 and ESP32-C61 would not work in the Arduino IDE anyway, those targets are not part of the package we build (C2 is hidden, C61 is skipped).

On top of that, both chips have a single general-purpose SPI and the Arduino SPI object already uses SPI2_HOST. Caching a strip on that bus means rgbLedWrite() and SPI.begin() cannot be used together.

I would close this. @me-no-dev @SuGlider What do you think?

me-no-dev:
> I would close this.

Agree

PR #12931, opened 2026-09-22
fix(zigbee): zero-initialize report_attr_cmd and set manuf_code in endpoint report helpers

### LINKED ISSUE #12917: Zigbee: report helpers pass an uninitialised manuf_code to esp_zb_zcl_report_attr_cmd_req()
### Board

M5Stack NanoH2 (ESP32-H2, SKU C149)

### Device Description

Bare module, USB-C, no shield

### Hardware Configuration

DS18B20 on a GPIO, 4k7 pull-up

### Version

v3.3.9

### Type

Bug

### IDE Name

Arduino IDE

### Operating System

Debin 13

### Flash frequency

64MHz

### PSRAM enabled

yes

### Upload speed

921600

### Description

Note, version also present unchanged on master today.

Every `report*()` helper in `libraries/Zigbee/src/ep/` declares
`esp_zb_zcl_report_attr_cmd_t` on the stack without initialising it, then sets every
field **except `manuf_code`**, and passes it to the stack. `manuf_code` is documented
in the SDK as *"The manufacturer code of the attribute to report"*, so the value the
stack receives is whatever was on the stack at that moment.

Representative case, `libraries/Zigbee/src/ep/ZigbeeAnalog.cpp:225` on master:

```cpp
bool ZigbeeAnalog::reportAnalogInput() {
  /* Send report attributes command */
  esp_zb_zcl_report_attr_cmd_t report_attr_cmd;
  report_attr_cmd.address_mode = ESP_ZB_APS_ADDR_MODE_DST_ADDR_ENDP_NOT_PRESENT;
  report_attr_cmd.attributeID = ESP_ZB_ZCL_ATTR_ANALOG_INPUT_PRESENT_VALUE_ID;
  report_attr_cmd.direction = ESP_ZB_ZCL_CMD_DIRECTION_TO_CLI;
  report_attr_cmd.clusterID = ESP_ZB_ZCL_CLUSTER_ID_ANALOG_INPUT;
  report_attr_cmd.zcl_basic_cmd.src_endpoint = _endpoint;
  report_attr_cmd.manuf_specific = 0x00U;    // Standard profile command. ...
  report_attr_cmd.dis_default_resp = 0x00U;  // Default response is enabled.

  if (!reportClusterAttribute(&report_attr_cmd)) { ... }
```

`manuf_code` is never assigned. The same is true of the padding bits around the
anonymous bitfield, and of `zcl_basic_cmd.dst_addr_u` / `zcl_basic_cmd.dst_endpoint`
(unused under this `address_mode`, but still passed as indeterminate bytes).

The struct, from `esp-zigbee-lib` 1.6.x, `zcl/esp_zigbee_zcl_command.h:169`:

```c
typedef struct esp_zb_zcl_report_attr_cmd_s {
    esp_zb_zcl_basic_cmd_t zcl_basic_cmd;
    esp_zb_zcl_address_mode_t address_mode;
    uint16_t clusterID;
    struct {
        uint8_t manuf_specific   : 2;
        uint8_t direction        : 1;
        uint8_t dis_default_resp : 1;
    };
    uint16_t manuf_code;   /*!< The manufacturer code of the attribute to report */
    uint16_t attributeID;
} esp_zb_zcl_report_attr_cmd_t;
```
### All affected call sites

17 declarations across 13 files in `libraries/Zigbee/src/ep/` on master, none of which
assigns `manuf_code`:

```
ZigbeeAnalog.cpp:227, 246
ZigbeeBinary.cpp:146, 284
ZigbeeCarbonDioxideSensor.cpp:123
ZigbeeElectricalMeasurement.cpp:278, 948
ZigbeeFlowSensor.cpp:115
ZigbeeIlluminanceSensor.cpp:100
ZigbeeMultistate.cpp:415, 434
ZigbeeOccupancySensor.cpp:72
ZigbeePM25Sensor.cpp:113
ZigbeePressureSensor.cpp:109
ZigbeeTempSensor.cpp:114, 161
ZigbeeWindSpeedSensor.cpp:123
```

### Why this looks like an oversight rather than a deliberate omission

The very same library sets the field, with the correct constant, on the *configure
reporting* path — e.g. `ZigbeeAnalog.cpp:277`, `ZigbeeTempSensor.cpp:94` and 192,
`ZigbeeElectricalMeasurement.cpp:247` and 908, and six other files all contain:

```cpp
reporting_info.manuf_code = ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC;
```

So the intended value for a standard attribute is known and spelled out in the library
already; only the one-shot report command path omits it.

### Expected vs actual

* Expected: the command struct handed to the stack is fully determined by the library.
* Actual: one documented field, plus bitfield padding and the unused destination
  fields, take whatever value the stack frame happened to hold. Behaviour depends on
  the call depth and the caller, which is exactly the kind of difference that makes a
  problem reproducible on one sketch and not another.

### Suggested fix

One line per helper:

```cpp
esp_zb_zcl_report_attr_cmd_t report_attr_cmd = {};
```

and, to be explicit rather than implicit, the same constant the reporting-info path
already uses:

```cpp
report_attr_cmd.manuf_code = ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC;
```

Note for whoever applies it: `ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC` is `0xFFFF`
in esp-zigbee-lib 1.x and `0x0000` in 2.x (where it is aliased to
`EZB_ZCL_STD_MANUF_CODE`, and `0xFFFF` becomes `EZB_ZCL_INV_MANUF_CODE`). The macro
name is portable across both; a literal is not. So `= {}` alone is *not* equivalent to
the named constant on 1.x — the assignment matters.

### Honest scope statement to include in the issue

> I do not have a runtime reproducer that isolates this field. I found it while
> debugging a device abort inside `esp_zigbee_zcl_command.c` and could not confirm
> that `manuf_code` was the cause — the stack may well ignore it while
> `manuf_specific` is 0, which would explain why the existing helpers work in
> practice. I am reporting it as an uninitialised value passed across an API boundary,
> which is a defect on its own terms and is cheap to close.

### Prior art to link

* espressif/esp-zigbee-sdk#728 — "Attribute reporting fails (TZ-2063)": another user
  reports the assert at `esp_zigbee_zcl_command.c:263` and notes it happens "in the
  self-report case".
* espressif/esp-zigbee-sdk#757 — "Zigbee reporting of custom cluster attributes
  (TZ-2212)": resolved by re-pairing after changing attributes, with the reporter
  asking for "a sensible error message rather than an unhelpful stack dump".

## Reason to file at all

* It is undefined behaviour in a library that thousands of sketches call, and it is
  one line per site to fix.
* The library is internally inconsistent about the same field, which is good evidence
  it was simply missed.
* These helpers are what everyone copies when they need to report an attribute the
  library does not wrap (the SDK documents no other pattern), so the omission
  propagates into third-party code — it did into ours.
* Low risk to accept: zero-initialising a stack struct cannot change the behaviour of
  any correct caller.

Note that I have asked Claude to help filing this issue. 

### Sketch

```cpp
https://github.com/mmattel/esp32/NanoH2_DS18B20_Zigbee
```

### Debug Message

```plain
No debug message available
```

### Other Steps to Reproduce

_No response_

pruthvi828:
Fixes #12917

### Summary
This PR zero-initializes the `esp_zb_zcl_report_attr_cmd_t` stack structure and explicitly assigns `report_attr_cmd.manuf_code = ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC;` across all Zigbee endpoint `report*()` helper methods in `libraries/Zigbee/src/ep/`.

### Problem
Previously, `esp_zb_zcl_report_attr_cmd_t` was declared uninitialized on the stack in 17 places across 12 endpoint files. Fields such as `manuf_code`, padding bits, and unused destination fields contained indeterminate stack values when passed to `reportClusterAttribute()`.

### Solution
1. Zero-initialize `esp_zb_zcl_report_attr_cmd_t report_attr_cmd = {};` to ensure clean padding and default fields.
2. Explicitly assign `report_attr_cmd.manuf_code = ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC;` matching the existing convention used in `ZigbeeEP.cpp` and the `reporting_info` configuration paths.

### Affected Files (12 files)
- `libraries/Zigbee/src/ep/ZigbeeAnalog.cpp`
- `libraries/Zigbee/src/ep/ZigbeeBinary.cpp`
- `libraries/Zigbee/src/ep/ZigbeeCarbonDioxideSensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeeElectricalMeasurement.cpp`
- `libraries/Zigbee/src/ep/ZigbeeFlowSensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeeIlluminanceSensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeeMultistate.cpp`
- `libraries/Zigbee/src/ep/ZigbeeOccupancySensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeePM25Sensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeePressureSensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeeTempSensor.cpp`
- `libraries/Zigbee/src/ep/ZigbeeWindSpeedSensor.cpp`


P-R-O-C-H-Y (changes requested):
Thanks for the PR.

I checked this against `esp_zb_zcl_report_attr_cmd_req()`. When `manuf_specific` is 0, the stack overwrites `manuf_code` with `0xFFFF` before it looks up the attribute or the reporting info. These helpers already set `manuf_specific = 0` (from #12325), so the uninitialized `manuf_code` is not what the lookup sees, and it is not what hits the assert.

`ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC` is that same 0xFFFF, so this assignment only repeats what the stack already does:

  `report_attr_cmd.manuf_code = ESP_ZB_ZCL_ATTR_NON_MANUFACTURER_SPECIFIC;`

Please drop that line and keep `manuf_specific = 0` and `dis_default_resp = 0`. The zero-init can stay, it clears the rest of the struct, but on this path the stack replaces manuf_code either way.

